PR #13, opened 2020-12-10
Initial proposal for a CRD that reasons over repos

viovanov:


repos/README.md
```diff
@@ -0,0 +1,8 @@
+# Repo CRD
+
+A custom resource definition for Kubernetes.
+
+- it allows the user to define a resource where they can store application sources and/or artifacts
+- once the resource is created, the user is presented with a protocol and location for uploading (in status)
```
mudler, line 6:
```
- once the resource is created, the user is presented with a protocol and location for uploading (in status)
```
Should we return an url to upload data, or just give ways to define where the sources are coming from?
Ideally, as an input I should be able to pick a supported `Repository` getter (that those can be extended to provide support to different source inputs, e.g. Github)

viovanov: I think we need to assume there's nothing else required outside of this.
If we need another source to pull from, we're shifting responsibility.
So given the user has sources, we provide a place to put them.

andreas-kupries:
I was confused by `that reasons over repos` ... I thought that the CRD provides some kind of reasoning capability for apps, etc, i.e. knows and/or can deduce dependencies and the like ...

Reading the description more closely this seems to propose a CRD/resource/controller which provides an abstract interface to the applications managed by carrier, providing access for uploading/querying sources, querying and manipulating state.

Behind that abstract interface we can then implement as we like, i.e for example use gitea as the storage engine.
Or something else, more lightweight.

Further unclear, who is the user in `... allows the user to define ...` ?
I.e. what role are we looking at here ?
  - carrier dev ?
  - carrier operator ?
  - app dev ?
  - app deployer ? (app operator, maybe a subset of carrier operators)

jimmykarily:
This is not part of any immediate goals. I'm closing this and we can revisit it when the time comes.

Issue #18, opened 2021-01-06
`carrier apps` output is (too) low-level (at the moment)


andreas-kupries:
`carrier apps` currently simply shows all the pods in the `eirini-workloads` namespace.

Not all of them may be associated with an app (Example: `el-staging-listener-...`), or may be remnants of the staging pipeline pods for an app. Although I suspect that the `staging-pipeline-...-run-...` pod is the app (when successfully deployed).

IMHO the `apps` method will need filtering and translation of the low-level pods into the actual active apps.

andreas-kupries:
With a successful deployment (`node-env`) I note that `staging-pipeline-...-run-...` is not the app. I see a separate pod for the app itself. So that has to be filtered out also.

andreas-kupries:
See #19 for a basic fix of the bash script. Nothing is done for the go code in development.

andreas-kupries:
Mooted by the go-based cli

PR #19, opened 2021-01-06
[fix] Remove display of internal pods from `apps` method

andreas-kupries:
See #18 in the `apps` bash script.

:warning: Note that the go code in the `installer` branch needs a different fix (assuming that the functionality is already present).

This is the reason this PR is not marked as fixing the issue.

### DIFF
--- a/apps
+++ b/apps
@@ -11,7 +11,7 @@ function apps() {
   app_name="$1"
   say "Getting app list"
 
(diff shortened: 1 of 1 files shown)


jimmykarily (approved):

andreas-kupries:
This is moot, the new go-based code does not have the problem.

PR #23, opened 2021-01-07
Installer

jimmykarily:
Create the basic structure of a Golang carrier cli. 

For now it only installs parts of Carrier. Need to merge this so more people can work on more aspects of it.


andreas-kupries (approved):
:+1: It is incomplete, that is not a barrier, we want more people to work concurrently. And any bugs still present will be fixed as part of the same.

PR #27, opened 2021-01-07
[feat] Debug/dev support. Script helper to inspect a cluster.

andreas-kupries:
Queries all available resource types, associated resources.
Dumps data into a directory containing resource lists, and
resource descriptions.

Does not query logs.

IOW this gives us the static cluster configuration.

Hack. Issues:
- Has problems with resource types from different vendors,
  with the same name. Examples:

  - events
  - ingresses
  - pods
  - nodes

k3s cluster with carrier/bash active.

### DIFF
--- a/scripts/kdump.sh
+++ b/scripts/kdump.sh
@@ -1,0 +1,88 @@
+#!/bin/bash
+
+dst="${1:-resources}"
+
+
+rm -rf   "${dst}"
+mkdir -p "${dst}"
+
+kubectl api-resources >> "${dst}/list-all"
+
+grep -v NAME  "${dst}/list-all" | while read name short api remainder
+do
+    case $short in
+	*v1*)
+	    # No shortname, api version is here. Shuffle around.
+	    api="${short}"
+	    short=""
+	    ;;
+	*) ;;
+    esac
+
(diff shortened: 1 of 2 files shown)


scripts/kdump.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash
+
+dst="${1:-resources}"
```
jimmykarily, line 3: gitignore this dir?

jimmykarily:
I feel inspection with k9s is usually quicker, because you can completely skip all the resources you don't care about. But, if you find this script useful for debugging then let's keep it.

jimmykarily (approved):

scripts/kdump.sh
andreas-kupries: Doen with commit 36b1ccf4d76edeca33eca62bd7c5c955b7315562.

andreas-kupries:
I believe both k9s and this have their place. Using this will work even if the cluster becomes unreachable (network issues, destroyed), etc. Can use as many editors and windows as my screen holds too, when havign to look at multiple things.
That said, I have made a note to self, to use `k9s` more also.

PR #28, opened 2021-01-07
GA workflow to validate PRs before merge.

### LINKED ISSUE #26: Need GA workflow for PRs
This workflow should run various linting tools over the go code.

I.e.
 - `go fmt`
 - `go vet`
 - `make test`
 - ...

These should all exist as Makefile targets, the workflow would then simply invoke them (or a single super target which calls on all)

Results requirements:

  - [x] Makefile targets for `fmt` and `vet`.
  - [x] Branch is clean with respect to `make test`, `... fmt`, and `... vet`.
  - [ ] PR workflow is specified, and working.

andreas-kupries:
Fixes #26 

The linting pieces I am confident in.
My changes to the GA flow, much less so.
Unclear how I could test them (beyond the PR having an approval and GA running things for it).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,6 +13,16 @@ jobs:
     steps:
       - name: Checkout
         uses: actions/checkout@v2
+      - name: Lint and Build Carrier
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.7'
+        run: make build
+      - name: Test Carrier
+        uses: actions/setup-go@v2
+        with:
+          go-version: '^1.13.7'
+        run: make test
       - name: Login to Docker Hub
         uses: docker/login-action@v1
         with:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -17,6 +17,8 @@ jobs:
         uses: actions/setup-go@v2
         with:
           go-version: '^1.13.7'
+      - name: Setup Ginkgo Test Framework
+        run: go get -u github.com/onsi/ginkgo/ginkgo
       - name: Lint and Build Carrier
         run: make build
       - name: Test Carrier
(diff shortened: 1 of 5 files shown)


mudler (approved):
Looks good here!

jimmykarily (approved):

andreas-kupries:
Found an issue, now that I found where the flow logs go ... From the log

> a step cannot have both the `uses` and `run` keys 

So, I guess I need a separate step to setup go, and the build/test steps can then run without anything.
Fix coming.

andreas-kupries:
Much better. This run of the flow made it to `Test Carrier` ... Need something to install `ginkgo` in GA, looks like.

andreas-kupries:
Ok. Flow passed, but it still used the bash-based carrier. Switch to go next, and disable the pieces we do not have yet!

andreas-kupries:
Failed ... Logs ... Ah! Of course.

I will now disable the step to install carrier with the go cli as well.
It has to wait for the cli to support non-interactive specification of config variables, notably the `system_domain`.
And we need a way to query the cluster's traefik's service's external ip directly. That should be doable with kubectl.

The flow error is clear:
```
🚢 Deploying Gitea
Couldn't install carrier: Failed installing Gitea: Error: Ingress.extensions "gitea" is invalid: spec.rules[0].host: Invalid value: "gitea.": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*')

Error: Process completed with exit code 1.
```

jimmykarily (approved):

Issue #24, opened 2021-01-07
carrier cli - system domain - traefik chicken and egg


andreas-kupries:
The description of the `system_domain` carrier configuration variable (*) contains the text
> Should be pointing to the traefik public IP

I believe that this pre-supposes that the underlying cluster already has `traefik` installed, and running, with an IP to use.

What happens for a cluster without `traefik` ?

While `carrier` then installs `traefik` for itself, at the point it does so the config variables have already been collected, from the user, who does not know what IP will be assigned to `carrier`'s `traefik`, right ? So what would the user enter at config collection time ?

Of course, if we go and decide that carrier should work only on traefik enabled clusters (**), then that deployment can be dropped from carrier itself.

----
(*) I decided to not call them options because of the confusion with cli options and the like.
(**) I currently only know of `k3s`/`k3d`.

PR #30, opened 2021-01-08
 Implemented support for non-interactive entry of config variables (system_domain)

### LINKED ISSUE #29: Need support for non-interactive entry of config variables (system_domain)
This was mentioned in #22. It has been split out of that because this feature is required in the tooling.
In other words, we cannot interactively enter the `system_domain` in the GA workflow testing PRs and `main`.
It needs a non-interactive way, be it cli option, or environment variable.

andreas-kupries:
Fixes #29.

  - Wrapped registration of install method into function.
    Moved into `install.go`, simplified `root.go`.
  - Made installer structure with deployments available to registration.
    (Early gathering of needed options, removed from command itself)
  - Added registration of cli options from the gathered config variables.
  - Extended `InstallationOption` structure with field `Default` to provide to cobra.
  - Updated gitea `system_domain` configuration to have a default.
  - Check for cobra option before going interactive.
  - Updated GA flow to trial carrier install again.

Note: Unclear how to test the new code paths. Existing code paths pass their unit tests.


.github/workflows/main.yml
```diff
@@ -50,9 +50,11 @@ jobs:
...
         run: |
           echo "`pwd`/output/bin" >> $GITHUB_PATH
-      # - name: Install Carrier
-      #   run: |
-      #     ./cli/dist/carrier install
+      - name: Check traefik service
+        run: kubectl get service -n kube-system traefik -o jsonpath='{.status.loadBalancer.ingress[0].ip}'
+      - name: Install Carrier
+        run: |
+          ./cli/dist/carrier install --system-domain "$(kubectl get service -n kube-system traefik -o jsonpath='{.status.loadBalancer.ingress[0].ip}').nip.io"
```
mudler, line 57: I think this somehow got lost - in the initial implementation on kubecfctl we used to generate this if not provided by the user, why not doing the same here?

mudler: For reference: https://github.com/mudler/kubecfctl/blob/master/pkg/kubernetes/installer.go#L31

cli/kubernetes/installer.go
```diff
@@ -54,15 +59,49 @@ type OptionsReader interface {
...
+func (i *Installer) PopulateNeededOptions(cmd *cobra.Command, reader OptionsReader) error {
 	if reader == nil {
 		reader = NewInteractiveOptionsReader(os.Stdout, os.Stdin)
 	}
 
 	var err error
 	newOptions := InstallationOptions{}
 	for _, opt := range i.NeededOptions {
-		opt.Value, err = reader.Read(opt)
+		var skipInteractive = false
```
jimmykarily, line 70: The original idea  of rthe OptionsReader was that you could make multiple calls to `PopulateNeededOptions` with a different OptionsReader each time. E.g.

```
...
PopulateNeededOptions(cobraReader)
PopulateNeededOptions(yamlReader)
PopulateNeededOptions(interactiveReader)
...
```
each call would fill in as many options as possible. This way you can have a set of options filled in  from command line, another set from a yaml file and some other interactively. This may be useful when for example the user stores most values in a yaml file but then wants to overwrite some on the command line. Secrets should probably never be written on disk thus a final interactive pass may be needed. (e.g. for the admin password)

With the above in mind, maybe we can wrap the cobra logic in a separate OptionsReader. What do you think?

.github/workflows/main.yml
andreas-kupries: Note #24

cli/kubernetes/installer.go
andreas-kupries: That sounds good. The main thing I see right now is the need for some kind flag indicating that the `opt.Value` is valid. That way each reader can skip the options already set by previous readers.

 Will look into this on Monday.

.github/workflows/main.yml
andreas-kupries: Ah! (Did not see the reference in the browser, only after refresh).
I will have a look if that can be added also.

andreas-kupries: Done with commits 118083a58150f52e11e39395b8b1f81a6a94d66d, ceeb8e1be9dc4c70db57e8318037dd65f78d20d3, and eb6d311ad65607705e12870fdcfdb76aab4be329

cli/kubernetes/installer.go
andreas-kupries: Done with commits 118083a58150f52e11e39395b8b1f81a6a94d66d, ceeb8e1be9dc4c70db57e8318037dd65f78d20d3, and eb6d311ad65607705e12870fdcfdb76aab4be329

cli/kubernetes/default_options_reader.go
```diff
@@ -0,0 +1,27 @@
...
+func NewDefaultOptionsReader() DefaultOptionsReader {
+	return DefaultOptionsReader{}
+}
+
+// Read attempts to fill the option with a default, dynamic or static.
+func (reader DefaultOptionsReader) Read(option *InstallationOption) error {
+	// Give priority to a function which provides the default
+	// value dynamically.
+	if option.DynDefault != nil {
+		err := option.DynDefault(option)
```
jimmykarily, line 17: This allows one to call a `DynDefault` function like this: 

```
option1.DynDefault(option2)
```
which of course makes little sense. We can make the invocation more intuitive by creating a "method" on the InstallationOption that calls the function set as an attribute:

```
func (option *InstallationOption) DynDefault() {
  option.DynDefaultFunc(option)
}
```
then [here](https://github.com/SUSE/carrier/pull/30/files#diff-c762dd02a47c7bb43d19bcf19914b1cc6294308fbcc4e1c2cfd4ecf4ddc3290aR71) you set the `option.DynDefaultFunc` and the invocation would then be:

```
  option.DynDefault()
```

It's more code for the same thing but I think it improves the readability of the invocation by wrapping the "ugly" part in a pre-defined function.

I like the function attribute idea for the default, +1.

andreas-kupries: Agreed. Missed that my code allows for option/argument mismatch.

cli/kubernetes/default_options_reader.go
```diff
@@ -0,0 +1,27 @@
...
+		if err != nil {
+			return err
+		}
+	} else if option.Default != nil {
+		option.Value = option.Default
+		option.Valid = true
```
jimmykarily, line 23: In the future we could have a  `Valid()` method on the InstallationOption that could have more complicated logic (e.g. checking that the system_domain actually looks like a domain). That function immediately return `true` if the Value is equal to the default one. Let's think about it when we need that validation logic.

andreas-kupries: I would call such a function `Validate()`, and would have it call a `ValidateFunc` function field of `InstallationOption`, like we have a `DynDefaultFunc`. Agreed that this can be deferred to until after the demo.

cli/kubernetes/default_options_reader.go
andreas-kupries: Done with commit 40a6401915c80f94e70b7575be98ab7224829041

cli/kubernetes/interactive_options_reader.go
```diff
@@ -21,10 +21,16 @@ func NewInteractiveOptionsReader(stdout io.Writer, stdin io.Reader) InteractiveO
...
-// Read aks the user what value should the given InstallationOption have and
+// Read asks the user what value should the given InstallationOption have and
 // returns that value validated and converted to the appropriate type as defined
 // by the Type field of the InstallationOption.
-func (reader InteractiveOptionsReader) Read(option InstallationOption) (interface{}, error) {
+func (reader InteractiveOptionsReader) Read(option *InstallationOption) error {
+	// Ignore anything which is already valid, be it default, or
+	// user-specified (cli option, etc.)
+	if option.Valid {
+		return nil
```
jimmykarily, line 31: I think the common case for cli's is that when you are in interactive mode and a default value exists something like this is printed:

```
$ system_domain [172.19.0.2.nip.io] :  # (Hit enter if default is fine)
```
and you can simply accept the default by hitting "Enter". This UX might be better because:

1. that's what people are used to
2. Allows you to override defaults in interactive mode
3. Doesn't need the `Valid` flag (which could be used then solely for actual Validation purposes)

Every other reader (other than the Interactive one) would implicitly/silently use the default value when another one is not available. 

If you agree with the above, then we don't need a DefaultOptionsReader but simply a `Default()` method which would try a hardcoded default and the dynamic default (the function). The dynamic default function should probably cache its result on the `Default` attribute so that its faster the next time the default value is requested.

andreas-kupries: I should possibly call the `Valid` flag: `UserSpecified`. In the current code this is nearer to how it is used. And in your proposal it would be exactly how it would be used. IOW interactive mode asks for any option but those already specified by the user.

My understanding of you idea for the `Default()` method is that it would be used by the interactive reader to show and set the value. Should also have cli option `--non-interactive`, (`-n`) where the interactive would simply accept all the defaults.

With respect to caching, each option/default is used only once at the moment. There is only the `install` command. And even in case of multiple commands the options (declarations) would not be not shared between them in the current setup.

Validation purposes: Would not need a `Valid` flag as a failed validation would terminate early, printing an error, not set a flag to abort late.

cli/cmd/install.go
```diff
@@ -4,26 +4,89 @@ package cmd
...
+		// reader or interactive entry
+		return nil
+	}
+
+	installer.PopulateNeededOptions(kubernetes.NewDefaultOptionsReader())
+	installer.PopulateNeededOptions(kubernetes.NewCLIOptionsReader(cmd))
+	installer.PopulateNeededOptions(kubernetes.NewInteractiveOptionsReader(os.Stdout, os.Stdin))
+	// At last, check that all options are valid.
+	fmt.Println("Configuration...")
+	installer.PopulateNeededOptions(kubernetes.NewCheckOptionsReader())
```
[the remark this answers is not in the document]
andreas-kupries: It is not truly "Validation" in the sense of "Is the value syntactically valid, etc". Only a check that the value got set by one of the preceding phases, and that nothing slipped through our nets.

Proper value validation I would put into `Validate()` method of the option, and have it called by each of the Readers, before they enter the user value (cli option or interactive) into the structures. I.e. early check in each phase instead of late check after all phases.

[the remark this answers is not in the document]
andreas-kupries: For interactive entry this would also loop the entry prompt, as you already do for boolean options. (Could maybe subsumed into a standard validator for booleans).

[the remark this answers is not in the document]
andreas-kupries: I will put the display of the entered configuration into a separate method of the installer.

cli/kubernetes/interactive_options_reader.go
jimmykarily: - +1 for the `--non-interactive`
- Maybe it's nit picking at the moment but we should be able to call `Default()` simply to check what the default value is anywhere in the code without worrying that this may trigger some slow `kubectl` command (e.g. find the external ip of the traefik service). Currently it wouldn't make any difference that's true but when it becomes important we may forget it needs refactoring. Feel free to postpone this for when it becomes an issue.
- Regarding Valid flag I was thinking that in the case of non-interactive readers (e.g. yaml configuration) we may want to check the validity of all options first and then print all the errors at once so the user doesn't have to call the installer again and again until she discovers all the problems.

jimmykarily: Feel free to postpone anything that can be postponed for later (e.g. `--non-interactive` flag, caching etc)

cli/cmd/install.go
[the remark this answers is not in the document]
andreas-kupries: Addressed with commit f1027e9040f094d29697aa5e7a16c88ccd9e55cc

cli/kubernetes/interactive_options_reader.go
andreas-kupries: Addressed with commit f1027e9040f094d29697aa5e7a16c88ccd9e55cc

andreas-kupries: I hate it when github does not update properly, hiding new comments from me

> Maybe it's nit picking at the moment but we should be able to call Default() simply to check what the default value is anywhere in the code without worrying that this may trigger some slow kubectl command (e.g. find the external ip of the traefik service). Currently it wouldn't make any difference that's true but when it becomes important we may forget it needs refactoring. Feel free to postpone this for when it becomes an issue.

This makes sense. I would place that into a separate field though, `CachedDefault` or some such. Would be needed because currently the func has precedence over `Default`, so caching into that field would simply not work.

> Regarding Valid flag I was thinking that in the case of non-interactive readers (e.g. yaml configuration) we may want to check the validity of all options first and then print all the errors at once so the user doesn't have to call the installer again and again until she discovers all the problems.

True. So, interactive reader does validation as user enters values, and a separate validation phase at the end, to catch everything from yaml files, options, env.variables, and so on, after everything is resolved to final values.

cli/kubernetes/interactive_options_reader.go
```diff
@@ -34,51 +39,79 @@ func (reader InteractiveOptionsReader) Read(option InstallationOption) (interfac
...
+	if userValue == "" {
+		// Keep the default set by (**).
+		return nil
+	}
+
+	// TODO (post MVP): String and integer types should be
+	// extended to call an option-specific validation function, if
+	// present, which would perform additional checks on the
+	// user's value. For example range limits, proper syntax of
+	// the string, etc. Thye would then loop similar to the entry
```
jimmykarily, line 69:
```suggestion
	// the string, etc. They would then loop similar to the entry
```

jimmykarily (approved):

cli/kubernetes/interactive_options_reader.go
andreas-kupries: Fixed, commit 37c23287068186b7dec35567e5e5129950ce972c

cli/kubernetes/cluster.go
```diff
@@ -12,10 +12,10 @@ import (
 	"github.com/kyokomi/emoji"
 	"github.com/pkg/errors"
 
-	generic "github.com/mudler/kubecfctl/pkg/kubernetes/platform/generic"
-	ibm "github.com/mudler/kubecfctl/pkg/kubernetes/platform/ibm"
-	k3s "github.com/mudler/kubecfctl/pkg/kubernetes/platform/k3s"
-	kind "github.com/mudler/kubecfctl/pkg/kubernetes/platform/kind"
+	generic "github.com/suse/carrier/cli/kubernetes/platform/generic"
+	ibm "github.com/suse/carrier/cli/kubernetes/platform/ibm"
+	k3s "github.com/suse/carrier/cli/kubernetes/platform/k3s"
```
mudler, line 17: 👍 went completely unnoticed!

andreas-kupries: Until I tried to fix the issue with the internal/external IPs for k3s and was wondering why my edits to the local files were ignored in the binary. To the point that I added panic statements to see if my code was actually run. It wasn't, and then the :bulb: came on after looking really closely.

PR #32, opened 2021-01-12
Tests and Fixes

andreas-kupries:
Followup on PR #30 for Ticket #29.

Mainly extending the testsuites for all the new functionality.
This also fixes issues in the CLIOptionsReader not seen before and luckily not breaking the usage of `install` in the GA flow.
Because the issue found was not triggered in the manual testing of cli processing.
:+1: for testsuites.
Generally added some more checks to the code itself, in various places.

### DIFF
--- a/cli/kubernetes/options_test.go
+++ b/cli/kubernetes/options_test.go
@@ -29,7 +29,7 @@ var _ = Describe("InstallationOption", func() {
 				return nil
 			},
 		}
-		It("it calls the DynDefaultFunc", func() {
+		It("calls the DynDefaultFunc", func() {
 			Expect(option.DynDefault()).To(BeNil())
 			Expect(option.Value).To(Equal("Hello"))
 		})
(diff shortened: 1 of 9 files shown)


cli/kubernetes/options_test.go
```diff
@@ -17,6 +19,67 @@ var _ = Describe("InstallationOption", func() {
...
+		optionError := InstallationOption{
+			Name:         "TheName",
+			DeploymentID: "SomeDeployment",
+			DynDefaultFunc: func(o *InstallationOption) error {
+				o.Value = "Hello"
+				return errors.New("an error")
+			},
+		}
+
+		It("it prefers the DynDefaultFunc over a static Default", func() {
```
jimmykarily, line 67: [minor] I think the "convention" is `It("prefers..."` (without an `it` in the text). In other places as well.

jimmykarily (approved):

cli/kubernetes/options_test.go
andreas-kupries: Yes. Tried to follow it, missed these. Will fix.

andreas-kupries: Fixed with commit `6a70a803f0b186c609364a3da9550d393a338e95`

PR #37, opened 2021-01-13
Add minikube platform 

rohitsakala:


andreas-kupries:
I am not convinced that it is working.
After fiddling with my system to get minikube working again (*) I ran the following session:

```
ork@tagetarl:~/SUSE/dev/carrier> time ./cli/dist/carrier-linux-amd64 install |& tee LOG.txt 
Carrier installing...
⚠️  No valid platform detected, trying general platform. Things might go wrong
⚓ Detected kubernetes platform: generic
🌎 ExternalIPs: []
➰ InternalIPs: [192.168.39.168]
[Shared] system_domain The domain you are planning to use for Carrier. Should be pointing to the traefik public IP []: 192.168.39.168.nip.io
Configuration...
  🧭 system_domain:	'192.168.39.168.nip.io'
🚢 Deploying Traefik Ingress
✔️  Traefik Ingress deployed
🚢 Deploying Quarks
✔️  Quarks deployed
🚢 Deploying Gitea
✔️  Gitea deployed
Carrier installation complete.

real	1m14.059s
user	0m1.603s
sys	0m0.403s
work@tagetarl:~/SUSE/dev/carrier> k get service :traefik
fetching service => 12
kubectl get service traefik --namespace traefik
NAME      TYPE           CLUSTER-IP     EXTERNAL-IP   PORT(S)                      AGE
traefik   LoadBalancer   10.98.24.186   <pending>     80:32661/TCP,443:31588/TCP   12m
work@tagetarl:~/SUSE/dev/carrier> k get nodes -o json|jq '.items[0].status.addresses'
kubectl get nodes -o json
[
  {
    "address": "192.168.39.168",
    "type": "InternalIP"
  },
  {
    "address": "m01",
    "type": "Hostname"
  }
]
```

andreas-kupries:
Added some scripts taken from kubecf for starting/stopping a minikube cluster, plus make targets.

cli/kubernetes/platform/minikube/minikube.go
```diff
@@ -0,0 +1,47 @@
...
+func (m *Minikube) String() string { return "minikube" }
+
+// Detect detects if it is a minikube platform.
+func (m *Minikube) Detect(kube *kubernetes.Clientset) bool {
+	nodes, err := kube.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+	if err != nil {
+		return false
+	}
+	for _, n := range nodes.Items {
+		if n.Name == "minikube" {
```
thardeck, line 32: The name of the node might be not a good identifier because you start a different cluster with the name `minikube`.
Since `ProviderID` is not there a better alternative might be scanning the Label keys for `minikube` instead.

thardeck (changes requested):
Label keys might be a safer indicator for recognizing Minikube.

cli/kubernetes/platform/minikube/minikube.go
viccuad: Labels on the nodes with keys in the form of `minikube.k8s.io/*` are present at least from minikube 1.9.2 to latest (1.16).

So, one could do:

```
		l := n.GetLabels()
		if _, found := l["minikube.k8s.io/version"]; found {
			return true
		}
```

andreas-kupries:
I confirm that with `minikube 1.16` the carrier code properly detects the minikube cluster.

cli/kubernetes/platform/minikube/minikube.go
andreas-kupries: And v 18.2 has the labels as well.
```
    minikube.k8s.io/commit: eb13446e786c9ef70cb0a9f85a633194e62396a1
    minikube.k8s.io/name: minikube
    minikube.k8s.io/updated_at: 2021_01_14T16_02_02_0700
    minikube.k8s.io/version: v1.8.2
```
Adding my vote to using labels as well.

viccuad:
Carrier finishes install on Minikube, but the traefik service externalIP never gets out of <pending>. @andreas-kupries saw something similar in the past too IIRC.

cli/kubernetes/platform/minikube/minikube.go
rohitsakala: Done. Changed to node labels

andreas-kupries:
> Carrier finishes install on Minikube, but the traefik service externalIP never gets out of <pending>. @andreas-kupries saw something similar in the past too IIRC.

Yes, I did see the same in my trials yesterday, see https://github.com/SUSE/carrier/pull/37#issuecomment-760208544, notes entry 2.

rohitsakala:
> > Carrier finishes install on Minikube, but the traefik service externalIP never gets out of . @andreas-kupries saw something similar in the past too IIRC.
> 
> Yes, I did see the same in my trials yesterday, see [#37 (comment)](https://github.com/SUSE/carrier/pull/37#issuecomment-760208544), notes entry 2.

LoadBalancer types are only for public clouds. So, we don't get an externalIP address for kind and minikube.

andreas-kupries:
Hm. Hm. We are seeing a LoadBalancer with external-IP when deploying via `k3s`/`k3d`. On the other side, this platform has `Traefik` pre-installed in itself, so carrier does not install any.

viccuad:
could we depend on ingress being installed on minikube? (`minikube addons enable ingress`)

andreas-kupries:
I have no idea how that minikube addon would interact with carrier's traefik.

andreas-kupries:
That said, I would be ok to merge, and revisit if testing with a working push shows that the missing external-ip is truly an issue for the platform and pushed apps. Or an issue with pushing apps (because push needs access to the gitea endpoint, and that may require access through traefik).

rohitsakala:
> That said, I would be ok with merging the PR, and revisit if testing with a working push shows that the missing external-ip is truly an issue for the platform and pushed apps. Or an issue with pushing apps (because push needs access to the gitea endpoint, and that may require access through traefik).

I checked how K3s is getting an external IP. Klipper-lb is installed with K3s which is helping the provision os externalIP. I found metallb addon in minikube which acts like klipper-lb. Added a commit to enable metallb. Danke.

andreas-kupries:
Looks ok, has an external-ip now
```
fetching service => 12
kubectl get service traefik --namespace traefik
NAME      TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)                      AGE
traefik   LoadBalancer   10.108.63.249   192.168.49.0   80:31115/TCP,443:31691/TCP   6m15s
```

andreas-kupries (approved):
:+1:

thardeck (approved):

PR #41, opened 2021-01-13
CLI for interacting with apps

viovanov:


cli/cmd/internal/client/create_org.go
```diff
@@ -0,0 +1,47 @@
...
+	SilenceErrors: true,
+	SilenceUsage:  true,
+}
+
+func init() {
+	pf := CmdCreateOrg.PersistentFlags()
+
+	argToEnv := map[string]string{}
+
+	cmd.KubeConfigFlags(pf, argToEnv)
```
jimmykarily, line 45: From [the docs](https://github.com/spf13/cobra#persistent-flags):

```
A flag can be 'persistent', meaning that this flag will be available to the command it's assigned to as well as every command under that command. For global flags, assign a flag as a persistent flag on the root.
```
Given this method only adds `kubeconfig` et al (and nothing specific to this command or any subcommands), I think we should only do this on the root cmd, the probably in `main` (not `init`).

cli/paas/client.go
```diff
@@ -0,0 +1,389 @@
...
+	c.giteaClient.CreateRepoHook(c.config.Org, name, gitea.CreateHookOption{
+		Active:       true,
+		BranchFilter: "*",
+		Config: map[string]string{
+			"secret":       HookSecret,
+			"http_method":  "POST",
+			"url":          StagingEventListenerURL,
+			"content_type": "json",
+		},
+		Type: "gitea",
```
andreas-kupries, line 215: What events is this hook triggering on ?
As I see no events listed at all I suspect this will not be triggered by anything.
Which would explain why nothing happens when I try to push the sample app.

andreas-kupries (changes requested):

cli/paas/client.go
andreas-kupries: Hm. Adding 
> Events: []string{"push"},
did not help. Same nothingness and same timeout.

Note: Local cluster is `k3s` based. (In case this is different from the setup used by @viovanov ).

andreas-kupries: Ok. I see that the event listener is triggered (not sure if because of the event decl I added, will be checked by removing.

```
{"level":"info","ts":"2021-01-18T11:22:38.972Z","logger":"eventlistener","caller":"sink/sink.go:237","msg":"ResolvedParams : [{Name:gitrevision Value:main} {Name:namespace Value:eirini-workloads} {Name:gitrepositoryurl Value:http://gitea.gitea:10080/workspace/sample4} {Name:appname Value:sample4}]","knative.dev/controller":"eventlistener","/triggers-eventid":"dhllt","/trigger":""}
{"level":"info","ts":"2021-01-18T11:22:38.974Z","logger":"eventlistener","caller":"resources/create.go:95","msg":"Generating resource: kind: &APIResource{Name:pipelineruns,Namespaced:true,Kind:PipelineRun,Verbs:[delete deletecollection get list patch create update watch],ShortNames:[pr prs],SingularName:pipelinerun,Categories:[tekton tekton-pipelines],Group:tekton.dev,Version:v1beta1,StorageVersionHash:RcAKAgPYYoo=,}, name: staging-pipeline-run-sample4-p5wps","knative.dev/controller":"eventlistener"}
{"level":"info","ts":"2021-01-18T11:22:38.974Z","logger":"eventlistener","caller":"resources/create.go:103","msg":"For event ID \"dhllt\" creating resource tekton.dev/v1beta1, Resource=pipelineruns","knative.dev/controller":"eventlistener"}
```

So there should be a `staging-pipeline-run-sample4-p5wps` somewhere. ... Seeing no pod however. ... Ah, a `pipelineruns` type ...

Look at that, all my attempts:

```
ork@tagetarl:~/SUSE/dev/carrier> k get pipelineruns :
kubectl get pipelineruns --all-namespaces
NAMESPACE          NAME                                 SUCCEEDED   REASON           STARTTIME   COMPLETIONTIME
eirini-workloads   staging-pipeline-run-sample-wzws7    False       CouldntGetTask   132m        132m
eirini-workloads   staging-pipeline-run-sample2-skg7m   False       CouldntGetTask   18m         18m
eirini-workloads   staging-pipeline-run-sample3-74fsc   False       CouldntGetTask   4m35s       4m35s
eirini-workloads   staging-pipeline-run-sample4-p5wps   False       CouldntGetTask   3m28s       3m28s
```

Even before I added "Events" to the structure, so that is indeed not required.
Retracting my change request here. Likely something in my setup now, not in the code.

andreas-kupries (approved):

cli/paas/client.go
andreas-kupries: Getting more details on the object I find

```
status:
  completionTime: "2021-01-18T09:13:12Z"
  conditions:
  - lastTransitionTime: "2021-01-18T09:13:12Z"
    message: 'Pipeline eirini-workloads/staging-pipeline can''t be Run; it contains
      Tasks that don''t exist: Couldn''t retrieve Task "buildpacks": tasks.tekton.dev
      "buildpacks" not found'
    reason: CouldntGetTask
    status: "False"
```

andreas-kupries: Now I wonder if the go-based tekton deployment is incomplete ? :bell: @jimmykarily

andreas-kupries: Ok. We have tekton's `tasks` CRD.
```
work@tagetarl:~/SUSE/dev/carrier> k api-resources|grep tekton|grep tasks
kubectl api-resources
clustertasks                                   tekton.dev                     false        ClusterTask
tasks                                          tekton.dev                     true         Task
```
We have no actual tasks running however, and something else expects a `buildpacks` task to be present.
```
work@tagetarl:~/SUSE/dev/carrier> k get tasks :
kubectl get tasks --all-namespaces
No resources found.
```

andreas-kupries: Ok, with the bash-based carrier the above commands return the same results. So that is not the exact difference.
That there is a difference between the go and bash tektons however is evident.
I am currently pushing to a cluster initialized from the bash scripts, and there pushing the sample-app does work.

I will see if I can dig out what the relevant differences are.

jimmykarily: That Task should be here: https://github.com/SUSE/carrier/blob/main/cli/embedded-files/tekton/staging.yaml#L4-L13

PR #43, opened 2021-01-14
Remove incluster kube config as it is unnecessary

rohitsakala:


andreas-kupries (approved):
I am approving.

While I first thought that it was not working I have determined that my issue is present on main as well. IOW the branch has inherited the issue, and not introduced it.

The change looks to be a simplification of the code, so I am ok.
I will make a separate ticket for the issue.

PR #49, opened 2021-01-19
Use correct port for registry and add missing Eirini configuration

jimmykarily:
We used the same port both for http and https (which is wrong). Also the

```
registry_secret_name: "cluster-registry-creds-http"
```

setting doesn't exist in the upstream Eirini yamls and should be added.


jimmykarily:
For all other embedded files, we have code in the Makefile (see `update_*` targets) that copy them from other directories or fetch them from the internet. For routing.yml I just copied that to the embedded_files directory because the source directory (https://github.com/SUSE/carrier/blob/main/eirini/eirini-release/deploy/core/routing.yaml) will be deleted after we cleanup the bash installer.

cli/statik/statik.go
```diff
@@ -6,7 +6,9 @@ import (
...
...fe\xc1\xd7fCT\xe4\xe0Y\x1c\xe9A\xb7\xe3\xda\xd4\"\xe1f@\xce\xfe}\xfe~\xc5\xa3\x97S\xaa\xd5\xe1\xfe\x1e\x90\x92\xffg\xcc`6\xcc\xe1\xeb\x90\xcb\xe2\x0eT[\x1d(\x90\xff\x05YL\xcb\xffC\xa1\x8d\xff\xbb\x1fH\xc8\xff\x0e\xca)\xcc\x8b\x18\xe0\xaf\x9a\xae\x13\xc3\xbc#\xd1\xac \x81@BB\x81\x0bv\x18\x95\x84\xf2\x10A\xa3\xa2\x03\x12\xc2\xe2F\xd5l\xc2Y\x18Z\xd4	X\xcd\x81\xbd\x00\xa9f\xea\xa7\x1c\x84\x0c\xc7D\x03\xdb\xab\x8bw\xc7\xcf\x84N\xc7\xf2\x9c\xb9\x1a\x14\xc0}\x0c\xa3\xf39\x8e\x8e\xc6D\xc2\xdd\xcb\xfc\x08\x98'\xbb\xc3Jl\x87\x8e%~4 (C\x98\xc0\xfa\xa2\xda\xb9\xe8\x0fl\x16\x17\xd8\xf58\xb4\x1c\x93\x83\xc3\xd3\x8bc?\xb2\x0bJ\x0c\x07\x84\xa3\xaa\x8a\xac\xb3\x1b\xda\x95Ot?\xaa\x129&\xd0\x9a\xf1\xb3\xa8](\x1d\xd0\x07\x87\xf4\x07~\xdc\x19\xd0P\xbe]#\x01\xc7\x84,_\xc8KHu\xff<T\xd9\xed\xa1\xe1\xc1 \x8a\xff\xfc\x0f!\xa6\x85\xa46j4\xc0\x85\x08@\x83\x82~\xd9de\xba\x8d\xd8x\xa0L,\xf7\x03\xff\xc46G\xb6\xf4\xc9\xa5))\x19\x93\x9b\xda\xe6\x82C2\x1d\x87\x03u\xdc;\\\x0eY\xec\xabB7|\x0c\x07<\xfc\n\xbf:\xbeRS~\x88ts\x06R\xe5\x8c\xe2\x05\xdc\xa2q\xda\xaa\xb7Y\xc4;\xefr@\xc7W\xa0\xce}\xdb\x05\xb9\xc12\xad/n\xdfL\xf1\xff\xe0\xcf\x86-@\x05\xfc_\x1c\xcai\xff\x7fI\xe2[\xfe\xbf\x17X\xc7\xffMbi\x16\x9bR\x0d\xe4\xd1\xab\
```
andreas-kupries, line 14: Nit: Newline at end of file is missing.

jimmykarily: This is a generated file: https://github.com/SUSE/carrier/blob/main/cli/Makefile#L59

jimmykarily: `make tools` always pulls the latest statik afaict. So this diff may have been the result of an updated statik. Should we stub the statik version?

jimmykarily: Last commit of statik is from April, so not the case.

andreas-kupries: I suspect that a `make fmt` may fix the issue.

Wrt latest statik ... For reproducible builds we really have to stub it, I believe.

andreas-kupries (approved):
Looks ok.

cli/statik/statik.go
jimmykarily: Missed this before I merged. I fixed it on `main` directly: https://github.com/SUSE/carrier/commit/6759a2cbbc9d74fda17d2ac20705b6a8d31ef8e4

I'll also have a look on how to stub the version.

PR #51, opened 2021-01-19
Add uninstall cli command

rohitsakala:
Co-authored-by: Tim Hardeck <thardeck@suse.com>


cli/kubernetes/installer.go
```diff
@@ -101,6 +101,18 @@ func (i *Installer) Install(cluster *Cluster) error {
 	return nil
 }
 
+func (i *Installer) Uninstall(cluster *Cluster) error {
+	// TODO: check if we should uninstall deployments in a
+	// specific order
```
andreas-kupries, line 106: Maybe simply in reverse order compared to `install` ?

jimmykarily:
I was wondering if uninstalling Carrier should also delete the workloads (the "Jumanji" style undo, everything that happened as a result of Carrier being installed gets reverted). Maybe a "soft" uninstall makes sense, let's see if anyone ever asks for it.

andreas-kupries:
Heh. and I was wondering if it would the full uninstall (i.e. apps as well) a moment ago, when plotting my test steps, and if I should add an app push to them before doing the uninstall.

The basic plan is, at the moment:

  1. k3s up
  2. `scripts/kdump.sh pre-install` # dump cluster contents, all resources
  3. `carrier install`
  4. dump again - `post-install`
  5. (push app ?)
  6. (dump - `post-push` ?)
  7. `carrier uninstall`
  8. dump - `post-uninstall`
  9. compare the resource lists `pre-install` vs `post-uninstall`. Should be same. Except if we do leave apps running.

cli/deployments/traefik.go
```diff
@@ -45,8 +47,32 @@ func (k Traefik) Describe() string {
...
+	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+	s.Suffix = emoji.Sprintf(" %s :zzz:", message)
+	s.Start()
+
+	currentdir, err := os.Getwd()
+	if err != nil {
+		return errors.New("Failed uninstalling Traefik: " + err.Error())
+	}
+
+	helmCmd := fmt.Sprintf("helm uninstall traefik --namespace '%s'", traefikDeploymentID)
```
jimmykarily, line 62: This one is tricky. If we find traefik already running on the cluster [we simply skip the installation](https://github.com/SUSE/carrier/blob/main/cli/deployments/traefik.go#L106) (as it happens on k3d for example).

We need a way to mark resources that belong to Carrier so we can decide whether to delete them or not when uninstalling. Maybe we can use a label on the namespace? There are other ways to achieve the same but I can't think of any better.

andreas-kupries: :+1: on that being tricky and needing a more careful approach. Well, I will see what happens on the k3s cluster I use. IIRC rohit uses minikube, where traefik is installed by carrier, so the issue would not be seen.

andreas-kupries: Other ways: Maybe a carrier-specific config map which lists the namespaces/resources carrier created/has ownership of ?  ... No, on further thought I do not like that. It has the information separate from the resources itself, allowing for desync. Labels on the namespace might really be best. (I do not see an easy way to label all the resources in the namespace, and it might be overkill also).

thardeck: Rohit already suggested to tag the resources and delete them but we decided against it to keep it simple for now.
The point does make sense, the question though is if we really want to support multi-tenants, because atm all namespaces and similar seem to be static and would potential conflict in such an environment.

thardeck: Rohit already suggested to tag the resources and delete them but we decided against it to keep it simple for now.
The point does make sense, the question though is if we really want to support multi-tenants, because atm all namespaces and similar seem to be static and would potential conflict in such an environment.

jimmykarily: Right now, the only component we are re-using if available is traefik. In the future we may decide to do the same with tekton or others. In any case, if we re-use something which exists, we shouldn't try to delete it on uninstall. To make sure we don't accidentally re-use an existing namespace, we could prefix all the namespaces we create with `carrier-` or something similar.

andreas-kupries: This is not about mutli-tenant. Ir is about differences in the foundations.

For a minikube cluster carrier installs traefik.
For a k3s cluster traefik comes pre-installed.

Trying to uninstall traefik there may break cluster.

Actually not true, as I just reached step 7 in my plan (above) and got the following error:

```
Carrier uninstalling...
⚓ Detected kubernetes platform: k3s
🌎 ExternalIPs: [192.168.176.2]
➰ InternalIPs: [192.168.176.2]
⣾ Deleting Traefik 💤  Couldn't install carrier: Failed uninstalling Traefik: Error: uninstall: Release not loaded: traefik: release: not found


real	0m0.177s
user	0m0.151s
sys	0m0.029s
```

I.e. the attempt to uninstall traefik ran into the fact that carrier had not installed it, so there was no proper helm release to act on.

cli/deployments/eirini.go
```diff
@@ -55,8 +57,45 @@ func (k Eirini) Describe() string {
...
+	out, err := helpers.Kubectl("delete -f " + file)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("%s failed:\n%s", message, out))
+	}
+	file = path.Join(releaseDir, "deploy", "workloads", "namespace.yml")
+	out, err = helpers.Kubectl("delete -f " + file)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("%s failed:\n%s", message, out))
+	}
+	s.Stop()
```
andreas-kupries, line 94: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

cli/deployments/gitea.go
```diff
@@ -52,8 +54,32 @@ func (k Gitea) Describe() string {
...
+	helmCmd := fmt.Sprintf("helm uninstall gitea --namespace %s", giteaDeploymentID)
+	if out, err := helpers.RunProc(helmCmd, currentdir, k.Debug); err != nil {
+		return errors.New("Failed uninstalling Gitea: " + out)
+	}
+
+	err = c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), giteaDeploymentID, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.New("Failed uninstalling Gitea: " + err.Error())
+	}
+	s.Stop()
```
andreas-kupries, line 78: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

cli/deployments/quarks.go
```diff
@@ -45,8 +47,32 @@ func (k Quarks) Describe() string {
...
+	helmCmd := fmt.Sprintf("helm uninstall quarks --namespace %s", quarksDeploymentID)
+	if out, err := helpers.RunProc(helmCmd, currentdir, k.Debug); err != nil {
+		return errors.New("Failed uninstalling Quarks: " + out)
+	}
+
+	err = c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), quarksDeploymentID, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.New("Failed uninstalling Quarks: " + err.Error())
+	}
+	s.Stop()
```
andreas-kupries, line 71: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

cli/deployments/registry.go
```diff
@@ -52,8 +54,32 @@ func (k Registry) Describe() string {
...
+	helmCmd := fmt.Sprintf("helm uninstall '%s' --namespace '%s'", registryDeploymentID, registryDeploymentID)
+	if out, err := helpers.RunProc(helmCmd, currentdir, k.Debug); err != nil {
+		return errors.New("Failed uninstalling Registry: " + out)
+	}
+
+	err = c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), registryDeploymentID, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.New("Failed uninstalling Registry: " + err.Error())
+	}
+	s.Stop()
```
andreas-kupries, line 78: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

cli/deployments/tekton.go
```diff
@@ -61,11 +64,26 @@ func (k Tekton) Describe() string {
...
+	message := "Deleting Tekton"
+	s := spinner.New(spinner.CharSets[11], 100*time.Millisecond)
+	s.Suffix = emoji.Sprintf(" %s :zzz:", message)
+	s.Start()
+
+	err := c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), tektonDeploymentID, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.New("Failed uninstalling Quarks: " + err.Error())
+	}
+	s.Stop()
```
andreas-kupries, line 78: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

cli/deployments/traefik.go
```diff
@@ -45,8 +47,32 @@ func (k Traefik) Describe() string {
...
+	if out, err := helpers.RunProc(helmCmd, currentdir, k.Debug); err != nil {
+		return errors.New("Failed uninstalling Traefik: " + out)
+	}
+
+	err = c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), traefikDeploymentID, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.New("Failed uninstalling Traefik: " + err.Error())
+	}
+	s.Stop()
```
andreas-kupries, line 71: Move up to `s.Start()` and use `defer` to ensure execution regardless of code path.

andreas-kupries (changes requested):
A few nits in the spinner handling. `defer` is our friend.

Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03. And then I did that I saw a few early returns where without the `defer` the spinner was not properly stopped.

cli/deployments/traefik.go
andreas-kupries: Ok. I disabled the `traefik` deployment in the uninstaller and redid things. Result:

```
arrier uninstalling...
⚓ Detected kubernetes platform: k3s
🌎 ExternalIPs: [192.168.176.2]
➰ InternalIPs: [192.168.176.2]
✔️  Quarks removed
✔️  Gitea removed
✔️  Eirini removed
✔️  Registry removed
⣾ Deleting Tekton 💤  Couldn't install carrier: Failed uninstalling Quarks: namespaces "tekton" not found
```

So there is something bogus in the tekton uninstall as well.

Further, the deletion of the eirini deployment killed the workload namespace as well, and thus will kill the deployed apps too. (In my case a failed deployed stuck in image pull for the sample app - That part is independent of the uninstall itself).

Should also fix the message of `Couldn't install carrier` to `Couldn't uninstall carrier`.

cli/deployments/traefik.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/tekton.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/registry.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/quarks.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/gitea.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/eirini.go
andreas-kupries: Done with commit 84de7f0c9c6094ee525ad5b8033ae42f11225d03.

cli/deployments/traefik.go
andreas-kupries: Message fixes done with commit 15444437e5b77dc153136bac370ec5d0495202c8

rohitsakala: Yes, Namespace labelling makes sense. Will implement it.

jimmykarily:
Note to myself: When this is ready, we should add the uninstall step to the CI pipeline.

jimmykarily:
Right now we took the following decisions:
- We remove everything from apps to CRDs as if we owned the whole cluster. This may not be what we want, especially when it comes to CRDs (read this to understand why: https://github.com/helm/community/blob/f9e06c16d89ccea1bea77c01a6a96ae3b309f823/architecture/crds.md)
- We don't wait until resources are actually deleted. That means an installation can't immediately follow then uninstall step because things are still being deleted even though success is reported from the uninstall command.

Let's see if any of these decisions should be changed based on how Carrier will be used.

andreas-kupries (approved):
LGTM.

PR #54, opened 2021-01-19
Better push behavior, listing apps and deleting orgs

viovanov:


cli/paas/client.go
```diff
@@ -75,13 +79,24 @@ func (c *CarrierClient) Apps() error {
...
-	msg := c.ui.Normal().WithTable("Name", "Status", "Route")
+	msg := c.ui.Normal().WithTable("Name", "Status", "Routes")
 
 	for _, app := range apps {
-		msg = msg.WithTableRow(app.Name)
+		status, err := c.kubeClient.StatefulSetStatus(
+			c.config.EiriniWorkloadsNamespace,
+			fmt.Sprintf("cloudfoundry.org/guid=%s", app.Name))
+		if err != nil {
+			return errors.Wrap(err, "failed to get app status")
```
andreas-kupries, line 89: I wonder if we should not abort here, but simply show it in the table, in the status cell.
Note also, right now it does not tell the user for which app we got in trouble, and at this point no output was done yet, at all, leaving the user to guess.

cli/paas/client.go
```diff
@@ -75,13 +79,24 @@ func (c *CarrierClient) Apps() error {
...
+		}
+
+		routes, err := c.kubeClient.ListIngressRoutes(
+			c.config.EiriniWorkloadsNamespace,
+			app.Name)
```
andreas-kupries, line 94: Here we seem to ignore the possible error completely ?
IOW neither showing in the table, nor abort ?

cli/paas/client.go
```diff
@@ -103,6 +118,20 @@ func (c *CarrierClient) CreateOrg(org string) error {
...
+	}
+
+	c.ui.Normal().WithStringValue("name", app).Msg("Deleted app repo.")
+
+	err = c.eiriniClient.EiriniV1().LRPs(c.config.EiriniWorkloadsNamespace).Delete(context.Background(), app, metav1.DeleteOptions{})
+	if err != nil {
+		return errors.Wrap(err, "failed to delete eirini lrp")
+	}
+
+	c.ui.Normal().WithStringValue("name", app).Msg("Deleted app LRP.")
```
andreas-kupries, line 133: Should we leak these interna (Repo and LRP) to the user ?
Maybe just tell that the application was deleted ?
In the failures I can see that it is useful to know which component made trouble.
In the success less so.

andreas-kupries (changes requested):
A few questions around error handling in `apps`, and messaging in `delete`.

cli/paas/client.go
viovanov: This should be an error when we can't talk to the cluster. I think that should be an immediate stop.

viovanov: I will include the app name in the error message though

cli/paas/client.go
viovanov: That's my mistake, thanks!

cli/paas/client.go
viovanov: Was thinking the same. I'll use more generic terms.

cli/paas/client.go
andreas-kupries: Thank you, that should be enough.

andreas-kupries (approved):

cli/cmd/internal/client/apps.go
```diff
@@ -1 +1,47 @@
...
+
+var ()
+
+// CmdApps implements the carrier app command
+var CmdApps = &cobra.Command{
+	Use:   "apps",
+	Short: "Lists all apps",
+	Args:  cobra.ExactArgs(0),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		app, cleanup, err := paas.BuildApp(cmd.Flags(), nil)
```
jimmykarily, line 18: [Minor] The use of "App" in this method's name is a bit confusing (initially I though it was a user's app). We could rename it to `BuildCarrier` or something like that. Could be another PR.

andreas-kupries: Maybe `NewCarrierClient` ? Because that is the main result in the `app` variable, a `CarrierClient` structure.

jimmykarily: :+1:

PR #56, opened 2021-01-20
Extended GA flow checking more commands.

andreas-kupries:
- Commands checked: `info`, `create-org`, `orgs`, `push`.
 - Inactive preps for `apps`, `delete`.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -69,6 +69,7 @@ jobs:
           ./cli/dist/carrier-linux-amd64 orgs
       - name: Push an App
         run: |
+          sleep 8m ;# Wait for tekton to be usable
           ./cli/dist/carrier-linux-amd64 create-org workspace ;# FUTURE: Part of install.
           git config --global user.email "ci@suse.dev"
           git config --global user.name "cibot"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -55,15 +55,33 @@ jobs:
       - name: Install Carrier
         run: |
           ./cli/dist/carrier-linux-amd64 install -n
+      - name: Carrier Setup Information
+        run: |
+          ./cli/dist/carrier-linux-amd64 info
+      - name: Default Organizations
+        run: |
+          ./cli/dist/carrier-linux-amd64 orgs
+      - name: Create Organization
+        run: |
+          ./cli/dist/carrier-linux-amd64 create-org myself
+      - name: Carrier Orgs
+        run: |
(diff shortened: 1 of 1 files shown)


andreas-kupries:
Yes, the sleep looks to be necessary. Not sure how small we can make it. The value set right now is based on my local experience, where I knew it had worked.

.github/workflows/main.yml
```diff
@@ -55,15 +55,35 @@ jobs:
...
+          ./cli/dist/carrier-linux-amd64 orgs
+      - name: Create Organization
+        run: |
+          ./cli/dist/carrier-linux-amd64 create-org myself
+      - name: Carrier Orgs
+        run: |
+          ./cli/dist/carrier-linux-amd64 orgs
+      - name: Push an App
+        run: |
+          sleep 8m ;# Wait for tekton to be usable
```
jimmykarily, line 72: Shouldn't we better add waiting loops in the install command to make sure Tekton is usable? We can do it on another PR if you prefer.

andreas-kupries: I agree that having some kind of waiting in the installer would be best. I have however no idea wait exactly we have to wait for in tekton, so for this one I simply did the hack to get around whatever this is. A new PR changing tekton (and removing the sleep here) would indeed be my preference.

jimmykarily (approved):

.github/workflows/main.yml
jimmykarily: ok let's do it like that

.github/workflows/main.yml
```diff
@@ -55,15 +55,35 @@ jobs:
...
+          ./cli/dist/carrier-linux-amd64 create-org workspace ;# FUTURE: Part of install.
+          git config --global user.email "ci@suse.dev"
```
jimmykarily, line 74: I guess this is a fix for this: https://github.com/SUSE/carrier/issues/12
I'm currently fixing that issue. If this is merged first I will remove these 2 lines so that we check if my fix works.
If this is merged last, we should remove these lines on this PR.

jimmykarily: https://github.com/SUSE/carrier/pull/58

andreas-kupries: This is now merged.

jimmykarily: So I can delete these lines I guess.

jimmykarily: https://github.com/SUSE/carrier/pull/60

PR #57, opened 2021-01-20
Print pod events when pod waiting fails

### LINKED ISSUE #53: When installer times out waiting for a Pod, print the relevant events
E.g. when dockerhub is rate limiting us, carrier installer will simply timeout but there is not way for the user to tell because the relevant events are never printed out.

The user could go and check manually on what's wrong but it would be useful to see the relevant events immediately.

It's also useful for CI (when things timeout) so we don't need to find a way to ssh to the CI container and check.

## Possible fix:

We need to run the equivalent of this command:

```
kubectl get event --namespace carrier-registry --field-selector involvedObject.name=$(kubectl get pods -o=jsonpath='{.items[0].metadata.name}'  --selector=app.kubernetes.io/name=container-registry -n carrier-registry)
```

somewhere here: https://github.com/SUSE/carrier/blob/main/cli/kubernetes/cluster.go#L213

jimmykarily:
so the user has an idea of what went wrong (e.g. Dockerhub rate limiting)

You can test it by setting this to `1`: https://github.com/SUSE/carrier/blob/main/cli/cmd/install.go#L14

Fixes #53

### DIFF
--- a/cli/kubernetes/cluster.go
+++ b/cli/kubernetes/cluster.go
@@ -114,7 +114,7 @@ func (c *Cluster) detectPlatform() {
 	}
 }
 
-// return a condition function that indicates whether the given pod is
+// IsPodRunning returns a condition function that indicates whether the given pod is
 // currently running
 func (c *Cluster) IsPodRunning(podName, namespace string) wait.ConditionFunc {
 	return func() (bool, error) {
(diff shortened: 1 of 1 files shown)


cli/kubernetes/cluster.go
```diff
@@ -210,12 +210,56 @@ func (c *Cluster) WaitForPodBySelectorRunning(namespace, selector string, timeou
...
+				return errors.Wrap(err, err2.Error())
+			} else {
+				return errors.New(fmt.Sprintf("Failed waiting for %s: %s\nPod Events: \n%s", pod.Name, err.Error(), events))
+			}
 		}
 	}
 	return nil
 }
 
+// GetPodEventsWithSelector tried to find a pod using the provided selector and
```
andreas-kupries, line 224: `tried` -> `tries`

cli/kubernetes/cluster.go
```diff
@@ -210,12 +210,56 @@ func (c *Cluster) WaitForPodBySelectorRunning(namespace, selector string, timeou
...
+// namespace. If found it returns the events on that Pod. If not found it returns
+// and error.
```
andreas-kupries, line 226: `an error`

cli/kubernetes/cluster.go
```diff
@@ -210,12 +210,56 @@ func (c *Cluster) WaitForPodBySelectorRunning(namespace, selector string, timeou
...
+// And equivalent kubectl command would look like this
```
andreas-kupries, line 227: `An equivalent ...`

andreas-kupries (changes requested):
Looks ok beyond the typo nits.

Question is, how can we test this, i.e. force a failure, to see that we get sensible information ?

cli/kubernetes/cluster.go
jimmykarily: fixed.

cli/kubernetes/cluster.go
jimmykarily: So many typos on one method description! :D Fixed.

jimmykarily:
> Looks ok beyond the typo nits.
> 
> Question is, how can we test this, i.e. force a failure, to see that we get sensible information ?

Check the description of the PR. You can set the DefaultTimeout to 1 second which will make it fails almost certainly on some Pod. You should then be able to see the events (e.g. Downloading Image or something like that).

andreas-kupries (approved):

PR #58, opened 2021-01-20
Set the git user when pushing to gitea

### LINKED ISSUE #12: carrier push with no user.name and user.email configured in git
When running `carrier push` but you don't have a user.email and user.name globally set for git you get an error:

```
From http://gitea.172.19.0.2.nip.io/dev/dizzy
 * [new branch]      main       -> carrier/main
Author identity unknown

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: no email was given and auto-detection is disabled
error: src refspec master does not match any
```

Carrier should assume any specific git configuration on the user's machine.

NOTE: If cf cli is supposed to do the pushing from here on, then we can close this issue.

jimmykarily:
[Fixes #12]

### DIFF
--- a/cli/paas/client.go
+++ b/cli/paas/client.go
@@ -314,6 +314,8 @@ func (c *CarrierClient) gitPush(name, tmpDir string) error {
 	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf(`
 cd "%s" 
 git init
+git config user.name "Carrier"
+git config user.email ci@carrier
 git remote add carrier "%s"
 git fetch --all
 git reset --soft carrier/main


cli/paas/client.go
```diff
@@ -314,6 +314,8 @@ func (c *CarrierClient) gitPush(name, tmpDir string) error {
 	cmd := exec.Command("/bin/sh", "-c", fmt.Sprintf(`
 cd "%s" 
 git init
+git config user.name "Carrier"
+git config user.email ci@carrier
```
jimmykarily, line 318: Since there is no validation for the domain I though a non-domain is fine. We don't have a relevant domain and it shouldn't matter much anyway.

andreas-kupries: In the future we may wish to insert information about the user doing the push, maybe the hostname as well, if we need a domain part. That info would have to come from the OS in some way, or we check the git config in some way (is there a go package for such git ops ?)

andreas-kupries (approved):
:+1:

PR #59, opened 2021-01-20
Fixed various CmdXXX mismatches in cli command initializations.

andreas-kupries:


jimmykarily:
I commented that on some other PR too, these methods only pass kubeconfig flags to the various subcommands. This looks very much like what Persistent Flags are supposed to be in cobra/viper. PersistentFlags are flags that are shared among all subcommands of a command and thus attached to the parent command. Since all subcommands need the kubeconfig information, we should simply create a PersistentFlag for it on the top level command and skip it in all the subcommands.

andreas-kupries:
I now remember that comment. Should make a ticket for it so that we will not forget.

jimmykarily (approved):

PR #64, opened 2021-01-21
Create default organization, `workspace`

andreas-kupries:
Modified the `install` method to run a `create-org workspace` at the end.
This enables the simple sequence of `install ; push app` (*), without the need for more work on the side of the user.

As a side effect of getting that done the code of the `install` method was moved around and refactored to fit into the setup of all the other cli methods implemented separately from it. This part is complete as well. By placing work on the spinners out of scope for this PR. The spinners will be handled in a separate PR.

----
(Ad *) Except for the current need to wait several minutes after `install`, for `tekton` to get its act together and be actually usable. This requires work in the `tekton` deployment to wait properly on whatever.

----
## Full notes

[feat] Added creation of default organization `workspace` to install.

[fix]  Reworked the `install` command to fit into the new cli setup of the other commands.

  - Shuffled files and package for `install` around to match new hierarchy.
  - Refactored `install` backend and moved into the `CarrierClient`.
  - Changed cluster setup to use a variant of `BuildApp` which works without `gitea` clients. These are not available at install time yet.
  - Switched messaging to use `ui` package.
      - Extended various interfaces, function signatures to have `ui` available.
  - Simplified various `Wrap/Sprintf` combos to `Wrapf`.
  - Extended `ui`
      - Added Note and sucess messages
      - Added emoji support for messages in general
      - Added emoji prefix as supplemental message type indicator, for when color is disabled (any kind of redirection).
      - Tweaked spinners (Central config, different char set)
      	Note: Will likely go away, or switchable.

- Dropped `workspace` org creation from GA flow, this is now handled by `install`.
- Regenerated the fakes.


cli/cmd/internal/client/install.go
```diff
@@ -0,0 +1,87 @@
+package client
```
jimmykarily, line 1: I though `internal` meant "everything that comes after installing carrier" thus things you do on Carrier itself after it's installed. With this move now I don't understand the role of the `internal` directory.

andreas-kupries: @viovanov Can you explain the `internal` directory to me ?
I am willing to move the `install` sources out of it if my move into it was wrong.

cli/deployments/eirini.go
```diff
@@ -60,7 +61,7 @@ func (k Eirini) Delete(c kubernetes.Cluster) error {
 	return c.Kubectl.CoreV1().Namespaces().Delete(context.Background(), eiriniDeploymentID, metav1.DeleteOptions{})
 }
 
-func (k Eirini) apply(c kubernetes.Cluster, options kubernetes.InstallationOptions) error {
+func (k Eirini) apply(c kubernetes.Cluster, ui *ui.UI, options kubernetes.InstallationOptions) error {
```
jimmykarily, line 64: Isn't it better if we just attach the UI to the Deployment object? This way it's easier to have a global UI without the need to pass it around in all methods that need it.

andreas-kupries: I tried to make it a field in the `Installer` structure IIRC.
That resulted in a circular import.
Not sure if that will happen for `Deployments` as well.

cli/kubernetes/installer.go
```diff
@@ -83,17 +81,26 @@ func (i *Installer) PopulateNeededOptions(reader OptionsReader) error {
 
 // ShowNeededOptions prints the options and their values to stdout, to
 // inform the user of the detected and chosen configuration
-func (i *Installer) ShowNeededOptions() {
-	fmt.Println("Configuration...")
+func (i *Installer) ShowNeededOptions(ui *ui.UI) {
```
jimmykarily, line 84: same things as with the Deployments. The UI could be an attribute of the Installer so it's available in all methods.

andreas-kupries: See above, rejected due to an import circle. Will re-check if still circular.

andreas-kupries: Ok. Falsified. The circular import is gone.
Will add an UI field to installer and reduce usage in function signatures.

cli/deployments/eirini.go
andreas-kupries: Hm. `Deployment` is an interface. I guess I can add `ui` fields to the various implementations, and add a `UI` interface method with which to set that information.

cli/paas/wire_gen.go
```diff
@@ -51,3 +52,27 @@ func BuildApp(flags *pflag.FlagSet, configOverrides func(*config.Config)) (*Carr
...
+	uiUI := ui.NewUI()
+	restConfig, err := config2.KubeConfig()
+	if err != nil {
+		return nil, nil, err
+	}
+	cluster, err := kubernetes.NewClusterFromClient(restConfig)
+	if err != nil {
+		return nil, nil, err
+	}
+	carrierClient := &CarrierClient{
```
jimmykarily, line 71: I have a feeling we are doing it wrong. This Injector (#BuildInstallApp function), returns the same struct as BuildApp Injector. They both return a CarrierClient but in once case that struct has more fields filled in. From now on, when you see a CarrierClient instance in the code, you can't know which "version" it is.
I think wire's (and more generally, the dependency injection's) thinking is that you have an object that depends on others and all wire is doing is helping you avoid boilerplate code for just initializing the object. But the question here is:
Does a CarrierClient depend on a gitea.Client or not?
If yes then why is there a method creating a crippled version of CarrierClient (this one)?

Personally I think wire was probably created for way more complicated initializers than the ones we ever going to need but we were are to stick with wire, we should at least create different object types out of each initializer.

For example in this case, we could have BuildInstallApp returning actually an "InstallApp" or simply an "Installer". After all the Installer struct was created as the Object that knows how to install Carrier. Now it seems that knowledge is shared between the CarrierClient and the Installer object (one calls the other eventually).

andreas-kupries: So, maybe an `InstallClient` which is the `CarrierClient` with the irrelevant fields gone ?

andreas-kupries: AFAICT the `CarrierClient` is the object which encapsulates all the backend actions you can perform on a carrier installation. ... In some sense `Uninstall` would be a method of that, while `Install` is less so, as it is about bringing up an installation. so from that perspective splitting of an `InstallClient` would also make sense.

Part of the problem right now is I mixed both the re-org for install with the creation of the default org in a single PR.

cli/deployments/eirini.go
jimmykarily: Probably a `SetUI` method

cli/paas/wire_gen.go
jimmykarily: I think it should be like:

- Install cmd initializes an Installer (or `InstallCliet` if you prefer, but there should be no need for a separate `Installer` object)
- Install cmd calls Install on the above object
- Install cmd initializes a CarrierClient
- Install cmd calls CreatOrg on the above object

andreas-kupries: I see, you are looking to merge the currently separate `Installer` structure of yours with a variant of the `CarrierClient` modified for install usage. Hm ... Should be doable. ... Heh, the addition of a `ui` field to `Installer` sort of goes in that direction already ... Add the `kubeClient` and we are there, I think. Maybe.

cli/deployments/eirini.go
andreas-kupries: Yes, that. To come soon, just finished the `Installer` change ...

cli/kubernetes/installer.go
andreas-kupries: Done with commit a3f7aded87a42a0a1ac2074c66bdac1e5414f43a.
Looking into `Deployment` now. That also gives me the time to have the back of my mind mull more on the merging of `Installer` and `CarrierClient` variant.

cli/deployments/eirini.go
andreas-kupries: Done with commit 5135955e4a0db387de755ea1bf73a4f8c85658b2

cli/paas/wire_gen.go
andreas-kupries: Split of `CarrierClient` done with commit 50fd1cb6c863ec90f13d9059fbe6e27aeaa4d6d6.

No work has been done on merging `kubernetes.Installer` into the `InstallClient` yet.
While it would be the logical next step I am not fully convinced that it cannot be defered at this point.

andreas-kupries: Ok. Continuing and looking further into the merge, flip/flopping. Fill the rest of the day.

Note: The `kubernetes.Installer` structure has some responsibilities which I believe are better left with it. Because they are things needed even before we have an `InstallClient`. I am talking about all the option handling. The only part of that we can move would be `ShowNeededOptions` ...

So ... Rename this to `kubernetes.DeploymentSet`, keep most of the option handling with it, move the `Show..` and `Install` methods over to the client. The `Show...` can become an internal function, even. That makes the `DS` mostly passive.

No UI facilities in the DeploymentSet. We do keep the global variable around, and hand that to the `Install` method as usual. Some renames in that area also for clarity.

andreas-kupries: Done with commit 30b4fb7285551c378763a243cc36e963d03eceff.

:warning: Undone commit 5135955e4a0db387de755ea1bf73a4f8c85658b2 attaching an `ui` field to Deployment implementations, and `SetUI` interface.

The `ui` reference is in the same category as the `cluster` reference, not inherent information of the deployment, but an externally supplied thing for use by the current operation. We could go and aggregate these two references into a single structure. That would be the `InstallClient`, actually, pretty much. Doing that would also force us to move the core deployment related definitions out of `kubernetes` into `paas`, to avoid circular imports/dependencies. Not to be done today, if at all.

jimmykarily (approved):

cli/cmd/internal/client/install.go
viovanov: `internal` is special in go. You can have public things without them being exported in an external package

PR #67, opened 2021-01-22
Add ability to target an org

viovanov:


cli/cmd/internal/client/target.go
```diff
@@ -1 +1,52 @@
...
+		if err != nil {
+			return errors.Wrap(err, "error initializing cli")
+		}
+
+		org := ""
+		if len(args) > 0 {
+			org = args[0]
+		}
+
+		err = client.Target(org)
```
andreas-kupries, line 34: Neither nor in the called function is `org` validated to exist.

UI nit: In my role as the user deploying an app I am unable to tell if the `target` command has an actual effect, as `push` the app does not tell which org the app is pushed into.

Only as a developer can I see that the code of the `Push` backend function was updated to take the org into account.

I am ok with fixing both these things after the PR is merged.

andreas-kupries (approved):
Mostly ok. Saw to UI experience nits I am willing to have fixed after the PR/functionality is merged. These are

- Validation of org existence when targeting an org, and
- Telling the user which org an app is pushed into (during push) (*).
- (*) New: App's org should be shown in app listing also IMHO.

cli/cmd/internal/client/target.go
andreas-kupries: Will be done in #68

PR #77, opened 2021-01-25
[feat] Install non-interactively by default

### LINKED ISSUE #72: `install --non-interactive` should be the default
The installation should be as quick and as non-interactive as possible so we should make `non-interactive` the default and add an option for `--interactive`.

Document the alternative of setting system_domain through a flag.

thardeck:
and introduce a new command flag `--interactive` to disable the non-ineractive mode again.

Fixes  #72


andreas-kupries (approved):
LGTM, short and sweet.

PR #78, opened 2021-01-25
[feat] Improve staging logs.

### LINKED ISSUE #75: Improve application staging logs
Right now we print pod names and other information but it's not obvious to the user where the logs come from and what they refer to (e.g. "pushing code", "staging" ?)

We should limit the output to only the useful stuff.

andreas-kupries:
Fixes #75.

Modified the staging logs to show an `Origin` (staging or app) for the log entries coming out of the tailers.
Dropped the full pod name in favor of it. Kept the container name.

Note: There are still log lines not captured by this.
These log lines have a green-colored leading `+` character.
At the moment it is unclear where these are coming from.

@viovanov do you have an idea ?

Example gathered by disabling tailer and progress indicator:
```
+ staging-pipeline-run-sample2-8tzkq-clone-cfv9t-pod-ndszb › step-git-source-source-repo-pghwv
+ staging-pipeline-run-sample2-8tzkq-clone-cfv9t-pod-ndszb › step-stage
+ staging-pipeline-run-sample2-8tzkq-stage-dt4jf-pod-hmxc5 › step-create
+ staging-pipeline-run-sample2-8tzkq-stage-dt4jf-pod-hmxc5 › step-create-dir-image-g6ckt
+ staging-pipeline-run-sample2-8tzkq-stage-dt4jf-pod-hmxc5 › step-image-digest-exporter-qbhwm
+ staging-pipeline-run-sample2-8tzkq-stage-dt4jf-pod-hmxc5 › step-prepare
+ staging-pipeline-run-sample2-8tzkq-run-wvqrr-pod-7lz55 › step-run
```

Just after the regular user message of 
```
🕞  Creating cloudfoundry.org/guid=sample2 in eirini-workloads
```

I wonder if that is some output from a `kubectl` command we are running here.

Ok, these lines are the stderr of something ... Looking at the uses of `helper.RunProc` (which allows stderr to be seen in the terminal) most of these are in the deployments, so do not apply to `Push`. The lone other is used in the `Kubectl` helper.

... Nope. Dead End. Switching to `RunProcNoErr` in `Kubectl` does not suppress these lines. Need a new hypothesis for the origin.

Completely disabling `logs()` prevents these lines.

... Found it

```
	if t.Options.Namespace {
		fmt.Fprintf(os.Stderr, "%s %s %s › %s\n", g("+"), p(t.Namespace), p(t.PodName), c(t.ContainerName))
	} else {
		fmt.Fprintf(os.Stderr, "%s %s › %s\n", g("+"), p(t.PodName), c(t.ContainerName))
	}
```

It is the part of the tailer telling us when it finds a new pod to track.
Not sure if we need that, actually.

### DIFF
--- a/cli/kubernetes/tailer/setup.go
+++ b/cli/kubernetes/tailer/setup.go
@@ -8,6 +8,7 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/suse/carrier/cli/kubernetes"
+	"github.com/suse/carrier/cli/paas/ui"
 	"k8s.io/apimachinery/pkg/labels"
 )
 
@@ -36,7 +37,7 @@ type Config struct {
 //   - For log entries `Exclude` is applied before `Include`.
 
 // Run starts the log watching
-func Run(ctx context.Context, config *Config, cluster *kubernetes.Cluster) error {
+func Run(ui *ui.UI, ctx context.Context, config *Config, cluster *kubernetes.Cluster) error {
 	var namespace string
 
 	if config.AllNamespaces {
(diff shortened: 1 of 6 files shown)


rohitsakala (approved):
Looks good to me.

PR #79, opened 2021-01-25
Generate statik.go during builds and test runs. Remove statik.go from the repository.

thardeck:
to prevent automatically generated binary changes from github history.


jimmykarily:
We should first try to stub the statik version we install to a specific version. I think that is the reason we see trafik.go changing although we didn't change any embedded files.

thardeck:
> We should first try to stub the statik version we install to a specific version. I think that is the reason we see trafik.go changing although we didn't change any embedded files.

It seems statik is already a stable version https://github.com/SUSE/carrier/blob/0ba379c6c4c3eb31c8126fd69f73a65931fa9d3e/cli/go.mod#L29, downgrading does not help.

But using the `-m` option might help. See #88

PR #82, opened 2021-01-26
Add release pipeline in github actions 

### LINKED ISSUE #65: [CI] Release when main branch is tagged
We need a GitHub workflow that creates a GitHub release whenever we tag the `main` branch.

An action like this one could be used: https://github.com/fnkr/github-action-ghr
(example usage: https://github.com/jimmykarily/fuzzygui/blob/master/.github/workflows/release.yml#L29)

There should be a new `make` target that creates a smaller binary for the release. With `-s -w` ldflags (as described here: https://stackoverflow.com/a/37468877), your binary size drops from 60M to 48M.
The release workflow should create binaries for all architectures.

rohitsakala:
This pipeline is triggered only when a tag is pushed on the main branch.

Fixes #65

### DIFF
--- a/cli/Makefile
+++ b/cli/Makefile
@@ -2,20 +2,23 @@ build: tools embed_files lint build-amd64
 
 build-all: tools embed_files lint build-amd64 build-arm64 build-arm32 build-windows build-darwin
 
+build-all-small:
+	@$(MAKE) LDFLAGS+="-s -w" build-all
+
 build-arm32: lint
-	GOARCH="arm" GOOS="linux" go build -o dist/carrier-linux-arm32
+	GOARCH="arm" GOOS="linux" go build -ldflags '$(LDFLAGS)' -o dist/carrier-linux-arm32
 
 build-arm64: lint
-	GOARCH="arm64" GOOS="linux" go build -o dist/carrier-linux-arm64
+	GOARCH="arm64" GOOS="linux" go build -ldflags '$(LDFLAGS)' -o dist/carrier-linux-arm64
 
 build-amd64: lint
-	GOARCH="amd64" GOOS="linux" go build -race -o dist/carrier-linux-amd64
+	GOARCH="amd64" GOOS="linux" go build -race -ldflags '$(LDFLAGS)' -o dist/carrier-linux-amd64
 
 build-windows: lint
-	GOARCH="amd64" GOOS="windows" go build -o dist/carrier-windows-amd64
+	GOARCH="amd64" GOOS="windows" go build -ldflags '$(LDFLAGS)' -o dist/carrier-windows-amd64
 
 build-darwin: lint
-	GOARCH="amd64" GOOS="darwin" go build -o dist/carrier-darwin-amd64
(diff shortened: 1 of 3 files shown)


cli/Makefile
```diff
@@ -2,20 +2,23 @@ build: tools embed_files lint build-amd64
...
 build-arm32: lint
-	GOARCH="arm" GOOS="linux" go build -o dist/carrier-linux-arm32
+	GOARCH="arm" GOOS="linux" go build -ldflags '$(LDFLAGS)' -o dist/carrier-linux-arm32
 
 build-arm64: lint
-	GOARCH="arm64" GOOS="linux" go build -o dist/carrier-linux-arm64
+	GOARCH="arm64" GOOS="linux" go build -ldflags '$(LDFLAGS)' -o dist/carrier-linux-arm64
 
 build-amd64: lint
-	GOARCH="amd64" GOOS="linux" go build -race -o dist/carrier-linux-amd64
```
thardeck, line 12: Was `-race` removed on purpose because it was added explicitly: 6467cd701cb99421f9c6f66356386496b8bdecca

rohitsakala: Yes, I removed by mistake. Corrected it.

thardeck (approved):
In principle looks good to me but the compression has quite a notable impact on the start time. It is still ok but might have an impact when our project gets bigger. What do the others think?
`$ ls -lah carrier-small*`
```
-rwxr-xr-x 1 thardeck users 48M Jan 27 10:41 carrier-small
-rwxr-xr-x 1 thardeck users 12M Jan 27 10:50 carrier-small-compressed
```

`$ time ./carrier-small help > /dev/null`
```
real    0m0.064s
user    0m0.026s
sys     0m0.030s
```

`$ time ./carrier-small-compressed help > /dev/null`
```
real    0m0.995s
user    0m0.960s
sys     0m0.024s
```

rohitsakala:
> In principle looks good to me but the compression has quite a notable impact on the start time. It is still ok but might have an impact when our project gets bigger. What do the others think?
> `$ ls -lah carrier-small*`
> 
> ```
> -rwxr-xr-x 1 thardeck users 48M Jan 27 10:41 carrier-small
> -rwxr-xr-x 1 thardeck users 12M Jan 27 10:50 carrier-small-compressed
> ```
> 
> `$ time ./carrier-small help > /dev/null`
> 
> ```
> real    0m0.064s
> user    0m0.026s
> sys     0m0.030s
> ```
> 
> `$ time ./carrier-small-compressed help > /dev/null`
> 
> ```
> real    0m0.995s
> user    0m0.960s
> sys     0m0.024s
> ```

@andreas-kupries @jimmykarily do you have any comments ?

thardeck:
Adding a simple checksum file with sha256 hashes might also be nice and should not be much effort.

andreas-kupries:
With respect to the startup time ... ... Most of our commands will/have to talk to the cluster. I suspect that the time needed for that will dominate the overall response time. On the other hand, this start time is the latency from the user hitting Enter to the first message from the cli, and a near-second is noticeable. ... On the third hand, disk is large and cheap these times.

Overall I would go __without compression__.

andreas-kupries (approved):

viovanov:
+1 on no compression

PR #86, opened 2021-01-27
[Fix] Ensure good organization for several commands, and after install

### LINKED ISSUE #76: BUE (bad user experience) After killing a cluster (not uninstall) target org information can screw with a new install
Basically I

  1. Had a cluster running, and targeted org `foo`.
  2. Tore the cluster down, created a new and installed carrier.
  3. At that point the `foo` org in carrier config was out of date, the only org was the default `workspace`.
  4. Then pushed some apps with an older binary which did not use the config, so pushed to `workspace` fine.
  5. After that, trying a new binary with tweaks to the tailer, it was unable to see anything because that then used the saved config and tried to access the non-existent org `foo`. 

After reading and understanding the low-level error message of `Unprocessable Entity: {"message":"org does not exist [id: 0, name: foo]"` things became more clear. `apps` only said `error listing apps: failed to list apps: 404 Not Found` which was not as helpful.

IMHO this can be fixed by

  - [x] When creating the default `workspace` org, target it immediately, properly. This will squash any leftover targeting config from whatever previous cluster.
  - [x] Rework `push`, `apps` to print a nicer error message when the targeted org they use does not exist.
  - [x] Rework `apps` to be clear about the org it is listing the apps for.

andreas-kupries:
Fixes #76.

Commands `push`, `apps`, and `target` now properly validate the organization, either argument, or from the config.
Command `install` now targets the newly created default workspace, squashing any bogus org information in the config left over from a previous installation.

### DIFF
--- a/cli/cmd/internal/client/install.go
+++ b/cli/cmd/internal/client/install.go
@@ -18,6 +18,10 @@ var NeededOptions = kubernetes.InstallationOptions{
 	},
 }
 
+const (
+	DefaultOrganization = "workspace"
+)
+
 var CmdInstall = &cobra.Command{
 	Use:           "install",
 	Short:         "install Carrier in your configured kubernetes cluster",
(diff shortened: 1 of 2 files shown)


thardeck (approved):
lgtm

rohitsakala (approved):

PR #87, opened 2021-01-27
Fix system domain

jimmykarily:
Fixes: https://github.com/SUSE/carrier/issues/24


cli/kubernetes/options.go
```diff
@@ -202,3 +213,21 @@ func (opts InstallationOptions) ForDeployment(deploymentID string) InstallationO
 
 	return result
 }
+
+// Populate will try to give values to the needed options
+// using the given OptionsReader. If none is given, the default is the
+// InteractiveOptionsReader which will ask in the terminal.
+// This method only populates what is possible and leaves the rest empty.
+// TODO: Implement another method to validate that all options have been set.
+func (opts *InstallationOptions) Populate(reader OptionsReader) (*InstallationOptions, error) {
```
jimmykarily, line 222: I tried to make this method to mutate opts itself but failed. I opted for returning a new pointer. This moves the responsibility of updating the old pointer to the caller, which is something one could forget. Maybe I'm doing something wrong, there must be a way to mutate the receiver (?).

andreas-kupries: Interesting. In the various option readers the mutated structure is a pointer argument, instead of pointer receiver. That certainly works. Maybe somewhere in your code the pointer receiver is a copy of the original structure, thus mutating it does not affect the original ? Because, on the other hand in `ui`, the Message mutators (`WithXXXValue`, etc.) all work.

andreas-kupries: Seeing the code, you are creating a new slice by going through the input slice. How did you try to mutate `opts` itself ?
Something like `opts = newOpts` ? Go has a builtin `copy` function: `copy(opts,newOpts)` ... https://yourbasic.org/golang/copy-explained/

cli/paas/install_client.go
```diff
@@ -21,88 +38,92 @@ type InstallClient struct {
...
 
 	// TODO (post MVP): Run a validation phase which perform
 	// additional checks on the values. For example range limits,
 	// proper syntax of the string, etc. do it as pghase, and late
 	// to report all problems at once, instead of early and
 	// piecemal.
 
-	err = c.deploySet(deployments)
+	deployment := deploymentSet["traefik"]
+	err = deployment.Deploy(c.kubeClient, c.ui, options.ForDeployment(deployment.ID()))
```
andreas-kupries, line 76: IMHO the whole `options.ForDeployment(deployment.ID())` can be moved into the `Deploy` function.
Plus removing the temp variable the section could be reduced to:
```
err = deploymentSet["traefik"].Deploy(c.kubeClient, c.ui, options)
```
This should simplify the coming loop over the remainder also.

cli/paas/install_client.go
```diff
@@ -21,88 +38,92 @@ type InstallClient struct {
...
+	if domain.Value.(string) == "" {
+		return errors.New("You didn't provide a system_domain and we were unable to setup a nip.io domain (couldn't find and ExternalIP)")
+	}
+
+	c.ui.Success().Msg("Created system_domain: " + domain.Value.(string))
+
+	for _, deploymentName := range []string{"quarks", "gitea", "eirini", "registry", "tekton"} {
+		deployment := deploymentSet[deploymentName]
+		options := options.ForDeployment(deployment.ID())
+		err := deployment.Deploy(c.kubeClient, c.ui, options)
```
andreas-kupries, line 100: See above, the entire thing can be moved into `Deploy`.

andreas-kupries: I believe this would be easier with a simple `[]deployment` instead of a map. Keep `traefik` separate, and then you can simply iterate directly over deployments here, without names and temp vars.

See the other simplification as well, noted above.

cli/paas/install_client.go
```diff
@@ -21,88 +38,92 @@ type InstallClient struct {
...
+		if err != nil {
+			return err
+		}
+	}
 
 	c.ui.Success().Msg("Carrier installed.")
+	c.ui.Success().Msg("Your system domain is: " + domain.Value.(string))
```
andreas-kupries, line 107: More in line with the existing use of ui in commands:
```
c.ui.Success().WithStringValue("System domain", domain.Value.(string)).Msg("Carrier installed.")
```

cli/paas/install_client.go
```diff
@@ -21,88 +38,92 @@ type InstallClient struct {
...
 
-	size := len(deploymentset.Deployments)
-	for index := range deploymentset.Deployments {
-		d := deploymentset.Deployments[size-index-1]
-		if d.ID() != eiriniDeploymentID {
-			err := d.Delete(c.kubeClient, c.ui)
-			if err != nil {
-				return err
-			}
+	for _, deploymentName := range []string{"tekton", "registry", "gitea", "quarks", "traefik"} {
```
andreas-kupries, line 122: Hm. To fit my idea of using `[]deployments` instead of the map ... All the deployments are separate variables, and then two slices are made for the install and uninstall sequences outside of the special cases, i.e. treafik in install, and eirini in uninstall.

cli/paas/install_client.go
```diff
@@ -129,14 +150,46 @@ func (c *InstallClient) showInstallConfiguration(ds *kubernetes.DeploymentSet) {
...
+func (c *InstallClient) fillInMissingSystemDomain(domain *kubernetes.InstallationOption) error {
+	if domain.Value.(string) == "" {
+		ip := ""
+		if !c.kubeClient.GetPlatform().HasLoadBalancer() {
+			ips := c.kubeClient.GetPlatform().ExternalIPs()
+			if len(ips) > 0 {
+				ip = ips[0]
+			}
+		} else {
+			c.ui.ProgressNote().Msg("Waiting for LoadBalancer IP on traefik service.")
```
andreas-kupries, line 162: Why not a
```
	s := ui.Progress("Waiting for LoadBalancer IP on traefik service.")
	defer s.Stop()
```
to also get the nice dot progress while we are waiting here ?

andreas-kupries (changes requested):
I have some comments which might require changes.

cli/kubernetes/options.go
jimmykarily: Yes I tried `opts = &newOpts`. I will try again when we get all other issues fixed (on this or other PR).

cli/paas/install_client.go
jimmykarily: true. The idea was that the caller would give each Deployment only the relevant options (so they wouldn't even have access to things they shouldn't). Imagine someone needing an option in a Deployment, and simply because the whole options object is available, they would bypass the `ForDeployment` method and read the option of another deployment directly.
In that case, we wouldn't be able to tell which option is needed by which deployment simply by looking at the options object.

Of course, with only one option currently being used maybe this measure is meaningless, but it could be useful in the future. What do you think?

cli/paas/install_client.go
jimmykarily: In the `Uninstall` method below I need to handle eirini first, like I do here with traefik. Since I need to have random access to deployments of the same list, a map is more suitable.

If I switch to a slice, I would need to keep eirini separate in the `Uninstall` method and then iterate over a different slice for the rest of them (different from the one used in `Install` I mean). This could lead to Deployments added/removed from the `Install` slice but not from the `Uninstall` and vise versa.

cli/paas/install_client.go
jimmykarily: I replied above before I saw this comment, but my comment is still valid I think (regarding the use of the same list of deployments in both cases).

cli/paas/install_client.go
jimmykarily: That's the one I was looking for! :D Thanks, changing.

cli/paas/install_client.go
andreas-kupries: Ok. I am not to fussed actually. It simply looked not as clean as I thought it could be. Leaving the argument as-is is fine. and the argument makes sense, so plus one on that too.

I would still go with slices and direct iteration over the map.
If we keep the map, I would at least drop the temp variable.

cli/paas/install_client.go
andreas-kupries: Yes to what going to slice entails. Danger of de-syncing the slices. True. (Although I believe it low, when the slices are defined side by side). We can keep the map. I would still get rid of the temp variables.

andreas-kupries: The argument re slices is not fully valid, I just realized. Because right now, even with the map you are handling slices. not of the deployments, but of the names, and these can desync as well. Especially as they are quite far apart, in the install and uninstall functions. With the deployments the plan was to have them together at the top.

Regardless, let us keep the map for now, and discuss later.

jimmykarily: Yes I realized that too and went with your suggestion. Will push a commit shortly.

andreas-kupries: Heh. Flip-flop :wink:

cli/paas/install_client.go
jimmykarily: Done

cli/paas/install_client.go
```diff
@@ -21,88 +29,99 @@ type InstallClient struct {
...
+	if err != nil {
+		return errors.Wrap(err, "Couldn't install carrier")
+	}
+	if domain.Value.(string) == "" {
+		return errors.New("You didn't provide a system_domain and we were unable to setup a nip.io domain (couldn't find and ExternalIP)")
+	}
+
+	c.ui.Success().Msg("Created system_domain: " + domain.Value.(string))
+
+	for _, deployment := range []kubernetes.Deployment{
```
andreas-kupries, line 88: The only thing I would do now, would be moving these slices, and the single deployments to the top of the file as variables.
That is most definitely a nit we can defer for a while.

andreas-kupries (approved):
Just a nit now (where to place some definitions). Not blocking merge.

PR #88, opened 2021-01-27
 Prevent statik.go being changed all the time

thardeck:
by ignoring the modification times of files.

Supersedes #79


jimmykarily (approved):

andreas-kupries (changes requested):
@thardeck There was still something going on.

  - I switched to the branch,
  - confirmed that `statik.go` was shown as unmodified, then
  - ran `make build`.
  - After that git claimed that `statik.go` was changed. Unfortunately it was not obvious where the difference was.
  - I then committed the changed file locally, and then
  - ran `make build` again.
  - Now git showed no difference anymore.

I suspect that the `statik.go` in the branch is not as up-to-date as it should have been, thus why it showed a change after the first `build`. With the update from that the file became current and no further changes happened.

I believe that with this change we have to be more careful about regenerating `statik.go`. I.OW about forgetting to do so. I propose that we add a `@$(MAKE) embed_files` to the `update_...` targets. I.e. automatically regenerate `statik.go` whenever we update any of the pieces going into it.

We can (should actually, I believe) keep the `embed_files` dependencies on `test` and `build` targets. That would catch if we forgot updating `statik.go` due to edits done outside of the `update_...` targets.

rohitsakala:
I tested this PR locally. I did two things

1. Checked out main branch. Ran `make build`. The statik.go file has changed.
2. Checked out this PR. Ran `make build`. The statik.go file didn't change.

andreas-kupries:
Weird. Under the suspicion that this then is something to do with my environment I retract my request for changes.

PR #91, opened 2021-01-28
[fix] Changed carrier to use the `--kubeconfig`/`-c` cli options.

### LINKED ISSUE #89: cli ignores option --kubeconfig/-c
Testing PR #43 for #40 I noticed that `carrier` ignores the `--kubeconfig` option to provide the path to the kubeconfig to use.
I approved the PR even so, because I further determined that this issue was present on main as well, and thus did not came from #43.

`carrier` looks to solely use the environment variable `KUBECONFIG` to determine where to find the cluster.

andreas-kupries:
Fixes #89.

:fireworks: It was a nice surprise that just the removal of the replicated setup of `--kubeconfig` in all the individual commands, and replacing it with a single centralized root level setup was enough to also fix the issue. Seems that the config code expected such a root level flag, and simply ignored it as `not set` when it was not present.

### DIFF
--- a/cli/cmd/internal/client/install.go
+++ b/cli/cmd/internal/client/install.go
@@ -1,7 +1,6 @@
 package client
 
 import (
-	"code.cloudfoundry.org/quarks-utils/pkg/cmd"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/suse/carrier/cli/kubernetes"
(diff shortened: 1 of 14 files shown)


cli/cmd/internal/client/install.go
```diff
@@ -33,15 +32,8 @@ var CmdInstall = &cobra.Command{
...
-	pf := CmdInstall.PersistentFlags()
-
-	argToEnv := map[string]string{}
-
-	cmd.KubeConfigFlags(pf, argToEnv)
-	cmd.AddEnvToUsage(CmdInstall, argToEnv)
-
-	CmdInstall.Flags().BoolP("verbose", "v", true, "Wether to print logs to stdout")
-	CmdInstall.Flags().BoolP("interactive", "i", false, "Whether to ask the user or not")
+	CmdInstall.Flags().BoolP("verbose", "v", true, "Whether to print logs (stdout) or not")
```
jimmykarily, line 35: Since you are at it, the description should say something like:

```
Print more detailed messages
```

because we do print logs to stdout anyway.

jimmykarily (approved):

cli/cmd/internal/client/install.go
andreas-kupries: Ok. Will do. Just a moment.

andreas-kupries: Done with commit 72726e959cd047b11a6b487091306ad48b80d7a5

PR #92, opened 2021-01-28
Add checksums to the release binaries

rohitsakala:
Co-authored-by: Tim Hardeck thardeck@suse.com

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,6 +27,9 @@ jobs:
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         run: |
           make build-all-small
+      - name: Create CHECKSUMS
+        if: steps.branch.outputs.BRANCH_NAME == 'main'
+        run: ( cd cli/dist ; sha256sum -b carrier* > SHA256SUM.txt )
       - name: Release Carrier
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         uses: fnkr/github-action-ghr@v1


.github/workflows/release.yml
```diff
@@ -27,6 +27,12 @@ jobs:
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         run: |
           make build-all-small
+      - name: Create CHECKSUMS
+        if: steps.branch.outputs.BRANCH_NAME == 'main'
+        run: |
+          pushd ./cli/dist/
+          sha256sum -b carrier* > SHA256SUM.txt
+          popd
```
andreas-kupries, line 35: Style nit, does not block approval. I would have written the command with a sub shell, i.e. as
`( cd cli/dist ; sha256sum -b carrier* > SHA256SUM.txt )`

andreas-kupries (approved):

.github/workflows/release.yml
rohitsakala: Done. Danke

andreas-kupries: :bell: Ah, not quite! It looks to me that you forgot the parentheses around the command, i.e. that which makes it a sub-shell (which limits the scope of the `cd`). As is the `cd` will affect the steps after it.

.github/workflows/release.yml
```diff
@@ -27,6 +27,9 @@ jobs:
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         run: |
           make build-all-small
+      - name: Create CHECKSUMS
+        if: steps.branch.outputs.BRANCH_NAME == 'main'
+        run: cd cli/dist ; sha256sum -b carrier* > SHA256SUM.txt
```
andreas-kupries, line 32: This is missing the `( )` of my original proposal. As is this `cd` affects the steps after. With the parens it will be a sub shell limiting the scope of the `cd` to that command and step.

andreas-kupries (changes requested):

.github/workflows/release.yml
rohitsakala: Danke. I didn't know that info.

andreas-kupries (approved):

PR #94, opened 2021-01-29
Remove HasLoadBalancer() func & Check ns ownership before uninstallation

### LINKED ISSUE #83: Carrier should not uninstall pre-existing components
I installed and uninstalled carrier into a cluster that already had tekton and traefik installed. When I ran `carrier uninstall` it removed my own tekton and traefik. 

It would be nice if carrier-installed tools had annotations so the uninstall could be more selective. 

As an aside: A pattern that I'm seeing a lot is to have a platform specific ingresses. I'd imagine that could extend eventually to all the components.

rohitsakala:
Fixes #83 

This PR does the following things

It removes HasLoadBalancer() concept as we cannot deterministically find if the underlying platform has the ability to provision IP address for `LoadBalancer` service type. Now, carrier will install traefik and wait for the the `LoadBalacner` IP. If it is not provisioned by the kube, then carrier will time out and point to a documentation which gives steps to enable this ability in local kubes.

This PR also checks ownership of the namespace before uninstalling components.

### DIFF
--- a/deployments/eirini.go
+++ b/deployments/eirini.go
@@ -65,6 +65,15 @@ func (k Eirini) Describe() string {
 func (k Eirini) Delete(c *kubernetes.Cluster, ui *ui.UI) error {
 	ui.Note().KeeplineUnder(1).Msg("Removing Eirini...")
 
+	existsAndOwned, err := c.NamespaceExistsAndOwned(eiriniCoreNamespace)
+	if err != nil {
+		return errors.Wrapf(err, "failed to check if namespace '%s' is owned or not", eiriniCoreNamespace)
+	}
+	if !existsAndOwned {
+		ui.Exclamation().Msg("Skipping Eirini because namespace either doesn't exist or not owned by Carrier")
+		return nil
+	}
+
 	releaseDir, err := k.ExtractRelease()
 	if err != nil {
 		return err
@@ -100,16 +109,13 @@ func (k Eirini) Delete(c *kubernetes.Cluster, ui *ui.UI) error {
 	// Delete namespaces last
 	for _, namespace := range []string{eiriniCoreNamespace, eiriniWorkLoadsNamespace, eiriniIngressNamespace} {
 		message := "Deleting Eirini namespace " + namespace
-		warning, err := helpers.WaitForCommandCompletion(ui, message,
+		_, err := helpers.WaitForCommandCompletion(ui, message,
 			func() (string, error) {
-				return c.DeleteNamespaceIfOwned(namespace)
+				return "", c.DeleteNamespace(namespace)
 			},
 		)
 		if err != nil {
 			return errors.Wrapf(err, "Failed deleting namespace %s", namespace)
-		}
-		if warning != "" {
-			ui.Exclamation().Msg(warning)
 		}
 	}
 
--- a/docs/install.md
+++ b/docs/install.md
@@ -1,0 +1,77 @@
+# Installation
+
+- [Installation](#installation)
+  - [Provision of External IP for LoadBalancer service type in Kubernetes](#provision-of-external-ip-for-loadbalancer-service-type-in-kubernetes)
+    - [K3s/K3d](#k3sk3d)
+    - [Minikube](#minikube)
+    - [Kind](#kind)
+    - [MircoK8s](#mircok8s)
+
+## Provision of External IP for LoadBalancer service type in Kubernetes
+
+Local kubernetes platforms do not have the ability to provide external IP address when you create a kubernetes service with `LoadBalancer` service type. The following steps will enable this ability for different local kubernetes platforms. Follow these steps before installing carrier.
+
+### K3s/K3d
+
+* Provision of LoadBalancer IP is enabled by default in K3s/K3d.
+
+### Minikube
+
+* Install and configure MetalLB
+```
+minikube addons enable metallb
+
+MINIKUBE_IP=($(minikube ip))
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: metallb-system
+  name: config
+data:
+  config: |
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - ${MINIKUBE_IP}/28
+EOF
+```
+
+### Kind 
+
+* Install JQ from https://stedolan.github.io/jq/download/
+
+* Install and configure MetalLB 
+```
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/namespace.yaml
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/v0.9.5/manifests/metallb.yaml
+kubectl create secret generic -n metallb-system memberlist --from-literal=secretkey="$(openssl rand -base64 128)"
+
+SUBNET_IP=`docker network inspect kind | jq -r '.[0].IPAM.Config[0].Subnet'`
+cat <<EOF | kubectl apply -f -
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  namespace: metallb-system
+  name: config
+data:
+  config: |
+    address-pools:
+    - name: default
+      protocol: layer2
+      addresses:
+      - ${SUBNET_IP}/28
+EOF
+```
+
+### MircoK8s
+
+* Install and configure MetalLB
+```
+INTERFACE=`ip route | grep default | awk '{print $5}'`
+IP=`ifconfig $INTERFACE | sed -n '2 p' | awk '{print $2}'`
+microk8s enable metallb:${IP}/16
+```
+
+
--- a/paas/install_client.go
+++ b/paas/install_client.go
@@ -4,12 +4,14 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 	"github.com/suse/carrier/cli/deployments"
+	"github.com/suse/carrier/cli/helpers"
 	"github.com/suse/carrier/cli/kubernetes"
 	"github.com/suse/carrier/cli/paas/config"
 	"github.com/suse/carrier/cli/paas/ui"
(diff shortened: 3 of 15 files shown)


cli/deployments/eirini.go
```diff
@@ -65,25 +65,6 @@ func (k Eirini) Describe() string {
...
-				dir := path.Join(releaseDir, "deploy", component)
-				return helpers.Kubectl("delete --ignore-not-found=true --wait=false -f " + dir)
-			},
-		)
-		if err != nil {
-			return errors.Wrapf(err, "%s failed:\n%s", message, out)
-		}
-	}
-
 	// Delete namespaces last
```
thardeck, line 68: This comment can be removed I suppose?

cli/deployments/eirini.go
```diff
@@ -97,6 +78,7 @@ func (k Eirini) Delete(c *kubernetes.Cluster, ui *ui.UI) error {
 		}
 		if warning != "" {
 			ui.Exclamation().Msg(warning)
+			return nil
```
thardeck, line 81: Warnings usually do not exit the method, maybe a comment explaining this behavior might help

cli/deployments/gitea.go
```diff
@@ -56,12 +56,26 @@ func (k Gitea) Describe() string {
...
+		func() (string, error) {
+			return c.DeleteNamespaceIfOwned(GiteaDeploymentID)
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "Failed deleting namespace %s", GiteaDeploymentID)
+	}
+	if warning != "" {
+		ui.Exclamation().Msg(warning)
+		return nil
```
thardeck, line 70: Same as above, why exit on warnings? A warning can be anything, this can result in unintended behavior.

cli/deployments/quarks.go
```diff
@@ -49,12 +49,26 @@ func (k Quarks) Describe() string {
...
+		func() (string, error) {
+			return c.DeleteNamespaceIfOwned(QuarksDeploymentID)
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "Failed deleting namespace %s", QuarksDeploymentID)
+	}
+	if warning != "" {
+		ui.Exclamation().Msg(warning)
+		return nil
```
thardeck, line 63: Same as above

cli/deployments/registry.go
```diff
@@ -57,12 +57,26 @@ func (k Registry) Describe() string {
...
+		func() (string, error) {
+			return c.DeleteNamespaceIfOwned(RegistryDeploymentID)
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "Failed deleting namespace %s", RegistryDeploymentID)
+	}
+	if warning != "" {
+		ui.Exclamation().Msg(warning)
+		return nil
```
thardeck, line 71: Same as above

cli/deployments/tekton.go
```diff
@@ -93,6 +80,20 @@ func (k Tekton) Delete(c *kubernetes.Cluster, ui *ui.UI) error {
 	}
 	if warning != "" {
 		ui.Exclamation().Msg(warning)
+		return nil
```
thardeck, line 83: Same as above, why exit on warnings? A warning can be anything, it should be more specific, otherwise this can result in unintended behavior.

cli/deployments/traefik.go
```diff
@@ -50,12 +50,26 @@ func (k Traefik) Describe() string {
...
+	warning, err := helpers.WaitForCommandCompletion(ui, message,
+		func() (string, error) {
+			return c.DeleteNamespaceIfOwned(TraefikDeploymentID)
+		},
+	)
+	if err != nil {
+		return errors.Wrapf(err, "Failed deleting namespace %s", TraefikDeploymentID)
+	}
+	if warning != "" {
+		ui.Exclamation().Msg(warning)
```
thardeck, line 63: Same as above

cli/paas/install_client.go
```diff
@@ -151,7 +151,11 @@ func (c *InstallClient) showInstallConfiguration(opts *kubernetes.InstallationOp
 func (c *InstallClient) fillInMissingSystemDomain(domain *kubernetes.InstallationOption) error {
 	if domain.Value.(string) == "" {
 		ip := ""
-		if !c.kubeClient.GetPlatform().HasLoadBalancer() {
+		foundLoadBalancer, err := c.kubeClient.GetPlatform().HasLoadBalancer(c.kubeClient.Kubectl)
```
thardeck, line 154: What do you think about showing a message that a loadbalancer was found? As general information and also for debugging purposes?

cli/kubernetes/platform/generic/generic.go
```diff
@@ -12,8 +12,17 @@ type Generic struct {
 	InternalIPs, ExternalIP []string
 }
 
-func (k *Generic) HasLoadBalancer() bool {
-	return true
+func (k *Generic) HasLoadBalancer(kube *kubernetes.Clientset) (bool, error) {
```
jimmykarily, line 15: The `HasLoadBalancer` method is supposed to return true when the cluster is deployed on a cloud (or other) where an external IP is automatically assigned (by an external load balancer). I couldn't find a way to introspect the cluster, I'm not sure that information is available. So I opted for hardcoded value on the platform. Default is `true` and we set it to `false` on the platforms we know there is no automatic IP assignment (e.g. kind).

jimmykarily:
Are you sure the mention of issue 83 in the description is correct? How does this fix that issue?

cli/deployments/eirini.go
rohitsakala: Ja.

cli/paas/install_client.go
rohitsakala: Removed this part of the code due to change in logic

cli/deployments/eirini.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/deployments/gitea.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/deployments/quarks.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/deployments/registry.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/deployments/tekton.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/deployments/traefik.go
rohitsakala: Added a comment and exiting only for a specific warning.

cli/kubernetes/platform/generic/generic.go
rohitsakala: As we discussed, we cannot hardcode the value as local kube platforms have addons which enable the ability of the local kube platforms to provision IP address.

rohitsakala:
> Are you sure the mention of issue 83 in the description is correct? How does this fix that issue?

This fixes the issue as there is a commit for re-ordering the uninstallation process. Namespaces should be deleted first which gives us the opportunity to check if it was created by carrier or not ?

paas/install_client.go
```diff
@@ -176,36 +176,32 @@ func (c *InstallClient) showInstallConfiguration(opts *kubernetes.InstallationOp
...
+		timeout := time.After(2 * time.Minute)
+		ticker := time.Tick(1 * time.Second)
+	Exit:
+		for {
+			select {
+			case <-timeout:
+				return errors.New("Timed out waiting for LoadBalancer IP on traefik service.\n" +
+					"Ensure your kubernetes platform has the ability to provision LoadBalancer IP address.\n\n" +
+					"Follow these steps to enable this ability\n" +
+					"https://github.com/SUSE/carrier/blob/main/docs/install.md#provision-of-external-ip-for-loadbalancer-service-type-in-kubernetes")
```
jimmykarily, line 190: Although it's easier for the user, you should probably avoid the anchor link (everything after `#`) because it will change if we simply reword the title. Just point to the file. Also put a TODO to update this link when we get a proper domain for carrier docs (e.g. docs.carrier.suse.dev or something like that)

deployments/eirini.go
```diff
@@ -110,6 +78,11 @@ func (k Eirini) Delete(c *kubernetes.Cluster, ui *ui.UI) error {
 		}
 		if warning != "" {
 			ui.Exclamation().Msg(warning)
+			// If the namespace is not created by carrier,
```
jimmykarily, line 81: Deleting the namespace first may lead to unpredictable results because we don't control the sequence of deletions. E.g. something is supposed to remove a finalizer from a deployment. If that "something" gets deleted before the deployment then the deployments will stay there not able to delete (We had that exact problem before).

Maybe create a method `NamespaceOwned` and call it instead of  `DeleteNamespceIfOwned` above? Something like:

- Check if we own the thing
- Delete components in order (one by one)
- Delete the namespace at the end (with `DeleteNamespaceIfOwned` or other method)
-

paas/install_client.go
rohitsakala: Done.

deployments/eirini.go
rohitsakala: Done.

deployments/eirini.go
```diff
@@ -65,83 +65,88 @@ func (k Eirini) Describe() string {
...
 
-	releaseDir, err := k.ExtractRelease()
+	owned, err := c.NamespaceOwned(eiriniCoreNamespace)
 	if err != nil {
-		return err
+		return errors.Wrapf(err, "failed to check if namespace '%s' is owned or not", eiriniCoreNamespace)
 	}
-	defer func() {
-		err := os.RemoveAll(releaseDir)
+	if owned {
```
jimmykarily, line 72: To make the code a bit more readable and avoid the additional identation, you can move the `else` part on top and return early. Something like:

```
if !c.NamespaceOwned(eiriniCoreNamespace)  {
  ui.Exclamation().Msg("Skipping Eirini because it is not owned by carrier")
  return nil
}

```

jimmykarily: The same in the other deployments.

docs/install.md
```diff
@@ -0,0 +1,72 @@
+# Installation
+
+- [Installation](#installation)
+  - [Provision of External IP for LoadBalancer service type in Kubernetes](#provision-of-external-ip-for-loadbalancer-service-type-in-kubernetes)
+    - [Minikube](#minikube)
```
jimmykarily, line 5: Maybe add k3s/k3d and simply say it's supported out of the box?

paas/install_client.go
```diff
@@ -176,36 +178,34 @@ func (c *InstallClient) showInstallConfiguration(opts *kubernetes.InstallationOp
...
-			defer s.Stop()
-			timeout := time.After(2 * time.Minute)
-			ticker := time.Tick(1 * time.Second)
-		Exit:
-			for {
-				select {
-				case <-timeout:
+		s := c.ui.Progressf("Waiting for LoadBalancer IP on traefik service.")
+		defer s.Stop()
+		timeout := time.After(2 * time.Minute)
```
jimmykarily, line 183: I think the code can be simplified if we use this: https://github.com/SUSE/carrier/blob/main/helpers/exec.go#L115

- Start the spinner
- ExecToSuccessWithTimeout
- Handle error or continue

deployments/eirini.go
rohitsakala: Done.

docs/install.md
rohitsakala: Done.

paas/install_client.go
rohitsakala: I created a new func RunToSuccessWithTimout since ExecToSuccessWithTimeout is used for running process.

Done.

jimmykarily (approved):

PR #95, opened 2021-01-29
Bump kubernetes libraries

jimmykarily:
Fixes https://github.com/SUSE/carrier/issues/93

We need to test if this gets rid of this errors as well:

```
error initializing cli: can't get gitea url: failed to list ingresses for gitea: the server could not find the requested resource
```

even if not, this PR gets rid of `replace` statements in go.mod and also bumps to latest kubernetes libraries so it should be good to merge anyway.


jimmykarily:
Installing and pushing an app works on gke 1.18 now (https://github.com/SUSE/carrier/pull/95) but the app route give us a 404.
The problem is that on gke (for some reason) the Ingress needs 2 annotations:

```
traefik.ingress.kubernetes.io/router.entrypoints: websecure
traefik.ingress.kubernetes.io/router.tls: "true"
```
as described here: https://community.traefik.io/t/kubernetes-ingress-cannot-create-service-subset-not-found/7723/3

I added those manually and now the application is accessible on gke too.

We can fix this additional thing on a new PR.

jimmykarily:
Added a new issue for the annotations: https://github.com/SUSE/carrier/issues/98

cli/deployments/tekton.go
```diff
@@ -329,6 +329,9 @@ func applyTektonStaging(c *kubernetes.Cluster, ui *ui.UI) (string, error) {
 func createTektonIngress(c *kubernetes.Cluster, subdomain string) error {
 	_, err := c.Kubectl.ExtensionsV1beta1().Ingresses("tekton-pipelines").Create(
 		context.Background(),
+		// TODO: Switch to networking v1 when we don't care about <1.18 clusters
+		// Like this (which has been reverted):
+		// https://github.com/SUSE/carrier/commit/7721d610fdf27a79be980af522783671d3ffc198
```
thardeck, line 334: Should we not support both calls, depending on the cluster version?

thardeck:
Besides that looks good to me.

thardeck (approved):

thardeck (changes requested):
I have verified the changes against different versions of Kubernetes on k3d and it seems that only 1.19 and 1.20 seem to work. In case of 1.16 Tekton does not come up and 1.17 and 1.18 have the issue `error initializing cli: can't get gitea url: failed to list ingresses for gitea: the server could not find the requested resource`.

thardeck:
1.18 seems to work too, there might be some hicckups in my local environment. I will look into it.

cli/deployments/tekton.go
jimmykarily: I think this one will work until v1beta1 is completely removed  (1.21 I think?). We can postpone conditionals until then

thardeck (approved):

PR #112, opened 2021-02-04
[fix] Removed old bash scripts and charts.

andreas-kupries:
Clean up, remove cruft


kube-config
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
jimmykarily, line 1: Added by mistake?

quark-secrets-log.txt
```diff
@@ -0,0 +1,275 @@
+2021-02-04T13:37:45.405Z	[34mINFO[0m	cmd/kubeconfig.go:16	Using in-cluster kube config
```
jimmykarily, line 1: also mistake?

kube-config
andreas-kupries: Likely yes. Will remove.

quark-secrets-log.txt
andreas-kupries: Very likely yes. Will remove.

It seems that I was in too much of a hurry when I made this :(

kube-config
andreas-kupries: Done with commit 0d380cfe9c5b24cdbc7eec2615976d3c88af3723 (rebase).

quark-secrets-log.txt
andreas-kupries: Done with commit 0d380cfe9c5b24cdbc7eec2615976d3c88af3723 (rebase).

jimmykarily:
We should probably update these make targets now: https://github.com/SUSE/carrier/blob/main/cli/Makefile#L44-L59

no files should be copied over anymore, they become first class citizens. The only exception is the registry which we have to `helm package`

jimmykarily:
This is a good first step. Next one would be to move the cli code outside the `cli` directory and make the root of the repo look like a go application project (so the contents of the `cli` directory move at the top level). Let's do it in steps. We can keep this PR small and merge it as it is. Maybe just update the Make targets for now.

andreas-kupries:
Ah. These targets pull files from the directories I removed. These files have to be restored and placed into the embedded-files hierarchy. Ok. Will be done ...

andreas-kupries:
Done with commit 8c0ce5a72fc6bf6536609fd5eb068f029144bb7b.
Files are already in the destination, so just removed the `cp` commands.

jimmykarily (approved):

PR #122, opened 2021-02-08
fix: Moved cli into the toplevel directory.

### LINKED ISSUE #119: Restructure project directories (cli is root dir)
After #112 we no longer have bash scripts around. The cli is the only thing left and makes no sense to have it in a subdirectory (Carrier *is* the cli). We should change the directory structure in Carrier to look like a go project the outcome of which is the cli itself (e.g. go.mod go.sum in the root dir etc).

andreas-kupries:
Fixes #119 

With the bash chaff gone the cli can take its place, making it a proper go project and application.
Merged the general targets of the top Makefile into the cli Makefile.
Updated the ignore clauses for the changed paths.


Makefile
```diff
@@ -1,88 +1,93 @@
...
-tidy:
-	@$(MAKE) -C cli tidy
+update_eirini:
+	mkdir -p embedded-files/eirini
+	wget https://github.com/cloudfoundry-incubator/eirini-release/releases/download/v2.0.0/eirini-yaml.tgz -O embedded-files/eirini/eirini-v2.0.0.tgz
 
-generate:
-	@$(MAKE) -C cli generate
+update_registry:
+	helm package ../container-registry/chart/container-registry/ -d embedded-files
```
jimmykarily, line 60: not `..` anymore?

andreas-kupries: Ggk. Checking ... Will fix.

andreas-kupries: @jimmykarily Huh. That is more difficult. The entire `container-registry` directory is gone, as part of removing the bash and charts. Do we have to restore it ? How was that directory filled ? Maybe we should put that command here, with the directory being temporary ?! ... Some kind of `helm fetch` ?

andreas-kupries: Fixed with commit ac32764269530c7676cb85093d5c2e4d83769199.

Makefile
```diff
@@ -1,88 +1,93 @@
...
-# Development
 
-tools-install:
-	@./scripts/tools-install.sh
+########################################################################
+## Development
 
-tools-versions:
-	@./scripts/tools-versions.sh
+compress:
```
jimmykarily, line 5: When I simply run `make` on my system, it complains about `upx` missing. I think make runs the first target by default. We should better move the `build` target at the top. Simply running `make` should run `build` I think.

jimmykarily (approved):

Makefile
andreas-kupries: Fixed with commit 15c69bf4567bb0da8d41a51bcdfb49434df1ac06

PR #124, opened 2021-02-08
Fix: Reduce default verbosity of `install`, `uninstall`, and `push`. Support verbosity levels now.

### LINKED ISSUE #117: Reduce command verbosity
The big commands (`install`, `uninstall`,`push`) are very verbose, telling the user about lots of stuff they are likely not interested in (kubernetes, gitea, ...) [at least when nothing goes wrong].

Consider adding log levels to the general user feedback. (Currently it supports only message categories (info, note, ok, exclamation, progress note). The levels would be separate from that ... Or maybe just for the `progress notes`.

We want to keep the progress indicator (dots).

andreas-kupries:
Fixes #117.

    feat: Extended ui package with verbosity levels.
    feat: Extended ui package with methods helping with keeping the output looking good regardless of level (newline control).
    fix:  Bumped the level of various progress messages (not the dots) to hide them at default verbosity.
    fix:  Tweaked the messaging in `push` for the different verbosity levels.
    fix:  Tweaked the messaging before various progress displays to be conditional on verbosity.
    
    The last fixes ensure that the dots of the dot-progress look nicely
    regardless of verbosity.  At standard verbosity, when the progress
    messages are hidden the dots attach to the last message before the
    progress indicator. With progress display active at higher verbosity
    the dots then attach to the individual progress messages.

TODO: Rebase after #122 is merged. :warning: __Do not merge before__ :warning:


cmd/root.go
```diff
@@ -40,6 +41,11 @@ func Execute() {
 	pf.StringVarP(&flagConfigFile, "config-file", "", "", "set path of configuration file")
 	config.KubeConfigFlags(pf, argToEnv)
 	config.LoggerFlags(pf, argToEnv)
+
+	pf.IntP("verbosity", "", 0, "Only print progress messages at or above this level")
```
thardeck, line 45: I think that we should provide some information what levels exist.

thardeck (changes requested):
It seems to work but I still get some verbose warnings without changing the verbosity nor the trace level, for example when pushing on kind (this might be more related to the trace pr though):
```
🚢 Deploying Tekton.............................................
W0210 09:34:28.187105   61839 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

Or when pushing an app:
```
../dist/carrier-linux-amd64 push test
W0210 09:34:55.604987   68575 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

🚢 Pushing application
Name: test
Organization: workspace
Sources: .

✔️  Application Repository created.

Creating webhook in the repo ...

Preparing code ...
W0210 09:34:56.268937   68575 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
```

cmd/root.go
andreas-kupries: The levels are numeric, zero and up. Currently only 0 and 1 used. Can try to place this in the help.

andreas-kupries:
The ingress stuff is printed from something low-level under no control of either verbosity, nor trace level. Something either from a called external command, or from the kube libraries.

I reported that in the cruel review friday ... It does not seem to have gotten an issue for it however.

With the reduced verbosity of the commands these messages now pop out much more than before.

cmd/root.go
andreas-kupries: Done with commit 6f842b54a673a184d1d728be9f37a580d056cdaa

thardeck (approved):

Issue #109, opened 2021-02-03
Remove Eirini?


jimmykarily:
Right now Eirini is only used in order to create a Statefulset out of an LRP. We don't use any other functionality from Eirini.

We should consider removing Eirini because:

- LRP is a CloudFoundry thing. Not known to anyone outside CloudFoundry
- It's equally easy to create a Deployment or Statefulset as it is to create an LRP. That means, we can directly create a Deployment instaed of an LRP
- There are other alternatives to abstract the Application concept if that's desirable (e.g. https://github.com/kubernetes-sigs/application). These tools could have bigger communities than Eirini in this world of Kubernetes.

satadruroy:
+1 

with regards to the "Application" concept, I found it interesting that someone asked a question about multiple microservices making up an application and how carrier would deal with that. FWIW, OAM formalizes the concept of an application supporting composites via components and workloads so I wonder if we need at least some basic abstraction as well.

https://github.com/oam-dev/spec/blob/master/3.component.md

https://github.com/oam-dev/spec/blob/master/4.workload_types.md

jimmykarily:
I've been thinking about that question. This seems to originate on [how Rio works (with riofile)](https://github.com/rancher/rio/blob/master/docs/riofile.md).

CloudFoundry, Heroku and buildpacks in general were created with a 12factor app in mind. That means, you have one app and various services your app needs. The PaaS deploys the app and gives you a way to provision the additional services. This model generally works (actually I would dare say it works in 99% of the apps out there).

Now, trying to find a case where this model doesn't work so well, we should think of an application where the services are not just external services. By external I mean services for which the developer doesn't write the code but simply wants to provision. We are talking about a microservices application on which mutliple "apps" (or "services") are developed and deployed separately and have to talk to each other.

The question is, how does the 12factor model of CloudFoundry/Heroku/buildpacks work for this case? I think, the various services of the "microservice" application are simply "apps" that are pushed on the same platform and they need to be able to talk to each other (with some additional features to not expose them to anybody else probably).

But even if this works, it still feels like a hack. The system doesn't recognize the application as one application and the features that block external access to the "microservices" are there only to support this case. This feels more like an afterthought than a real feature. On the other hand, looking at a riofile or a dockercompose or similar, the complexity of deploying your app seems to increase. Maybe so much, that using Kubernetes directly to deploy your app isn't that much harder anymore.

[...]

PR #128, opened 2021-02-09
Bump statik.go

thardeck:


thardeck:
We will wait until #109 is done so we do not have to update the file again.

andreas-kupries (approved):

PR #132, opened 2021-02-11
Accept path to app sources, and delay push, with ability to abort action

### LINKED ISSUE #120: Where should I be to push?
From the feedback we got so far, when the time comes to push an app, the users seem to get confused on where they should be when calling `carrier push myapp`.

Probably because of previous experience with other PaaS/clis they sometimes expect it to work by specifying some command line argument or other way.

We do explain how it works in the command's help but in any case, to avoid pushing things that are not the application code we could add an confirmation before pushing the application code. To stick to the "one command to deploy" scheme we could have a timer as in:

- user calls `carrier push myapp`
- A message is shown:
  `Carrier will now deploy the application in the current directory (/home/myuser/myapp) using the name "myapp". Hit Enter to continue or Ctrl+C to abort (deployment will continue automatically in 5 seconds)`
- The deployment continues after 5 seconds or after the user hits Enter. Deployment is aborted if Ctrl+C is hit.

If you can think of a better flow, please comment.

andreas-kupries:
Fixes #120 

I am not fully happy with the new initial message for push.
I still like the original message somewhat better.

May the blood-letting^Wbike-shedding begin :wink:


cmd/internal/client/push.go
```diff
@@ -10,9 +12,9 @@ var ()
 
 // CmdPush implements the carrier orgs command
 var CmdPush = &cobra.Command{
-	Use:   "push NAME",
+	Use:   "push NAME [PATH_TO_APPLICATION_SOURCES]",
 	Short: "Push an application from the current working directory",
```
jimmykarily, line 16: + `... or the path passed as argument"

andreas-kupries: Yes, definitely. Will fix.

andreas-kupries: Done with commit 28ff611d2058a50b69563b5e9285968ec205c187

andreas-kupries:
It seems that my change to `push` works badly with the acceptance tests ... 3 runs in CI, all failing in the same way.
The push does go forward, but then something breaks at the end, and we see an exit error ?!

That said ... I just had it working locally, although it was only on the second try.

jimmykarily (approved):

Issue #15, opened 2020-12-11
Carrier Acceptance Tests


jimmykarily:
We need a test suite we can run to be sure we didn't break any existing functionality.

Maybe it's too early but better start early than having to backfill the suite with tests later.

# Acceptance Criteria

- It should be possible to parallelize (each instance uses it's own cluster ? cluster pool ?)
- Should be possible to run in GitHub actions
- The test suite should complete in <= 10 minutes (and stay this way, either by parallelizing more or making the suite faster)
- For now it should check the basic functionality: (installation, application pushing, uninstallation etc)

Some tests could assume Carrier is running (e.g. tests that push apps) but some others can't (e.g. the tests that check install/uninstall). These 2 groups may not be able to run parallelized, sharing the same cluster. Needs some thought.

andreas-kupries:
Fixed by #135

PR #135, opened 2021-02-11
Add acceptance tests

thardeck:
Sometimes pushing/deleting an app fails but this seems to be related to our implementation, not the test.

We should squash the commits before merging them to main.


thardeck:
The github actions run seems to take now 10 minutes instead of 20.

.github/workflows/main.yml
```diff
@@ -20,101 +20,24 @@ jobs:
...
-          kubectl get all -A -o wide
-
-          # echo "-----------------------------------------------"
-          # echo "-----------------------------------------------"
-          # echo "DESCRIBE EIRINI-WORKLOADS PODS"
-          # echo "-----------------------------------------------"
-          # echo "-----------------------------------------------"
-
-          # kubectl get pod -n eirini-workloads -o json | jq -r .items[].metadata.name | xargs -I{} kubectl describe pod -n eirini-workloads {}
+          make test-acceptance
```
andreas-kupries, line 43: Nit, no newline at end of file.

acceptance/acceptance_suite_test.go
```diff
@@ -0,0 +1,208 @@
...
+var nodeSuffix, nodeTmpDir string
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	if os.Getenv("REGISTRY_USERNAME") == "" || os.Getenv("REGISTRY_PASSWORD") == "" {
+		panic("REGISTRY_USERNAME and REGISTRY_PASSWORD environment variables can't be empty")
+	}
+
+	fmt.Printf("Compiling Carrier on node %d\n", config.GinkgoConfig.ParallelNode)
+
+	buildCarrier()
```
andreas-kupries, line 38: It is unclear to me why `carrier` is built here, given that it is also built in the flow (main.yml#L23).

acceptance/acceptance_suite_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+func installCarrier() (string, error) {
+	return Carrier("install", "")
+}
+
+func uninstallCarrier() (string, error) {
+	return Carrier("uninstall", "")
+}
+
+// Carrier invoces the `carrier` binary, running the specified command.
```
andreas-kupries, line 174: Nit: `invokes` (`k`, not `c`)

acceptance/acceptance_suite_test.go
```diff
@@ -0,0 +1,208 @@
...
+// It returns the command output and/or error.
+// dir parameter defines the directory from which the command should be run.
+// It default to the current dir if left empty.
```
andreas-kupries, line 177: `defaults` - Trailing `s` missing.

acceptance/acceptance_suite_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	kubeconfig, err := RunProc("k3d kubeconfig get "+name, nodeTmpDir, false)
+	if err != nil {
+		panic("Getting kubeconfig failed: " + err.Error())
+	}
+	err = ioutil.WriteFile(path.Join(nodeTmpDir, "kubeconfig"), []byte(kubeconfig), 0644)
+	if err != nil {
+		panic("Writing kubeconfig failed: " + err.Error())
+	}
+}
```
andreas-kupries, line 106: Future: Should wait here for the cluster to be properly present. I saw `carrier install` getting run while the cluster's `traefik` was still in creation. It ended with `carrier` installing its own traefik, and k3d never able to start its own.

acceptance/apps_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+var _ = Describe("Apps", func() {
+	var org = "apps-org"
+	BeforeEach(func() {
+		_, err := Carrier("create-org "+org, "")
+		Expect(err).ToNot(HaveOccurred())
+		_, err = Carrier("target "+org, "")
+		Expect(err).ToNot(HaveOccurred())
+	})
+	Describe("push and delete", func() {
```
andreas-kupries, line 21: Future: Add `listing of pushed args`.

acceptance/orgs_test.go
```diff
@@ -0,0 +1,36 @@
...
+			out, err := Carrier("target", "")
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(MatchRegexp("Currently targeted organization: mycreatedorg"))
+
+			By("switching org back to default")
+			_, err = Carrier("target workspace", "")
+			Expect(err).ToNot(HaveOccurred())
+		})
+	})
+})
```
andreas-kupries, line 36: Future: Todo: `listing orgs`

scripts/tools/k3d.sh
```diff
@@ -0,0 +1,16 @@
+# shellcheck shell=bash
```
andreas-kupries, line 1: What is this for ? It looks incomplete, and has typos.

scripts/tools/k3d.sh
```diff
@@ -0,0 +1,16 @@
...
+
+K3D_VERSION="4.1.1"
+
+function k3d_version { k3d version; }
+
+K3D_SHA256_DARWIN="40fe273f567bad7e031636fe896a05637f173f047ee41f9566a6ad933e7da34f"
+K3D_SHA256_LINUX="4148b8035774a2765969898cc73066491ac718d1b2890415a4625931aaacd3ac"
+K3D_SHA256_WINDOWS="dc13acba412b473ab5b602420bf61cb2149a8b99cc24a0261574e3f965f03560"
+
+K3D_URL_DARWIN="https://github.com/rancher/k3d/releases/download/v{version}/k3d-darwin-amd64"
```
andreas-kupries, line 14: What is `{version}` ? Looks to be a typo. See no definition.

andreas-kupries (changes requested):
Mostly nits, a few things to do in the future.
The `request changes` is more for the two questions I had.

acceptance/apps_test.go
thardeck: Listing is already tested in this test to confirm that apps are pushed or deleted.

acceptance/orgs_test.go
thardeck: Same as above, we use the `carrier orgs` command to verify org creation and targeting.

scripts/tools/k3d.sh
thardeck: https://github.com/koalaman/shellcheck/blob/master/shellcheck.1.md#environment-variables

scripts/tools/k3d.sh
thardeck: Some scripting magic, it does properly install k3d. Maybe we should make it simpler.

acceptance/apps_test.go
andreas-kupries: Ah. I missed that. I believe because it was part of a larger block, not visually separate.

acceptance/orgs_test.go
andreas-kupries: Ditto. Missed, and is visually part of the create block.

scripts/tools/k3d.sh
andreas-kupries: Actually `this` refered to the entire script, not the particular line.

scripts/tools/k3d.sh
andreas-kupries: Ok. So this script is used somewhere ... I can't find a use however.

thardeck: This make target https://github.com/SUSE/carrier/blob/e758a47b4bfa305bac1fd74f703d6ba7fcfb0686/Makefile#L83 is called in the main.yml.

andreas-kupries: Digging ... Ah, name is dynamic, via globbing ... and it is some kind of configuration file. I understand now.

andreas-kupries (approved):

andreas-kupries:
Fixed #15

PR #145, opened 2021-02-15
Allow Acceptance tests to run in multiple ginkgo nodes

### LINKED ISSUE #142: Allow Acceptance tests to run in multiple ginkgo nodes
Currently the currently targeted org is always looked up in the config file in the home directory of the user. Although we use a different cluster per ginkgo node, they both use that file to refer to the targeted org. This results in flaky tests (e.g. https://github.com/SUSE/carrier/blob/main/acceptance/apps_test.go#L14) because the test may target an org but the other ginkgo node may target another in the meantime so when the first test tries to list the deployed apps, it lists those of the org of the other ginkgo node. 

This may not be a problem for the developer since they will probably use only one carrier instance per machine per user but it is a problem for tests. 

One solution would be to be able to override the default config location with a command line arg or an environment variable.

After fixing this, we should revert back to using more ginkgo nodes (https://github.com/SUSE/carrier/pull/137/commits/cf01b573a601beb0fa1c6aaa727814f0fa1df22d).

- [x] Environment variable for config directory.
- [x] Global cli option for the same.
- :warning: Check for ordering issues with the backend clients and their usage of config.
- [x] Document location of configuration, and how to influence it.
- [x] Change :cat2: to use the above to place config into node-specific directories
- [x] Change :cat2:  back to use of 2 ginkgo nodes

andreas-kupries:
Fixes #142.

  - Added support for environment variable `CARRIER_CONFIG` for config location.
  - Global cli option for config location already existed (`--config-file`).
  - Updated README, documenting the above
  - Use `CARRIER_CONFIG` in the :cat2: setup to place the config into ginkgo-node specific directories.
  - Re-enabled multi-node execution of :cat2:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,9 +37,7 @@ test: lint
 test-acceptance:
 	# Consider less nodes to avoid creating too many clusters which are not really needed
 	#ginkgo -p -stream acceptance/.
-	#ginkgo -nodes 2 -stream acceptance/.
-	ginkgo acceptance/.
-
+	ginkgo -nodes 2 -stream acceptance/.
 
 generate:
 	go generate ./...
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ $ carrier install
 ```bash
 $ carrier uninstall
 ```
----
+
 ### Push an application
 
 Run the following command for any supported application directory (e.g. inside [sample-app directory](sample-app)).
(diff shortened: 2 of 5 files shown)


thardeck:
I ran this a couple of times an still had two failures in the org test and one in the app test, of five to six runs I suppose.

thardeck:
> I ran this a couple of times an still had two failures in the org test and one in the app test, of five to six runs I suppose.

With the latest fixes in regards to  #146 the issue might gone - it ran fine four times in a row at least.

jimmykarily (approved):

PR #147, opened 2021-02-17
Add document to describe the process of deploying Wordpress on Carrier

### LINKED ISSUE #144: Deploy wordpress (no db)
We need to check if the PHP buildpack we are using is capable of running Wordpress. This story doesn't include the setup of Wordpress which comes after deployment.

jimmykarily:
[Fixes #144]

### DIFF
--- a/docs/wordpress.md
+++ b/docs/wordpress.md
@@ -1,0 +1,59 @@
+# Installing Wordpress with Carrier
+
+## Create a directory for your application:
+
+```bash
+mkdir wordpress
+cd wordpress
+```
+
+## Get the code:
+
+https://wordpress.org/download/#download-install
+
+```bash
+wget https://wordpress.org/latest.tar.gz
+tar xvf latest.tar.gz
+mv wordpress htdocs
+```
+
+## Create a buildpack.yml for your application
+
+```bash
+cat << EOF > buildpack.yml
+---
+php:
+  version: 7.3.x
+  script: index.php
+  webserver: nginx
+  webdirectory: htdocs
+EOF
+```
+
+## Enable needed php extensions
+
+The PHP buildpack supports additional ini files for PHP through
+[the PHP_INI_SCAN_DIR mechanism](https://paketo.io/docs/buildpacks/language-family-buildpacks/php/#php_ini_scan_dir).
+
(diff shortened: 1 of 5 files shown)


docs/wordpress.md
```diff
@@ -0,0 +1,59 @@
...
+The PHP buildpack supports additional ini files for PHP through
+[the PHP_INI_SCAN_DIR mechanism](https://paketo.io/docs/buildpacks/language-family-buildpacks/php/#php_ini_scan_dir).
+
+We need zlib and mysqli extensions enabled:
+
+```bash
+mkdir .php.ini.d
+cat << EOF > .php.ini.d/extensions.ini
+extension=zlib
+extension=mysqli
```
andreas-kupries, line 44: `EOF` missing.

docs/wordpress.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Deploy
+
+```
+carrier push wordpress
```
andreas-kupries, line 50: Should we remove the tarball from the cwd first ? I don't think that it is required, right ?

docs/wordpress.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+## Additional steps
+
+Wordpress needs a database to work. After visiting the route of your deployed
+application you will have to set the connection details to the database.
+
+You can install a MySQL database on your cluster or use an external one. One
+option is using a helm chart like this one: https://bitnami.com/stack/mysql/helm
```
andreas-kupries, line 59: IMHO adding the steps for setting up the database would be helpful too, while we do not have general services support.

andreas-kupries (changes requested):
Nits, and (maybe) a bit more information at the end (about the database).

docs/wordpress.md
jimmykarily: Added.

docs/wordpress.md
jimmykarily: Removed.

docs/wordpress.md
andreas-kupries: Retracting, the linked page has the necessary docs.

andreas-kupries (approved):

PR #153, opened 2021-02-19
Fix changelog generation

thardeck:


.github/workflows/release.yml
```diff
@@ -38,11 +30,21 @@ jobs:
...
+          issuesWoLabels: "true"
+          stripHeaders: "true"
       - name: Release Carrier
         uses: softprops/action-gh-release@v1
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         with:
           files: ./dist/*
-          body: ${{steps.changelog.outputs.changelog}}
+          body_path: ./CHANGELOG.md
+          prerelease: "true"
```
jimmykarily, line 48: Why a prerelease?  Should we manually check an convert to a final release? I'm ok with that, I'm just making sure I got this right.

jimmykarily (approved):

.github/workflows/release.yml
thardeck: Since the pipeline does automatic releases and all our other releases were pre-releases I thought we should stick with it until we decided differently. So yes, I think we should manually switch it when we think we are ready for beta.

PR #161, opened 2021-02-22
fix: Placed app errors into the table of apps, instead of aborting

### LINKED ISSUE #155: Listing apps: Any failure for an app listed, aborts listing
Issue:
Listing apps seems to abort listing apps, if there's a failure in listing e.g. the first app - this implies we cannot list all apps, if one of them can't be listed.

How to reproduce:
- push sample apps, e.g. sample-0, and sample-1 (while currently, in my environment those apps can't be listed, due to https://github.com/SUSE/carrier/issues/154)
- run e.g. `./carrier-linux-amd64-0.0.3 apps`

Output:
```
> ./carrier-linux-amd64-0.0.3 apps
🚢 Listing applications
Organization: suse
error listing apps: failed to get routes for app 'sample-0': failed to list ingresses: the server could not find the requested resource
```

... see that app `sample-1` isn't mentioned/listed at all, while it does exist:

```
> kubectl get deployments.apps -n carrier-workloads 
NAME                  READY   UP-TO-DATE   AVAILABLE   AGE
el-staging-listener   1/1     1            1           43m
suse.sample-0         1/1     1            1           29m
suse.sample-1         1/1     1            1           24m
```

In addition there's currently no way in carrier to list a specific app -> will report that as enhancement.

andreas-kupries:
Fixes #155 

When encountering issues with an app while listing all, it is better
to show the errors as part of the table, instead of blocking the
entire listing. This makes it easier to see if the cause is likely
more general, or more specific to some apps.

### DIFF
--- a/paas/client.go
+++ b/paas/client.go
@@ -15,6 +15,7 @@ import (
 	"time"
 
 	"code.gitea.io/sdk/gitea"
+	"github.com/fatih/color"
 	"github.com/go-logr/logr"
 	"github.com/otiai10/copy"
 	"github.com/pkg/errors"
@@ -136,7 +137,7 @@ func (c *CarrierClient) Apps() error {
 			fmt.Sprintf("carrier/app-guid=%s.%s", c.config.Org, app.Name),
 		)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get status for app '%s'", app.Name)
+			status = color.RedString(err.Error())
 		}
 
 		details.Info("kube get ingress", "App", app.Name)
@@ -144,7 +145,7 @@ func (c *CarrierClient) Apps() error {
 			c.config.CarrierWorkloadsNamespace,
 			app.Name)
 		if err != nil {
-			return errors.Wrapf(err, "failed to get routes for app '%s'", app.Name)
+			routes = []string{color.RedString(err.Error())}
 		}
 
 		msg = msg.WithTableRow(app.Name, status, strings.Join(routes, ", "))


jimmykarily:
I pushed 2 apps, I manually deleted the ingress of the first one and listed apps with carrier:

```
~/workspace/carrier (issue-155-app-listing)*$ ./dist/carrier-linux-amd64 apps
W0223 07:14:18.296624  733959 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

🚢 Listing applications
Organization: workspace

✔️  Carrier Applications:
|   NAME    | STATUS |                 ROUTES                  |
|-----------|--------|-----------------------------------------|
| wordpress | 1/1    | failed to list ingresses:               |
|           |        | ingresses.networking.k8s.io             |
|           |        | "wordpress" not found                   |
| worpress2 | 1/1    | worpress2.10.86.4.38.omg.howdoi.website |
~/workspace/carrier (issue-155-app-listing)*$ 

```

not bad :) The output could become weird if there is a lot to print. But for now, it's better than before.

jimmykarily (approved):

andreas-kupries:
> The output could become weird if there is a lot to print. But for now, it's better than before.

Yes. That is something in the underlying table implementation, I would think.
IOW the https://github.com/olekukonko/tablewriter
It might be as simple as adding code detecting terminal width to it, so that it can stretch the table.
Still, out of scope for now.

PR #163, opened 2021-02-22
fix: binary version of `carrier`

### LINKED ISSUE #152: carrier binary version not updated (always at version 0.1)
Carrier version 0.0.1 through 0.0.3 - the binaries do alwas report version 0.1:

```
cap@jump:~/caasp4/tf1-cluster> ./carrier-linux-amd64-0.0.1 --version
carrier version 0.1
cap@jump:~/caasp4/tf1-cluster> ./carrier-linux-amd64-0.0.2 --version
carrier version 0.1
cap@jump:~/caasp4/tf1-cluster> ./carrier-linux-amd64-0.0.3 --version
carrier version 0.1
```

How to reproduce:
Download e.g. the linux binary for carrier (https://github.com/SUSE/carrier/releases/download/v0.0.3/carrier-linux-amd64) and run `carrier-linux-amd64 --version`.

Verified the 3 given binaries do actually differ, but all do report version 0.1:
```
> md5sum carrier-linux-amd64-0.0.*
5d98b161492d4739e7b6fc3998c788ae  carrier-linux-amd64-0.0.1
1560d957fff7fbe56354720a6a9039cb  carrier-linux-amd64-0.0.2
b83ea99655b29a8858da2e988d03d635  carrier-linux-amd64-0.0.3
```

thardeck:
by using the result of `git describe --tags`.

Fixes #152

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,3 @@
-
 ########################################################################
 ## Development
 
@@ -40,7 +39,7 @@ test-acceptance:
 generate:
 	go generate ./...
 
-lint:	fmt vet tidy
+lint:	prepare_version fmt vet tidy
 
 vet:
 	go vet ./...
@@ -53,6 +52,9 @@ fmt:
 
 gitlint:
 	gitlint --commits "origin..HEAD"
+
+prepare_version:
+	echo -e "package cmd\n\nconst Version = \"$$(git describe --tags)\"" > cmd/version.go
 
 .PHONY: tools
 tools:
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,9 @@ gitlint:
 	gitlint --commits "origin..HEAD"
 
 prepare_version:
-	echo -e "package cmd\n\nconst Version = \"$$(git describe --tags)\"" > cmd/version.go
+	echo "package cmd" > cmd/version.go
+	echo "const Version = \"$$(git describe --tags)\"" >> cmd/version.go
+	cat cmd/version.go
 
 .PHONY: tools
 tools:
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -54,6 +53,9 @@ fmt:
 gitlint:
 	gitlint --commits "origin..HEAD"
 
+prepare_version:
+	echo -e "package cmd\n\nconst Version = \"$$(git describe --tags)\"" > cmd/version.go
```
andreas-kupries, line 57: Would be useful to `cat` the generated file here, to see that the contents are ok.
Especially good for CI logs.

jimmykarily (approved):

Makefile
```diff
@@ -40,7 +39,7 @@ test-acceptance:
 generate:
 	go generate ./...
 
-lint:	fmt vet tidy
+lint:	prepare_version fmt vet tidy
```
kkaempf, line 42: If `lint` depends on `prepare_version`, the changes in lines 11,14,17,20, and 23 are superfluous.

thardeck: Yes, I realized that yesterday after fixing linting, but I thought it is confusing if linting is required for building to be successful. Running the step twice should not heart so I left it but I can remove it if the dependency is not an issue.

Makefile
andreas-kupries: Looks to be done with commit 08951e38e856650fa6720bc18ab90956885b2eac

Makefile
andreas-kupries: > I  thought it is confusing if linting is required for building to be successful.

Based on how the Makefile is written we are requiring lint to be successful before we are even starting to build.  That is IMHO a good thing, i.e. as `vet` tells us about issues we might encounter first. Not just plain syntax errors, also more general things, like outdated constructions. I do not see the confusion.

Having the `prepare_version` required for lint simply makes that even more clear.

... So, I would go with @kkaempf here.

PR #164, opened 2021-02-22
fix: Wait for app pod to be removed before returning to caller

### LINKED ISSUE #131: `carrier delete` does sometimes exit before the deletion is finished
While writing the acceptance tests it occurred that `carrier apps` still listed the deleted app after `carrier delete` exited successfully.

Since this can result in race conditions we should make sure that `carrier` does not exit before the app and everything related to it was successfully deleted.

andreas-kupries:
Fixes #131 

Added support functions to detect pod non-existence, and to wait for
such. Used that new support to wait for a deleted application's pod to
be gone before continuing / returning to the user.

### DIFF
--- a/kubernetes/cluster.go
+++ b/kubernetes/cluster.go
@@ -170,6 +170,19 @@ func (c *Cluster) PodExists(namespace, selector string) wait.ConditionFunc {
 	}
 }
 
+func (c *Cluster) PodDoesNotExist(namespace, selector string) wait.ConditionFunc {
+	return func() (bool, error) {
+		podList, err := c.ListPods(namespace, selector)
+		if err != nil {
+			return true, nil
+		}
+		if len(podList.Items) > 0 {
+			return false, nil
+		}
+		return true, nil
+	}
+}
+
 // Poll up to timeout seconds for pod to enter running state.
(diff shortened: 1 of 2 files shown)


jimmykarily:
I just realized how unnatural it feels to not have the concept of the "Application" in Carrier (after all it's all about deploying applications right?). We deploy apps, we delete apps, we will soon bind services to apps but there is no "app" object to be found in our code.

I wonder if it is the right place to introduce that. The code on this PR looks fine so we could do that on a separate issue.

There are 2 sides to Applications, the object in our go code (which can be introduced at any time) and the representation of an app (or the `app abstraction`) in the cluster. For the latter this may be a good option: https://github.com/kubernetes-sigs/application

Feel free to do some refactoring to introduce the App concept in the code here if you want or we can do it in a separate PR.
I'm imagining code like:

```
  application := Application.New("wordpress")
  // Do things with the application
  application.Start()
  application.Stop()
  application.Delete()
  application.Logs() // stream?
  // etc
```

jimmykarily (approved):

andreas-kupries:
I would definitely go with a separate ticket and PR, simply so that we can discuss design first, instead of trying something ad hoc.

PR #170, opened 2021-02-24
Install service catalog

### LINKED ISSUE #157: As a carrier user I should be able to provision an SQL service on public cloud
This is our first take on the Services milestone and the next step on the "Carrier can run Wordpress" deliverable. We decided to give [crossplane](https://crossplane.io/) a try for the public cloud providers (it doesn't seem to support any "local" in-cluster provisioning like minibroker).

In some examples in the docs, crossplane is using an almost similar yaml for all providers, e.g. here: https://crossplane.io/docs/v1.0/getting-started/compose-infrastructure.html
but in some other places (e.g. here: https://crossplane.io/docs/v1.0/getting-started/provision-infrastructure.html) the yamls are very different for each provider.
Part of this issue is to see if using crossplane will allow us to minimize the provider-specific code we write in Carrier itself. With that in-mind, if crossplane provides multiple ways to provision services, we should pick the one that allows us to be provider-agnostic. If that's not possible then it should be reported on the issue and we should go back to the drawing board.

## Acceptance Criteria

- When the feature is enabled, the user should be able to:

  - list available services
  - provision services and bind them to applications (in 1 step or 2, TBD)
  - delete provisioned services

### LINKED ISSUE #169: Install/Uninstall service catalog deployment
This issue supersedes #159

See comments on the referenced issue on why we switched.

jimmykarily:
Fixes #169 

Doesn't add any functionality to Carrier on its own. Will keep as a draft PR and continue working on it to also fix #157 and maybe add minibroker to it (for local provisioning).


jimmykarily:
This PR is already big enough. Let merge this first and continue on new branch for #157

paas/client.go
```diff
@@ -50,6 +52,38 @@ type CarrierClient struct {
 	Log           logr.Logger
 }
 
+func NewCarrierClient(flags *pflag.FlagSet) (*CarrierClient, func(), error) {
```
jimmykarily, line 55: This is exactly the code `wire` generated. For now I just moved it to a place where it is usually expected (after the struct definition). We can simplify the method later.

paas/install_client.go
```diff
@@ -27,13 +29,40 @@ const (
 // installing Carrier on it.
 type InstallClient struct {
 	kubeClient *kubernetes.Cluster
+	options    *kubernetes.InstallationOptions
 	ui         *ui.UI
 	config     *config.Config
 	Log        logr.Logger
 }
 
+func NewInstallClient(flags *pflag.FlagSet, options *kubernetes.InstallationOptions) (*InstallClient, func(), error) {
```
jimmykarily, line 38: same here but with the addition of the `options` parameter. This parameter was the main reason for the removal of wire in the PR. Maybe there are workarounds but they certainly look uglier than removing wire.

paas/install_client.go
```diff
@@ -150,6 +174,12 @@ func (c *InstallClient) Uninstall(cmd *cobra.Command) error {
 	return nil
 }
 
+// InstallDeployment installs one single Deployment on the cluster
+func (c *InstallClient) InstallDeployment(deployment kubernetes.Deployment, logger logr.Logger) error {
```
jimmykarily, line 178: I will use this method to install minibroker (and other brokers) as separate carrier subcommands (in future PRs).

jimmykarily:
Pushed one more commit to install/uninstall Minibroker using `carrier enable services-incluster` and `carrier disable services-incluster` commands.

I will add another set of subcommands for gcloud broker.

cmd/internal/client/disable.go
```diff
@@ -0,0 +1,78 @@
...
+
+func init() {
+	CmdDisable.AddCommand(CmdDisableInCluster)
+	CmdDisable.AddCommand(CmdDisableGoogle)
+}
+
+func DisableInCluster(cmd *cobra.Command, args []string) error {
+	return UninstallDeployment(
+		cmd, &deployments.Minibroker{Timeout: paas.DefaultTimeoutSec},
+		"in-cluster services functionality has been disabled")
```
andreas-kupries, line 50: When uninstalling the minibroker I noticed that the command returns before the deployment is gone.
Fix might to have a `wait` flag for this function, and have the deployment then wait.

paas/install_client.go
```diff
@@ -132,15 +155,14 @@ func (c *InstallClient) Uninstall(cmd *cobra.Command) error {
...
 		&deployments.Tekton{Timeout: DefaultTimeoutSec},
 		&deployments.Registry{Timeout: DefaultTimeoutSec},
 		&deployments.Gitea{Timeout: DefaultTimeoutSec},
 		&deployments.Quarks{Timeout: DefaultTimeoutSec},
 		&deployments.Traefik{Timeout: DefaultTimeoutSec},
 	} {
-		details.Info("remove", "Deployment", deployment.ID())
-		err := deployment.Delete(c.kubeClient, c.ui)
-		if err != nil {
+		if err := c.UninstallDeployment(deployment, details); err != nil {
```
andreas-kupries, line 165: Uninstalling carrier with `in-cluster services` enabled leaves the relevant pod, i.e. `minibroker` active and running.
Pretty sure that is a general thing for the various services deployments.
Should possible add them here, unconditionally.
Anything not enabled will be skipped.

andreas-kupries (changes requested):
- Wait for xxx to be deleted before returning from `disable xxx`.
- Delete all the service xxx when uninstalling carrier.
-

paas/install_client.go
jimmykarily: Good catch! I will add them to the Uninstall of  `carrier uninstall`

cmd/internal/client/disable.go
jimmykarily: I thought about doing something just for this one but this is still the case with all Deployments. When deleting a namespace with `kubectl delete namespace` there is a `-wait` argument to wait until the namespaces is gone I think. I looked for something like this in the go client library and couldn't find it. I thought we may want to fix all `Delete` methods to wait in a separate PR. That's why I left it like this here.
I will have one more look to see if there is an easy way to fix at least this Deployment.

paas/install_client.go
jimmykarily: Done

cmd/internal/client/disable.go
andreas-kupries: ok.

andreas-kupries:
Issue with uninstall:

```
Removing Minibroker....Failed uninstalling helm release minibroker: Error: uninstallation completed with 1 error(s): unable to build kubernetes objects for delete: unable to recognize "": no matches for kind "ClusterServiceBroker" in version "servicecatalog.k8s.io/v1beta1"
: exit status 1
```

Looks to be an order issue. We have to remove the brokers before we remove the service catalog and its CRDs.

jimmykarily:
> Issue with uninstall:
> 
> ```
> Removing Minibroker....Failed uninstalling helm release minibroker: Error: uninstallation completed with 1 error(s): unable to build kubernetes objects for delete: unable to recognize "": no matches for kind "ClusterServiceBroker" in version "servicecatalog.k8s.io/v1beta1"
> : exit status 1
> ```
> 
> Looks to be an order issue. We have to remove the brokers before we remove the service catalog and its CRDs.
> 
> Weirdly enough, running `carrier disable ...` for the minibroker then works.

Fixed with: https://github.com/SUSE/carrier/pull/170/commits/ae4c219d9f0df863032644d47ad4563ff3ededa3

jimmykarily:
>     * Wait for xxx to be deleted before returning from `disable xxx`.
> 
>     * Delete all the service xxx when uninstalling carrier.

Fixed with https://github.com/SUSE/carrier/pull/170/commits/85d30e4b1188eeb95b5c1af829873c7cee48caa3

andreas-kupries (approved):

PR #171, opened 2021-02-24
Basic service management

### LINKED ISSUE #160: As a user I should be able to bind services to applications
## Acceptance Criteria
The user should be able to:

- Provision services
- Bind services to applications.
  This should make the service information (e.g. credentials) accessible to the application code somehow (e.g. env var or secret mounted as files).
- See what services an application has attached
- Unbind services from applications
- List service (bound or non-bound)

andreas-kupries:
Fixes #160.

The implementation is currently able to:

  - Create custom service from name and key/value dictionary.
  - Delete service by name
  - List services

Not yet implemented:

  - bind service to application
  - unbind service from application
  - list bindings per service
  - list bindings per application

Each service is represented by a secret, as described in #160.


andreas-kupries:
Based on my thoughts written down at https://github.com/SUSE/carrier/issues/160#issuecomment-785174422
I believe I have check myself here and research more before continuing.
The point to research is the abilities and schemata used by SvCat and how that influences integration and the CUPS representations.

paas/client.go
```diff
@@ -50,6 +50,176 @@ type CarrierClient struct {
...
+	details.Info("validate")
+	err := c.ensureGoodOrg(c.config.Org, "Unable to list applications.")
+	if err != nil {
+		return err
+	}
+
+	secrets, err := c.kubeClient.Kubectl.CoreV1().
+		Secrets("carrier-workloads").
+		List(context.Background(), metav1.ListOptions{})
+	// TODO: See if we can prefilter for service + org secrets
```
jimmykarily, line 73: I think using a label or annotation will allow easier filtering.

andreas-kupries: Likely yes. Have to learn labeling ... There is also the info about the SvCat resources you added to the ticket in response to my thoughts ... Thanks for that too.

jimmykarily (approved):

PR #175, opened 2021-02-25
fix: Check for existing webhook before creating it

flaviodsr:
When pushing an app, check if the webhook already exists before creating
it or it will crate a webhook for every push and consequently will trigger
the pipeline multiple times.

### DIFF
--- a/.github/workflows/lint_commit_messages.yml
+++ b/.github/workflows/lint_commit_messages.yml
@@ -12,7 +12,8 @@ jobs:
       - name: Setup gitlint
         run: sudo apt-get install gitlint
       - name: Checkout
-        run: git clone https://github.com/SUSE/carrier.git && cd carrier && git checkout ${GITHUB_HEAD_REF}
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
       - name: Lint new commits
-        working-directory: ./carrier
-        run: gitlint --commits "origin..HEAD"
+        run: gitlint --commits "origin/main..HEAD"
--- a/paas/client.go
+++ b/paas/client.go
@@ -474,7 +474,20 @@ func (c *CarrierClient) createRepo(name string) error {
 }
 
 func (c *CarrierClient) createRepoWebhook(name string) error {
-	c.ui.Normal().Msg("Creating webhook in the repo ...")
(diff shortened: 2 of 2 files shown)


paas/client.go
```diff
@@ -474,7 +474,20 @@ func (c *CarrierClient) createRepo(name string) error {
 }
 
 func (c *CarrierClient) createRepoWebhook(name string) error {
-	c.ui.Normal().Msg("Creating webhook in the repo ...")
+	hooks, resp, err := c.giteaClient.ListRepoHooks(c.config.Org, name, gitea.ListHooksOptions{})
+	if resp == nil && err != nil {
+		return errors.Wrap(err, "failed to list webhooks")
```
jimmykarily, line 479: Not sure if that can even happen, but if resp is not nil and we have an error, then this code ignores the error. Maybe just:

```
if err != nil {
  return errors.Wrap(err, "failed to list webhooks")
}
```

jimmykarily (changes requested):
Thanks for submitting this PR Flavio! Good catch. Just one comment, the rest works as expected.

jimmykarily:
CI is breaking because forked PRs don't have access to GitHub secrets. I created a PR which will hopefully fix it: https://github.com/SUSE/carrier/pull/176/files

this one should be rebased when #176 is merged.

jimmykarily:
ok CI is good, you just have to put a `fix:` in-front of your commit message to make the conventional commit linter happy: https://github.com/SUSE/carrier/blob/main/docs/development.md

paas/client.go
flaviodsr: thanks, TIL :smile:

.github/workflows/lint_commit_messages.yml
```diff
@@ -12,7 +12,10 @@ jobs:
       - name: Setup gitlint
         run: sudo apt-get install gitlint
       - name: Checkout
-        run: git clone https://github.com/SUSE/carrier.git && cd carrier && git checkout ${GITHUB_HEAD_REF}
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
jimmykarily, line 17: oh nice! @thardeck I think this should solve the issue you had (which is why you manually cloned).

jimmykarily (approved):

jimmykarily:
All good, merged! Thanks again.

PR #176, opened 2021-02-26
fix: Acceptance tests should run for PRs coming from forks

jimmykarily:
fix: Acceptance tests should run for PRs coming from forks  but Github secrets are not available in that case:
    
https://github.community/t/make-secrets-available-to-builds-of-forks/16166
    
We will make authenticated requests to GitHub only when the secrets are available.
We could use this instead:

https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target
https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks
    
but it comes with a big warning.

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -31,7 +31,7 @@ var nodeSuffix, nodeTmpDir string
 
 var _ = SynchronizedBeforeSuite(func() []byte {
 	if os.Getenv("REGISTRY_USERNAME") == "" || os.Getenv("REGISTRY_PASSWORD") == "" {
-		panic("REGISTRY_USERNAME and REGISTRY_PASSWORD environment variables can't be empty")
+		fmt.Println("REGISTRY_USERNAME or REGISTRY_PASSWORD environment variables are empty. Pulling from dockerhub will be subject to rate limiting.")
 	}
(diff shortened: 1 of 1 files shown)


andreas-kupries (commented):
What are the security implications ?
Can this be limited to forks from a suse repository ?

andreas-kupries (dismissed):
Ok.

jimmykarily:
According to the first link in the PR description, their main concern was that someone could change the workflow in a PR in a way that allows them to simply access the secret. The second link describes the implementation Github chose afaict and it allows running the workflow for PRs but using the version of the workflow in the upstream repo. So a malicious contributor can't just simply print your secrets in public.

This change has the implication that we won't be able to test workflow changes by simply opening a PR (even our PRs). But we can't have everything I guess.

jimmykarily:
Hmm I see a big warning here: https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target

jimmykarily:
Changed the code to use credentials when they are available and just make unauthenticated requests to Dockerhub when not.
The PR description is also updated.

andreas-kupries (approved):
ok

PR #180, opened 2021-03-02
Brokers acceptance test

jimmykarily:
This PR is adding tests for the previously added functionality (https://github.com/SUSE/carrier/pull/170)

It also simplifies the way we check for running pods (and makes it more reliable).


andreas-kupries (approved):
LGTM. Trialing local ...

PR #183, opened 2021-03-03
Remove tmp dir at the end of the prepareCode func

### LINKED ISSUE #173: Pushing apps leaves tmp dirs behind
We don't seem to delete the tmp directory we create when pushing and app: https://github.com/SUSE/carrier/blob/main/paas/client.go#L506

We should cleanup when pushing is done

rohitsakala:
fixes #173


andreas-kupries (dismissed):
looks good, seems to be working

jimmykarily (approved):

PR #193, opened 2021-03-05
test: Use multiple local clusters when provided to run acceptance tests

jimmykarily:
this should allow us to use a pool of warmed up clusters to run the
tests faster.


acceptance/acceptance_suite_test.go
```diff
@@ -53,11 +54,13 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	copyCarrier()
-	// NOTE: Don't set CARRIER_ACCEPTANCE_KUBECONFIG when using multiple ginkgo
-	// nodes because they will all use the same cluster. This will lead to flaky
-	// tests.
-	if kubeconfigPath := os.Getenv("CARRIER_ACCEPTANCE_KUBECONFIG"); kubeconfigPath != "" {
+
+	if kubeconfigDir := os.Getenv("CARRIER_ACCEPTANCE_KUBECONFIG_DIR"); kubeconfigDir != "" {
+		files, err := ioutil.ReadDir(kubeconfigDir)
+		Expect(err).ToNot(HaveOccurred())
+		kubeconfigPath := path.Join(kubeconfigDir, files[config.GinkgoConfig.ParallelNode-1].Name())
```
jimmykarily, line 61: This assumes that the `test-acceptance.sh` script below is doing the right thing and that the number of kubeconfig files match the number of ginkgo nodes. This should be a safe assumption.

thardeck (approved):

PR #201, opened 2021-03-08
Fix: Centralize carrier time durations into an internal package everything else imports.

### LINKED ISSUE #195: Carrier timeouts should be configurable
We have various timeouts with arbitrary values (e.g. https://github.com/SUSE/carrier/blob/main/paas/install_client.go#L25). In some cases (e.g. acceptance tests) we may need higher timeouts and it would be desirable to be able to multiply the default value by a factor using an environment variable.

## Acceptance Criteria

- The user should be able to multiple the default values for all timeouts by a factor using an environment variable. This could be on a per context basis too (e.g. network, cpu etc)

We will probably need to extract all timeouts to a separate package and import it when a timeout value is needed. To be able to use a multiplier, we will access the timeout values through getter methods that will make sure the multiplier is applied when available.

andreas-kupries:
Fixes #195 .

### DIFF
--- a/internal/duration/duration.go
+++ b/internal/duration/duration.go
@@ -1,0 +1,68 @@
+// Package duration defines the various durations used throughout
+// Carrier, as timeouts, and other.
+package duration
+
+import (
+	"time"
+)
+
+const (
+	multiplier    = 1
+	systemDomain  = 2 * time.Minute
+	deployment    = 5 * time.Minute
+	serviceSecret = 5 * time.Minute
+	podReady      = 5 * time.Minute
+	appBuilt      = 33 * time.Minute
+
+	// Fixed. __Not__ affected by the multiplier.
+	pollInterval = 3 * time.Second
+	userAbort    = 5 * time.Second
+	logHistory   = 48 * time.Hour
+)
+
+// LogHistory returns the duration to reach into the past for tailing logs.
+func LogHistory() time.Duration {
+	return logHistory
+}
+
+// PollInterval returns the duration to use between polls of some kind
+// of check.
+func PollInterval() time.Duration {
+	return pollInterval
+}
+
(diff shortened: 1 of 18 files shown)


internal/services/catalog_service.go
```diff
@@ -280,7 +280,8 @@ func (s *CatalogService) CreateBinding(bindingName, org, serviceName string) (in
 // to an Application.
 func (s *CatalogService) GetBindingSecret(bindingName string) (*corev1.Secret, error) {
 	// TODO: Replace hardcoded timeout with a constant
```
jimmykarily, line 282: remove TODO?

internal/duration/duration.go
```diff
@@ -0,0 +1,68 @@
+// Package duration defines the various durations used throughout
+// Carrier, as timeouts, and other.
+package duration
+
+import (
+	"time"
+)
+
+const (
+	multiplier    = 1
```
jimmykarily, line 10: I don't see a way to change the multiplier are run time. Maybe we can create a `Multiplier()` method that return either the value of `CARRIER_TIMEOUT_MULTIPLIER` environment variable or `1` if the variable is not set. What do you think?

andreas-kupries: Heh. I had defered that yesterday evening to today. The commit is currently waiting to be made (just a bit of testing).
The variable in the commit is `CARRIER_DURATION_MULTIPLIER` ... Hm. Timeout fits better, all the affected durations are timeouts (poll interval and user abort are not affected, and not timeouts per se). ... Will change.

internal/services/catalog_service.go
andreas-kupries: yes.

jimmykarily (approved):

PR #207, opened 2021-03-09
fix: Do not delete Services with bound apps

### LINKED ISSUE #198: Users shouldn't be able to delete services when those are bound to applications
## Acceptance Criteria

- When the user tries to delete a service but the service is bound some application, an error should be returned listing the bound applications
- The user should have a way to list the bound applications without trying to delete the Service (e.g. `carrier service mydb`)

andreas-kupries:
Fixes #198 .

- Do not delete services with bound apps. Shows the bound apps.
- Force delete services with bound apps, via `--unbind`. Service is unbound from apps before deletion. Shows the bound apps.
- Deleting an app unbinds all bound services first. Shows the bound services.
- Show bound apps per service when listing services.
- Show bound services per app, when listing apps.

Chores:
- Switched cli backends involving app listing to the application package
- Switched to `deployments.workloadsDeploymentID` for the app namespace. Removing the configuration element.

:heavy_check_mark: local :cat2:

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -1,12 +1,6 @@
 package acceptance_test
 
 import (
-	"fmt"
-	"os"
-	"path"
-	"strconv"
-	"time"
-
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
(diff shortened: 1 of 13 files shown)


andreas-kupries:
Multiple rebases
- Compressed history
- To main, with conflict resolution.

acceptance/utilities_test.go
```diff
@@ -0,0 +1,154 @@
...
+	Expect(out).To(MatchRegexp(serviceName))
+
+	out, err = helpers.Kubectl(fmt.Sprintf("get deployment -n carrier-workloads %s.%s -o=jsonpath='{.spec.template.spec.containers[0].volumeMounts}'", org, appName))
+	Expect(err).ToNot(HaveOccurred(), out)
+	Expect(out).To(MatchRegexp("/services/" + serviceName))
+}
+
+func deleteApp(appName string) {
+	out, err := Carrier("delete "+appName, "")
+	Expect(err).ToNot(HaveOccurred(), out)
```
jimmykarily, line 116: we use these  `delete*` methods on AfterEach blocks as cleanups. Sometimes when the test fails, one of these cleanups fail (e.g. trying to delete an app when the tests failed to create one). This make the AfterEach block to fail early and skip the rest of the cleanups. That's why I made them to not panic here: https://github.com/SUSE/carrier/commit/0c472ce52d7c8c9fd4b9b54b52a34c2421119143

so that cleanup can continue. If we can't to keep this functionality, we should probably return an error from these methods here, and run the `Expect(err).ToNot(HaveOccurred())` only where needed.

acceptance/apps_test.go
```diff
@@ -1,46 +1,46 @@
...
-		Expect(err).ToNot(HaveOccurred(), out)
-		out, err = Carrier("target "+org, "")
-		Expect(err).ToNot(HaveOccurred(), out)
+		setup_and_target_org(org)
 	})
 	Describe("push and delete", func() {
 		var appName string
 		BeforeEach(func() {
-			appName = "apps-" + strconv.Itoa(int(time.Now().Nanosecond()))
+			appName = app_name()
```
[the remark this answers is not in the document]
andreas-kupries: All done with the last rebase. Commit aac7c68847f17563b0748825114b3b064c39caa9

acceptance/utilities_test.go
andreas-kupries: I believe I saw these changes yesterday. And today I clearly missed them in my conflict resolution. Will fix.

andreas-kupries: Looking over them ... We have two different "delete service" ... One with the 10m timeout, and then those which should not fail because they are cleanup ... Likely similar for "delete app" ... I will separate these two uses n the utilities, rewrite the after blocks, and then see if both forms are still needed.

andreas-kupries: Done with commit ad4b4b67fa48464f0e5b8e2624ae74342d2034fc

acceptance/custom_services_test.go
```diff
@@ -1,128 +1,110 @@
...
 		It("creates a custom service", func() {
-			out, err := Carrier(fmt.Sprintf("create-custom-service %s username carrier-user", serviceName), "")
-			Expect(err).ToNot(HaveOccurred(), out)
-			out, err = Carrier("services", "")
-			Expect(err).ToNot(HaveOccurred(), out)
-			Expect(out).To(MatchRegexp(serviceName))
+			makeCustomService(serviceName)
+		})
+		AfterEach(func() {
+			cleanupService(serviceName)
```
jimmykarily, line 20: iiuc, `delete*` methods will raise on error while `cleanup*` versions will simply write it to stdout. Correct?

paas/client.go
```diff
@@ -523,50 +587,66 @@ func (c *CarrierClient) CreateOrg(org string) error {
 	return nil
 }
 
-// Delete deletes an app
-func (c *CarrierClient) Delete(app string) error {
+// Delete removes the named application from the cluster
+func (c *CarrierClient) Delete(appname string) error {
 	// TODO: lookup app, (get object), invoke action.
 	// TODO: Move action here into `internal/application`.
```
jimmykarily, line 593: TODOs done?

acceptance/custom_services_test.go
andreas-kupries: Yes.

paas/client.go
andreas-kupries: will check

andreas-kupries: correct, are done. removed with commit d4b8ed95ef3a382859c88fec943960c22130e9d8

jimmykarily (approved):

PR #209, opened 2021-03-11
Fix: Wait for service to be provisioned

### LINKED ISSUE #199: Users should be able to see the status of a provisioned service
RIght now, as soon as we create a ServiceInstance we return success. The service is not available yet though. We can see that status of the service by describing the ServiceInstance.

The user should be able to tell what the status is without having to use kubectl with something like: `carrier service mydb`

andreas-kupries:
Fixes #199 

- New command `service` to show service details, and status
- Extended `create-service` to wait for the new service to be provisioned by default
- Added option `--dont-wait` to tell `create-service` that it shall not wait as above.
- Added acceptance tests around `--dont-wait`, and for the new `service` command.

:heavy_check_mark: of local :cat2: (extended test suite)

### DIFF
--- a/acceptance/catalog_services_test.go
+++ b/acceptance/catalog_services_test.go
@@ -15,9 +15,14 @@ var _ = Describe("Catalog Services", func() {
 	})
 
 	Describe("create-service", func() {
-		It("creates a catalog based service", func() {
+		It("creates a catalog based service, with waiting", func() {
 			makeCatalogService(serviceName)
 		})
+
+		It("creates a catalog based service, without waiting", func() {
+			makeCatalogServiceDontWait(serviceName)
+		})
+
 		AfterEach(func() {
 			cleanupService(serviceName)
 		})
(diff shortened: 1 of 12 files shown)


acceptance/catalog_services_test.go
```diff
@@ -15,9 +15,14 @@ var _ = Describe("Catalog Services", func() {
 	})
 
 	Describe("create-service", func() {
-		It("creates a catalog based service", func() {
+		It("creates a catalog based service, with waiting", func() {
 			makeCatalogService(serviceName)
 		})
+
+		It("creates a catalog based service, without waiting", func() {
+			makeCatalogServiceDontWait(serviceName)
```
thardeck, line 23: Can't this test run into a race condition with the `AfterEach` cleanup?

acceptance/catalog_services_test.go
```diff
@@ -109,4 +114,23 @@ var _ = Describe("Catalog Services", func() {
...
+			makeCatalogService(serviceName)
+		})
+
+		It("it shows service details", func() {
+			out, err := Carrier("service "+serviceName, "")
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(MatchRegexp("Service Details"))
+			Expect(out).To(MatchRegexp(`Status .*\|.* Provisioned`))
+			Expect(out).To(MatchRegexp(`Class .*\|.* mariadb`))
+			Expect(out).To(MatchRegexp(`Plan .*\|.* 10-3-22`))
```
thardeck, line 129: How stable are the plans? Can we expect the number to be there for quite some time? Maybe we should do a regex for the numbers to keep it stable?

thardeck (dismissed):
Looks good to me besides the comments.

acceptance/catalog_services_test.go
andreas-kupries: No. The `without waiting` is for the core `create-service` command. The wrapper then waits explicitly on its own, using the `service command`.

acceptance/catalog_services_test.go
andreas-kupries: If the number is wrong the `create-service` will already break, because that is the plan we are asking for.
The check here simply verifies that the plan we asked for is the plan we got.

Issue #194, opened 2021-03-05
`carrier install` fails on minikube with podman


kkaempf:
Installing Carrier on minikube with podman. 

`carrier install` finishes with gitea network error:

> &gt; carrier install
> 
> 🚢 Carrier installing...
> 
> Configuration...
>   🧭  system_domain: 
> 
> 🚢 Deploying Traefik Ingress.......................
> ✔️  Traefik Ingress deployed
> .
> ✔️  Created system_domain: 192.168.49.0.omg.howdoi.website
> 
> 🚢 Deploying Quarks......................
> ✔️  Quarks deployed
> 
> 🚢 Deploying Workloads....
> ✔️  Workloads deployed
> 
> 🚢 Deploying Gitea........................................................................
> ✔️  Gitea deployed
> 
> 🚢 Deploying Registry..................................
> ✔️  Registry deployed
> 
> 🚢 Deploying Tekton..............................................................W0305 17:59:21.599096   27966 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
> .......................
> ✔️  Tekton deployed
> 
> 🚢 Deploying ServiceCatalog..................................................W0305 18:00:24.075369   27966 warnings.go:70] apiextensions.k8s.io/v1beta1 CustomResourceDefinition is deprecated in v1.16+, unavailable in v1.22+; use apiextensions.k8s.io/v1 CustomResourceDefinition
> 
> ✔️  ServiceCatalog deployed
> 
> ✔️  Carrier installed.
> System domain: 192.168.49.0.omg.howdoi.website
> W0305 18:00:24.129974   27966 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
> error initializing cli: gitea client failed: Get "http://gitea.192.168.49.0.omg.howdoi.website/api/v1/version": dial tcp 192.168.49.0:80: connect: network is unreachable
>

kkaempf:
`192.168.49.0` is `cni-podman0`:

> &gt; ip a
> ...
> 8: cni-podman0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default qlen 1000
>     link/ether 0e:cb:00:23:02:c4 brd ff:ff:ff:ff:ff:ff
[...]

PR #210, opened 2021-03-11
docs: Discourage minikube with podman driver

kkaempf:
Temporary workaround for #194


jimmykarily (approved):
Based on the comments on #194 we may need to revisit our documentation on the other clusters as well but for now it's good to warn people about podman not working. At least until we know why and either fix it and document it or explain what's wrong.

Issue #186, opened 2021-03-04
namespaces still exists immediately after `carrier uninstall`


kkaempf:
After an incomplete install[1], I did `carrier uninstall` which finished successfully, and a `carrier install` following immediately.

This resulted in
```
🚢 Carrier installing...

Configuration...
  🧭  system_domain: 
error installing Carrier: Namespace traefik present already
```

A `kubectl get namespace` did _not_ show `traefik` and a subsequent `carrier install` re-starts the installation.

It seems as if `carrier uninstall` returns too early, while the namespace deletion isn't completed yet.

---
¹ I tried `carrier install` on minikube and it failed with `error installing Carrier: Timed out waiting for LoadBalancer IP on traefik service.` - but that's a different problem ;-)

kkaempf:
This is true for all namespaces

kkaempf:
> This is true for all namespaces

No, it's not.  Carrier does

> 🕞   Waiting for workloads namespace to be gone ......

for workloads, but no other namespaces

jimmykarily:
The reason we waited for `workloads` is because there are some race conditions that prevent it from being deleted if some other components are removed first. iirc it was some finalizers that were left there. So we remove it first and after it's done we remove the rest of the components. It part the domain knowledge Carrier has that makes it useful in the first place.

jimmykarily:
There have been a lot of changes on how `uninstall` works the last months. This issue is about a rather old version. Please re-open is you encounter issues with the latest version. Closing.

PR #211, opened 2021-03-11
chore: wait for traefik namespace to be actually removed

kkaempf:
this is a copy & paste from deployments/workloads.go and partially
fixes issue #186

### DIFF
--- a/deployments/traefik.go
+++ b/deployments/traefik.go
@@ -12,6 +12,7 @@ import (
 	"github.com/suse/carrier/helpers"
 	"github.com/suse/carrier/kubernetes"
 	"github.com/suse/carrier/paas/ui"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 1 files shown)


kkaempf:
A lot of code below `deployments/` has similar patterns and is a prime candidate for refactoring. I don't have sufficient Go knowledge to do this myself :-/

jimmykarily:
At some point we implemented this: https://github.com/SUSE/carrier/blob/main/deployments/minibroker.go#L88 `WaitForNamespaceMissing`. I think it could be used for both for Traefik and Workloads.

kkaempf:
Or we close #186 and make it an explicit decision to uninstall as fast as possible

jimmykarily:
We need #186 but we instead of uninstalling everything sequentially, we need to use go routines to parallelize what is possible. Some deployments need to be uninstalled before others but some others could be uninstalled in parallel.

kkaempf:
Closing this PR due to above discussion.

PR #212, opened 2021-03-12
test: Disable workflows on forks

jimmykarily:
because we want to use self hosted GitHub Actions workers. There is no
way to use shared (GHA provided) workers just for forks and stick to
self hosted for our repository. This is a security problem because
anyone can just open a PR and start running artbitrary code on our
workers.

With this in place, in order to tests PRs from forks, a contributor on
our repository should create a copy of the forked PR (after reviewing
the code) in order to trigger tests. That copy should be updated with
every update of the original PR until it's merged.

https://github.community/t/stop-github-actions-running-on-a-fork/17965/2

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,6 +8,7 @@ on:
 
 jobs:
   build:
+    if: github.repository == "SUSE/carrier"
     runs-on: ubuntu-latest
 
     steps:


kkaempf:
How can we test this ?

kkaempf (dismissed):
lgtm

jimmykarily:
I guess I could open a PR from a fork using this branch as a base. Given we don't use `pull_request_target` (described [here](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/#improvements-for-public-repository-forks)), it should use my fork version of the workflow and not run anything. Let me try that.

jimmykarily:
I didn't disable the linting pipeline. Maybe that's safe to keep?

jimmykarily:
turns our `github.repository` is the repository where the workflow runs on (the target one then). There is nothing for the source repository. Doesn't help.

PR #213, opened 2021-03-12
Test forks

jimmykarily:
This PR is used to test this one: https://github.com/SUSE/carrier/pull/212/files

no build should be triggered for this PR


jimmykarily:
This doesn't work, `github.repository` is the target one, the one where the workflow runs. This doesn't help.

PR #214, opened 2021-03-12
feat: Warm up Carrier cluster and reduce app push timeout

jimmykarily:
otherwise the first time the user pushes an app the (big) buildpack
buidler image will have to be pulled for the first time.

On a warmed up cluster 30 minutes to stage an app should be too much. It
was introduced because test cluster were not warmed up and they timed
out. We need them to fail earlier now in order to retry flakes with
Ginkgo.

### DIFF
--- a/deployments/workloads.go
+++ b/deployments/workloads.go
@@ -8,11 +8,14 @@ import (
 	"github.com/kyokomi/emoji"
 	"github.com/pkg/errors"
 	"github.com/suse/carrier/helpers"
+	"github.com/suse/carrier/internal/duration"
 	"github.com/suse/carrier/kubernetes"
 	"github.com/suse/carrier/paas/ui"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	typedbatchv1 "k8s.io/client-go/kubernetes/typed/batch/v1"
 )
 
 type Workloads struct {
@@ -120,6 +123,13 @@ func (k Workloads) Deploy(c *kubernetes.Cluster, ui *ui.UI, options kubernetes.I
 	if err != nil {
 		return err
 	}
+
+	s := ui.Progress("Warming up cluster with builder image")
+	err = k.warmupBuilder(c)
+	if err != nil {
+		return err
+	}
+	s.Stop()
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


deployments/workloads.go
```diff
@@ -121,6 +124,13 @@ func (k Workloads) Deploy(c *kubernetes.Cluster, ui *ui.UI, options kubernetes.I
 		return err
 	}
 
+	s := ui.Progress("Warming up cluster with builder image")
```
jimmykarily, line 127: @andreas-kupries this message doesn't print :confused: any idea?

andreas-kupries: Crank the verbosity level.

andreas-kupries: Spinners show only the dots at the base level, and not their message.

andreas-kupries: For things where I always want a line I use
```
		c.ui.Note().KeeplineUnder(1).Msg("Provisioning...")
		s := c.ui.Progressf("Provisioning")
		defer s.Stop()
```
(CreateService in the wait for provision branch)

jimmykarily:
I run acceptance test locally and 1 test failed twice: 

```
[1] Summarizing 2 Failures:
[1] 
[1] [Fail] Apps push and delete [It] unbinds bound services when deleting an app 
[1] /home/opensuse/workspace/go/pkg/mod/github.com/onsi/gomega@v1.10.3/internal/assertion/assertion.go:43
[1] 
[1] [Fail] Apps push and delete [It] unbinds bound services when deleting an app 
[1] /home/opensuse/workspace/go/pkg/mod/github.com/onsi/gomega@v1.10.3/internal/assertion/assertion.go:43
[1] 
```

both times for the same reason:

```
[1]     🕞  Creating application resources
[1]                                                                                                                                                                                                                
[1]     🕞  Creating cloudfoundry.org/guid=apps-org.apps-886990501 in carrier-workloads ...........................................................................................................................
...................................................................................................................................................................................................................
...................................................................................................................................................................................................................
.......................................................                                                                                                                                                            
[1]     error pushing app: waiting for app failed: waiting for app to be created failed: timed out waiting for the condition  
```

It seems as if Tekton pipeline never triggered. We push the code but nothing happens. What I've seen happening on some manual tests is that the webhook endpoint is not accessible for some reason and the Gitea never retries to contact it after failing. It works if you go to Gitea setting and try the webhook manually (there is a button for that). I wonder if there is some Gitea setting to automate that.

jimmykarily:
others seem to ask for a "retry" feature: https://github.com/go-gitea/gitea/issues/5470#issuecomment-468594944 
no answer though.

deployments/workloads.go
jimmykarily: Done with #05053b1

andreas-kupries (approved):

PR #222, opened 2021-03-16
feat: add kubernetes app labels on servicebinding and its secrets

### LINKED ISSUE #218: Use Kubernetes common labels for services created for carrier apps.
This is the remaining work to be done on issue #150. 
The task for this issue is to ensure that generic labels are put on carrier app services.

rohitsakala:
fixes #218

### DIFF
--- a/internal/services/catalog_service.go
+++ b/internal/services/catalog_service.go
@@ -208,7 +208,7 @@ func (s *CatalogService) GetBinding(appName string) (*corev1.Secret, error) {
 		return nil, err
 	}
 	if binding == nil {
-		_, err = s.CreateBinding(bindingName, s.OrgName, s.Service)
+		_, err = s.CreateBinding(bindingName, s.OrgName, s.Service, appName)
 		if err != nil {
 			return nil, err
 		}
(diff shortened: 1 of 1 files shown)


internal/services/catalog_service.go
```diff
@@ -275,8 +285,34 @@ func (s *CatalogService) CreateBinding(bindingName, org, serviceName string) (in
...
 
-	return dynamicClient.Resource(serviceBindingGVR).Namespace(deployments.WorkloadsDeploymentID).
+	serviceBinding, err := dynamicClient.Resource(serviceBindingGVR).Namespace(deployments.WorkloadsDeploymentID).
 		Create(context.Background(), obj, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+
+	// Update the binding secret with kubernetes app labels
+	secret, err := s.GetBindingSecret(bindingName)
```
andreas-kupries, line 295: This code seems to assume that the secret is immediately available after the binding resource is created.
I do not believe that we can rely on that. After creating the binding resource
  - It has to be seen by the catalog operator
  - The operator has to talk to the responsible broker (i.e. request and get binding info)
  - And only after all of the above can the operator create the secret with the information it got from the broker.

I would extend the line 295ff to check if the error is `Not Found` and then loop until the secret is actually present
(Maybe a limit of 2 minutes ?) Or maybe place this loop in `GetBindingSecret`.

andreas-kupries (changes requested):
Looks mostly good. One place where I am concerned about timing.

internal/services/catalog_service.go
andreas-kupries: Retracting, now that I saw https://github.com/SUSE/carrier/blob/a07454c02a3705ba74f10be936b9b5d1f5dd6194/internal/services/catalog_service.go#L320

IOW, `GetBindingSecret` already wait.

andreas-kupries (dismissed):

andreas-kupries (approved):

PR #223, opened 2021-03-16
Fix Un/bind service in conflict

### LINKED ISSUE #221: CI fails with 4 Gingko runners
See https://github.com/SUSE/carrier/actions/runs/654956612.
Fails during `doesn't delete a bound service [It]`

Previous runs (with 2 Ginkgo runners) were successful.

@andreas-kupries :point_up_2: :wink:

andreas-kupries:
Fixes #221 

It is expected that the loop will resolve quickly. The most likely reason for it is that the service is bound/unbound concurrently. In that case the loop breaks at the recheck for service un/bound.

### DIFF
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -13,6 +13,7 @@ import (
 	"github.com/suse/carrier/kubernetes"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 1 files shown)


internal/application/application.go
```diff
@@ -131,44 +139,53 @@ func (a *Application) Bind(service interfaces.Service) error {
...
+		deployment.Spec.Template.Spec.Volumes = volumes
+		deployment.Spec.Template.Spec.Containers[0].VolumeMounts = volumeMounts
+
+		_, err = a.kubeClient.Kubectl.AppsV1().Deployments(deployments.WorkloadsDeploymentID).Update(
+			context.Background(),
+			deployment,
+			metav1.UpdateOptions{},
+		)
+
+		if err == nil || !apierrors.IsConflict(err) {
```
kkaempf, line 181: This is essentially the same exit criteria as in lines 113 - 118 above. I would use the same code (two separate `if`` statements) at both places

kkaempf (changes requested):
Ok except for the one comment

internal/application/application.go
andreas-kupries: Ok. A moment ...

andreas-kupries: Addressed with commit 7ae5bb8452eda05992bb49eb97d524efdb6b8fd1

kkaempf (approved):
lgtm

PR #225, opened 2021-03-16
perf: Use "lite" image for quarks

jimmykarily:
fixes: https://github.com/SUSE/carrier/issues/158


thardeck (approved):
LGTM, tests were green on my machine, so I will merge it.

Issue #217, opened 2021-03-15
Spike: Improve CI performance and stability


thardeck:
Our CI often takes quite some time to finish and runs into timeouts and other various issues.

### Potential improvements:
* warm up cluster
* more gingko nodes
* local/public cloud runners
...

## Acceptance criteria
* Findings are documented
* Issues are fixed or Github items are created
* Time box: one work week

kkaempf:
The `safe to test` label, as suggested [by github](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) seems to be a good (initial) safeguard for 'external' PRs.

kkaempf:
re-running on a "warmed up" cluster doesn't make a huge difference:
- first run ~ 15 mins
- second run ~ 14 mins
- third run ~ 13 mins

kkaempf:
> The `safe to test` label, as suggested [by github](https://securitylab.github.com/research/github-actions-preventing-pwn-requests) seems to be a good (initial) safeguard for 'external' PRs.

Confirmed. The trigger is on 'adding event to PR', so the action runs once after the label has been set.
Subsequent edits of the PR do not re-trigger the action run. Only removal and re-adding of the label (needs repo write access) does.

kkaempf:
By checking `

> re-running on a "warmed up" cluster doesn't make a huge difference:
> 
>     * first run ~ 15 mins
> 
>     * second run ~ 14 mins
> 
>     * third run ~ 13 mins

Adding more hardware (+50% cpu cores, +100% main memory) does not improve these numbers.

kkaempf:
Wow, warmed cluster and 4 parallel gingko nodes - down to 8 minutes !

https://github.com/SUSE/carrier/runs/2120683393

PR #226, opened 2021-03-16
Issue 217: restrict self hosted to known users

kkaempf:
Run `main` CI task on self-hosted runners for known users

Also bump GINGKO_NODES to 6. The current tests don't make use of more than 6 nodes.

This brings the CI time down to approx. 8 minutes.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,8 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
+    if: contains(["kkaempf", "jimmikarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
 
     steps:
       - name: Checkout
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -9,7 +9,7 @@ on:
 jobs:
   build:
     runs-on: self-hosted
-    if: contains(["kkaempf", "jimmikarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
+    if: contains(["kkaempf", "jimmykarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
 
     steps:
       - name: Checkout
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -41,4 +41,4 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
         run: |
-          make test-acceptance
+          GINKGO_NODES=6 make test-acceptance


.github/workflows/main.yml
```diff
@@ -8,7 +8,8 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
+    if: contains(["kkaempf", "jimmikarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
```
jimmykarily, line 12:
```suggestion
    if: contains(["kkaempf", "jimmykarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
```

jimmykarily:
Can this be combined with the labeling method inplemented in #219 ? As in "run tests if the actor is known or someone labeled the PR".

Also, do we know who the "actor" is? Is it the committer of the last commit, the person opening the PR or someone else? Who is the actor when we simply merge to master?

.github/workflows/main.yml
kkaempf: :facepalm: :wink:

kkaempf:
The actor, in the github action sense, is the one triggering the action. As `workflow/main.yml` triggers on `push` and `pull_request`, we _should_ be safe.

#219 is the correct solution for _outside_ PRs, where someone with write access to the repo must explicitly add a label to trigger the action run. I haven't found a way to combine both inside `workflow/main.yml` (would need an `else` expression). So I'd guess we need another action for this. (PR follows :wink:)

kkaempf:
#228 opened

.github/workflows/main.yml
```diff
@@ -8,7 +8,8 @@ on:
 
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: self-hosted
+    if: contains(["kkaempf", "jimmykarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}")
```
jimmykarily, line 12: wouldn't this combine both #217 and #228 ? : 

```
if : contains(["kkaempf", "jimmykarily", "andreas-kupries", "rohitsakala", "svollath", "thardeck"], "${{ github.actor }}") || contains(github.event.pull_request.labels.*.name, 'safe to test')
```

https://docs.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions#operators

kkaempf: :thinking: 

I also don't know how much importance the `types: [ labeled ]` line has.

jimmykarily: I'm reading [this relatively irrelevant page](https://github.com/marketplace/actions/label-actions#examples) but afaict `types: [ labeled]` refers to the PR events that will trigger the workflow. That means the workflow it only triggered when a label is added. Combined with the label check it does what we want.

kkaempf: Yes, `types: [labeled]` is the trigger. But this trigger would be valid for any actor. Which doesn't match my intention:
- if known actor, run unconditionally
- if unknown actor, run only when `safe to test` label is added

I fear this cannot be combined in a single action.

jimmykarily: Now I get it. You are right. I would still prefer to have them combined in one like this:

- Run PRs only when tagged. This allows us to skip builds when we are working on a PR, pushing commits but when we are not done yet. I don't mind manually tagging the PR again to trigger a build.
- Run main branch only when the actor is one of the team members (when I merge a PR, I'm the actor right?).

jimmykarily (approved):

PR #231, opened 2021-03-18
Feature for creating production certificates when user specifies a domain address

### LINKED ISSUE #179: User has a domain but no certificate for it
Buying a domain is a quite simple task nowadays. Buying a certificate for it is both more complex and unnecessary, because there are free options (like let's encrypt).

Carrier needs to be able to create proper CA signed certificates for a domain the user owns.

## Acceptance Criteria

- When the user specifies a domain during Carrier installation, we should use let's encrypt to create production ready certificates for the applications (probably using cert-manager)

## Useful links

- https://docs.cert-manager.io/en/release-0.7/tasks/acme/configuring-http01.html

rohitsakala:
fixes #179


acceptance/wordpress_test.go
```diff
@@ -110,13 +110,11 @@ var _ = Describe("Wordpress", func() {
...
-		Expect(resp.StatusCode).To(Equal(http.StatusOK))
-		defer resp.Body.Close()
-
-		bodyBytes, err := ioutil.ReadAll(resp.Body)
-		Expect(err).ToNot(HaveOccurred())
-		Expect(string(bodyBytes)).To(MatchRegexp("Select a default language"))
+		Eventually(func() int {
+			resp, err := client.Do(request)
+			Expect(err).ToNot(HaveOccurred())
+			defer resp.Body.Close()
```
andreas-kupries, line 116: The `Body` is not read/used, contrary to before the change. Why is the `Close()` still needed ?

andreas-kupries (changes requested):
Have a question actually.

Note further, this and PR #241 are in conflict, both modify the `Delete` function of the CarrierClient. Whichever is merged last needs a rebase to resolve the conflict.

andreas-kupries:
Another question: How can we test this facility. While I managed to do the basic install/uninstall and it looks good, the cert create/delete on app push/delete is triggered only for a non `omg` domain. And I do not recall us having tests using such in any way. So this part of the change cannot be tested ?

rohitsakala:
> Another question: How can we test this facility. While I managed to do the basic install/uninstall and it looks good, the cert create/delete on app push/delete is triggered only for a non `omg` domain. And I do not recall us having tests using such in any way. So this part of the change cannot be tested ?

I tested it using a personal domain and a VM from the google cloud. I added google cloud dns entry mapping the wildcard of personal domain to the VM IP address. So, maybe we can use the google cloud account of carrier to test PR.

acceptance/wordpress_test.go
rohitsakala: Removed.

andreas-kupries (approved):

rohitsakala:
Created an issue about rate limiting - https://github.com/SUSE/carrier/issues/256.

PR #232, opened 2021-03-18
Added command to show app details

### LINKED ISSUE #156: Carrier to list information on a single app specified
Use case:
Given there is more that one app deployed to an org -
As a user, I like to get information on a single app, I do specify -
e.g. by running `carrier app sample-1`

So, e.g.:
`./carrier-linux-amd64-0.0.3 apps` does list information on "sample-0" and "sample-1" (-> all apps in the targeted org)
`./carrier-linux-amd64-0.0.3 app sample-1` would only list information on "sample-1"

while `carrier app <app-name>` could list more detailed information on a singe app, in addition.

((Related to https://github.com/SUSE/carrier/issues/155))

andreas-kupries:
Fixes #156 

Beyond the new command to show details of a single app this PR reorgs the `apps` command and the new command into a new hierarchy:
```
    apps
        list
        show
```

While the `push` command was not moved into this hierachy, it could, maybe should.

Existing tests were updated to match the new command name for `apps list`.
Tests added to cover the new `apps show` command.

Fix to the workpress test, it was not setting up an org for itself. Seen only when focusing on it alone.

Extended the Makefile with a `showfous` target to tell us which tests, if any, are focused.
This is now a dependency of the `test-acceptance` target.

Mostly :heavy_check_mark: for a local :cat2: Wordpress test fails in pushing the app, does not seem to be related.

And :heavy_check_mark: :cat2: on our gh runner :fireworks:

### DIFF
--- a/cmd/internal/client/apps.go
+++ b/cmd/internal/client/apps.go
@@ -8,9 +8,25 @@ import (
 
 var ()
 
-// CmdApps implements the carrier app command
-var CmdApps = &cobra.Command{
-	Use:   "apps",
+// CmdApp implements the carrier -app command
+var CmdApp = &cobra.Command{
+	Use:           "apps",
+	Aliases:       []string{"app"},
+	Short:         "Carrier application features",
+	Long:          `Manage carrier application`,
+	Args:          cobra.ExactArgs(0),
+	SilenceErrors: true,
+	SilenceUsage:  true,
+}
+
+func init() {
+	CmdApp.AddCommand(CmdAppsShow)
+	CmdApp.AddCommand(CmdAppsList)
+}
+
+// CmdAppsList implements the carrier `apps list` command
+var CmdAppsList = &cobra.Command{
+	Use:   "list",
 	Short: "Lists all applications",
 	Args:  cobra.ExactArgs(0),
 	RunE: func(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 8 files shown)


thardeck:
It might be confusing that we have `carrier service` (singular) but `carrier apps` (plural). I know that there is the alias but the default should be similar I think.

thardeck:
I would also adapt the acceptance tests to represent the name change, they should test the default name, not the alias, at least not only the alias.

andreas-kupries:
Will do

andreas-kupries:
@thardeck Done

cmd/internal/client/apps.go
```diff
@@ -35,3 +51,51 @@ var CmdApps = &cobra.Command{
 	SilenceErrors: true,
 	SilenceUsage:  true,
 }
+
+// CmdAppsShow implements the carrier `apps show` command
+var CmdAppsShow = &cobra.Command{
```
thardeck, line 56: Nitpick: the var names and comments also still use the plural form.

cmd/internal/client/apps.go
```diff
@@ -8,9 +8,25 @@ import (
...
+	SilenceErrors: true,
+	SilenceUsage:  true,
+}
+
+func init() {
+	CmdApp.AddCommand(CmdAppsShow)
+	CmdApp.AddCommand(CmdAppsList)
+}
+
+// CmdAppsList implements the carrier `apps list` command
```
thardeck, line 27: Same as mentioned below.

thardeck (changes requested):
Besides that lgtm

cmd/internal/client/apps.go
andreas-kupries: Done with commit 1823b167f79f1efa05871a307e3afdcb601cdf6b

cmd/internal/client/apps.go
andreas-kupries: Done with commit 1823b167f79f1efa05871a307e3afdcb601cdf6b

PR #234, opened 2021-03-24
fix: Remove thim binary from gitignore (we have no shim anymore)

jimmykarily:


jimmykarily:
the clusters in the self-hosted runners need to be re-created. They must have failed at some point and some namespace didn't get the label. Now they are not properly cleaned up. That's a bug we need to fix but it's irrelevant to this PR. This PR passes locally.

PR #235, opened 2021-03-24
Api

jimmykarily:
Create an HTTP api and a dashboard to consume it.

This is just a prototype just to introduce the idea. Future work:

- cli should be talking to the same api. This should happen in 2 steps. At first when a cli command is run, we will spin up the server on a local port and the cli will talk to the server to fetch the needed information. After we switch all commands to consume an API endpoint we will move the `carrier server` inside the cluster (as part of the `carrier install` command). Then the carrier cli will need to be pointed to the carrier service (probably using an ingress).
- Extend the web dashboard to show more information (e.g. available service plans)


internal/api/v1/applications.go
```diff
@@ -0,0 +1,86 @@
...
+	if handleError(w, err, 500) {
+		return
+	}
+	defer func() {
+		if cleanup != nil {
+			cleanup()
+		}
+	}()
+
+	apps, err := application.List(client.KubeClient, client.GiteaClient, org)
```
jimmykarily, line 32: This is a bit of a "hack". I wanted to re-use code as much as possible so I made `KubeClient`  and `GiteaClient` public in order to be able to use those here. 

After the refactoring explained in the PR description, the cli itself should not need to talk to gitea or kube. It will only talk to the carrier server. The carrier server will create its own instances of the needed clients.

Right now CarrierClient has 2 responsibilities. It's both the cli implementation (logs to stdout etc) and the thing that gathers the information from gitea, kube etc. After refactoring there will be the cli, talking only to the carrier server and presenting the results in stdout and the information gathering will be done on the server side (probably wrapped in some new structs/classes).

jimmykarily:
image

image

jimmykarily:
I open this PR for review now because the next step would be to refactor the cli to:

- start the server on a local port
- make a request to the server
- present the results
- stop the server

This requires some heavy refactoring of the `application` and `service` packages along with changes on the CarrierClient (which will no longer need to talk to either kubernetes or gitea).

Doing the above in this PR will make it impossible to review.

internal/web/info_controller.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/suse/carrier/internal/cli/clients"
+	"github.com/suse/carrier/version"
+)
+
+type InfoController struct {
+}
+
+func (hc InfoController) Index(w http.ResponseWriter, r *http.Request) {
+	// TODO: Second return value is always nil (the cleanup function)?
+	client, _, err := clients.NewCarrierClient(nil)
```
rohitsakala, line 15: I think this logic of finding information about version, platform should be in the internal/api/v1/info.go just like how logic is in internal/api/v1/applications.go for fetching applications info. Also, I think internal/api/v1/version.go should return the version info.

internal/api/v1/applications.go
rohitsakala: Ok. Gotcha 👍🏽

internal/web/applications_controller.go
```diff
@@ -0,0 +1,76 @@
...
+		}
+	}
+
+	if handleError(w, err, 500) {
+		return
+	}
+	w.WriteHeader(http.StatusOK)
+
+	// Inject the request in the data. It can be used in the templates to find the
+	// current url
```
rohitsakala, line 61: I cannot find where the current_url is being used in the templates.

internal/web/applications_controller.go
```diff
@@ -0,0 +1,76 @@
...
+	"path"
+
+	"github.com/suse/carrier/internal/filesystem"
+)
+
+type ApplicationsController struct {
+}
+
+func (hc ApplicationsController) Index(w http.ResponseWriter, r *http.Request) {
+	Render([]string{"main_layout", "icons", "applications_index"}, w, r, map[string]interface{}{"serverUrl": r.Host})
```
rohitsakala, line 17: I cannot understand how r.Host information is used ?

rohitsakala (dismissed):
LGTM. My review can be addressed in the next PR. Approving it.

Makefile
```diff
@@ -83,6 +84,8 @@ update_tekton:
 
 embed_files:
 	statik -m -f -src=./embedded-files
+	statik -m -f -src=./embedded-web-files/views -ns webViews -p statikWebViews
```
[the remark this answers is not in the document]
jimmykarily: It depends a bit on how we plan to run the Carrier services inside the cluster. Currently the same server that serves the API endpoints also serves the HTML/CSS/Javascript files. When we move those inside the cluster maybe those 2 can be 2 different services and can be split in 2 binaries. Not sure if that's better. Let's discuss it then.

embedded-web-files/views/icons.html
```diff
@@ -0,0 +1 @@
+[[define "icon-plus-rectangle"]]<svg xmlns="http://www.w3.org/2000/svg" viewBox="-2 -2 24 24" width="24" height="24" preserveAspectRatio="xMinYMin" class="icon__icon"><path d="M10 20C4.477 20 0 15.523 0 10S4.477 0 10 0s10 4.477 10 10-4.477 10-10 10zm0-2a8 8 0 1 0 0-16 8 8 0 0 0 0 16zm1-7v4a1 1 0 0 1-2 0v-4H5a1 1 0 0 1 0-2h4V5a1 1 0 1 1 2 0v4h4a1 1 0 0 1 0 2h-4z"></path></svg>[[end]]
```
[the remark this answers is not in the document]
jimmykarily: :D it came with copy-paste from another project of mine. I kept it around in case I wanted a `plus` icon for a "New application" button. I didn't make it that far though. I kept the file around as a placeholder for icons.

internal/web/applications_controller.go
jimmykarily: `serverUrl` is a left-over. I though `r.Host` would be what I would need to compare to the various link  `href`s but it turns out it was `r.Url`: https://github.com/SUSE/carrier/pull/235/files#diff-ecc48de753581438ba869ceff1df561c7fcf30b3baf38d6e19b20a658e11f6a8R73

I will remove it.

internal/web/applications_controller.go
jimmykarily: https://github.com/SUSE/carrier/pull/235/files#diff-ecc48de753581438ba869ceff1df561c7fcf30b3baf38d6e19b20a658e11f6a8R73

I thought about Rails and how the request information is available inside the views. We could simply inject just the r.URL.Path but I thought other attributes of the request may be useful in the future.

internal/web/info_controller.go
jimmykarily: True, I just got lazy and avoided vuejs just for the info page. But you are right, we should be consistent. Do it in a new PR?

acceptance/utilities_test.go
```diff
@@ -17,6 +21,10 @@ import (
 // I.e. create/delete applications/services, bind/unbind services, etc.
 // This is done in the hope of enhancing the readability of various before/after blocks.
 
+func newOrgName() string {
+	return "apps-" + strconv.Itoa(int(time.Now().Nanosecond()))
```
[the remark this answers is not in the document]
jimmykarily: changing.

acceptance/utilities_test.go
```diff
@@ -25,6 +33,33 @@ func newServiceName() string {
 	return "service-" + strconv.Itoa(int(time.Now().Nanosecond()))
 }
 
+// func Curl is used to make requests against a server
+func Curl(uri string, requestBody *strings.Reader) (*http.Response, error) {
```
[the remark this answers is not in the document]
jimmykarily: Hm I think it's more likely that this function is changed to also make `POST` or other requests as soon as we have such endpoints. It's just that we only have 2 GET endpoints for now.

embedded-web-files/views/main_layout.html
```diff
@@ -0,0 +1,118 @@
...
+    <!-- <link rel="icon" href="/assets/favicons/favicon-32x32.png" sizes="32x32" type="image/png"> -->
+    <!-- <link rel="icon" href="/assets/favicons/favicon-16x16.png" sizes="16x16" type="image/png"> -->
+    <!-- <link rel="manifest" href="/assets/favicons/manifest.json"> -->
+    <!-- <link rel="mask-icon" href="/assets/favicons/safari-pinned-tab.svg" color="#563d7c"> -->
+    <!-- <link rel="icon" href="/assets/favicons/favicon.ico"> -->
+    <!-- <meta name="msapplication-config" content="/assets/favicons/browserconfig.xml"> -->
+    <!-- <meta name="theme-color" content="#563d7c"> -->
+
+
+    <style>
```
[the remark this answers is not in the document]
jimmykarily: Done: https://github.com/SUSE/carrier/pull/235/commits/3df061d3ddee817b8e63a9c91d44a746a8893329#diff-f5efcccc4dedeb3995b640254c6035044f83ed6f8398a73c5325e06ae568ed74

andreas-kupries (approved):
+1

PR #236, opened 2021-03-24
Refactor packages

jimmykarily:
This PR hopefully simplifies our package names and imports with the following logic:

- anything that is Carrier specific and shouldn't be used by external projects is moved to `internal`  ([docs](https://golang.org/doc/go1.4#internalpackages))
- packages that are re-usable are left at the top level.
- `paas` namespace is removed. Everything around carrier is "paas", no need to namespace some packages.
- client package is now named "cli" because that's what it is, the cli interface of Carrier. we will soon introduce a web interface.
- ui is renamed to termui because we are going to have a web ui soon.

Although there are many files changed in this PR, it's mostly files moved around and the import statements changed accordingly. Reviewing commit by commit may help.


Makefile
```diff
@@ -31,7 +31,7 @@ compress:
 	upx --brute -1 ./dist/carrier-darwin-amd64
 
 test: lint
-	ginkgo ./cmd/internal/client/ ./tools/ ./helpers/ ./kubernetes/
+	ginkgo ./cli/internal/client/ ./tools/ ./helpers/ ./kubernetes/
```
andreas-kupries, line 34: This looks to be out of date. The full set of directories inside of which I found `*_test.go` (*) files is

  - acceptance
  - helpers
  - internal/cli
  - internal/services
  - kubernetes

(*) Command used: `find . -type f -name '*_test.go'|xargs -n1 dirname | uniq | sort`

andreas-kupries (changes requested):
Test nit. Will do the necessary change and commit

Makefile
andreas-kupries: Done with commit 73aa2f0b7413d271dc64a99759e9ea1d9f3909c2

andreas-kupries (approved):
+1 now

PR #251, opened 2021-04-05
fix: Don't allow triggering of the staging pipeline without a password

jimmykarily:
As part of the research for mTLS, we discovered that it's possible to
trigger the staging of an application from another application's
container. More here:

https://github.com/SUSE/carrier/issues/130#issuecomment-813314112

This PR closes this hole by using a CEL interceptor
(https://github.com/tektoncd/triggers/blob/v0.11.1/docs/eventlisteners.md#cel-interceptors).

We still don't verify who the caller is, like mTLS would do but it's
better than nothing since we control who has access to that secret.


embedded-files/tekton/triggers.yaml
```diff
@@ -77,6 +77,14 @@ spec:
       - ref: staging-pipelinebinding
       template:
         ref: staging-triggertemplate
+      interceptors:
+        - cel:
```
rohitsakala, line 81: I think we should use a secret here instead of exposing the value? like explained here [0]

[0] https://tekton.dev/docs/triggers/eventlisteners/#github-interceptors

jimmykarily: Yes though your link is only for the github interceptor. For `cel` I think we need something like `compareSecret` from here: https://tekton.dev/docs/triggers/cel_expressions/

I didn't try that now because the password we use is already hardcoded here: https://github.com/SUSE/carrier/blob/main/internal/cli/clients/client.go#L1134 . 

A proper solution needs to handle secret generation/rotation and access (who can read it etc). This was a quick "fix" to make sure we don't forget to lock the triggering with a password. Right now, although we used a (hardcoded) secret when sending the request from Gitea, it was completely ignored on the Tekton side. I will create a new issue for the proper fix. Is that ok?

jimmykarily: Maybe I can create a test for this fix at least.

rohitsakala: Ok, No problem. I see a `TODO:` . That should do.

rohitsakala (approved):

embedded-files/tekton/triggers.yaml
jimmykarily: Created a new issue: https://github.com/SUSE/carrier/issues/255

PR #258, opened 2021-04-06
fix: use networking.k8s.io/v1 Ingress instead of extensions/v1beta1

### LINKED ISSUE #172: extensions/v1beta1 Ingress is deprecated
When pushing or listing apps w/ e.g. kubernetes-v1.20.4, carrier output includes warnings for

`W0225 11:53:19.188157   11295 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress`

So if we don't intend to support kubernetes older than v1.14, it seems we could go for `networking.k8s.io/v1` already.
I only see `c.Kubectl.ExtensionsV1beta1().Ingresses(<name>)` in "deployments/tekton.go" and "kubernetes/cluster.go".

kkaempf:
Removes warnings like
> W0402 10:55:31.309809   17207 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress

Fixes #172

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -80,10 +80,10 @@ update_google_service_broker:
 	@./scripts/update-google-service-broker.sh
 
 update_tekton:
-	mkdir -p assets/embedded-files/tekton
-	wget https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.19.0/release.yaml -O assets/embedded-files/tekton/pipeline-v0.19.0.yaml
-	wget https://storage.googleapis.com/tekton-releases/triggers/previous/v0.11.1/release.yaml -O assets/embedded-files/tekton/triggers-v0.11.1.yaml
-	wget https://github.com/tektoncd/dashboard/releases/download/v0.11.1/tekton-dashboard-release.yaml -O assets/embedded-files/tekton/dashboard-v0.11.1.yaml
+	mkdir -p embedded-files/tekton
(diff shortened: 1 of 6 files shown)


rohitsakala (changes requested):
Are we not caring about kube clusters <1.18 now ? [0]

[0] https://github.com/SUSE/carrier/blob/main/kubernetes/cluster.go#L513

kkaempf:
> Are we not caring about kube clusters <1.18 now ?

I wouldn't care about <1.18, see [here](https://github.com/SUSE/carrier/issues/172#issuecomment-814815835) for the arguments.

kkaempf:
Hmm, 1.18.17 also has problems with `networking/v1` :thinking:

manno:
> networks.k8s.io/v1 was introduced in Kubernetes v1.15+, the Ingress resource was promoted to v1 only in Kubernetes v1.19+

So, we need to bump the requirement to 1.19 to use v1.


However, with the way versioning works in k8s (stored version, served version), you can use any version to read it: 

> When you create an ingress object, it can be read via any version (the server handles converting into the requested version).  (https://github.com/kubernetes/kubernetes/issues/94761#issuecomment-691982480)

jimmykarily:
The end goal is for the epinio cli to only talk to the epinio server and not to the Kube api directly. That means, any warnings will only appear on the epinio server pod and the user won't see them. I'm closing this.

jimmykarily:
For as long as we still support <=1.18 : https://github.com/epinio/epinio/pull/354

Issue #253, opened 2021-04-06
Implement `services` command through the api [read endpoints]


jimmykarily:
## Acceptance Criteria

The following commands go through the api:

- list          Lists all services
- list-classes  Lists the available service classes
- list-plans    Lists all plans provided by the named service class
- show          Service information

Issue #239, opened 2021-03-29
Implement `org` cli command through the api


jimmykarily:
# Acceptance Criteria

- command `create-org` becomes `orgs create`
- `orgs create` -> `POST api/v1/orgs`
- `orgs` -> `GET api/v1/orgs`

PR #260, opened 2021-04-08
252 deploy carrier server

### LINKED ISSUE #252: Move carrier server inside the cluster
While we refactor the various commands to talk to the Carrier server, it make sense to have the Carrier server running inside the cluster as a deployment instead of starting it locally.

This simplifies the tests (they don't have to start and stop external processes) and is closer to the final result. As soon as all commands are migrated, there is nothing left to do (compared to the local server solution we are using now).

During development, it still makes sense to manually start a local server instead of generating a new image, pushing it and deploying Carrier again. We should make sure this development workflow, with the local server, is still possible.

## Acceptance Criteria

1. `carrier install` creates a deployment running the Carrier server
1. There is a pipeline that builds and pushes the carrier server image to a public registry.
1. The release pipeline tags the image with the version of the release.
1. All migrated commands talk to the server running inside the cluster instead of starting a local one.
1. The carrier server is accessible through an ingress and its url is discovered automatically if not set explicitly through the command line.
1. [Optional] The first time we discover the url, we write it to the carrier config so we don't have to do that again every time. Can be moved to a separate issue to keep the scope small.

jimmykarily:
It partly fixes #252 . We still need to change all the implemented cli commands to use the deployed server instead of starting their own. We will do it in another PR to avoid stepping on each other with on stories #239 and #253


.github/workflows/release.yml
```diff
@@ -27,6 +27,10 @@ jobs:
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         run: |
           make build-all-small
+      - name: Build Carrier images
+        if: steps.branch.outputs.BRANCH_NAME == 'main'
+        run: |
+          make build-images
```
jimmykarily, line 33: We need to check if we are logged in to Docker at this point

jimmykarily:
Local run:

```[2] Ran 19 of 19 Specs in 2526.299 seconds
[2] SUCCESS! -- 19 Passed | 0 Failed | 1 Flaked | 0 Pending | 0 Skipped
[2] PASS
[1] Deleting tmpdir on node 1
[1] 
[1] Ran 16 of 16 Specs in 2577.045 seconds
[1] SUCCESS! -- 16 Passed | 0 Failed | 0 Flaked | 0 Pending | 0 Skipped
[1] PASS

```
CI is probably being rate limited by dockerhub

jimmykarily:
There is one change I keep locally to make the patching work:

```
-       GOARCH="amd64" GOOS="linux" go build -race -ldflags '$(LDFLAGS)' -o dist/carrier-linux-amd64
+       GOARCH="amd64" GOOS="linux" CGO_ENABLED=0 go build -ldflags '$(LDFLAGS)' -o dist/carrier-linux-amd64
```
the reason is that my host is tumbleweed and the image is leap so the librabries don't match. If I compile carrier statically then it works (that's what the above change does). We could change the static compilation in general but we have to live without the `-race` flag that requires `CGO_ENABLED` to be true.

.github/workflows/release.yml
jimmykarily: Fixed  in https://github.com/SUSE/carrier/pull/260/commits/042e13441ce26224ac00de2f0f083377a76d4ec7

rohitsakala (approved):

PR #272, opened 2021-04-12
Add initial swagger spec generation for API docs

manno:
Screenshot 2021-04-12 at 16-50-40 API documentation

This adds swagger comments into a separate doc package, so they don't
interfere with existing comments.  Models are scanned automatically, but
routes are not and need to be added manually.

https://goswagger.io/use/spec.html


manno:
Swagger specs have to be manually created. Linking to the [pkg reference docs](https://pkg.go.dev/github.com/epinio/epinio/internal/api/v1) for now

PR #282, opened 2021-04-13
Convert mutating service commands to an API

### LINKED ISSUE #254: Implement `services` command through the api [mutating endpoints]
## Acceptance Criteria

The following commands go through the api:

- bind          Bind a service to an application
- create        Create a service
- create-custom Create a custom service
- delete        Delete a service
- unbind        Unbind service from an application

andreas-kupries:
Fixes #254

### DIFF
--- a/acceptance/api_v1_services_mutation_test.go
+++ b/acceptance/api_v1_services_mutation_test.go
@@ -126,7 +126,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("Organization 'bogus' does not exist\n"))
 		})
 
-		Context("On conflict", func() {
+		Context("with conflicting service", func() {
 			var service string
 
 			BeforeEach(func() {
@@ -138,7 +138,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 				cleanupService(service)
 			})
 
-			It("returns a 'conflict' when the service already exists", func() {
+			It("returns a 'conflict'", func() {
 				response, err := Curl("POST",
 					fmt.Sprintf("%s/api/v1/orgs/%s/services",
 						serverURL, org),
@@ -199,7 +199,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("Service plan 'meh' does not exist for class 'mariadb'\n"))
 		})
 
-		Context("Creation", func() {
+		Describe("Creation", func() {
 			AfterEach(func() {
 				cleanupService(service)
 			})
@@ -333,7 +333,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("Organization 'bogus' does not exist\n"))
 		})
 
-		Context("On conflict", func() {
+		Context("with conflicting service", func() {
 			var service string
 
 			BeforeEach(func() {
@@ -345,7 +345,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 				cleanupService(service)
 			})
 
-			It("returns a 'conflict' when the service already exists", func() {
+			It("returns a 'conflict'", func() {
 				response, err := Curl("POST",
 					fmt.Sprintf("%s/api/v1/orgs/%s/custom-services",
 						serverURL, org),
@@ -364,7 +364,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			})
 		})
 
-		Context("Creation", func() {
+		Describe("Creation", func() {
 			var service string
 
 			BeforeEach(func() {
@@ -461,7 +461,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("service 'bogus' not found\n"))
 		})
 
-		Context("With bound applications", func() {
+		Context("with bound applications", func() {
 			var app string
 			var service string
 
@@ -509,7 +509,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			})
 		})
 
-		Context("Without bound applications", func() {
+		Context("without bound applications", func() {
 			var service string
 
 			BeforeEach(func() {
@@ -599,7 +599,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("Organization 'bogus' does not exist\n"))
 		})
 
-		Context("Needs application", func() {
+		Context("with application", func() {
 			var app string
 			var service string
 
@@ -630,12 +630,12 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 				Expect(string(bodyBytes)).To(Equal("service 'bogus' not found\n"))
 			})
 
-			Context("Already bound", func() {
+			Context("and already bound", func() {
 				BeforeEach(func() {
 					bindAppService(app, service, org)
 				})
 
-				It("returns a 'conflict' when the service is already bound", func() {
+				It("returns a 'conflict'", func() {
 					response, err := Curl("POST",
 						fmt.Sprintf("%s/api/v1/orgs/%s/applications/%s/services",
 							serverURL, org, app),
@@ -707,7 +707,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			Expect(string(bodyBytes)).To(Equal("application 'bogus' not found\n"))
 		})
 
-		Context("Needs application", func() {
+		Context("with application", func() {
 			var app string
 
 			BeforeEach(func() {
@@ -734,7 +734,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 				Expect(string(bodyBytes)).To(Equal("service 'bogus' not found\n"))
 			})
 
-			Context("Needs service", func() {
+			Context("with service", func() {
 				var service string
 
 				BeforeEach(func() {
@@ -746,7 +746,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 					cleanupService(service)
 				})
 
-				Context("Bound service", func() {
+				Context("already bound", func() {
 					BeforeEach(func() {
 						bindAppService(app, service, org)
 					})
--- a/acceptance/api_v1_services_mutation_test.go
+++ b/acceptance/api_v1_services_mutation_test.go
@@ -617,7 +617,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			bodyBytes, err := ioutil.ReadAll(response.Body)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(response.StatusCode).To(Equal(http.StatusNotFound), string(bodyBytes))
-			Expect(string(bodyBytes)).To(Equal("application 'bogus' does not exist\n"))
+			Expect(string(bodyBytes)).To(Equal("application 'bogus' not found\n"))
 		})
 
 		Context("with application", func() {
--- a/acceptance/api_v1_services_mutation_test.go
+++ b/acceptance/api_v1_services_mutation_test.go
@@ -53,7 +53,7 @@ var _ = Describe("Services API Application Endpoints, Mutations", func() {
 			bodyBytes, err := ioutil.ReadAll(response.Body)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(response.StatusCode).To(Equal(http.StatusBadRequest), string(bodyBytes))
-			Expect(string(bodyBytes)).To(Equal("json: cannot unmarshal array into Go value of type v1.CatalogCreateRequest\n"))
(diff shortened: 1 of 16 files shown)


acceptance/api_v1_services_mutation_test.go
```diff
@@ -0,0 +1,787 @@
...
+					    "waitforprovision": true
+					}`, service)))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response).ToNot(BeNil())
+
+				defer response.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(response.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response.StatusCode).To(Equal(http.StatusCreated), string(bodyBytes))
+				Expect(string(bodyBytes)).To(Equal(""))
```
jimmykarily, line 224: the expectations are similar to the test below. Maybe we should check that the service is there indeed? On the other hand I can't think of a simple way to make this kind of test reliable, it will be "racy" unless we find some way to control (hold and release) the provisioning process on the server.

acceptance/api_v1_services_mutation_test.go
```diff
@@ -0,0 +1,787 @@
...
+			AfterEach(func() {
+				cleanupApp(app)
+				cleanupService(service)
+			})
+
+			It("returns 'bad request'", func() {
+				response, err := Curl("DELETE",
+					fmt.Sprintf("%s/api/v1/orgs/%s/services/%s",
+						serverURL, org, service),
+					strings.NewReader(`{ "unbind": false }`))
```
jimmykarily, line 485: Can we make `"unbind": false` the default? Maybe we can also allow an empty body too. E.g. if body is an empty string we don't even try to parse it. If it's not empty then it should be JSON.

acceptance/api_v1_services_mutation_test.go
```diff
@@ -0,0 +1,787 @@
...
+				strings.NewReader(`{ "name": "meh" }`))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(response).ToNot(BeNil())
+
+			defer response.Body.Close()
+			bodyBytes, err := ioutil.ReadAll(response.Body)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(response.StatusCode).To(Equal(http.StatusNotFound), string(bodyBytes))
+			Expect(string(bodyBytes)).To(Equal("Organization 'bogus' does not exist\n"))
+		})
```
jimmykarily, line 600: a test is missing to check what happens when the org exists but the application doesn't.

acceptance/api_v1_services_mutation_test.go
```diff
@@ -0,0 +1,787 @@
...
+			Expect(response).ToNot(BeNil())
+
+			defer response.Body.Close()
+			bodyBytes, err := ioutil.ReadAll(response.Body)
+			Expect(err).ToNot(HaveOccurred())
+			Expect(response.StatusCode).To(Equal(http.StatusNotFound), string(bodyBytes))
+			Expect(string(bodyBytes)).To(Equal("application 'bogus' not found\n"))
+		})
+
+		Context("Needs application", func() {
```
jimmykarily, line 710: maybe `With("existing application", func() {` reads better?

jimmykarily: same for the rest of the `Needs...` occurrences in this file

internal/api/v1/services.go
```diff
@@ -107,6 +110,7 @@ func (sc ServicesController) Index(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	appsOf, err := servicesToApps(cluster, gitea, org)
+
```
jimmykarily, line 113: [nit] you probably wanted to add the empty line above `appsOf` not below.

andreas-kupries: Agreed.

acceptance/api_v1_services_mutation_test.go
andreas-kupries: I should add checks for service presence to both indeed. Should be able to take that from the cli service creation tests.

acceptance/api_v1_services_mutation_test.go
andreas-kupries: Hm. It feels like a false optimization to me at the moment. What we have here is trivial to document, and trivial to implement cli side. Adding the exemptions makes the docs more complex at least. We could keep the cli side unchanged, i.e. simple, but then the exemption is just ballast server side.

In short, I do not see the value here, only the problems.

So, coming down on the no side here.

acceptance/api_v1_services_mutation_test.go
andreas-kupries: Will fix.

internal/api/v1/services.go
```diff
@@ -136,6 +140,284 @@ func (sc ServicesController) Index(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	cluster, err := kubernetes.GetCluster()
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	// Verify that the requested name is not yet used by a different service.
+	_, err = services.Lookup(cluster, org, createRequest.Name)
+	if err == nil {
```
jimmykarily, line 193: the opposite is not always safe to ignore. I mean, if there is an error, it doesn't mean the error is related to the service not being there, it may be some irrelevant error which shouldn't be ignored.

You need to check what kind of error this was.

acceptance/api_v1_services_mutation_test.go
andreas-kupries: Can you point me to the documentation for `With()` blocks ? I am unable to find any.
I see only `Describe()` and `Context()`. I guess I could use `with` in the block descriptions, where pertinent.

internal/api/v1/router.go
```diff
@@ -10,16 +10,28 @@ import (
...
 	router := httprouter.New()
 	router.HandlerFunc("GET", "/api/v1/info", InfoController{}.Info)
+
+	// List, show, and delete applications
 	router.HandlerFunc("GET", "/api/v1/orgs/:org/applications", ApplicationsController{}.Index)
 	router.HandlerFunc("GET", "/api/v1/orgs/:org/applications/:app", ApplicationsController{}.Show)
 	router.HandlerFunc("DELETE", "/api/v1/orgs/:org/applications/:app", ApplicationsController{}.Delete)
 
+	// Bind and unbind services to/from applications
+	router.HandlerFunc("POST", "/api/v1/orgs/:org/applications/:app/services", ApplicationsController{}.Bind)
```
jimmykarily, line 20: I'm not sure but, strictly speaking, the resource we are creating is neither a Service not an Application, it's the thing in between (ServiceBinding?) Thus the url should be something like:

```
POST /api/v1/orgs/:org/applications/:app/servicebindings
```
the controller would be `ServiceBindingsController` and the action would be `Create`. The end result is the same but we keep the API resource-based and "semantic". A POST to an endpoint that ends with `services` should generally create a Service resource.
What do you think?

internal/api/v1/services.go
```diff
@@ -136,6 +140,284 @@ func (sc ServicesController) Index(w http.ResponseWriter, r *http.Request) {
...
```
jimmykarily, line 268: same as above, need to check what the error is

acceptance/api_v1_services_mutation_test.go
jimmykarily: ok agreed.

acceptance/api_v1_services_mutation_test.go
jimmykarily: ok try that but be careful of race conditions. Eventually the service will be there and depending on how long that takes to happen, you may get different results when you check.

internal/api/v1/router.go
andreas-kupries: While I can see changing the name to `service bindings`, I am less source about using a separate controller.
I am modeling it as a sub-collection resource in the application
See https://restfulapi.net/resource-naming .
In this sense I am creating an element in the sub-collection of the application resource.
Or maybe I misunderstood that page.

@manno Can you weigh in ?

acceptance/api_v1_services_mutation_test.go
jimmykarily: Turns out `With` only existed in my mind ( :brain: ), sorry for the confusion. Maybe `Context("when an application exists")` or `When("application exists")`, I don't know. "Needs application" immediately brings the question "who needs it?", it took me a while to understand it's the test that needs it (is it?).

acceptance/api_v1_services_mutation_test.go
andreas-kupries: I am not sure where you see the race. This test has `waitforprovision: true` in the request. The curl will literally not return before the service exists, because the server waits for it to exist. The next test has it `wfp:false`, so the server returns immediately, and thus the test has to wait for the service to exist (using `Eventually`) before it can continue.

internal/api/v1/services.go
```diff
@@ -136,6 +140,284 @@ func (sc ServicesController) Index(w http.ResponseWriter, r *http.Request) {
...
+			http.StatusConflict)
+		return
+	}
+
+	// Verify that the requested class is supported
+	serviceClass, err := services.ClassLookup(cluster, createRequest.Class)
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+	if serviceClass == nil {
```
jimmykarily, line 279: [nit] This is inconsistent. Other lookup methods return an error when the resource is not found, this one returns nil. I prefer the `nil` implementation to be honest. We can change this later, feel free to keep it as is here.

andreas-kupries: Yes. Our low-level lookups are inconsistent how `X not found` is reported.
Some use an error, others no error and object is nil.

acceptance/api_v1_services_mutation_test.go
andreas-kupries: Commit 45223beed41f7017fdc69428d33cdfdf3abc97a3

acceptance/api_v1_services_mutation_test.go
andreas-kupries: Commit 0b5fb5a2e8dbb8af0982f37b5ef03cf786d6129b

internal/api/v1/services.go
andreas-kupries: Commit 17f989c20b754a2dac2b7be2a629e7b9c699079d

internal/cli/server.go
```diff
@@ -80,18 +82,41 @@ func startEpinioServer(wg *sync.WaitGroup, port int, ui *termui.UI) (*http.Serve
...
+
+	// Recreate body for the actual handler
+	r.Body = ioutil.NopCloser(bytes.NewBuffer(bodyBytes))
+
+	if len(bodyBytes) == 0 {
+		msg.Msg("RequestBody: n/a")
+		return
+	}
+
+	msg.Msgf("RequestBody\n%s\n/RequestBody", string(bodyBytes))
```
jimmykarily, line 121: In some cases the request body could have sensitive data in it. For example when creating custom services with credentials to external databases and such. Logging to stdout maybe result in those credentials being written in more places than needed (e.g. local filesystem  unencrypted).

Among other problems, people may need to send out the server logs for debugging when they have issues. It's easy to leak credentials without them noticing. If you decide to keep it as is in this PR, please create an issue to find a solution to this problem. Maybe we can disable logging the request body in some actions, when we expect sensitive data as part of the request body.

acceptance/api_v1_services_mutation_test.go
jimmykarily: According to the test description ("creates the catalog service and waits for it to be provisioned"), this test also checks that waiting is indeed happening. Let's say we change the server code in the future and somehow we break the part that waits for the service to be created and instead it always returns immediately. This test wouldn't fail in that case. Even if you tested that the service was created in the test, you won't know if this happened before or after your request was completed.

internal/api/v1/services.go
andreas-kupries: Checking...

internal/api/v1/services.go
andreas-kupries: Checking...

internal/api/v1/services.go
andreas-kupries: Done. Commit 3477a08cded8b3d70df5e707d00bb0d461afa91f

internal/cli/server.go
andreas-kupries: Note that these messages are limited to `--verbosity 1`. By default the server runs at level 0.
IOW the standard install will not log anything at all.
I see this as something for dev only at the moment.
Where I can bump the verbosity level in the server deployment code.

jimmykarily: increasing verbosity would be the first thing we would ask a user to do if they were trying to report an issue. We don't want people to accidentally leak their credentials during some debugging. It's ok for now but we need to think about a safer solution. I still think this deserves an issue in the backlog.

jimmykarily (approved):

internal/cli/server.go
andreas-kupries: Will make an issue for it.

PR #287, opened 2021-04-14
Change service commands to use api 

### LINKED ISSUE #253: Implement `services` command through the api [read endpoints]

rohitsakala:
fixes #253


internal/api/v1/serviceclasses.go
```diff
@@ -0,0 +1,102 @@
...
+		return
+	}
+	w.Header().Set("Content-Type", "application/json")
+	_, err = w.Write(js)
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+}
+
+func (scc ServiceClassesController) ServicePlanList(w http.ResponseWriter, r *http.Request) {
```
jimmykarily, line 55: ServicePlans are another resources. It would be better if they had their own controller. Then the action would be:

```
func (sp ServicePlansController) List(...)
```

internal/api/v1/services.go
```diff
@@ -0,0 +1,124 @@
...
+	"github.com/epinio/epinio/internal/cli/clients"
+	"github.com/epinio/epinio/internal/services"
+	"github.com/epinio/epinio/kubernetes"
+	"github.com/julienschmidt/httprouter"
+)
+
+type ServicesController struct {
+}
+
+func (sc ServicesController) ServiceInfo(w http.ResponseWriter, r *http.Request) {
```
jimmykarily, line 17: shouldn't this be the `show` action? Also the route shouldn't need the `/info` part

rohitsakala: Fixed it hopefully.

internal/api/v1/serviceclasses.go
rohitsakala: Fixed it.

rohitsakala:
I forgot to write tests. I will make a new PR for that.

rohitsakala:
Tests passed locally

rohitsakala:
Tests passed locally.
Screenshot 2021-04-16 at 12 40 28

acceptance/api_v1_serviceplans_test.go
```diff
@@ -0,0 +1,76 @@
...
+					servicePlanDescs = append(servicePlanDescs, servicePlan.Description)
+					servicePlanFrees = append(servicePlanFrees, servicePlan.Free)
+				}
+				Expect(servicePlanNames).Should(ContainElements("10-1-26", "10-1-28", "10-3-20", "10-3-16", "10-3-17"))
+				Expect(servicePlanDescs).Should(ContainElements("Fast, reliable, scalable, and easy to use open-source relational database system. MariaDB Server is intended for mission-critical, heavy-load production systems as well as for embedding into mass-deployed software. Highly available MariaDB cluster."))
+				Expect(servicePlanFrees).Should(ContainElements(true, true))
+			})
+
+			It("returns a 404 when the org does not exist", func() {
+				response, err := Curl("GET", fmt.Sprintf("%s/api/v1/orgs/:org/idontexist/:serviceclass/serviceplans", serverURL), strings.NewReader(""))
```
manno, line 65: Doesn't that URI need to match the URL pattern from the router?

```suggestion
				response, err := Curl("GET", fmt.Sprintf("%s/api/v1/orgs/idontexist/serviceclasses/someclass/serviceplans", serverURL), strings.NewReader(""))
```

Nit: not a big fan of the existing `Curl` func, it's not really curl and we could get rid of a param by having `Get()`, `Post()` etc. 
Also while you can send a request body with a GET requests, it's not often (never) used. So that param is also useless. (https://stackoverflow.com/a/978094)

internal/api/v1/router.go
```diff
@@ -17,6 +17,12 @@ func Router() *httprouter.Router {
 	router.HandlerFunc("GET", "/api/v1/orgs", OrganizationsController{}.Index)
 	router.HandlerFunc("POST", "/api/v1/orgs", OrganizationsController{}.Create)
 
+	router.HandlerFunc("GET", "/api/v1/orgs/:org/services", ServicesController{}.Index)
+	router.HandlerFunc("GET", "/api/v1/orgs/:org/services/:service", ServicesController{}.Show)
+
+	router.HandlerFunc("GET", "/api/v1/orgs/:org/serviceclasses", ServiceClassesController{}.Index)
+	router.HandlerFunc("GET", "/api/v1/orgs/:org/serviceclasses/:serviceclass/serviceplans", ServicePlansController{}.Index)
```
andreas-kupries, line 24: Neither service classes nor plans are linked to orgs in any way.
IMHO the routes should be `/api/v1/serviceclasses` and `/api/v1/serviceclasses/:serviceclass/serviceplans`.

internal/api/v1/serviceclasses.go
```diff
@@ -0,0 +1,53 @@
...
+	exists, err := gitea.OrgExists(org)
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+	if !exists {
+		http.Error(w, fmt.Sprintf("Organization '%s' does not exist", org),
+			http.StatusNotFound)
+		return
+	}
+	serviceClasses, err := services.ListClasses(cluster)
```
andreas-kupries, line 39: Note here how listing the service classes does not make use of the org information at all.

internal/api/v1/serviceplans.go
```diff
@@ -0,0 +1,64 @@
...
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+	if !exists {
+		http.Error(w, fmt.Sprintf("Organization '%s' does not exist", org),
+			http.StatusNotFound)
+		return
+	}
+
+	serviceClass, err := services.ClassLookup(cluster, serviceClassName)
```
andreas-kupries, line 41: and again the org does not play any role at all in listing the plans of a class.

internal/api/v1/services.go
```diff
@@ -0,0 +1,157 @@
...
+		return
+	}
+	appsOf, err := servicesToApps(cluster, gitea, org)
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	responseData := map[string]interface{}{
+		"Services":    orgServices,
+		"ServiceApps": appsOf,
```
andreas-kupries, line 115: Interesting ... You are returning the list of services, and the bound apps as separate keys of an object ...

I would have used an array of objects, with the info about bounds app added to the service objects directly.

Hm. Another point here. While the marshalling and transmission is easy, trivial even, it also transmits everything the server knows, even stuff we may consider internal (Instance name for catalog services, and secret name for the custom services).

andreas-kupries: And looking at the client side, with all the casting, that makes me nervous.
I am coming to the opinion that a custom struct type should be used for the information, instead of the internal service structures. Something like

```
type struct ServiceListReponse {
    Name: string
    BoundApp: []string
}
```

While using that would require the server to convert from the internal form to this, the cli side unmarshalling becomes much simpler/easier, and no casting will be required at all for get the information to print.

andreas-kupries (changes requested):

internal/api/v1/services.go
manno: A custom response struct would also fit nicely with generated API docs (The wrappers can reference those structs https://github.com/epinio/epinio/pull/272/files#diff-e36e12b03db99680c15305336f9147bfac43be780989ad2833633a457839f4c8R36)

jimmykarily: I like the special struct idea. Adheres to the Presenter pattern and it separates the view logic.

internal/api/v1/router.go
rohitsakala: Done.

internal/api/v1/serviceclasses.go
rohitsakala: Done.

internal/api/v1/serviceplans.go
rohitsakala: Done.

internal/api/v1/services.go
rohitsakala: Done.

acceptance/api_v1_serviceplans_test.go
rohitsakala: That might be a good suggestion. I will create a new issue for this.

andreas-kupries (approved):

PR #292, opened 2021-04-15
fix: Use dockerhub credentials when they exist for registry mirror

jimmykarily:
because on CI we do have the credentials and dockerhub limits us base on
a (shared) IP address. This results in the mirror not being able to pull
images too soon.


jimmykarily:
I want to check what happens when you don't specify the env variables.

jimmykarily:
> I want to check what happens when you don't specify the env variables.

works as expected (unauthenticated)

manno (approved):
Nice, better to handle the errors in the caller.

PR #293, opened 2021-04-15
fix: binary path matching in acceptance tests

thardeck:
before the tests failed when more then one binary was available.


jimmykarily:
Please fix the golangci-lint comment. Other than that, it looks ok.

jimmykarily (approved):

manno (approved):
Sure, seems to match the Makefile

Issue #270, opened 2021-04-12
Cleanup repo structure


jimmykarily:
# Acceptance Criteria

- Remove non-needed files (e.g. scripts that download tools etc).
- Move files to where they belong (e.g. sample app could be under fixtures or assets etc)

jimmykarily:
done

PR #297, opened 2021-04-15
chore: Generate statik.go files

manno:
Since we keep gettings conflicts for those.

refs #270


manno:
Had to update and removed the 'embed_files' from test-acceptance target, since it's already called from acceptance_suite.go

.gitignore
```diff
@@ -16,3 +16,6 @@ personal-setup
 /resources
 acceptance/acceptance.test
 .vscode
+statik/statik.go
```
andreas-kupries, line 19: Why a file here, versus directories below ?

andreas-kupries (changes requested):
Just a question/nit.

.gitignore
manno: 👍

manno:
Talked with @jimmykarily and changed the getstatik to get a specific version of statik.

andreas-kupries (approved):

PR #299, opened 2021-04-16
feat: Uninstall k3d clusters after acceptance test run

rohitsakala:
Co-authored-by: Mario Manno <mario.manno@suse.com>

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -133,6 +133,9 @@ var _ = AfterSuite(func() {
 	err := uninstallCluster()
 	Expect(err).NotTo(HaveOccurred())
 
+	out, err = cleanupTmp()
+	Expect(err).NotTo(HaveOccurred(), out)
+
 	fmt.Printf("Deleting tmpdir on node %d\n", config.GinkgoConfig.ParallelNode)
 	deleteTmpDir()
 })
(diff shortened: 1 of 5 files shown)


rohitsakala:
TODO:

1. Shift builderpack asgartech image to dockerhub.
2. Report time difference.

acceptance/acceptance_suite_test.go
```diff
@@ -284,6 +294,13 @@ func uninstallEpinio() (string, error) {
 	return Epinio("uninstall", "")
 }
 
+// Remove all tmp directories from /tmp/epinio-* . Test should try to cleanup
+// after themselves but that sometimes doesn't happen, either because we forgot
+// the cleanup code or because the test failed before that happened.
+func cleanupTmp() (string, error) {
+	return RunProc("rm -rf /tmp/epinio-*", "", false)
```
jimmykarily, line 301: This would be a problem is someone was running 2 instances of the acceptance test suite in parallel on the same machine. As soon as the first instance would be done, it would delete all directories of  both instances. I don't think that is very likely to happen but it's good to keep in mind.

jimmykarily:
13 minutes to complete. Not very different from the last successful build without these changes: https://github.com/epinio/epinio/actions/runs/763282426 (12 minutes)

jimmykarily (dismissed):

acceptance/acceptance_suite_test.go
```diff
@@ -280,8 +295,11 @@ func installEpinio() (string, error) {
 	return Epinio("install --skip-default-org", "")
 }
 
-func uninstallEpinio() (string, error) {
-	return Epinio("uninstall", "")
+// Remove all tmp directories from /tmp/epinio-* . Test should try to cleanup
+// after themselves but that sometimes doesn't happen, either because we forgot
+// the cleanup code or because the test failed before that happened.
```
manno, line 300: Since this PR also fixes the leaking tests, we're not sure if we  need that extra clean up?
Is the "that sometimes doesn't happen" part still true?

I would add a warning:

> This code will create problems if more than one acceptance_suite_test.go is run in parallel (e.g. two PRs on one worker). However we keep it as an extra measure.

jimmykarily: added

PR #305, opened 2021-04-19
fix: Login to dockerhub before running tests

jimmykarily:
because if a registry mirror is not running, the acceptance tests will
try to start it. If the image is not yet there (like it happens on new
github runners), docker command will try to pull it. On self-hosted
runners which share an external IP with lots of other machines, this
could immediately hit the rate limiting of dockerhub.

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -39,6 +39,8 @@ const (
 )
 
 var _ = SynchronizedBeforeSuite(func() []byte {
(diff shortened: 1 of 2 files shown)


acceptance/acceptance_suite_test.go
```diff
@@ -39,6 +39,8 @@ const (
 )
 
 var _ = SynchronizedBeforeSuite(func() []byte {
+	fmt.Printf("I'm running on runner = %s\n", os.Getenv("HOSTNAME"))
```
kkaempf, line 42: :+1:

manno (approved):

PR #261, opened 2021-04-08
fix: drop deprecated apiextensions.k8s.io/v1beta1 CustomResourceDefinition

kkaempf:
It was used in Tekton and kubernetes/cluster.go

fixes #204

PR #318, opened 2021-04-23
fix: drop deprecated apiextensions.k8s.io/v1beta1 CRD

### LINKED ISSUE #20: carrier cli install - Should it roll back on failure ?
Tried to use the `cli install` in a k3d cluster I used for a bash-based carrier before. The bash carrier was deleted.
Result:
```
work@tagetarl:~/SUSE/dev/carrier> cli/dist/carrier install
Carrier installing...
[Shared] system_domain The domain you are planning to use for Carrier. Should be pointing to the traefik public IP () : 
🚢 Traefik already installed, skipping
🚢 Deploying Quarks
✔️  Quarks deployed
Couldn't install carrier: Namespace gitea present already
```

So, wondering if the `install` should roll back the pieces it has already done when it fails in some later stage.
Or, maybe have `uninstall` capable of cleaning up a partial installation ?

This all might be for post MVP.

jimmykarily:
It was used in Tekton and kubernetes/cluster.go

fixes #20

Supersedes #261  (rebased and non-forked so it has access to dockerhub secrets)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -81,9 +81,9 @@ update_google_service_broker:
 
 update_tekton:
 	mkdir -p assets/embedded-files/tekton
-	wget https://storage.googleapis.com/tekton-releases/pipeline/previous/v0.19.0/release.yaml -O assets/embedded-files/tekton/pipeline-v0.19.0.yaml
-	wget https://storage.googleapis.com/tekton-releases/triggers/previous/v0.11.1/release.yaml -O assets/embedded-files/tekton/triggers-v0.11.1.yaml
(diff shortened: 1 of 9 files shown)


jimmykarily:
this bump somehow made the uuid bigger (?) here: https://github.com/epinio/epinio/blob/main/assets/embedded-files/tekton/triggers.yaml#L21

because  now we can't use any app name bigger than 5 characters. See here why: https://github.com/epinio/epinio/pull/322

we can remove the app name completely. The UUID is 36 characters long (https://github.com/google/uuid/blob/master/uuid.go#L174) .

jimmykarily:
https://github.com/tektoncd/triggers/commit/1066d18ecb7b1e4ca88472251a2561e5c684046c#diff-ae88b5cfc4e3878426b81d01b3b8e7f4a40cb05f264986652af77a694c141cc5

thardeck (approved):

PR #323, opened 2021-04-23
Skip cleanup or sleep if files in tmp/ exist

manno:
```
echo 10 >  tmp/after_each_sleep
touch tmp/skip_cleanup
make test-acceptance
```


thardeck (changes requested):
We should add the repo tmp dir to `.gitignore`, besides that lgtm.

manno:
> We should add the repo tmp dir or the two files to to `.gitignore`, besides that lgtm.

It's already in there.

thardeck (approved):

PR #324, opened 2021-04-23
Use backoff when communicating with gitea

manno:


manno:
This was just to illustrate how we could use retry when calling out to external APIs.

PR #325, opened 2021-04-23
Simplify acceptance tests

### LINKED ISSUE #263: Split tests ...
... for parallelization, more focussed test areas, and speedup.

To be tackled post-mvp

jimmykarily:
fixes #263 though it's not splitting the tests but rather making splitting not necessary. This PR makes it so that we only need one cluster


jimmykarily:
Acceptance tests on this branch took [18m44sec](https://github.com/epinio/epinio/pull/325/checks?check_run_id=2425966198) using only one cluster and 10 ginkgo nodes while on `main` they took [18m30s](https://github.com/epinio/epinio/runs/2420998551) using 6 clusters and equal number of ginkgo nodes.

I expected the time to drop significantly on this branch, especially since we can increase the number of ginkgo nodes without increasing the number of clusters. It doesn't seem to be that easy after all but still one cluster takes less disk space on the runner, consumes less bandwidth, makes less requests to Dockerhub (see rate limiting), make test setup less complex and  makes it easier to inspect the test cluster (no need to figure out which cluster to connect to and then switch back and forth between clusters) and finally it tests Epinio in a more real world setup where people may deploy at the same time on the same cluster (though any failure that is caused by this will be hard to debug).

With these in mind, I would say that this PR is and improvement and we should consider merging it.

andreas-kupries (dismissed):

manno:
```
  -randomizeAllSpecs
        If set, ginkgo will randomize all specs together.  By default, ginkgo only randomizes the top level Describe, Context and When groups.
```

andreas-kupries:
I am trying the `randomizeAllSpecs` now.

@jimmykarily Will you (have time to) fix the conflict ?

andreas-kupries:
Being random it does not fully balance things. Somewhat however, in my single run with it. IMHO good enough to add.

andreas-kupries (dismissed):

jimmykarily:
> Being random it does not fully balance things. Somewhat however, in my single run with it. IMHO good enough to add.

Done here: https://github.com/epinio/epinio/pull/325/commits/2541b1e4f39fcc51ebdb1b79ea3f2c4f2f42bcc6

jimmykarily:
The new flag doesn't seem to improve distribution. We got from 5 test on one node to 19 on another. Anyway, keeping it.

PR #326, opened 2021-04-24
Parallel install

jimmykarily:
Installs Epinio in 3 minutes on a warmed up cluster:

```
real    3m8.877s
user    0m15.747s
sys     0m3.206s
```

May need some improvements on how we print messages from the various go routines.

### DIFF
--- a/deployments/cert_manager.go
+++ b/deployments/cert_manager.go
@@ -142,11 +142,15 @@ func (cm CertManager) apply(c *kubernetes.Cluster, ui *termui.UI, options kubern
 		return err
 	}
 
+	if err := c.CreateLabeledNamespace(CertManagerDeploymentID); err != nil {
+		return err
+	}
+
 	// Setup CertManager helm values
 	var helmArgs []string
 
 	helmArgs = append(helmArgs, `--set installCRDs=true`)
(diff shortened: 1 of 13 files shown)


jimmykarily:
A similar approach can be taken for `uninstall` although special care is needed to not remove things that act as finalizers too early thus preventing other objects from being deleted.

jimmykarily:
I'm amazed by how experiment sometimes contradicts the theory's expectations :D . I tried installing from `main` on a warmed up cluster and it was actually faster:

```
real    2m20.477s
user    0m14.101s
sys     0m2.304s
```

I don't know how to interpret this. Generally the time it takes is not deterministic but I expected some improvement when doing things in parallel. On the other hand, the difference may be noticeable on fresh cluster so I will try that too.

jimmykarily:
I tried again from this branch (with a warmed up cluster) and these are today's results:

```
real    0m48.312s
user    0m14.945s
sys     0m2.726s
```
definitely faster than the `main` branch. I'm using a VM which is running on a machine with shared resource so maybe generally today things are faster. I'd say now theory and experiment agree :).

jimmykarily:
Fresh cluster, parallel install:

```
real    2m39.405s
user    0m15.617s
sys     0m2.983s
```

jimmykarily:
Fresh cluster, main branch (non parallel install):

```
real    6m6.295s
user    0m14.434s
sys     0m2.916s
```

andreas-kupries (approved):

PR #338, opened 2021-04-29
Use Upload API for App Source

### LINKED ISSUE #328: Add upload API endpoint

manno:
Fixes #328 

This changes how apps are pushed. Up to now the CLI would 
* copy the app directory to a temporary directory,
* write deployment metadata files (app.yml) for tekton to the directory,
* finally issue a 'git init; ...; git push' to epinio's Gitea server.

The CLI would also talk to the k8s cluster to query and create additional resources.

This PR changes that flow. Now the CLI 
* creates a tarball from the app source directory,
* uploads that tarball to the API server,
* but the CLI still creates the prod/local certificates in k8s.

Creating the git repo from the source and adding the deployment metadata files is now handled by the API server.


internal/api/v1/upload.go
```diff
@@ -0,0 +1,94 @@
...
+
+	_, err = io.Copy(f, file)
+	if err != nil {
+		internalError(w, "failed to copy app sources to temp location")
+		return
+	}
+
+	log.V(2).Info("unpacking temp dir")
+	appDir := path.Join(tmpDir, "app")
+	err = archiver.Unarchive(blob, appDir)
```
jimmykarily, line 66: To prevent bad actors from performing a DoS attacks on Epinio we should do such things in a queue in the background. Obviously that's not possible right now, because our storage solution (gitea) is supposed to store the code after extraction but that's something to keep in mind when we investigate #347 . I'll add this comment on that story too.

andreas-kupries (approved):
LGTM.

internal/api/v1/upload.go
manno: This endpoint should not be accessible without authentication. I'm still hopeful, we can use a traefik middleware for that.

There is an option to validate the overall request size before reading to protect against uploads that are "too large". But I wasn't able to get it to work and I don't know what values (523mb?) to put. I think we should postpone that, I can create an issue so we don't forget though.

Regarding the unarchive, I'm not sure how expensive that is, the data is not compressed, so it's only IO and data in a tarball is ordered sequentially. Since this is a http upload, we need to move the data somewhere anyhow (read+insert/upload). The worker could then read the data again to write it to the final destination.

We could add a story to support "large" apps. I'm not sure what's considered large and how much resources that would effectively use.

PR #337, opened 2021-04-28
329 go native or go home

jimmykarily:
Fixes #329

PR #339, opened 2021-04-29
Fix logs by monitoring pods based on labels

jimmykarily:
This is a follow-up PR of this: https://github.com/epinio/epinio/pull/337#issuecomment-829083348

We broke the staging logs when we removed the application name from the staging pods. The above PR would break it anyway because the staging pods now run on a dedicated namespace (tekton-staging). This PR is fixing it by adding labels to the staging pods and then when collecting logs, we filter pods using those labels.

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
+	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/client-go/dynamic"
 )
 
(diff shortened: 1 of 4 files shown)


internal/cli/clients/client.go
```diff
@@ -1461,12 +1474,12 @@ func (c *EpinioClient) logs(appName, org string) (context.CancelFunc, error) {
...
 		Since:                 duration.LogHistory(),
-		AllNamespaces:         false,
-		LabelSelector:         labels.Everything(),
+		AllNamespaces:         true,
+		LabelSelector:         selector,
 		TailLines:             nil,
 		Template:              tailer.DefaultSingleNamespaceTemplate(),
-		Namespace:             org,
-		PodQuery:              regexp.MustCompile(fmt.Sprintf(".*-%s-.*", appName)),
+		Namespace:             "",
```
thardeck, line 1481: Can we not remove the Namespace entry because we scan all?

jimmykarily: no it's either all or one. It doesn't seem to accept a list of Namespaces

andreas-kupries (approved):

PR #340, opened 2021-04-29
Remove openssl dependency

### LINKED ISSUE #298: Remove dependency to openssl
In order to make the tekton pods trust our registry CA and push the generated image there, we need to place the ca in a directory inside the pod using a very specific name. The name of the file has to be similar to whatever this command generates:

```
openssl x509 -hash -noout
```
more here: https://github.com/epinio/epinio/blob/8fd8ba151c814670289831aa7c94c991a2001e6d/helpers/exec.go#L151-L157

While migrating from bash scripts to go code, we couldn't find an way to produce that SHA in go so we [shelled out to `openssl`](https://github.com/epinio/epinio/blob/8fd8ba151c814670289831aa7c94c991a2001e6d/helpers/exec.go#L164). This is a dependency that seems unreasonable and we should try to get rid of it.

I found a (random) link which explains a bit what's going on: https://docs.infor.com/ln/10.5/it-it/lnolh/help/tt/onlinemanual/https_soap_generate_hash.html

andreas-kupries:
Fix #298


thardeck (approved):
LGTM, tests might be helpful for the methods at the same time errors would directly appear I suppose.

thardeck:
`Note that the cert `registy-tls` is not removed. It is still used by some other parts of the system.`
Should we not rename the certificate then according to what function it is used by?

andreas-kupries:
> > Note that the cert `registy-tls` is not removed. It is still used by some other parts of the system.

> Should we not rename the certificate then according to what function it is used by?

I did not investigate what parts still use it. I only confirmed that removing it breaks deployment of the registry.

PR #343, opened 2021-04-30
328 Prepare for Upload API

manno:
Several changes to the API (refers to #328), prepares PR #338

* Gitea related code was moved into a separate gitea package, so it can be reused without introducing cyclic imports
* Added a helper struct to register named routes and retrieve them later (better than constants, not as complicated as verbose as registered URLs in mux)

### DIFF
--- a/helpers/routes/routes.go
+++ b/helpers/routes/routes.go
@@ -1,0 +1,49 @@
+// Package routes implements registered urls and parameter substitution
+package routes
+
+import (
+	"fmt"
+	"net/http"
+	"regexp"
+	"strings"
+)
+
+// Route describes a route for httprouter
+type Route struct {
+	Method  string
+	Path    string
+	Format  string
+	Handler http.HandlerFunc
+}
+
+var formatRegex = regexp.MustCompile(`:\w+`)
+
+// NewRoute returns a new route, which can be added to NamedRoutes and used with
+// httprouter. Trailing and leading slashes are removed.
+func NewRoute(method string, path string, h http.HandlerFunc) Route {
+	format := formatRegex.ReplaceAllString(path, "%s")
+	format = strings.Trim(format, "/")
+	return Route{method, path, format, h}
+}
+
+// NamedRoutes is a map of all named routes, to provide something like
+// https://github.com/gorilla/mux#registered-urls
+type NamedRoutes map[string]Route
+
+// Path returns a route's path with params substituted, panics if
+// used inproperly.
+func (n NamedRoutes) Path(name string, params ...interface{}) string {
+	r, ok := n[name]
+	if !ok {
+		// this means you have a typo or you allowed a user controlled
+		// string to reach this method, don't.
+		panic(fmt.Sprintf("route not found for '%s'", name))
+	}
+
+	// make sure to pass the right amount of params
+	// otherwise this method might fail
+	if r.Format == "" || len(params) == 0 {
+		return strings.Trim(r.Path, "/")
+	}
+	return fmt.Sprintf(r.Format, params...)
+}
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -17,7 +17,7 @@ import (
 	"strings"
 	"time"
 
-	"code.gitea.io/sdk/gitea"
+	giteaSDK "code.gitea.io/sdk/gitea"
 	"github.com/epinio/epinio/deployments"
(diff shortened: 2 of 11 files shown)


helpers/routes/routes.go
```diff
@@ -0,0 +1,49 @@
...
+type Route struct {
+	Method  string
+	Path    string
+	Format  string
+	Handler http.HandlerFunc
+}
+
+var formatRegex = regexp.MustCompile(`:\w+`)
+
+// NewRoute returns a ne route, which can be added to NamedRoutes and used with
```
andreas-kupries, line 21: Nit: `ne` - `new`

internal/cli/clients/client.go
```diff
@@ -111,7 +112,7 @@ func (c *EpinioClient) ServicePlans(serviceClassName string) error {
 		Msg("Listing service plans")
 
 	// todo: sort service plans by name before display
-	jsonResponse, err := c.curl(fmt.Sprintf("api/v1/serviceclasses/%s/serviceplans", serviceClassName), "GET", "")
+	jsonResponse, err := c.curl(api.Routes.Path("ServicePlans", serviceClassName), "GET", "")
```
andreas-kupries, line 115: Should that be `c.get`, like the other changed places ?

internal/cli/clients/client.go
```diff
@@ -341,7 +341,7 @@ func (c *EpinioClient) UnbindService(serviceName, appName string) error {
 		WithStringValue("Organization", c.Config.Org).
 		Msg("Unbind Service from Application")
 
-	_, err := c.curl(fmt.Sprintf("api/v1/orgs/%s/applications/%s/servicebindings/%s",
+	_, err := c.curl(api.Routes.Path("ServiceBindingDelete",
```
andreas-kupries, line 344: Should this be `c.delete` ?

internal/cli/clients/client.go
```diff
@@ -741,7 +750,7 @@ func (c *EpinioClient) Delete(appname string) error {
 	s := c.ui.Progressf("Deleting %s in %s", appname, c.Config.Org)
 	defer s.Stop()
 
-	jsonResponse, err := c.curl(fmt.Sprintf("api/v1/orgs/%s/applications/%s", c.Config.Org, appname), "DELETE", "")
+	jsonResponse, err := c.curl(api.Routes.Path("AppDelete", c.Config.Org, appname), "DELETE", "")
```
andreas-kupries, line 753: `c.delete` ?

internal/cli/server.go
```diff
@@ -37,11 +41,12 @@ var CmdServer = &cobra.Command{
...
 		port := viper.GetInt("port")
 		ui := termui.NewUI()
-		_, listeningPort, err := startEpinioServer(httpServerWg, port, ui)
+		logger := tracelog.NewServerLogger()
+		_, listeningPort, err := startEpinioServer(httpServerWg, port, ui, logger)
 		if err != nil {
 			return err
 		}
-		fmt.Println("listening on localhost on port " + listeningPort)
+		ui.Normal().V(0).Msg("listening on localhost on port " + listeningPort)
```
andreas-kupries, line 49: `V(0)` is a no-op. Argument is a delta, not a level.

andreas-kupries (changes requested):
Only small stuff.

internal/cli/server.go
```diff
@@ -95,17 +108,15 @@ func logRequestHandler(h http.Handler, ui *termui.UI) http.Handler {
...
 }
 
-func logRequest(r *http.Request, ui *termui.UI) {
+func logRequest(r *http.Request, log logr.Logger) {
 	uri := r.URL.String()
 	method := r.Method
-	ui.Normal().V(1).Msgf("%s %s", method, uri)
-
-	msg := ui.Normal().V(1).Compact()
+	log.V(1).Info(fmt.Sprintf("%s %s", method, uri))
```
andreas-kupries, line 114: With a logger, should we split the data into proper fields for the log message ?

manno: I'm not going to change this as part of this PR.
I'm not sure what these fields would look like or how you'd configure them with logr. 
Someone has to convince me of Verbosity-levels and structured logging. 

Maybe we can have a story to make API server request logging better.

andreas-kupries: Yes, we can do that in a separate ticket.

andreas-kupries (approved):

PR #351, opened 2021-05-04
fix: explicity handle org-namespace already exist case

### LINKED ISSUE #344: Creating an org with the name `epinio` is not possible anymore since the latest refactoring
If I run org create epinio  after patching the local instance, I get the following:

```
🚢 Creating organization...
Name: epinio
error creating org: Bad Gateway: Bad Gateway
$ ./dist/epinio-linux-amd64 org list
W0430 08:37:29.920622   28311 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0430 08:37:29.926065   28311 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0430 08:37:29.970442   28311 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress🚢 Listing organizations✔️  Epinio Organizations:
| NAME |
|------|
```

Other orgs I have tested work fine, and the name `epinio` was also possible in the past. So we should make sure that all name conflicts are verified and handled with a proper error message.

rohitsakala:
fixes #344

### DIFF
--- a/internal/organizations/organizations.go
+++ b/internal/organizations/organizations.go
@@ -11,6 +11,7 @@ import (
 	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -68,6 +69,9 @@ func Create(ctx context.Context, kubeClient *kubernetes.Cluster, gitea GiteaInte
 		},
 		metav1.CreateOptions{},
 	); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return errors.Errorf("Org '%s' name cannot be used. Please try another name", org)
+		}
 		return err
 	}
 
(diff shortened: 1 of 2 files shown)


internal/organizations/organizations.go
```diff
@@ -68,6 +69,9 @@ func Create(ctx context.Context, kubeClient *kubernetes.Cluster, gitea GiteaInte
 		},
 		metav1.CreateOptions{},
 	); err != nil {
+		if apierrors.IsAlreadyExists(err) {
+			return errors.Errorf("Org '%s' name cannot be used. Please try another name", org)
+		}
```
andreas-kupries, line 74: We do not have an acceptance test covering this code path, don't we ?

andreas-kupries (changes requested):

internal/organizations/organizations.go
rohitsakala: Added a test.

PR #355, opened 2021-05-05
Fix reloading of updated apps

thardeck:
We use the previous git commit of the app storage to have an unique
identifier in the image url which changes every time the app is
repushed.

Even after the first push there are at least two commits because we have
an initial commit and one for the first push.

Even if the app is not changed, only pushed again the app will
retrigger, because our current implementation creates a new git commit
every time we push.

I think we should rather not create new git commits during push if the app source has not change then to work around this issue here.


andreas-kupries (commented):
I am ok with it.
However having it seen before, and having been part of writing it I am not sure that I should be a reviewer/approver.

rohitsakala (approved):

PR #356, opened 2021-05-05
Fix upload breakage for push with app directory

### LINKED ISSUE #353: Pushing an app using a directory path no longer works
Doing a `cf push sample assets/sample-app` no longer works:

```
~/workspace/epinio (main)*$ ./dist/epinio-linux-amd64 push --verbosity 1 sample assets/sample-app/
W0505 06:04:52.502202    6064 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0505 06:04:52.507256    6064 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
W0505 06:04:52.523119    6064 warnings.go:70] extensions/v1beta1 Ingress is deprecated in v1.14+, unavailable in v1.22+; use networking.k8s.io/v1 Ingress
🚢 About to push an application with given name and sources into the specified organization
Name: sample
Sources: assets/sample-app/
Organization: workspace
Compressing application code ...
error pushing app to server: can't create archive: walking buildpack.yml: buildpack.yml: stat: stat buildpack.yml: no such file or directory
```

the reason must be that the `f.Name()` method here:
https://github.com/epinio/epinio/blob/main/internal/cli/clients/client.go#L881
returns the base path for the file, skipping the directory.

Doing a push from within the app directory still works.
We need to make sure the files are read from the app directory when one is provided. Still we may want to keep the file paths "flat" inside the tarball. A quick look in the archiver package code shows there is a function `NameInArchive` which could be useful.

## Acceptance Criteria

- Bug is fixed
- There is a test for this

andreas-kupries:
Fixes #353 

Added app source dir back to list of dir entries
Because the dir may not be the local cwd.
Added test pushing an app via name and path, instead of name and cwd.

### DIFF
--- a/acceptance/utilities_test.go
+++ b/acceptance/utilities_test.go
@@ -82,7 +82,28 @@ func makeApp(appName string) string {
 	ExpectWithOffset(1, err).ToNot(HaveOccurred())
 	appDir := path.Join(currentDir, "../assets/sample-app")
 
+	// Note: appDir is handed to the working dir argument of Epinio().
+	// This means that the command runs with it as the CWD.
 	pushOutput, err := Epinio(fmt.Sprintf("apps push %s --verbosity 1", appName), appDir)
+	ExpectWithOffset(2, err).ToNot(HaveOccurred(), pushOutput)
+
+	// And check presence
+
+	out, err := Epinio("app list", "")
+	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
+	ExpectWithOffset(1, out).To(MatchRegexp(appName + `.*\|.*1\/1.*\|.*`))
+
+	return pushOutput
+}
+
+func makeApp2(appName string) string {
+	currentDir, err := os.Getwd()
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	appDir := path.Join(currentDir, "../assets/sample-app")
+
+	// Note: appDir is handed as second argument to the epinio cli.
+	// This means that the command gets the sources from that directory instead of CWD.
+	pushOutput, err := Epinio(fmt.Sprintf("apps push %s %s --verbosity 1", appName, appDir), "")
 	ExpectWithOffset(2, err).ToNot(HaveOccurred(), pushOutput)
 
 	// And check presence
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -871,14 +871,21 @@ func (c *EpinioClient) Push(app string, source string) error {
 		return fmt.Errorf("%s: %s", "app name incorrect", strings.Join(errorMsgs, "\n"))
 	}
 
-	c.ui.Normal().Msg("Compressing application code ...")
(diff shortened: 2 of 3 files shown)


acceptance/utilities_test.go
```diff
@@ -94,6 +96,25 @@ func makeApp(appName string) string {
 	return pushOutput
 }
 
+func makeApp2(appName string) string {
```
thardeck, line 99: We could make the makeApp method more flexible instead of duplicating it but I am fine with the current implementation.

thardeck (approved):

acceptance/utilities_test.go
andreas-kupries: Things, we only need these two variants. As the second is used only once I could inline it into the test.

andreas-kupries:
Re-checked the generated tarball (i.e. path structure), and git structure.

The current setup looks ok, for both relative and absolute path. The tarball has everything proper in the top directory. No separate directory for the app. Example:

```
% tar tvf /tmp/epinio-app200774214/blob.tar
-rw-rw-r-- work/users       29 2021-04-16 16:37 buildpack.yml
drwxrwxr-x work/users        0 2021-05-04 15:14 htdocs/
-rw-rw-r-- work/users       21 2021-05-04 15:14 htdocs/index.php
-rw-rw-r-- work/users       21 2021-05-05 11:17 index.php
```

Using the application directory/path directly, instead of a list of its contents however inserts an additional directory into the tarball. And that breaks deployment:

```
% tar tvf /tmp/epinio-app686598867/blob.tar
drwxrwxr-x work/users        0 2021-05-05 11:17 sample-app/
-rw-rw-r-- work/users       29 2021-04-16 16:37 sample-app/buildpack.yml
drwxrwxr-x work/users        0 2021-05-04 15:14 sample-app/htdocs/
-rw-rw-r-- work/users       21 2021-05-04 15:14 sample-app/htdocs/index.php
-rw-rw-r-- work/users       21 2021-05-05 11:17 sample-app/index.php
```

Conclusion: The current code is :heavy_check_mark:, no changes required.

internal/cli/clients/client.go
```diff
@@ -871,14 +871,21 @@ func (c *EpinioClient) Push(app string, source string) error {
 		return fmt.Errorf("%s: %s", "app name incorrect", strings.Join(errorMsgs, "\n"))
 	}
 
-	c.ui.Normal().Msg("Compressing application code ...")
+	c.ui.Normal().Msg("Collecting the application sources ...")
 	files, err := ioutil.ReadDir(source)
 	if err != nil {
 		return errors.Wrap(err, "canot read the apps source files")
```
jimmykarily, line 877: there was a little typo here `canot` -> `cannot` . I wasn't fast enough.

PR #359, opened 2021-05-05
348 json errors

### LINKED ISSUE #348: API returns error objects
API Clients can always expect a JSON response.

Along the lines of https://jsonapi.org/format/#error-objects and https://jsonapi.org/examples/#error-objects-basics
(We don't need to support all the fields.)

* uses a [wrapper handler](https://stackoverflow.com/questions/32485021/simplifying-repetitive-error-handling-with-julienschmidt-httprouter), API actions can just return the error
* return codes indicate if it's a user error (4xx) or server failure (5xx) 
* CLI client understands JSON responses

```
{
  "errors": [
    {
      "status": "422",
      "title":  "Invalid Attribute",
    }
  ]
}
```

jimmykarily:
Fixes #348

### DIFF
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -16,6 +16,10 @@ import (
 	"github.com/pkg/errors"
 )
 
+// APIError struct is meant to host an error as described here:
+// https://jsonapi.org/examples/#error-objects-basics
+type APIError map[string]([]map[string]string)
+
 type ApplicationsController struct {
 }
 
@@ -33,8 +37,8 @@ func (hc ApplicationsController) Index(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if !exists {
-		http.Error(w, fmt.Sprintf("Organization '%s' does not exist", org),
-			http.StatusNotFound)
+		err := errors.Errorf("Organization '%s' does not exist", org)
+		handleError(w, err, http.StatusNotFound)
 		return
 	}
 
@@ -79,8 +83,8 @@ func (hc ApplicationsController) Show(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if app == nil {
-		http.Error(w, fmt.Sprintf("application '%s' not found", appName),
-			http.StatusNotFound)
+		err := errors.Errorf("application '%s' not found", appName)
+		handleError(w, err, http.StatusNotFound)
 		return
 	}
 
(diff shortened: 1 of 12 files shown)


internal/api/v1/applications.go
```diff
@@ -12,8 +14,13 @@ import (
 	"github.com/epinio/epinio/internal/organizations"
 	"github.com/epinio/epinio/internal/services"
 	"github.com/julienschmidt/httprouter"
+	"github.com/pkg/errors"
 )
 
+// APIError struct is meant to host an error as described here:
+// https://jsonapi.org/examples/#error-objects-basics
+type APIError map[string]([]map[string]string)
```
andreas-kupries, line 22: I believe this could do with separate type names for the inner elements, for higher readability.

internal/api/v1/applications.go
```diff
@@ -176,9 +183,31 @@ func (hc ApplicationsController) Delete(w http.ResponseWriter, r *http.Request)
 }
 
 // Write the error to the response writer and return  true if there was an error
-func handleError(w http.ResponseWriter, err error, code int) bool {
+func handleError(w http.ResponseWriter, err error, code int, details ...string) bool {
 	if err != nil {
-		http.Error(w, err.Error(), code)
```
andreas-kupries, line 181: Nit. More go-ish:

```
if err == nil {
  return false
}

... [all the big error ops, outdented by a level] ...
```

internal/api/v1/applications.go
andreas-kupries: And it will make it easier to add functionality for returning multiple errors in one response.
I.e. function to create a single error entry, assemble multiple entries, wrap the response around such.

internal/api/v1/upload.go
```diff
@@ -27,68 +28,62 @@ func (hc ApplicationsController) Upload(w http.ResponseWriter, r *http.Request)
...
 		return
 	}
 
 	log.V(2).Info("parsing multipart form")
 
 	file, _, err := r.FormFile("file")
 	if err != nil {
-		userError(w, "can't read multipart file input")
+		err = errors.Wrapf(err, "can't read multipart file input")
+		handleError(w, err, http.StatusBadRequest)
```
andreas-kupries, line 40: Why removing user/internal error ? Could have been re-implemented using `handleError`. Less work, just two places to change.

andreas-kupries: Maybe change the signature, to provide the `err` for wrapping.

andreas-kupries (changes requested):

andreas-kupries (approved):

jimmykarily:
Superseded by #365

Issue #358, opened 2021-05-05
Remove Gitea webhook


jimmykarily:
Right now, application staging is triggered when Gitea talks to the Tekton event-listener pod. Then by using an EventBinding and a PipelineTemplate, eventually a PipelineRun resource is created.

This has the following problems:
- when gitea fails to talk to event-listener (e.g. because of network hickup or other reason), it never retries.
- when the above happens, the user is not notified in any way about the failure. Will simply wait until `epinio push` times out.
- We rely too much on a component we don't control (the gitea webhook)
- can event-listener be HA?

Since we know when the user asked for staging (`epinio push` , remember?), we can simply create the PipelineRun ourselves and not rely on all this wiring.

## Acceptance Criteria

- There is no event-listener deployed on the cluster anymore
- There are no TriggerBindings or PipelineTemplates on the cluster
- We don't create a webhook on Gitea anymore
- Staging still works.

## Additional steps (new stories?)

- Remove `.kube` directory creation in the repo itself. Store these some place else (ConfigMap?)
- Implement git flow by having something else monitoring git repos and asking the epinio server for staging like the user would. (e.g. using [Rancher git-job](https://github.com/rancher/gitjob))
- Implement Web UI deployment using tarball upload
- Implement Web UI deployment using external repo configuration.

Issue #347, opened 2021-05-04
Application abstraction


jimmykarily:
Because:

- You can create apps that are not started (like in cf), bind services to them and then start them (fixing: #303)
- You can pull codes from different locations (e.g. gitea, s3, minio, github, other) and you can store this in for on the application abstraction
- You can store the epinio deployment information on it (e.g. which "revision" or "deployment id" is currently being run). We can rollback based on this info.
- You can run expensive operations like tarball extraction in a queue in the background (see also: https://github.com/epinio/epinio/pull/338/files#r625631405).

## Acceptance Criteria

1. There is a Custom Resource that represents an Epinio application
1. Every other resource belonging to the application (Deployment/Ingress/Service/Secret) has an owner reference to the CR.
1. All resources share the same label(s) identifying the application
1. Any metadata (e.g. stageID of the running app, commit id of the running app etc) are stored on the CR.
1. When the CR is deleted, so do all the other resources (owner relationship).
1. Epinio commands that list or show applications use this new CRD (instead of Deployments that they do now).
1. `epinio push` should create a CR for the app
1. [can be moved to a new issue] A new command `epinio app create` exists that simply creates the application without doing staging or pushing of the code or anything else. Just creates the CR.

## Stories

We discussed this more and created the following stories (TBD): #508  #509 #510 #511 #512 #513 #514 #515 #516 

## (Likely) Related Bugs

#536

manno:
We have two(three) different kinds of data for an application:

* the **application data** itself, which is just files and directories
* **application meta data**, like the name and organzitation it belongs to (user provided)
  * **deployment info**, we generate, the webhook, the k8s resources in the `.kube` folder (generated)
 
### Idea 1 - blob store + crd

We can split the application data (files) from the meta data and store them separately.
We create a CRD that holds the meta data and references the application data.
When deploying, we generate the tekton staging inputs and finally the k8s resources (deployment) from that CRD.

It's enough to support one source (git?) initially.

jimmykarily:
#303 could also be fixed by implementing an "instances" parameter on `cf push`. E.g. `epinio push sample -i 0` (this should create a deployment with 0 instances. After connecting the services the user could do something ilke this to start it: `epinio app update sample -i 1`

jimmykarily:
This may also be the solution to #375 . For some applications, it's necessary that the environment variables are set before the app can be successfully staged. We would need to store the environment variables somewhere, before a Deployment even exists.

[...]

PR #363, opened 2021-05-06
feat: Enable the binding of services to an app as part of `push`

### LINKED ISSUE #303: Binding Service on push
The UX of pushing an app, waiting for it to fail, then binding the service is really painful. It'd be nice if there was a way to push with a binding as a single step.

andreas-kupries:
Fixes #303

INCOMPATIBILITY: API CHANGE
	Modified service bind requests to carry an array of service
	names, instead of a single name.

Enables the binding of multiple services to one app in a single API
call.

Modified the receiver to match.
Note that the receiver still performs the binds sequentially.

TODO: Collect as many errors as possible to return in a single
response.

Updated the `service bind` command to match.

Extended the `push` command to accept services to be bound, via a new
option `--bind`/`-b` for `push`. Note that the option allows the
specification of multiple services (1). The code removes duplicates in
the specified set of services.

Added command completion for `--bind`.

__Simple implementation__: Internally does what the user would otherwise
do manually, i.e.

  - push the app
  - bind the specified services to the app, one after the other

The only optimization right now is that the entire set of services is
transfered to the server in a single bind call.

__Optimizations__ like adding the services to the app kube deployment as
part of starting the app, instead of via separate updates, are __defered__
until we have separated uploading the app sources from starting the
app (2).

----

  1. Example: `... --bind service1,service2 --bind service3`

  2. Currently app start is automatically triggered by the source
upload (into gitea). See issues #347 and #358 for the requisite work
needed.

### DIFF
--- a/internal/api/v1/errors.go
+++ b/internal/api/v1/errors.go
@@ -1,0 +1,96 @@
+package v1
+
+import (
+	"fmt"
+	"net/http"
+	"strings"
+)
+
+type APIError struct {
+	Status  int
+	Title   string
+	Details string
+}
+
+// Satisfy the error interface
+func (err APIError) Error() string {
+	return err.Title
+}
+
+func NewAPIError(message, details string, status int) APIError {
+	return APIError{
+		Title:   message,
+		Details: details,
+		Status:  status,
+	}
+}
+
+type APIErrors []APIError
+
+// All our actions match this type. They can return a list of errors.
+// The "Status" of the first error in the list becomes the response Status Code.
+type APIActionFunc func(http.ResponseWriter, *http.Request) APIErrors
+
+func InternalError(err error) APIError {
+	return NewAPIError(err.Error(), "", http.StatusInternalServerError)
+}
+
+func BadRequest(err error, details ...string) APIError {
+	return NewAPIError(err.Error(), strings.Join(details, ", "), http.StatusBadRequest)
+}
+
+func OrgIsNotKnown(org string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Organization '%s' does not exist", org),
+		"",
+		http.StatusNotFound)
+}
+
+func AppIsNotKnown(app string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Application '%s' does not exist", app),
+		"",
+		http.StatusNotFound)
+}
+
+func ServiceIsNotKnown(service string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Service '%s' does not exist", service),
+		"",
+		http.StatusNotFound)
+}
+
+func ServiceClassIsNotKnown(serviceclass string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("ServiceClass '%s' does not exist", serviceclass),
+		"",
+		http.StatusNotFound)
+}
+
+func OrgAlreadyKnown(org string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Organization '%s' already exists", org),
+		"",
+		http.StatusConflict)
+}
+
+func ServiceAlreadyKnown(service string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Service '%s' already exists", service),
+		"",
+		http.StatusConflict)
+}
+
+func ServiceAlreadyBound(service string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Service '%s' already bound", service),
+		"",
+		http.StatusConflict)
+}
+
+func ServiceIsNotBound(service string) APIError {
+	return NewAPIError(
+		fmt.Sprintf("Service '%s' is not bound", service),
+		"",
+		http.StatusBadRequest)
+}
--- a/internal/api/v1/servicebindings.go
+++ b/internal/api/v1/servicebindings.go
@@ -2,13 +2,13 @@ package v1
 
 import (
 	"encoding/json"
-	"fmt"
 	"io/ioutil"
 	"net/http"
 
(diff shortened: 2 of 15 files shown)


internal/cli/push.go
```diff
@@ -35,7 +73,12 @@ var CmdPush = &cobra.Command{
 			return errors.Wrap(err, "path not accessible")
 		}
 
-		err = client.Push(args[0], path)
+		services, err := cmd.Flags().GetStringSlice("bind")
```
jimmykarily, line 76: And I'm doing my own thing here: https://github.com/epinio/epinio/compare/368-application-scale#diff-9d5877b4c4b9701570001c32b04a95d163025ade9671e9c5add3a32d98934e9dR38-R45 :D  

Since your PR is ready, I will rebase when we merge this and come up with some refactoring to work for all flags. Maybe we can pair on that after this one is merged.

andreas-kupries: Heh. Sure.

internal/api/v1/errors.go
```diff
@@ -0,0 +1,96 @@
+package v1
```
jimmykarily, line 1: Good refactoring :+1:

internal/api/v1/servicebindings.go
```diff
@@ -26,71 +31,90 @@ func (hc ServicebindingsController) Create(w http.ResponseWriter, r *http.Reques
...
+
+	if len(theIssues) > 0 {
+		return theIssues
 	}
 
 	w.Header().Set("Content-Type", "application/json")
 	_, err = w.Write([]byte{})
 	if err != nil {
-		return APIErrors{NewAPIError(err.Error(), "", http.StatusInternalServerError)}
+		return append(APIErrors{InternalError(err)}, theIssues...)
```
jimmykarily, line 117: with the current code, there is no way for `theIssues` to be non-empty at this point (see line 110 above). Should we keep this just in case someone adds code between line 110 and 117 that populates `theIssues`?

internal/api/v1/servicebindings.go
```diff
@@ -26,71 +31,90 @@ func (hc ServicebindingsController) Create(w http.ResponseWriter, r *http.Reques
...
+				continue
+			}
+
+			return append(APIErrors{InternalError(err)}, theIssues...)
 		}
-	}
-	if err != nil {
-		return APIErrors{NewAPIError(err.Error(), "", http.StatusInternalServerError)}
+
+		theServices = append(theServices, service)
```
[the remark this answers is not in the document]
andreas-kupries: Checking...

internal/api/v1/servicebindings.go
andreas-kupries: Checking...

andreas-kupries: Right. I forgot to simplify that when I fixed an issue, i.e. adding the `return theIssues`. Will fix.

internal/cli/clients/client.go
```diff
@@ -939,6 +949,30 @@ func (c *EpinioClient) Push(app string, source string) error {
 		return errors.Wrap(err, "waiting for app failed")
 	}
 
+	if len(services) > 0 {
+		c.ui.Note().Msg("Binding Services")
+
+		// Application is up, bind the services.
+		// This will restart the application.
+		// TODO: See #347 for future work
```
jimmykarily, line 957: I'm not sure we need #347 to fix this. We can simply send over the desired services as part of the `upload` request and let the server do the bindings. This way the client does as little as possible. That's the approach I took in #368 to handle the `instances` parameter. 
Since we are going to have conflicts with #368, let's see if we can improve this code there. Let's keep it as it is for now.

internal/api/v1/servicebindings.go
andreas-kupries: Done with commit c86ec5ad0204734e53a780788412a84c284b10d6

internal/api/v1/servicebindings.go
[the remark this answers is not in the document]
jimmykarily: Oh true, I got confused with the various blocks. This is inside the `for` loop. All good.

internal/cli/clients/client.go
andreas-kupries: Thanks. Hm ... `upload` provides the sources as tarball, in the request body ... I guess if the framework does the correct thing for us when we add additional parameters like services and #instances, then it should work fine. I guess the `upload` is then more of an `app create` API. And Mario's work removing the event listener and web hook, etc. then gets a separate `app start` API.

internal/cli/push.go
```diff
@@ -10,7 +11,44 @@ import (
...
+			return matches, cobra.ShellCompDirectiveNoFileComp
+		}
+
+		// Expand the last segment. The returned matches are
+		// the string with its last segment replaced by the
+		// expansions for that segment.
+
+		matches := []string{}
+		for _, match := range app.ServiceMatching(values[len(values)-1]) {
+			values[len(values)-1] = match
```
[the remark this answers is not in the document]
andreas-kupries: It changes the list of (for example `['fox','d']` to `['fox','dog']` when `dog` is valid expansion/match for `d`.
The join then generates `fox,dog` from it, which goes into the slice of matches.
This is done for all expansions (dog, delphin, ...) of `d`...

Now that I know a bit more about `append` it could be done without overwriting, I guess.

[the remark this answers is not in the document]
andreas-kupries: Leaving as is for now.

jimmykarily (approved):

internal/cli/push.go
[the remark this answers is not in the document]
jimmykarily: gotcha, looks good.

PR #364, opened 2021-05-06
Local testing works without full re-install

### LINKED ISSUE #362: CI - option to skip k3d install when testing locally
While developing, sometimes we need to run a set of tests (e.g. using "focus") again and again. Most of the time the installation part of Epinio is irrelevant to what is being tested. Still it costs a lot of time because it is happening on a fresh cluster so images have to be downloaded etc.

We need to implement a flag (e.g. environment variable) that allows us to completely skip the installation and uninstallation of Epinio on the cluster. This should be off by default but can be enabled locally when trying out things.

Since this flag will also skip the uninstallation of Epinio, we should consider adding a deletion step to the BeforeSuite block (if a cluster is still there) so that the next time we run the tests without the flag, the cluster is cleaned up (since last built left it running).

manno:
fixes #362

This PR adds the features mentioned in the commit titles.

To be able to re-use the existing install:

* use `tmp/skip_cleanup` to avoid cluster deletion after the tests complete 
* run `k3d cluster delete --all` manually, when a fresh install is needed 
* use `k3d kubeconfig merge -a -d -u` to update kube config with the new cluster(s)

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -77,12 +77,10 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 		))
 	}
 
-	fmt.Println("Installing Epinio")
-	// Allow the installation to continue
-	out, err = RunProc("../dist/epinio-linux-amd64 install --skip-default-org", "", false)
-	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
+	ensureEpinio()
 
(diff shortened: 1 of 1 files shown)


manno:
I see now that all of the `ensure*` funcs in acceptance_suite are idempotent, however "epinio install" is not.

manno:
I updated the description

jimmykarily (approved):

PR #366, opened 2021-05-06
Add delete organisation api & cmd

### LINKED ISSUE #103: Add a command to delete organizations
Complete the set of operations on organizations.

We should wait until #329 is done first because that would change the implementation a lot.

rohitsakala:
fixes #103

### DIFF
--- a/acceptance/api_v1_orgs_test.go
+++ b/acceptance/api_v1_orgs_test.go
@@ -182,8 +182,7 @@ var _ = Describe("Orgs API Application Endpoints", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(string(bodyBytes)).To(Equal(""))
 
-				_, err = helpers.Kubectl(fmt.Sprintf("get namespace %s", org))
-				Expect(err).To(HaveOccurred())
+				verifyOrgNotExist(org)
 			})
 		})
 	})
--- a/acceptance/api_v1_orgs_test.go
+++ b/acceptance/api_v1_orgs_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/epinio/epinio/helpers"
 	apiv1 "github.com/epinio/epinio/internal/api/v1"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -146,6 +147,43 @@ var _ = Describe("Orgs API Application Endpoints", func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(response.StatusCode).To(Equal(http.StatusCreated), string(bodyBytes))
 				Expect(string(bodyBytes)).To(Equal(""))
+			})
+		})
+
+		Describe("DELETE api/v1/orgs/:org", func() {
+			It("deletes an organization", func() {
+				response, err := Curl("DELETE", fmt.Sprintf("%s/api/v1/orgs/%s", serverURL, org),
+					strings.NewReader(``))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response).ToNot(BeNil())
+				defer response.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(response.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response.StatusCode).To(Equal(http.StatusOK), string(bodyBytes))
+				Expect(string(bodyBytes)).To(Equal(""))
+
+				_, err = helpers.Kubectl(fmt.Sprintf("get namespace %s", org))
+				Expect(err).To(HaveOccurred())
+			})
+
+			It("deletes an organization including apps and services", func() {
+				app1 := newAppName()
+				makeApp(app1)
+				svc1 := newServiceName()
+				makeCustomService(svc1)
+				bindAppService(app1, svc1, org)
+
+				response, err := Curl("DELETE", fmt.Sprintf("%s/api/v1/orgs/%s", serverURL, org),
+					strings.NewReader(``))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response).ToNot(BeNil())
+				defer response.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(response.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(bodyBytes)).To(Equal(""))
+
+				_, err = helpers.Kubectl(fmt.Sprintf("get namespace %s", org))
+				Expect(err).To(HaveOccurred())
 			})
 		})
 	})
--- a/acceptance/orgs_test.go
+++ b/acceptance/orgs_test.go
@@ -41,7 +41,8 @@ var _ = Describe("Orgs", func() {
 			setupAndTargetOrg(org)
 
 			By("deleting organization")
-			out, err := Epinio("org delete "+org, "")
+			out, err := Epinio("org delete -f "+org, "")
+
 			Expect(err).ToNot(HaveOccurred(), out)
 		})
 	})
--- a/acceptance/orgs_test.go
+++ b/acceptance/orgs_test.go
@@ -34,4 +34,15 @@ var _ = Describe("Orgs", func() {
 			Expect(out).To(MatchRegexp(fmt.Sprintf("Organization '%s' already exists", org)))
 		})
 	})
+
+	Describe("org delete", func() {
+		It("deletes an org", func() {
+			org := newOrgName()
+			setupAndTargetOrg(org)
+
+			By("deleting organisation")
+			out, err := Epinio("org delete "+org, "")
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+	})
 })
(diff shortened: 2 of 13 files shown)


acceptance/orgs_test.go
```diff
@@ -34,4 +34,15 @@ var _ = Describe("Orgs", func() {
 			Expect(out).To(MatchRegexp(fmt.Sprintf("Organization '%s' already exists", org)))
 		})
 	})
+
+	Describe("org delete", func() {
+		It("deletes an org", func() {
+			org := newOrgName()
+			setupAndTargetOrg(org)
+
+			By("deleting organisation")
```
andreas-kupries, line 43: Nit: organi __z__ ation

internal/api/v1/organizations.go
```diff
@@ -93,3 +96,73 @@ func (oc OrganizationsController) Create(w http.ResponseWriter, r *http.Request)
...
+
+	for _, app := range apps {
+		err = application.Delete(cluster, gitea, org, app)
+		if handleError(w, err, http.StatusInternalServerError) {
+			return
+		}
+
+		servicesToBeDeleted = append(servicesToBeDeleted, app.BoundServices...)
+	}
+
```
andreas-kupries, line 139: Would it not be simpler (and more correct) to ask for all services in the org, bound or not ?

With the above it will only try to delete services in the org bound to an app.
Services in the org not bound to an app are not deleted.

Nit: If a service is bound to multiple apps the code will record it multiple times, and try to delete multiple times.
Simply asking for and deleting all services in the org trivially avoids that.

andreas-kupries (changes requested):

acceptance/orgs_test.go
rohitsakala: Done.

acceptance/api_v1_orgs_test.go
```diff
@@ -132,5 +133,42 @@ var _ = Describe("Orgs API Application Endpoints", func() {
...
+				response, err := Curl("DELETE", fmt.Sprintf("%s/api/v1/orgs/%s", serverURL, org),
+					strings.NewReader(``))
+				Expect(err).ToNot(HaveOccurred())
+				Expect(response).ToNot(BeNil())
+				defer response.Body.Close()
+				bodyBytes, err := ioutil.ReadAll(response.Body)
+				Expect(err).ToNot(HaveOccurred())
+				Expect(string(bodyBytes)).To(Equal(""))
+
+				_, err = helpers.Kubectl(fmt.Sprintf("get namespace %s", org))
```
jimmykarily, line 169: we never checked that the service lived in this namespace thus checking that a namespace doesn't exist doesn't tell us anything about the service. Imaging if someone changed the code an introduced a bug where the service appears in another namespace. This test would continue to be green, though based on the description one would expect it to fail.

Also consider checking the actual output of the command to make sure the error is about the namespace not being there anymore.

internal/api/v1/organizations.go
rohitsakala: Done

internal/api/v1/organizations.go
```diff
@@ -93,3 +96,73 @@ func (oc OrganizationsController) Create(w http.ResponseWriter, r *http.Request)
...
+	apps, err := application.List(cluster, org)
+	if handleError(w, err, http.StatusInternalServerError) {
+		return
+	}
+
+	servicesToBeDeleted := []string{}
+
+	for _, app := range apps {
```
jimmykarily, line 129: I wonder if we gain anything by deleting things manually. If everything we delete lives inside the org's namespace, can't we simply delete the namespace and wait until it's gone? Is there anything we need to delete that won't be deleted this way?

internal/duration/duration.go
```diff
@@ -74,6 +75,11 @@ func ToDeployment() time.Duration {
 	return Multiplier() * deployment
 }
 
+// ToNamespace returns the duration to wait for deletion of namespace
+func ToNamespace() time.Duration {
```
jimmykarily, line 79: `ToNamespaceDeletion` maybe?

internal/cli/orgs.go
```diff
@@ -66,3 +67,25 @@ var CmdOrgCreate = &cobra.Command{
 	SilenceErrors: true,
 	SilenceUsage:  true,
 }
+
+// CmdOrgDelete implements the epinio `orgs delete` command
+var CmdOrgDelete = &cobra.Command{
+	Use:   "delete NAME",
+	Short: "Deletes an organization",
+	Args:  cobra.ExactArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
```
jimmykarily, line 76: This is a dangerous spell! Should we maybe show a warning and require confirmation? E.g.

```
You are about to delete organization "myorg" and everything included in it (applications, services etc). Are you sure? (Y/n)
```

internal/duration/duration.go
rohitsakala: Done.

internal/api/v1/organizations.go
rohitsakala: Yes, @thardeck suggested that. I thought we should delete things and then delete the namespace since this is nice way step by step to do it. Sometimes I see terminating namespace for more than 20 min waiting for some resources/finalizers ... By this, we can be sure, that in the future such problems don't occur or are easy to identify.

acceptance/api_v1_orgs_test.go
rohitsakala: Right, will add checking for service & app presence too before and after.

internal/cli/orgs.go
rohitsakala: Right, will add this.

internal/api/v1/organizations.go
jimmykarily: True, but this could be really slow if you had, let's say, 1000 apps running in an org. Even if we manually delete things, we should probably do it in parallel, unless there is a reason for some actions to happen sequentially. Feel free to improve this on a new PR, since this is optimization.

kkaempf: How would we test this (and scaling in general) ?

jimmykarily: Do you mean one-off test or as part of the test suite? As a test suite I think this has to be a new one (a performance test suite or something). Deploying 1000 apps may need us to take some shortcuts to avoid staging etc. As an one off it should be relatively easy.

Generally, scaling involves more concepts like HA-ability of components, pod-affinity rules [and others](https://opsani.com/blog/10-kubernetes-performance-tuning-tips-best-practices/). We haven't even started working on optimizations like this since we are at the `make it work` stage yet. That's why I think we could postpone fixing my comment for another (later) PR. Unless  it bites us earlier (e.g. make tests or development slow).

internal/cli/orgs.go
rohitsakala: Done.

acceptance/api_v1_orgs_test.go
rohitsakala: Service and app presence is checked in the `makeApp` and `makeService` functions using `Epinio` func.

Regarding testing after calling deletion of the namespace, I have added a function that checks for the presence of org using `Epinio` func.

internal/api/v1/organizations.go
jimmykarily: I would recommend a smaller-scope issue. Just to delete orgs by deleting objects in parallel. Otherwise it will be too open ended.

acceptance/api_v1_orgs_test.go
jimmykarily: as long as org "show" action is doing the right thing, it should be ok. An alternative is to check on the kube level (using the `kubectl` helper), that the org namespace is actually gone.

rohitsakala: Kewl. The `org list` calls kubectl.

internal/api/v1/organizations.go
rohitsakala: Done.

internal/cli/orgs.go
```diff
@@ -66,3 +77,57 @@ var CmdOrgCreate = &cobra.Command{
...
+	Use:   "delete NAME",
+	Short: "Deletes an organization",
+	Args:  cobra.ExactArgs(1),
+	PreRunE: func(cmd *cobra.Command, args []string) error {
+		force, err := cmd.Flags().GetBool("force")
+		if err != nil {
+			return err
+		}
+		if !force {
+			cmd.Printf("You are about to delete organization %s and everything included in it (applications, services etc). Are you sure? (Y/n): ", args[0])
```
jimmykarily, line 92: When the user simply hits `Enter` the see the message:

```
You are about to delete organization workspace and everything included in it (applications, services etc). Are you sure? (Y/n): 
Please enter Y or n: 
```
That means `Y` is not the default option thus it shouldn't be capitalized. Should be `y/n` instead.

jimmykarily: also in the verification message (^ above), it should also be in lower case.

jimmykarily (dismissed):

internal/cli/orgs.go
rohitsakala: Done. Changed it to lowercase y.

jimmykarily (approved):

andreas-kupries (approved):

PR #371, opened 2021-05-07
368 application scale

### LINKED ISSUE #368: Application scale command
Right now all applications are deployed with only one instance/replica

The user should be able to:

- Push a new application and define the desired number of instances
- Update a running application and set the number of desired instances.
- Scale to zero instances
- Goes without saying but everything should be implemented through the API. We may want to demonstrate its use in the web interface.

Example commands:

```
$ epinio push sample --instances 3
```

```
$ epinio app update sample --instances 2
```

## TBD

- Should there be an upper limit to the number of desired instances?

jimmykarily:
Fixes #368

### DIFF
--- a/acceptance/api_v1_applications_test.go
+++ b/acceptance/api_v1_applications_test.go
@@ -35,15 +35,6 @@ var _ = Describe("Apps API Application Endpoints", func() {
 	})
 
 	Context("Apps", func() {
-		var app1, app2 string
-
-		BeforeEach(func() {
-			app1 = newAppName()
-			app2 = newAppName()
-			makeApp(app1, 1, true)
-			makeApp(app2, 1, true)
-		})
-
 		Describe("POST /orgs/:org/applications/:app", func() {
 			It("deploys an application with the desired number of instances", func() {
 				Skip("TODO")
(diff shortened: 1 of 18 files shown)


acceptance/api_v1_applications_test.go
```diff
@@ -28,22 +28,81 @@ var _ = Describe("Apps API Application Endpoints", func() {
 	})
 
 	Context("Apps", func() {
-		var app1, app2 string
+		Describe("POST /orgs/:org/applications/:app", func() {
+			It("deploys an application with the desired number of instances", func() {
```
jimmykarily, line 32: Maybe these new tests belong to the `apps_test.go` instead?

manno: I think this file is meant  to test the API directly, whereas apps_test.go uses the CLI to talk to the API

andreas-kupries: Correct. all the `api_v1_*` files poke directly at the API.

jimmykarily: So given `makeApp` is using the cli, these tests should be moved. Will do.

andreas-kupries: :bulb: And here I thought from the `It` description that you were directly poking the API.
That should be changed, I believe.
And maybe additional tests which do work directly with the API ?
Around supplication bogus data for instances (negative numbers, non-integer numbers, not-number strings, ...)

jimmykarily: Done

rohitsakala (approved):

rohitsakala (dismissed):

PR #374, opened 2021-05-10
feat: internal dev docs/notes about cat configuration elements

andreas-kupries:


acceptance/acceptance_suite_test.go
```diff
@@ -32,6 +32,27 @@ var nodeSuffix, nodeTmpDir string
 var serverURL string
 var registryMirrorName = "epinio-acceptance-registry-mirror"
 
+// Configuration Notes
+//
+// - skipCleanupPath: Path relative to the test directory.  IOW
```
manno, line 37: As far as I have seen, ginkgo always changes the working dir to the directory the test suite is in.

I'm not certain what this documentation tries to achieve: 

* If it's a manual on how to configure tests, it might be better to have it in `docs/running-tests.md`, drop all the variable names and mention the files and env vars directly.

* If this is meant to document the test_suite code, it would be nice to move the comments onto the corresponding variables, so they show up when hovering over the variable name and in docs.
Also starting the comment with the name of the thing [seems to help with go doc](https://golang.org/doc/effective_go#commentary): https://pkg.go.dev/github.com/epinio/epinio?utm_source=godoc

andreas-kupries: I wanted some documentation which helps me as an epinio dev to remember how I can configure our :cat2: .
So I guess the first point. Although the second point is a good idea also IMHO.

I see no `docs/running-tests.md` in the current `main` commit.
Guessing now that you are proposing to create such a doc file, right ?

Will let this bake a bit more in the back of my mind, focus now is the auth, and this PR is not an urgent thing/

andreas-kupries: Yep. Pushed reworked commit

docs/acceptance_tests.md
```diff
@@ -0,0 +1,64 @@
...
+     tests across. The default is `2`. The CI flows use `8`.
+
+  1. `FLAKE_ATTEMPTS`: The number of tries to perform when a test
+     fails, to ensure that the failure is real, and not a flake,
+     i.e. caused by a transient environmental condition. The default
+     is `2`.
+
+  1. `EPINIO_SKIP_PATCH`: When present and not empty the test startup
+     will skip over patching the epinio server pod of the test
+     cluster. This should be used when the test cluster still has a an
```
jimmykarily, line 31:
```suggestion
     cluster. This should be used when the test cluster still has an
```

jimmykarily: Actually having an epinio server running is a prerequisite. The "reason" to do it is to run tests faster when you don't need to patch. For example, when the only thing you changed is the tests themselves and not the actual epinio code.

docs/acceptance_tests.md
```diff
@@ -0,0 +1,64 @@
...
+the epinio checkout running the acceptance tests.
+
+  1. `tmp/skip_cleanup`: If present system shutdown will not uninstall
+     epinio from the tests cluster, nor will it tear down the test
+     cluster. This allows future test runs to bypass the costly setup
+     of a cluster and epinio installation.
+
+     Note however that this should not be done when testing changed to
+     epinio's server side. This always requires a new epinio
+     installation to test the latest changes.
```
jimmykarily, line 55: well... not always, if your changes don't change the installation part but only the epinio server, you can still avoid re-installing and rely on patching to update. Maybe that's too detailed to explain here? I don't know.

docs/acceptance_tests.md
```diff
@@ -0,0 +1,64 @@
...
+
+  1. `tmp/after_each_sleep`: If present, readable, and containing an
```
jimmykarily, line 57: can we also put these files in `.gitignore`? Maybe the whole `tmp` dir?

andreas-kupries: `/tmp` is in `.gitignore`.

docs/acceptance_tests.md
andreas-kupries: Tweaked, see commit c87eb2fca2646a1425bbed92df6b521696b38fe4

jimmykarily (approved):

Issue #61, opened 2021-01-21
Generate all passwords.


jimmykarily:
Right now there are hardcoded passwords in various places:

  - [x] [api password](https://github.com/epinio/epinio/blob/main/internal/cli/install.go#L33)
  - [x] [tekton gitea access](https://github.com/epinio/epinio/blob/main/deployments/tekton.go#L373) :arrow_up_down: [gitea setup](https://github.com/epinio/epinio/blob/main/deployments/gitea.go#L135)
  - [x] [tekton registry access](https://github.com/epinio/epinio/blob/main/deployments/tekton.go#L409) :arrow_up_down: [registry setup](https://github.com/epinio/epinio/blob/main/assets/container-registry/chart/container-registry/templates/container-registry.yaml#L8)

Unsure about the tokens below. Are they related to the gitea user/pass ? Derived ?

  - [-] [gitea internal token](https://github.com/epinio/epinio/blob/main/deployments/gitea.go#L152)
  - [-] [gitea jwt token](https://github.com/epinio/epinio/blob/main/deployments/gitea.go#L160)

Some of these services are exposed publicly through Ingress resources and thus make Carrier installation open to malicious users.

Every password we use should be generated and possible to rotate.

If it's too much work, consider splitting the generation and the rotation in 2 different stories.

jimmykarily:
Let's also fix this: https://github.com/epinio/epinio/pull/376#discussion_r632360177

andreas-kupries:
Updated description, links were broken due to development having moved on and moving files.
- gitea went to a slightly different directory.
- eirini is gone in full

andreas-kupries:
# Generation

  - [api password](https://github.com/epinio/epinio/blob/main/internal/cli/install.go#L33)

    Change default to a DynDefault. The func generates a hexified random string
   Need a generator for random strings - https://gist.github.com/dopey/c69559607800d2f2f90b1b1ed4e550fb

  - [tekton gitea access](https://github.com/epinio/epinio/blob/main/deployments/tekton.go#L373) :arrow_up_down: [gitea setup](https://github.com/epinio/epinio/blob/main/deployments/gitea.go#L135)

    Create gitea user/pass deployment options. DynDefaults. Note that doing this also makes them available to the person doing the `ep install`. Rework the gitea and tekton deployments to use the new options.

[...]

PR #376, opened 2021-05-11
feat: apply basic auth to API ingress

### LINKED ISSUE #350: Epinio API has Authentication
Instead of implementing auth in the API, we put the API behind traefik and use a [traefik middleware](https://doc.traefik.io/traefik/middlewares/basicauth/) to only allow authenticated connections.

The CLI knows how to authenticate. 
We support basic auth first, storing the CLI's user and password in the shell environment (viper).

andreas-kupries:
Fixes #350 

Note: Uses fixed credentials at the moment.
Todo:
  - setup user-chosen credentials during install,
  -  save to config,
  -  get from config.

The auth is working, i.e. traefik v1 (1) processed the annotations and activates auth handling.
API requests with the auth credentials work. Without they get `401 unauthorized`.

----
  1. That is k3d/k3s builtin traefik. The epinio traefik is v2 and has to checked to be working. If not, setup for it will be required also.

### DIFF
--- a/acceptance/api_v1_applications_test.go
+++ b/acceptance/api_v1_applications_test.go
@@ -218,6 +218,7 @@ var _ = Describe("Apps API Application Endpoints", func() {
 
 			// make the request
 			request, err := http.NewRequest("POST", url, body)
+			request.SetBasicAuth("me", "change")
 			if err != nil {
 				return nil, errors.Wrap(err, "failed to build request")
 			}
--- a/acceptance/api_v1_applications_test.go
+++ b/acceptance/api_v1_applications_test.go
@@ -218,7 +218,7 @@ var _ = Describe("Apps API Application Endpoints", func() {
 
 			// make the request
 			request, err := http.NewRequest("POST", url, body)
-			request.SetBasicAuth("me", "change")
+			request.SetBasicAuth(epinioUser, epinioPassword)
 			if err != nil {
 				return nil, errors.Wrap(err, "failed to build request")
 			}
--- a/deployments/epinio.go
+++ b/deployments/epinio.go
@@ -2,12 +2,15 @@ package deployments
 
 import (
 	"context"
+	"encoding/base64"
 	"fmt"
 	"io/ioutil"
 	"os"
 	"regexp"
+	"strings"
 	"time"
 
+	"github.com/GehirnInc/crypt/apr1_crypt"
 	"github.com/epinio/epinio/helpers"
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/helpers/termui"
@@ -25,9 +28,10 @@ type Epinio struct {
 }
 
 const (
-	EpinioDeploymentID = "epinio"
-	epinioServerYaml   = "epinio/server.yaml"
-	epinioStagingYAML  = "epinio/staging.yaml"
+	EpinioDeploymentID  = "epinio"
+	epinioServerYaml    = "epinio/server.yaml"
+	epinioStagingYAML   = "epinio/staging.yaml"
+	epinioBasicAuthYaml = "epinio/basicauth.yaml"
 )
 
 func (k *Epinio) ID() string {
@@ -70,6 +74,16 @@ func (k Epinio) Delete(c *kubernetes.Cluster, ui *termui.UI) error {
 		return errors.Wrap(err, fmt.Sprintf("Deleting %s failed:\n%s", epinioStagingYAML, out))
 	}
 
+	// (yyy) Note: We ignore deletion errors due to a mising
+	// Middleware CRD. That indicates that a traefik v1 controller
+	// was used, and the object was not applied. See also (xxx).
+
+	if out, err := helpers.KubectlDeleteEmbeddedYaml(epinioBasicAuthYaml, true); err != nil {
+		if !strings.Contains(out, `no matches for kind "Middleware"`) {
+			return errors.Wrap(err, fmt.Sprintf("Deleting %s failed:\n%s", epinioServerYaml, out))
+		}
+	}
+
 	message := "Deleting Epinio namespace " + EpinioDeploymentID
 	_, err = helpers.WaitForCommandCompletion(ui, message,
 		func() (string, error) {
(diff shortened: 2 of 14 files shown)


andreas-kupries:
The issue with :cat2:  looks to be that the chosen user/pass for API auth is not saved to the config file.
:bulb: It is saved. Just to the wrong file. The generic location.
It is later in the :cat2: startup sequence that `EPINIO_CONFIG` is set to a custom file, the per-node config.
That file cannot have the user/pass info from the install, it did not exist at the time.
So it contains the default config, empty user/pass, know to not work.

I'll have to see if forcing install to use that file can be done, and easy.

deployments/epinio.go
```diff
@@ -167,8 +193,25 @@ func (k Epinio) Upgrade(c *kubernetes.Cluster, ui *termui.UI, options kubernetes
...
+	// traefik v1 controller. See also (yyy).
+
+	yamlPathOnDisk, err := helpers.ExtractFile(epinioBasicAuthYaml)
+	if err != nil {
+		return "", errors.New("Failed to extract embedded file: " + epinioBasicAuthYaml + " - " + err.Error())
+	}
+	defer os.Remove(yamlPathOnDisk)
+
+	out, err := helpers.Kubectl(fmt.Sprintf("apply -n %s --filename %s", EpinioDeploymentID, yamlPathOnDisk))
+	if err != nil && !strings.Contains(out, `no matches for kind "Middleware"`) {
```
jimmykarily, line 209: I wonder if different k8s versions could produce different error messages. If that happens we should probably stop supporting traefik v1 rather than trying to catch everything. Let's see.

deployments/epinio.go
```diff
@@ -244,3 +305,14 @@ func (k *Epinio) createIngress(c *kubernetes.Cluster, subdomain string) error {
 
 	return err
 }
+
+// hashApr1 generates an Apache MD5 hash for a password.
```
jimmykarily, line 309: md5 is not secure as encryption for passwords but we do store this in a secret. I'm not sure if it's worth it switching to SHA1 or bcrypt (both supported: https://doc.traefik.io/traefik/middlewares/basicauth/#configuration-options).
If someone can read the md5 hash, it means they can read the k8s secrets, they can probably do more then without needing the password itself to talk to the epinio API. 

otoh, generating a sha1 sum is anything but difficult: https://golang.org/pkg/crypto/sha1/#New so why not?

docs/install.md
```diff
@@ -88,7 +88,6 @@ microk8s enable metallb:${IP}/16
 - git 2.22 or later
```
jimmykarily, line 88: remove git too? (since you are at it)

jimmykarily: and maybe bash too?

internal/cli/config/config.go
```diff
@@ -42,6 +44,10 @@ func Load() (*Config, error) {
 	v.SetDefault("epinio_protocol", "http")
 	v.SetDefault("org", "workspace")
 
+	// The defaults are a non-working user/pass combination
```
jimmykarily, line 47: what does "non-working" mean here? Maybe change it to `Use empty defaults in viper to allow NeededOptions defaults to apply` ?

internal/cli/install.go
```diff
@@ -22,6 +22,20 @@ var NeededOptions = kubernetes.InstallationOptions{
...
+		Description: "The user name for authenticating all API requests",
+		Type:        kubernetes.StringType,
+		Default:     "me",
+		Value:       "",
+	},
+	{
+		Name:        "password",
+		Description: "The password for authenticating all API requests",
+		Type:        kubernetes.StringType,
+		Default:     "change",
```
jimmykarily, line 36: I would prefer a "secure by default" approach. We can generate a password when the user doesn't provide one instead of using a hardcoded one. The user may forget to provide a password and given we don't have a way for the user to change it (yet), they will be forced to re-install. Even when changing the password will be possible, it's still better if we generated a good password by default so the user doesn't have to come up with a good password.

We have 2 personas, the cluster admin installing Epinio and the application developer.
The admin needs to have access to the cluster with kubectl/kubeconfig because that's how we create the various resources when installing Epinio.
The app developer, may or may not have kubectl access to the cluster. Right now we still try to access the cluster directly in some places (e.g. when trying to figure out what the domain is) but we should get to the point where it's possible to use Epinio with no direct access to the cluster, unless through Epinio server.

We can then have a group of subcommands (e.g. under the `admin` command) that are meant to be used by the administrator. Those are the commands that need direct access to the cluster.
Examples

- `epinio admin install` (the current `epinio install`)
- `epinio admin uninstall` (the current `epinio uninstall`)
- `epinio admin get-credentials`
- `epinio admin change-credentials`
- etc

With this in mind, I imagine the final process as:

- Admin installs Epinio on the cluster
- After installation is done, a command is printed that the user can run to read the credentials from the kube secret (e.g. `epinio admin get-credentials`). Alternatively, when the admin user is also the application developer, a command like `epinio admin login` can be used that will not require the user to know the credentials. It will simply copy the credentials from the secret into the Epinio config.
- If the application developer is a different user, then the admin will give her the credentials so the app dev doesn't need to have direct access to the cluster. 

This is just a proposal, we need to discuss it a bit. Let's keep things as they are here and improve on another PR. We still have to see if printing and storing credentials in plain text is ok.

acceptance/api_v1_applications_test.go
```diff
@@ -218,6 +218,7 @@ var _ = Describe("Apps API Application Endpoints", func() {
 
 			// make the request
 			request, err := http.NewRequest("POST", url, body)
+			request.SetBasicAuth("me", "change")
```
jimmykarily, line 221: maybe instead of relying on the defaults in tests, we can (indirectly) test that the installation parameters work by specifying some credentials when we install Epinio in the SynchronizedBeforeSuite block.

deployments/epinio.go
jimmykarily: [I read in various places](https://codahale.com/how-to-safely-store-a-password/) that bcrypt is the safest option but it's very slow (0.3 secs to hash a password?! though it depends on the work factor). That could make every request slower by 300ms which is not a negligible delay. Let's stick to MD5 for now. Securing k8s secrets is the right way to protect this password. After all, we store it in plain text on the cli side.

docs/install.md
jimmykarily: I updated the commit to remove these too.

internal/cli/config/config.go
jimmykarily: I changed the text and updated the commit.

acceptance/api_v1_applications_test.go
jimmykarily: I did so in a separate commit.

jimmykarily (approved):

internal/cli/install.go
jimmykarily: Let's fix it as part of this: https://github.com/epinio/epinio/issues/61#issuecomment-841161324

jimmykarily (changes requested):
We need to add a test to check that when credentials are not supplied the response is 401

jimmykarily (dismissed):

jimmykarily:
> We need to add a test to check that when credentials are not supplied the response is 401

I added a test.

PR #381, opened 2021-05-11
Remove cli dependencies that are no longer needed from README.md

jimmykarily:
- we no longer need git on the cli side (since #338)
- we no longer need sh (since #338)
- we no longer need openssl (since #298)


andreas-kupries (approved):
Easy enough, and looking correct as well.

PR #385, opened 2021-05-11
Always use cert-manager to generate workload certs

manno:
Tests didn't test cert-manager since it was only used in prod.

This adds the missing RBAC for the server to create certificates. 
I don't have a production env to test, but https://github.com/epinio/epinio/pull/360 moved the certificate creation into the server and didn't add permissions. Tests couldn't detect it.

Now dev/CI and prod are closer.

### DIFF
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -179,7 +179,7 @@ func createCertificates(ctx context.Context, cfg *rest.Config, app gitea.App) er
 			return errors.Wrap(err, "create production ssl certificate failed")
 		}
 	} else {
-		err = createCertificate(ctx, cfg, app, c.Domain, "selfsigned-issuers")
+		err = createCertificate(ctx, cfg, app, c.Domain, "selfsigned-issuer")
 		if err != nil {
 			return errors.Wrap(err, "create local ssl certificate failed")
 		}
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -17,7 +17,6 @@ import (
 	"github.com/tektoncd/pipeline/pkg/apis/resource/v1alpha1"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	corev1 "k8s.io/api/core/v1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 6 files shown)


internal/api/v1/stage.go
```diff
@@ -175,20 +174,20 @@ func createCertificates(ctx context.Context, cfg *rest.Config, app gitea.App) er
...
 	// else create a local cluster self-signed tls secret.
 	if !strings.Contains(c.Domain, "omg.howdoi.website") {
-		err = createProductionCertificate(ctx, cfg, app, c.Domain)
+		err = createCertificate(ctx, cfg, app, c.Domain, "letsencrypt-production")
 		if err != nil {
 			return errors.Wrap(err, "create production ssl certificate failed")
 		}
 	} else {
-		err = createLocalCertificate(ctx, cfg, app, c.Domain)
+		err = createCertificate(ctx, cfg, app, c.Domain, "selfsigned-issuers")
```
jimmykarily, line 182: Shouldn't this match this line: https://github.com/epinio/epinio/pull/385/files#diff-765c4fca879b550830d6c022bb21b9d959b31ca945b7b17853f64fc6fe1006deR244 ?

`selfsigned-issuers` -> `selfsigned-issuer` ?

internal/api/v1/stage.go
```diff
@@ -235,71 +234,3 @@ func createProductionCertificate(ctx context.Context, cfg *rest.Config, app gite
...
-					},
-					"commonName" : "%s.%s",
-					"isCA" : false,
-					"alternativeNames": [
-						"%s.%s"
-					],
-					"signerType" : "local"
-				}
-			},
-			"secretName" : "%s-tls",
```
jimmykarily, line 266: The application ingress [is still expecting a secret named like this](https://github.com/epinio/epinio/blob/main/internal/cli/clients/gitea/upload.go#L312) (e.g. `sample-tls`) but the new code generates a secret named something like `sample-6f2kd`. This creates an ingress that doesn't work. Visiting the app route with `https` results in a 404.

jimmykarily: Turns out the wrong name was due to the typo above (https://github.com/epinio/epinio/pull/385/files#r631744787). I fixed it and now I get the correct `*-tls` name for the generated secret. cert-manager seems to handle the wrong issuer name with some fallback (?). I wonder why no test failed. We should probably add a test that would fail if this happens again.

internal/api/v1/stage.go
jimmykarily: fixed

internal/api/v1/stage.go
jimmykarily: I added a test and fixed the issue

jimmykarily (approved):

PR #388, opened 2021-05-12
feat: drop ingress for tekton dashboard

### LINKED ISSUE #378: Tekton dashboard should be either not accessible or password protected
If people are deploying Epinio in production, we should not expose more services than we need to the public.

The Tekton dashboard is accessible to anybody without a password (Unless #376 changes that).

If we decide we want the dashboard to be available, we should make sure it's password protected and served over tls.

If we don't need it, we should not deploy it at all.

andreas-kupries:
Fixes #378

dashboard not needed by users, nor devs
keep private


jimmykarily:
I rebased on top of main and added  a commit to completely remove the dashboard installation.

jimmykarily (approved):

jimmykarily (approved):

PR #390, opened 2021-05-12
feat: remove remaining irrelevant client side org checks

### LINKED ISSUE #389: Remove client side `ensureGoodOrg`. Should all be server side now
We should remove the calls to `ensureGoodOrg` in the client.
These validations should by now all be done server side.

andreas-kupries:
Fixes #389

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -893,13 +893,6 @@ func (c *EpinioClient) Push(app, source string, services []string) error {
 		Timeout(duration.UserAbort()).
 		Msg("Hit Enter to continue or Ctrl+C to abort (deployment will continue automatically in 5 seconds)")
 
-	// todo: fix, remove, move to server
-	details.Info("validate")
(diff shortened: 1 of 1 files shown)


internal/cli/clients/client.go
```diff
@@ -1030,16 +1023,15 @@ func (c *EpinioClient) Target(org string) error {
...
-	details.Info("validate")
-	err := c.ensureGoodOrg(org, "Unable to target.")
-	if err != nil {
-		return err
-	}
+	// TODO: Validation of the org name removed. Proper validation
+	// of the targeted org is done by all the other commands using
+	// it anyway. If we really want it here and now, implement an
+	// `org show` command and API, and then use that API for the
+	// validation.
```
jimmykarily, line 1030: Although an org can be deleted after being targeted (by another user for example), I think it is good UX to let the user know that the org she is targeting doesn't exist. Let's keep this node around until we have the `org show` endpoint.

jimmykarily (approved):

PR #397, opened 2021-05-17
fix: regression for long app names breaking staging

### LINKED ISSUE #392: Don't use app name in staging pod names
This is a regression of #322 . We introduced the same bug here: https://github.com/epinio/epinio/blob/main/internal/api/v1/stage.go#L120 

When the app name is used together with the uid for a pod name, it limits the available characters by the number of characters in the uid. We have labels that allow us to filter the staging pods of the app, the pod doesn't need to have the app name in it.

andreas-kupries:
Fixes #392

Removed app name from pipeline run, uid is enough.
Keep it short

### DIFF
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -31,6 +31,7 @@ import (
 	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/epinio/epinio/internal/api/v1/models"
 	"github.com/epinio/epinio/internal/cli/clients/gitea"
+	"github.com/epinio/epinio/internal/names"
 )
 
 const (
@@ -222,7 +223,7 @@ func createCertificate(ctx context.Context, cfg *rest.Config, app models.App, sy
 			"namespace": "%s"
 		},
 		"spec": {
-			"commonName" : "%s.%s",
+			"commonName" : "%s",
 			"secretName" : "%s-tls",
 			"dnsNames": [
 				"%s.%s"
@@ -232,7 +233,7 @@ func createCertificate(ctx context.Context, cfg *rest.Config, app models.App, sy
 				"kind" : "ClusterIssuer"
 			}
 		}
-        }`, app.Name, app.Org, app.Name, systemDomain, app.Name, app.Name, systemDomain, issuer)
+        }`, app.Name, app.Org, names.TruncateMD5(fmt.Sprintf("%s.%s", app.Name, systemDomain), 64), app.Name, app.Name, systemDomain, issuer)
 
 	decoderUnstructured := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
 	obj := &unstructured.Unstructured{}
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -215,6 +215,22 @@ func createCertificates(ctx context.Context, cfg *rest.Config, app models.App) e
 }
 
 func createCertificate(ctx context.Context, cfg *rest.Config, app models.App, systemDomain string, issuer string) error {
+	// Notes:
+	// - spec.CommonName is length-limited.
+	//   At most 64 characters are allowed, as per [RFC 3280](https://www.rfc-editor.org/rfc/rfc3280.txt).
+	//   That makes it a problem for long app name and domain combinations.
+	// - The spec.dnsNames (SAN, Subject Alternate Names) do not have such a limit.
+	// - Luckily CN is deprecated with regard to DNS checking.
+	//   The SANs are prefered and usually checked first.
+	//
+	// As such our solution is to
+	// - Keep the full app + domain in the spec.dnsNames/SAN.
+	// - Truncate the full app + domain in CN to 64 characters,
+	//   replace the tail with an MD5 suffix computed over the
+	//   full string as means of keeping the text unique across
+	//   apps.
+
+	cn := names.TruncateMD5(fmt.Sprintf("%s.%s", app.Name, systemDomain), 64)
 	data := fmt.Sprintf(`{
 		"apiVersion": "cert-manager.io/v1alpha2",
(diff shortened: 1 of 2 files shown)


jimmykarily:
There is a 63 characters limit on labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

To make sure the app name length never breaks staging, I think the tests should be deploying an app with 63 characters long name.

andreas-kupries:
:construction: on it

andreas-kupries:
Hm.

Error with apname+uid:
```
    error pushing app to server: can't stage app: Internal Server Error: {"errors":[{"status":500,"title":"admission webhook \"validation.webhook.pipeline.tekton.dev\" denied the request: validation failed: Invalid resource name: length must be no more than 63 characters: metadata.name","details":"failed to create pipeline run: \u0026v1beta1.PipelineRun{TypeMeta:v1.TypeMeta{Kind:\"\", APIVersion:\"\"}, ObjectMeta:v1.ObjectMeta{Name:\"\", GenerateName:\"\", Namespace:\"\", SelfLink:\"\", UID:\"\", ResourceVersion:\"\", Generation:0, CreationTimestamp:v1.Time{Time:time.Time{wall:0x0, ext:0, loc:(*time.Location)(nil)}}, DeletionTimestamp:(*v1.Time)(nil), DeletionGracePeriodSeconds:(*int64)(nil), Labels:map[string]string(nil), Annotations:map[string]string(nil), OwnerReferences:[]v1.OwnerReference(nil), Finalizers:[]string(nil), ClusterName:\"\", ManagedFields:[]v1.ManagedFieldsEntry(nil)}, Spec:v1beta1.PipelineRunSpec{PipelineRef:(*v1beta1.PipelineRef)(nil), PipelineSpec:(*v1beta1.PipelineSpec)(nil), Resources:[]v1beta1.PipelineResourceBinding(nil), Params:[]v1beta1.Param(nil), ServiceAccountName:\"\", ServiceAccountNames:[]v1beta1.PipelineRunSpecServiceAccountName(nil), Status:\"\", Timeout:(*v1.Duration)(nil), PodTemplate:(*pod.Template)(nil), Workspaces:[]v1beta1.WorkspaceBinding(nil), TaskRunSpecs:[]v1beta1.PipelineTaskRunSpec(nil)}, Status:v1beta1.PipelineRunStatus{Status:v1beta1.Status{ObservedGeneration:0, Conditions:v1beta1.Conditions(nil), Annotations:map[string]string(nil)}, PipelineRunStatusFields:v1beta1.PipelineRunStatusFields{StartTime:(*v1.Time)(nil), CompletionTime:(*v1.Time)(nil), TaskRuns:map[string]*v1beta1.PipelineRunTaskRunStatus(nil), Runs:map[string]*v1beta1.PipelineRunRunStatus(nil), PipelineResults:[]v1beta1.PipelineRunResult(nil), PipelineSpec:(*v1beta1.PipelineSpec)(nil), SkippedTasks:[]v1beta1.SkippedTask(nil)}}}"}]}
```

Error with uid only.

```
Stage application ...
    error pushing app to server: can't stage app: Internal Server Error: {"errors":[{"status":500,"title":"create local ssl certificate failed: admission webhook \"webhook.cert-manager.io\" denied the request: spec.commonName: Too long: must have at most 64 bytes","details":""}]}
```

We seem to have limits somewhere else also.
Trying to find that limit.
Looks to be for some kind of cert.

andreas-kupries:
See https://github.com/epinio/epinio/blob/main/internal/api/v1/stage.go#L200

commonName, created from app name and system domain.
With an OMG domain the app name has to be quite short, I believe.
What is the common name for ? (I could try to limit)

andreas-kupries:
An OMG domain is 27 to 35 characters long, leaving 29 to 37 for the app.

However, it seems that the common name is for the primary domain, and the dnsNames are alternate names.
So, I could truncate an overly long common name and the subject alternate name should still provide the proper info for the tls validation.

Will attempt that now.

internal/api/v1/stage.go
```diff
@@ -207,7 +208,7 @@ func createCertificate(ctx context.Context, cfg *rest.Config, app gitea.App, sys
 				"kind" : "ClusterIssuer"
 			}
 		}
-        }`, app.Name, app.Org, app.Name, systemDomain, app.Name, app.Name, systemDomain, issuer)
+        }`, app.Name, app.Org, names.TruncateMD5(fmt.Sprintf("%s.%s", app.Name, systemDomain), 64), app.Name, app.Name, systemDomain, issuer)
```
jimmykarily, line 211: Can we put a comment (in code) on why we do this here? Something about the character limit in the CN field which is not a problem in the dnsNames. Maybe also mention that generally the CN field is deprecated and the SANs are preferred.

andreas-kupries: Will do.

andreas-kupries: Done with commit 4a778220aa55258c8507c6fe61944dfcf63632b8

andreas-kupries: And rebased due to conflict with new tekton staging work.

manno (approved):

internal/api/v1/stage.go
```diff
@@ -232,7 +249,7 @@ func createCertificate(ctx context.Context, cfg *rest.Config, app models.App, sy
 				"kind" : "ClusterIssuer"
 			}
 		}
-        }`, app.Name, app.Org, app.Name, systemDomain, app.Name, app.Name, systemDomain, issuer)
+        }`, app.Name, app.Org, cn, app.Name, app.Name, systemDomain, issuer)
```
manno, line 252: One more variable to that format string and we should turn it into a go template ;)

andreas-kupries: Well, I removed a variable ;) But yes, the counting is quite onerous.

PR #398, opened 2021-05-17
fix: Add missing template to info controller

### LINKED ISSUE #332: Server/Dashboard: Error showing "About" (/info) page
Installed Epinio to RKE2 cluster with `./epinio-0.0.11 install --system-domain epinio.paas.suse.de`.
Accessing "http://epinio.epinio.paas.suse.de/" works, showing the Epinio dashboard (https will not work).

If I click on "About" (-> http://epinio.epinio.paas.suse.de/info), I'll get the following error message:

`html/template:page_template:113:15: no such template "new-org-modal"`

and the server logs:

`2021/04/27 08:21:33 http: superfluous response.WriteHeader call from github.com/epinio/epinio/internal/web.handleError (applications_controller.go:163)`

This hould not happen, and the info page should be shown.

andreas-kupries:
Fixes #332

### DIFF
--- a/internal/web/info_controller.go
+++ b/internal/web/info_controller.go
@@ -34,5 +34,5 @@ func (hc InfoController) Index(w http.ResponseWriter, r *http.Request) {
 		"giteaVersion": giteaVersion,
 	}
 
-	Render([]string{"main_layout", "icons", "info"}, w, r, data)
+	Render([]string{"main_layout", "icons", "info", "modals"}, w, r, data)
 }


jimmykarily:
This PR fixes the error but UX-wise, something is still missing:
image

The info url doesn't have an org in it so the drop down shows no org selected and has no orgs to select from. The modal still works though. Anyway, that's a different issue, I'm approving this PR.

jimmykarily (approved):

PR #399, opened 2021-05-17
fix: restrict app service to app pods

### LINKED ISSUE #395: Service selector is overlybroad
It looks like the service description also points at the staging pods as well. I don't know if this would cause any issue but it might be worth refining the selector a bit

image

andreas-kupries:
Fixes #395


agracey (approved):
LGTM, simple fix!

PR #400, opened 2021-05-17
Wait for specific app to run

### LINKED ISSUE #246: If staging fails, the CLI doesn't stop waiting
I tried to run codimd in carrier, and the buildpack doesn't like the nodejs engine version causing the stage to fail. (I found this out via the tekton dashboard)

Due to this failure, the pipeline failed in the staging task but the dots never stopped on the CLI.

### LINKED ISSUE #357: Handle first and followup pushes the same way - wait for the app to be up[dated]
As the issue #262 demonstrated the user experience is different when an app is pushed for the first time in contrast to when it is updated later.

At the moment `epinio app push` does wait until the app is online when pushed for the first time, but when pushed again it directly exits so the user might expect the app to be updated while it is not - for quite some time.

So we should handle both cases the same way.

manno:
This PR stops a new staging run from happening, if one is already running for that image.

The CLI client now waits for the exact pipelinerun, that was triggered by the staging endpoint, to finish (fixes #246).

The CLI push commands waits for the app to start. Now it looks for an app, with the exact image ID from the upload endpoint (fixes #357).

- This should be improved so we wait for the right staging ID (from the staging endpoint), instead of the image ID. 
Even though the code now returns a random staging ID, we can't wait for it, because it's unknown by the time of the upload. This will change, once the `.kube` files are not created in upload anymore.

Also, minor changes:
* document logger
* add request/response structs for upload/staging endpoint
* remove app part from pipelinerun name

### DIFF
--- a/internal/api/v1/models/app.go
+++ b/internal/api/v1/models/app.go
@@ -1,0 +1,68 @@
+package models
+
+import "fmt"
+
+const (
+	EpinioImageIDLabel = "epinio.suse.org/image-id"
+	EpinioStageIDLabel = "epinio.suse.org/stage-id"
+)
+
+// App has all the app properties, like the image, repo, route and staging information
+type App struct {
+	AppRef
+	Image     ImageRef
+	Git       *GitRef
+	Route     string
+	Stage     StageRef
+	Instances int32
+}
+
+// NewApp returns a new app for name and org
+func NewApp(name string, org string) *App {
+	return &App{AppRef: AppRef{Name: name, Org: org}}
+}
+
+// GitURL returns the git URL by combining the server with the org and name
+func (a *App) GitURL(server string) string {
+	return fmt.Sprintf("%s/%s/%s", server, a.Org, a.Name)
+}
+
+// ImageURL returns the URL of the image, using the ImageID. The ImageURL is
+// later used in app.yml.  Since the final commit is not know when the app.yml
+// is written, we cannot use Repo.Revision
+func (a *App) ImageURL(server string) string {
+	return fmt.Sprintf("%s/%s-%s", server, a.Name, a.Image.ID)
+}
+
+// AppRef references an App by name and org
+type AppRef struct {
+	Name string `json:"name"`
+	Org  string `json:"org"`
+}
+
+// StageRef references a tekton staging run by ID, currently randomly generated
+// for each POST to the staging endpoint
+type StageRef struct {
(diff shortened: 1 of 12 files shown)


internal/api/v1/models/app.go
```diff
@@ -0,0 +1,68 @@
...
+	return &App{AppRef: AppRef{Name: name, Org: org}}
+}
+
+// GitURL returns the git URL by combining the server with the org and name
+func (a *App) GitURL(server string) string {
+	return fmt.Sprintf("%s/%s/%s", server, a.Org, a.Name)
+}
+
+// ImageURL returns the URL of the image, using the ImageID. The ImageURL is
+// later used in app.yml.  Since the final commit is not know when the app.yml
```
andreas-kupries, line 31: Typo: know / known.

internal/api/v1/models/models.go
```diff
@@ -1,7 +1,6 @@
-package models
-
-// The structures in this package encapsulate the requested and
+// Package models contains structures to encapsulate the requested and
```
andreas-kupries, line 1: IIRC you have to start with the package name for the go docs to work ?
Also:
```
- models contains structures to encapsulate the requested and 
+ models contains structures encapsulating the request and
```

internal/api/v1/stage.go
```diff
@@ -79,6 +80,25 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		return singleInternalError(err, "failed to get access to a tekton client")
 	}
 
+	// return if another run for this is imageID is running; one imageID == one stagingID at the same time
```
andreas-kupries, line 83:
```
- this is ImageID
+ this ImageID
```

andreas-kupries (changes requested):
Only a few comment typos and phrasing issues seen. Code looks ok.

internal/api/v1/models/models.go
manno: It's recommended to start with 'Package' https://golang.org/doc/effective_go#commentary

internal/api/v1/stage.go
manno: Thanks!

internal/api/v1/models/models.go
andreas-kupries: Ok

andreas-kupries (approved):

PR #401, opened 2021-05-17
Don't write .kube files in the app  source repo

### LINKED ISSUE #394: go sample app doesn't work
I was trying to find a sample app that maybe stages faster than the one we have in `assets/sample-app` to make the acceptance tests run faster.

The first application I tried was this: https://github.com/paketo-buildpacks/samples/blob/main/go/no-imports/main.go

but it fails at the `run` step with this error:

```
🕞  Now tracking tekton-staging sample2a75fb0001b01c303-run-v4mtg-pod-cv2hr › step-run ..
🕞  [APP] step-run /workspace/source/app 
🕞  [APP] step-run total 0 
🕞  [APP] step-run drwxr-xr-x    2 1000     1000             6 May 14 13:45 . 
🕞  [APP] step-run drwxrwxrwx    3 1000     1000            17 May 14 13:44 .. ..
🕞  [APP] step-run error: the path "./.kube/" does not exist .
🕞  [APP] affinity-assistant 2021/05/14 13:46:18 received signal: terminated 
```

(in the output above, I've added a `pwd` and an `ls -la` in the `run` step to see what's going on).

It seems that the app directory is empty for some reason so there is no `.kube` directory to use to apply the kube resources.

What I did:

- Created a `tmp/sample` directory at the root of Epinio sources dir
- Placed the linked file in that directory as `main.go`
- Run `go mod init github.com/epinio/sample` to create a go.mod and a go.sum file
- Pushed with `epinio push --verbosity 1 sample tmp/sample/`

jimmykarily:
Fixes #394


internal/cli/clients/gitea/upload.go
```diff
@@ -30,20 +27,21 @@ func (c *Client) Upload(app *models.App, tmpDir string) error {
...
 	if err != nil {
-		return err
+		return errors.Wrap(err, "failed to parse gitea url")
 	}
+	u.User = url.UserPassword(c.Username, c.Password)
+	u.Path = path.Join(u.Path, app.Org, app.Name)
 
-	// sets repo.revision
-	err = c.gitPush(app, tmpDir)
+	// mutates app and sets repo.revision
```
manno, line 37:
```suggestion
```

internal/cli/apps.go
```diff
@@ -86,6 +86,7 @@ var CmdAppShow = &cobra.Command{
 	},
 }
 
+// TODO: When pushing the app again after scaling the instances is reset (?)
```
manno, line 89: I think that's ok? The user executed the push with that number of instances. 

How do you tell the difference between default value and intentional change?
> epinio push --instances 1
> epinio push
(Well, yes, we could make instances a pointer and if it's nil try to do the right thing)

manno (approved):

internal/cli/apps.go
jimmykarily: I created a new issue for this: https://github.com/epinio/epinio/issues/403

I don't know what the implementation will look like, probably what you suggested with the pointer. In any case, if there is no values explicitly set for instances, we should maintain the current value.

acceptance/acceptance_suite_test.go
```diff
@@ -453,22 +453,22 @@ func checkDependencies() error {
...
-	}
-
-	fmt.Println("The cluster deployments: ")
-	out, err = helpers.Kubectl("get deployments --all-namespaces")
-	if err != nil {
-		fmt.Print(err.Error())
-	} else {
-		fmt.Print(out)
-	}
+	// fmt.Println("\nA test failed. You may find the following information useful for debugging:")
```
jimmykarily, line 456: I accidentally committed these changes. On the other hand, this output hasn't been very helpful lately, should we maybe remove it? I can keep the method as a placeholder for local debugging if we need to add debugging output and such.

PR #402, opened 2021-05-17
Add app logs cmd

### LINKED ISSUE #367: Application logs endpoint and command
The only time we show logs are during staging. There is no way for the user to read the application logs in the command line.

It's very common for developers to want to look at they application's logs to debug an issue or generally see what's going on.

## Acceptance Criteria

- There should be an endpoint in the API that streams application logs from all the application instances (pods).
- There should be a command to read the logs of the application. E.g. `epinio app logs sample`
- There should be an optional flag to stream the staging logs 
  e.g. `epinio app logs sample --staging`
- The logs from the various instances/pods should be tagged with the name of the pod or some unique identifier (e.g. a prefix: `[uuid_here] Listening on port 8080`)

rohitsakala:
fixes #367 

This PR only implements the basic functionality of the `app logs {appName}` command. It also implements the `--follow` option for the command. 

The next PR's will include
- Watch for addition and removal of Pods while log streaming is done
- color the output of the `app logs` command for different instances
- Implement the `--staging` option for the command

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -433,7 +433,7 @@ func Epinio(command string, dir string) (string, error) {
 	return RunProc(cmd, commandDir, false)
 }
 
-// GetKCommand retuns the KCommand without running it
+// GetKCommand returns the KCommand without running it
 func GetKCommand(command string, dir string) (*kexec.KCommand, error) {
 	var err error
 
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -443,7 +443,7 @@ func GetKCommand(command string, dir string) (*kexec.KCommand, error) {
 			return nil, err
 		}
 	}
-	cmd := fmt.Sprintf(nodeTmpDir+"/epinio-darwin-amd64 %s", command)
+	cmd := fmt.Sprintf(nodeTmpDir+"/epinio %s", command)
 
 	p := kexec.CommandString(cmd)
 	p.Dir = dir
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -32,7 +32,7 @@ func TestAcceptance(t *testing.T) {
 var nodeSuffix, nodeTmpDir string
 
 // serverURL is the URL of the epinio API server
-var serverURL string
+var serverURL, websocketURL string
 var registryMirrorName = "epinio-acceptance-registry-mirror"
 
 const (
@@ -193,6 +193,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
 
 	serverURL = "http://" + out
+	websocketURL = "ws://" + out
 })
 
 var _ = SynchronizedAfterSuite(func() {
@@ -430,6 +431,24 @@ func Epinio(command string, dir string) (string, error) {
 	cmd := fmt.Sprintf(nodeTmpDir+"/epinio %s", command)
 
 	return RunProc(cmd, commandDir, false)
+}
+
+// GetKCommand retuns the KCommand without running it
+func GetKCommand(command string, dir string) (*kexec.KCommand, error) {
+	var err error
+
+	if dir == "" {
+		dir, err = os.Getwd()
+		if err != nil {
+			return nil, err
+		}
+	}
+	cmd := fmt.Sprintf(nodeTmpDir+"/epinio-darwin-amd64 %s", command)
+
+	p := kexec.CommandString(cmd)
+	p.Dir = dir
+
+	return p, nil
 }
 
 func checkDependencies() error {
--- a/acceptance/api_v1_applications_test.go
+++ b/acceptance/api_v1_applications_test.go
@@ -10,11 +10,14 @@ import (
 	"net/http"
 	"os"
 	"path/filepath"
+	"regexp"
 	"strings"
+	"time"
 
 	v1 "github.com/epinio/epinio/internal/api/v1"
 	"github.com/epinio/epinio/internal/api/v1/models"
 	"github.com/epinio/epinio/internal/application"
+	"github.com/gorilla/websocket"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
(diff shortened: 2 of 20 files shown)


acceptance/apps_test.go
```diff
@@ -227,4 +228,60 @@ var _ = Describe("Apps", func() {
...
+
+		AfterEach(func() {
+			deleteApp(appName)
+		})
+
+		It("follows logs", func() {
+			p, err := GetKCommand("app logs --follow "+appName, "")
+			Expect(err).NotTo(HaveOccurred())
+
+			go p.Run()
```
jimmykarily, line 263: this is an external command, I wonder if it stops after tests are done. Maybe we should Kill that process manually.

internal/api/v1/applications.go
```diff
@@ -157,6 +172,165 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
 
 	return nil
 }
+func (hc ApplicationsController) Logs(w http.ResponseWriter, r *http.Request) APIErrors {
+	params := httprouter.ParamsFromContext(r.Context())
+	org := params.ByName("org")
```
jimmykarily, line 177: I think you need to check the validity of org and app before you start doing websocket things. Check how other actions return 404 when org or app don't exist.

internal/cli/apps.go
```diff
@@ -20,11 +20,19 @@ var CmdApp = &cobra.Command{
 }
 
 func init() {
+	flags := CmdAppLogs.Flags()
+	flags.Bool("follow", false, "follow the logs of the application")
+
+	flags = CmdAppUpdate.Flags()
+	flags.IntP("instances", "i", 1, "The number of instances the application should have")
+	cobra.MarkFlagRequired(flags, "instances")
```
jimmykarily, line 28: how did this change make it in this branch? Rebase issue?

jimmykarily: oh duplicate init (below)! all good.

internal/cli/clients/client.go
```diff
@@ -734,6 +741,74 @@ func (c *EpinioClient) AppUpdate(appName string, instances int) error {
...
+				if websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+					c.ui.Normal().Msg("-- Logs End --")
+					webSocketConn.Close()
+					done <- true
+					return
+				}
+				errorChan <- err
+				return
+			}
+			fmt.Println(string(message))
```
jimmykarily, line 789: I'm not sure we read line by line (https://github.com/epinio/epinio/pull/402/files#diff-168b4a07f391b22def39fd870b0373811deb114cd513793f9cf5a3182f8452ceR249), maybe we shouldn't print new line characters?

acceptance/acceptance_suite_test.go
```diff
@@ -426,6 +427,24 @@ func Epinio(command string, dir string) (string, error) {
 	return RunProc(cmd, commandDir, false)
 }
 
+// GetKCommand returns the KCommand without running it
+func GetKCommand(command string, dir string) (*kexec.KCommand, error) {
```
andreas-kupries, line 431: Could this be used to simplify the preceding `Epinio` function ?

acceptance/api_v1_applications_test.go
```diff
@@ -524,4 +527,76 @@ var _ = Describe("Apps API Application Endpoints", func() {
...
+
+	Context("Logs", func() {
+		Describe("GET api/v1/orgs/:orgs/applications/:app/logs", func() {
+			logLength := 0
+			var (
+				route string
+				app   string
+			)
+
+			BeforeEach(func() {
```
andreas-kupries, line 539: From the content of the function I am wondering if that should be an `It` instead of a `BeforeEach` ?!
If it is the latter, maybe hiding the complexity in a utility function ?

acceptance/apps_test.go
```diff
@@ -227,4 +228,60 @@ var _ = Describe("Apps", func() {
...
+			logLength      int
+		)
+
+		BeforeEach(func() {
+			appName = newAppName()
+			out := makeApp(appName, 1, true)
+			routeRegexp := regexp.MustCompile(`https:\/\/.*omg.howdoi.website`)
+			route = string(routeRegexp.Find([]byte(out)))
+
+			out, err := Epinio("app logs "+appName, "")
```
andreas-kupries, line 244: From here on the code feels again to be something more for `It` instead of `BeforeEach` ?!

acceptance/apps_test.go
andreas-kupries: Why should it stop ? AFAIK the `--follow` tells it to not stop. On the other hand, if the application it is following is deleted, then it will stop ? ... Checked, no, it does not stop even then.

So, right now `log --follow` can only be stopped by a manual Kill.

helpers/kubernetes/tailer/tailer.go
```diff
@@ -2,217 +2,112 @@ package tailer
...
-
-	colors := colorList[idx]
-	return colors[0], colors[1]
-}
-
-func originOf(podName string) string {
-	if strings.HasPrefix(podName, "staging-pipeline-run-") {
-		return "[STAGE]"
+// Start writes log lines to the logChan. It can be stopped using the ctx.
+// It's the calles responsibility to close the logChan (because there may be more
```
andreas-kupries, line 48: Nit
```
-calles
+callers
```

andreas-kupries (changes requested):
Nit, questions, and wondering if it would be possible to get a diagrammatic overview of the structures, go-routines and channels involved in all the logging.

internal/api/v1/applications.go
```diff
@@ -157,6 +166,133 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
...
+		if err != nil {
+			jsonErrorResponse(w, InternalError(err))
+		}
+		if app == nil {
+			jsonErrorResponse(w, AppIsNotKnown(appName))
+		}
+	}
+
+	if appName == "" && stageID == "" {
+		jsonErrorResponse(w, BadRequest(errors.New("You need to speficy either the stage id or the app")))
```
andreas-kupries, line 201:
```suggestion
		jsonErrorResponse(w, BadRequest(errors.New("You need to specify either the stage id or the app")))
```

andreas-kupries (changes requested):

acceptance/apps_test.go
jimmykarily: I added the `defer` above to manually kill it but when I simply run `p.Process.Kill()` I got an error because `p.Process` was nil. Maybe the library we use doesn't properly expose the underlying process for killing?

jimmykarily:
> Nit, questions, and wondering if it would be possible to get a diagrammatic overview of the structures, go-routines and channels involved in all the logging.

There is logging as the user sees it (e.g. the `epinio app logs` command), there is logging as the API user sees it (e.g. how do I connect to the websockets, what responses should I expect etc) and finally there is logging as we (epinio developers) see it.

I guess the epinio developer is going to face the most trouble among those 3. It would be nice to document this somehow but we don't have developer docs (yet). Generally we don't seem to need them because most of our code is straightforward but this is a special case. I will try to explain how the various methods should be consumed in comments (in code) on another PR. I will also address the remaining comments about test refactoring in a new PR. This one is green and working so let's merge it.

acceptance/apps_test.go
jimmykarily: p.Process it not nil after calling `p.Run()` so the handling we did is fine. Also fixed the order of things here: https://github.com/epinio/epinio/commit/bdffb6edbf4b293e8fc1dff39381b52bac9fcd40

acceptance/acceptance_suite_test.go
jimmykarily: Fixed here: https://github.com/epinio/epinio/commit/414aad727d43d33268d80c057487aed2b21aeacc

acceptance/apps_test.go
jimmykarily: In this case I think it's fine. This code simply calculates the number of lines already printed in order to be used later in the test. It is not testing anything in the BeforeEach block (except for unexpected error handling).

acceptance/api_v1_applications_test.go
jimmykarily: Done here: https://github.com/epinio/epinio/commit/30bc46fcf0e4734d83a0a47fb91c62266046c652

PR #405, opened 2021-05-18
Generate credentials

### LINKED ISSUE #379: Gitea service should be either password protected or not accessible
See also #378 and #376

If we decide we want to keep gitea accessible to the world, it should be protected (password, ssh keys etc) and served using encryption.

If not, then we shouldn't expose it at all with an ingress. `git push` is happening on the server side since v0.0.12 (#338) so we don't need to access it from the client side unless we want to support git ops.

NOTE: The gitea ingress is used in various places to detect what the app domain is so removing it will need changes to other parts of the code.

andreas-kupries:
Fixes, in part, #61 

Fixes #379


helpers/uuid/uuid.go
```diff
@@ -0,0 +1,23 @@
+package uuid
+
+import (
+	"crypto/rand"
+	"encoding/hex"
+	"hash/fnv"
+)
+
+func UUID() (string, error) {
```
manno, line 9: This is not a UUID as defined in several standards. https://en.wikipedia.org/wiki/Universally_unique_identifier.

Can we call this something else? I expect UUIDs to have a certain length and format.

andreas-kupries: Sure. ... Package rename to `randstr` and `UUID` becomes `Hex16`. Noted.

manno: Nice, that means it doesn't "stutter"   (https://golang.org/doc/effective_go#package-names) anymore. Instead of `uuid.UUID()`, we have `randstr.Hex16()` 👍

andreas-kupries: Done with commit 28f8364ac94c1a734c0d50c71034a6e3be51b337

deployments/registry.go
```diff
@@ -28,6 +29,23 @@ const (
...
+var registryAuthMemo *auth.PasswordAuth
+
+func RegistryInstallAuth() (*auth.PasswordAuth, error) {
+	if registryAuthMemo == nil {
+		auth, err := auth.RandomPasswordAuth()
+		if err != nil {
+			return nil, err
+		}
+		registryAuthMemo = auth
+		// registryAuthMemo = &auth.PasswordAuth{
```
jimmykarily, line 41: Can we remove these comments now?

deployments/tekton.go
```diff
@@ -360,18 +361,29 @@ func applyTektonStaging(c *kubernetes.Cluster, ui *termui.UI) (string, error) {
...
 func (k Tekton) createGiteaCredsSecret(c *kubernetes.Cluster) error {
-	_, err := c.Kubectl.CoreV1().Secrets(TektonStagingNamespace).Create(context.Background(),
+	// See internal/cli/clients/gitea/gitea.go, func
+	// `getGiteaCredentials` for where the cli retrieves the
+	// information for its own gitea client.
+	//
+	// See deployments/gitea.go func `apply` where `install`
+	// configures gitea for the same credentials.
+
+	giteaAuth, err := GiteaInstallAuth()
```
jimmykarily, line 371: The memoization of this method will be a problem if we ever want to support resuming an incomplete installation but let's solve this problem when (and if) we get there.

helpers/randstr/randstr.go
```diff
@@ -0,0 +1,23 @@
+package randstr
```
jimmykarily, line 1: Maybe add a comment on what this package is all about?

internal/cli/clients/client.go
```diff
@@ -580,12 +573,14 @@ func (c *EpinioClient) Info() error {
...
 	giteaVersion := "unavailable"
 
-	version, resp, err := c.GiteaClient.Client.ServerVersion()
-	if err == nil && resp != nil && resp.StatusCode == 200 {
-		giteaVersion = version
-	}
+	// version, resp, err := c.GiteaClient.Client.ServerVersion()
+	// if err == nil && resp != nil && resp.StatusCode == 200 {
+	// 	giteaVersion = version
+	// }
```
jimmykarily, line 583: delete the comments? The TODO above should be enough to fix this.

deployments/registry.go
andreas-kupries: oops. yes

deployments/tekton.go
andreas-kupries: Agreed. My handling of a partial install is to `uninstall` and then start fresh.

internal/cli/clients/client.go
andreas-kupries: Right. I left quite a bit in regarding the gitea version for info ...

helpers/randstr/randstr.go
andreas-kupries: Yes

deployments/registry.go
andreas-kupries: Done, 2ceb59f79c32d91a623818ebb927e5e855a6f709

helpers/randstr/randstr.go
andreas-kupries: Done, 2ceb59f79c32d91a623818ebb927e5e855a6f709

internal/cli/clients/client.go
andreas-kupries: Done, 2ceb59f79c32d91a623818ebb927e5e855a6f709

andreas-kupries:
Also rebased to resolve the go.mod conflict.

jimmykarily (approved):

PR #406, opened 2021-05-18
Really exit on first pipelinerun failure

manno:
Previous PR did allow waitForApp to start, even though the pipelinerun failed.

Also no longer prints progress dots into the staging log, when using 'verbosity=1'.
Also cleans up client.Push() a little, by moving all push related funcs into a separate file 'push.go'.

This depends on https://github.com/epinio/epinio/pull/401

### DIFF
--- a/internal/cli/clients/push.go
+++ b/internal/cli/clients/push.go
@@ -159,6 +159,13 @@ func (c *EpinioClient) waitForPipelineRun(app models.AppRef, id string) error {
 				return false, nil
 			}
 			for _, pr := range l.Items {
+				// any failed conditions, throw an error so we can exit early
+				for _, c := range pr.Status.Conditions {
+					if c.IsFalse() {
+						return false, errors.New(c.Message)
+					}
+				}
+				// it worked
 				if pr.Status.CompletionTime != nil {
 					return true, nil
 				}
--- a/internal/cli/clients/push.go
+++ b/internal/cli/clients/push.go
@@ -17,6 +17,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/mholt/archiver/v3"
 	"github.com/pkg/errors"
+	"github.com/spf13/viper"
 	tekton "github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


internal/cli/clients/push.go
```diff
@@ -0,0 +1,198 @@
...
+	// returns staging ID
+	stage := &models.StageResponse{}
+	if err := json.Unmarshal(b, stage); err != nil {
+		return nil, err
+	}
+
+	return stage, nil
+}
+
+func (c *EpinioClient) logs(app models.AppRef, stageID string) (context.CancelFunc, error) {
```
jimmykarily, line 96: I guess in #402 this will have to be changed to `staging-logs` or something. Let's figure it out there though.

internal/cli/clients/push.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+			// pr exists, but still running
+			return false, nil
+		})
+}
+
+func (c *EpinioClient) waitForApp(app models.AppRef, id string) error {
+	c.ui.ProgressNote().KeeplineUnder(1).Msg("Creating application resources")
+	err := c.KubeClient.WaitUntilPodBySelectorExist(
+		c.ui, app.Org, fmt.Sprintf("app.kubernetes.io/name=%s,%s=%s", app.Name, models.EpinioStageIDLabel, id),
```
jimmykarily, line 181: Shouldn't this selector make use of this additional label? https://github.com/epinio/epinio/pull/399/files

Also the selector below (line 109) needs to filter the app pod. Now it's matching the staging pods too.

Otoh, the application should be considered "Running" when the deployment rollout is finished. I'm not sure what happens if the application has more than 1 pods. Does this code wait for all of them?

manno: I thought the selector in 109 wants to match the staging pod's logs? Not sure, I just moved the code to the new file.

I agree, waitForApp should wait for the deployment. The existing code waits for the pod to exist, then it waits for it to be running. (I think that could be one check.)
It checks all the pods that exist, so if the deployment wants ten pods and only 5 are there, it will check those 5.
Making sure the deployment is available would be the better check? https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment

Maybe, we do that in another story?

jimmykarily: yes let's do that in another story

jimmykarily (approved):

PR #409, opened 2021-05-19
Retry applying Tekton resources when webhook fails

manno:
The pod becomes ready and receives traffic, before the https service works.

If we try to apply a Tekton resource immediately after the pod is ready,
the service will fail. This happens even though all components are ready. The pod never receives any request.

> `Post "https://tekton-pipelines-webhook.tekton.svc:443/defaulting?timeout=10s": dial tcp 10.43.110.8:443: connect: connection refused`

Tried workarounds:
* wait for deployment to be available
* wait for webhook-cert secret to appear and have the right keys
* remove HorizontalPodAutoscaler
* add a startup probe
* point probes to https ports
* change probe endpoint to admissionwebhooks path (`/defaulting`)

The only thing that worked, was waiting ~15s before applying the resources.

I also tried minikube on Mac (virtualbox driver). That works as expected, even though it had less resources. The pod becomes ready and is immediately able to serve admission requests.

It's also working with k3d on Linux and more specifically k3d on Mac using a remote Docker on Linux via `DOCKER_HOST`.

I suspect the bug only happens with Docker-Desktop on Mac with K3d.
Might be related to: https://github.com/tektoncd/pipeline/issues/2207

### DIFF
--- a/deployments/cert_manager.go
+++ b/deployments/cert_manager.go
@@ -126,6 +126,11 @@ func (cm CertManager) Delete(c *kubernetes.Cluster, ui *termui.UI) error {
 		return errors.Wrapf(err, "Failed deleting namespace %s", CertManagerDeploymentID)
 	}
 
+	err = c.WaitForNamespaceMissing(ui, CertManagerDeploymentID, cm.Timeout)
+	if err != nil {
+		return errors.Wrap(err, "failed to delete namespace")
+	}
+
 	ui.Success().Msg("CertManager removed")
 
 	return nil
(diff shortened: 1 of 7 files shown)


andreas-kupries (approved):
Looks good. And the uninstall is now pretty blazing, compared to before.

  - Unable to fully test the new retry logic, as the issue does not happen on my machine.
    Code looks ok tough.

  - Watching the uninstall I believe I saw `epinio` to exit before the namespace for the service catalog stuff was gone. That may need an additional WaitForNamespaceMissing. The extra broker pieces may also, I did not have them active when I tested.

    I am ok with having that in a separate PR tough.

andreas-kupries:
Confirmed the behavior and created ticket #412 for it.

PR #414, opened 2021-05-21
Updating workload keeps replica count

### LINKED ISSUE #403: Scale app and then restage resets instances
I didn't check but I'm 99% sure this is the case (because I worked on that feature and I don't remember doing anything about it). 

If you scale an app to more than 1 instance and then re-stage it (without providing the instance number again), it will be reset to 1 instance. That's because we simply re-apply the deployment with `kubectl apply -f` and that resets it to the new version.

Let's check if that happens and fix it. There is a possibility that this works for some reason and this issue can simply be closed. Let's check.

manno:
Fixes #403 

`epinio push`  keeps workload's replica count if not specified


internal/cli/apps.go
```diff
@@ -86,8 +87,8 @@ var CmdAppShow = &cobra.Command{
 	},
 }
 
-// TODO: When pushing the app again after scaling the instances is reset (?)
-// CmdAppShow implements the epinio `apps show` command
+// CmdAppUpdate is used by the epinio `apps show` command to show information
```
rohitsakala, line 90: This should be `apps update` command instead of `apps show` command.

internal/api/v1/applications.go
```diff
@@ -135,22 +134,13 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
...
-	}
-
-	instances, err := strconv.Atoi(updateRequest.Instances)
-	if err != nil {
-		return APIErrors{BadRequest(err, "")}
-	}
-
-	if instances < 0 {
+	if updateRequest.Instances < 0 {
 		return APIErrors{NewAPIError(
```
andreas-kupries, line 138: Could this use the `BadRequest` utility function ? (See sibling file `errors.go`).

internal/api/v1/stage.go
```diff
@@ -62,7 +64,7 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		return singleNewError("name parameter from URL does not match name param in body", http.StatusBadRequest)
 	}
 
-	if req.Instances < 0 {
+	if req.Instances != nil && *req.Instances < 0 {
 		return APIErrors{NewAPIError(
```
andreas-kupries, line 68: Ditto `BadRequest` utility usable here ?

internal/api/v1/stage.go
```diff
@@ -100,11 +102,22 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
...
 	}
 
+	// find out the instances
+	var instances int32
+	if req.Instances != nil {
+		instances = int32(*req.Instances)
+	} else {
+		instances, err = existingReplica(ctx, cluster.Kubectl, req.App)
+		if err != nil {
+			return singleError(err, http.StatusInternalServerError)
```
andreas-kupries, line 112: Maybe use `singleInternalError` ?

internal/cli/apps.go
andreas-kupries: And the description is for `show` also, i.e. incorrect for `update`.

internal/cli/push.go
```diff
@@ -48,6 +50,23 @@ func init() {
 	})
 }
 
+// instances checks if the user provide instances. If they don't, then we'll
```
andreas-kupries, line 53:
```suggestion
// instances checks if the user provided an instance count. If they didn't, then we'll
```

internal/cli/push.go
```diff
@@ -48,6 +50,23 @@ func init() {
...
+// pass nil and eihter use the default or whatever is deployed in the cluster.
```
andreas-kupries, line 54:
```suggestion
// pass nil and either use the default or whatever is deployed in the cluster.
```

internal/cli/push.go
```diff
@@ -48,6 +50,23 @@ func init() {
...
+func instances(cmd *cobra.Command) (*int32, error) {
+	var i *int32
+	instances, err := cmd.Flags().GetInt32("instances")
+	if err != nil {
+		return i, errors.Wrap(err, "could not read instances parameter")
+	}
+	cmd.Flags().Visit(func(f *pflag.Flag) {
+		if f.Name == "instances" {
+			n := int32(instances)
+			i = &n
```
andreas-kupries, line 64: We seem to search here the flag set ?! I do not understand why.
Can we not directly use lines 63-64 without the search ?

andreas-kupries (changes requested):
Some typos, and a number of questions.

internal/cli/apps.go
manno: Ah, I only fixed half of the message...

internal/api/v1/stage.go
manno: I do want to change how errors are handled in the API and have some ideas, but changing line 68 is not part of this PR.

internal/api/v1/stage.go
manno: No, that needs a message.

internal/cli/push.go
manno: I need a spellchecker

internal/api/v1/stage.go
andreas-kupries: The message can be empty

internal/cli/push.go
manno: The user might not have provided the flag.
This PR changes the instances argument to push to a pointer, so we can later decide if we should use the default value, or the value from the cluster.
The user could explicitly push with the default value, e.g. to scale down. So we can't match on the default number:
"1" could be intentional, or just the default from flag.

There are no optional arguments in flag,`Visit()` solves that problem, it visits only those flags that have been set.

andreas-kupries: Ah. Now I understand. Maybe add a small comment about this ?

andreas-kupries (approved):

PR #415, opened 2021-05-21
feat: use gitea connection credentials for the gitea database as well

andreas-kupries:
Fixes, in part, #61

The gitea database now uses the same dynamic random credentials as network access to gitea itself.

### DIFF
--- a/deployments/gitea.go
+++ b/deployments/gitea.go
@@ -188,7 +188,17 @@ gitea:
     oauth2:
       ENABLE: true
       JWT_SECRET: HLNn92qqtznZSMkD_TzR_XFVdiZ5E87oaus6pyH7tiI
-`, subdomain, GiteaPort, giteaAuth.Username, giteaAuth.Password, subdomain, GiteaProtocol+"://"+subdomain)
+
+postgresql:
+  global:
+    postgresql:
+      postgresqlDatabase: epinio-gitea
+      postgresqlUsername: "%s"
+      postgresqlPassword: "%s"
+`, subdomain, GiteaPort,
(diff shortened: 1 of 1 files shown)


manno (approved):

deployments/gitea.go
```diff
@@ -188,7 +188,17 @@ gitea:
     oauth2:
       ENABLE: true
       JWT_SECRET: HLNn92qqtznZSMkD_TzR_XFVdiZ5E87oaus6pyH7tiI
```
jimmykarily, line 190: yet another hardcoded secret? I also see and `INTERNAL_TOKEN` some lines above.

jimmykarily: oh, it's already in the description of #61 . So, work in progress.

andreas-kupries: Yes. However it is unclear where they are used. I originally them linked to the regular gitea credentials, i.e. derived from them. Apparently not, as changing the gitea user/password does not require these to change to keep things working.

We could change/remove them and then see what breaks, if any.

jimmykarily: These are passed directly to gitea as config (https://docs.gitea.io/en-us/config-cheat-sheet/#oauth2-oauth2).

andreas-kupries: Do we actually use the oauth parts of gitea ?

andreas-kupries: See also https://docs.gitea.io/en-us/config-cheat-sheet/#security-security for the `INTERNAL_TOKEN`.

andreas-kupries: Now testing which of these we can do without.

PR #416, opened 2021-05-21
Makefile and Releases

manno:
This adds a new task for the release pipeline 'build-releases'.
The target just generates all embedded files and builds all the binaries.
It does not call 'lint' to validate and/or modify the source.

Also changes the names of the workflows to match their filename.
And bumps the Go version used for building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,6 +12,9 @@ build-all: embed_files lint build-amd64 build-arm64 build-arm32 build-windows bu
 build-all-small:
 	@$(MAKE) LDFLAGS+="-s -w" build-all
 
+build-releases:
+	@$(MAKE) LDFLAGS+="-s -w" -o lint embed_files build-all
+
 build-arm32: lint
 	GOARCH="arm" GOOS="linux" CGO_ENABLED=$(CGO_ENABLED) go build $(BUILD_ARGS) -ldflags '$(LDFLAGS)' -o dist/epinio-linux-arm32
 
--- a/images/Dockerfile
+++ b/images/Dockerfile
@@ -12,7 +12,7 @@ WORKDIR /go/src/github.com/epinio
 COPY . .
 COPY --from=tools /helm /usr/bin/helm
 
-RUN make build-all-small
+RUN make -o lint embed_files build-amd64 LDFLAGS="-s -w"
 
 FROM $BASE_IMAGE
 COPY --from=build /go/src/github.com/epinio/dist/epinio-linux-amd64 /epinio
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -12,6 +12,9 @@ build-all: embed_files lint build-amd64 build-arm64 build-arm32 build-windows bu
 build-all-small:
 	@$(MAKE) LDFLAGS+="-s -w" build-all
 
+build-releases:
+	@$(MAKE) LDFLAGS+="-s -w" -o lint embed_files build-all
```
jimmykarily, line 16: build-all is the same as build-all-small but building smaller binaries. Also, build-all (and build-all-small) both run line and embed_files so no need to include it here?

jimmykarily:
`helm` was missing from the docker image that runs `make build` not from the Github Action runner. I tried this PR's approach but reverted it because it didn't work. I ended up doing this instead: https://github.com/epinio/epinio/compare/faede315af7336ab1e3d7a3af3c2a32927ce4cf7..a764daad47769c1adf29f1651479d201f67e59fc . 

It's a bit ugly but our tools scripts have too many requirements, they need ruby they need python venv etc. I chose to call then in a different "stage" and simply copy over helm on the go container that builds the binaries. We should clean this up after cleaning up our tools scripts.

Makefile
manno: This one does not run lint, because of `-o  lint`.

We don't run tests in that pipeline, why would we run all the things `lint` is doing? Well, except, to my confusion, `lint` actually runs:

1. `go fmt`, to format the code (I think that's provided by the Go version from the workflow).
1. `go mod tidy`, which will modify our go.mod (using the Go version from the workflow, not the one we tested earlier).
1. as a dependency of `vet`, everything that's in `embed_files`, which is needed for compiling.
1.  `go vet`, even though we don't run tests in this pipeline.

So my point is, most of those we don't want in this pipeline, if we want them, we probably want them in another order: generate, fmt, tidy, vet

images/Dockerfile
```diff
@@ -12,7 +12,7 @@ WORKDIR /go/src/github.com/epinio
 COPY . .
 COPY --from=tools /helm /usr/bin/helm
 
-RUN make build-all-small
+RUN make -o lint embed_files build-amd64 LDFLAGS="-s -w"
```
manno, line 15: I couldn't find a nice way to put this into the existing Makefile targets.

Makefile
jimmykarily: Oh I missed the `-o` part. Gotcha.

manno:
We discussed this and will try to have a flat dependency hierarchy for the Makefile in the future.

jimmykarily:
moved to #451

PR #420, opened 2021-05-25
feat: enable user to force non-deployment of traefik

### LINKED ISSUE #380: Skip traefik deployment flag
In some cases, the Epinio user may have already installed Traefik in way we can't detect (we only look in the `kube-system` namespace).

In those cases, we want to allow the user to manually skip the installation of traefik from the Epinio side. A flag `--skip-traefik` should be enough for that.

## Accetance Criteria

- There is a flag `epinio install --skip-traefik` that skips the traefik deployment in exactly the same way it does when it detects traefik already installed.

andreas-kupries:
Fixes #380 

epinio's detection method is simplicistic and limited.
further, fixed termui bool display

### DIFF
--- a/deployments/traefik.go
+++ b/deployments/traefik.go
@@ -136,7 +136,16 @@ func (k Traefik) GetVersion() string {
 
 func (k Traefik) Deploy(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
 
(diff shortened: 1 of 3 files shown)


manno (approved):
Why do the options use snake_case, when the command line arguments are kebab-case... 🤷

andreas-kupries:
It started with `system_domain` and I followed now. I do not remember the why any longer.

PR #421, opened 2021-05-25
feat: wait for app's deployment, not pod(s).

### LINKED ISSUE #408: Wait for app should wait for the deployment to be available instead
The application should be considered "Running" when the deployment rollout is finished.

This also removes an existing problem with the selectors not being strict enough.

_Originally posted by @jimmykarily in https://github.com/epinio/epinio/pull/406#discussion_r635049422_

andreas-kupries:
Fixes #408 

This lets kube wait for the pod(s).
And is correct for --instances > 1.

### DIFF
--- a/helpers/kubernetes/cluster.go
+++ b/helpers/kubernetes/cluster.go
@@ -18,6 +18,7 @@ import (
 	minikube "github.com/epinio/epinio/helpers/kubernetes/platform/minikube"
 	"github.com/epinio/epinio/helpers/termui"
 
+	appsv1 "k8s.io/api/apps/v1"
 	apibatchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/api/extensions/v1beta1"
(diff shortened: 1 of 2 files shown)


helpers/kubernetes/cluster.go
```diff
@@ -260,6 +261,32 @@ func (c *Cluster) WaitForJobCompleted(namespace, jobName string, timeout time.Du
 	return wait.PollImmediate(time.Second, timeout, c.IsJobCompleted(client, jobName, namespace))
 }
 
+// IsDeploymentCompleted returns a condition function that indicates whether the given
+// Deployment is in Completed state or not.
+func (c *Cluster) IsDeploymentCompleted(deploymentName, namespace string) wait.ConditionFunc {
+	return func() (bool, error) {
+		deployment, err := c.Kubectl.AppsV1().Deployments(namespace).Get(context.Background(),
```
manno, line 268: I feel I should note that from what I saw, idiomatic Go uses shorter variable names, than it's normally the case in Epinio.

* https://github.com/golang/go/wiki/CodeReviewComments#variable-names
* https://talks.golang.org/2014/names.slide#19

manno (approved):

helpers/kubernetes/cluster.go
```diff
@@ -260,6 +261,32 @@ func (c *Cluster) WaitForJobCompleted(namespace, jobName string, timeout time.Du
...
+			if condition.Type == appsv1.DeploymentAvailable && condition.Status == v1.ConditionTrue {
+				return true, nil
+			}
+		}
+		return false, nil
+	}
+}
+
+func (c *Cluster) WaitForDeploymentCompleted(ui *termui.UI, namespace, deploymentName string, timeout time.Duration) error {
+	s := ui.Progressf("Starting %s in %s", deploymentName, namespace)
```
jimmykarily, line 284: I'm late to the party but anyway, this method could be used when updating the Deployment (thus not "Starting" but "Updating") or simply in a place where we don't want to print anything. I think the caller should decide whether to print any text or not because the caller has better context. This method shouldn't need a `termui.UI`.

PR #430, opened 2021-05-28
Ignore verbosity when printing staging logs

jimmykarily:
It was only used to decide whether to print a message and then dots.
Since we always print the staging logs now, we shouldn't be printing any
dots. Also the message reveals Epinio internals (pipelinerun) and is not
useful to the user.


andreas-kupries (approved):
:+1:

PR #443, opened 2021-05-31
Extended `app logs` to support `--staging`

### LINKED ISSUE #431: `epinio app logs --staging` command
The user should be able to see the latest staging logs for an application with this command.

The command should find the latest PipelineRun, extract the stageID from it and stream the logs for that staging process.

andreas-kupries:
Fixes #431.

Beyond support for `epinio app logs --staging` this also implements:

  - Delete pipelinerun(s) for app when deleting app
  - Delete old pipelinerun(s) for app when pushing new revision for app

TODO: Documentation for --staging, and tests.

### DIFF
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -30,6 +30,7 @@ import (
 	"github.com/epinio/epinio/helpers/randstr"
 	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/epinio/epinio/internal/api/v1/models"
+	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/cli/clients/gitea"
 	"github.com/epinio/epinio/internal/domain"
 	"github.com/epinio/epinio/internal/names"
@@ -118,6 +119,11 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		Git:       req.Git,
 		Route:     req.Route,
 		Instances: instances,
+	}
+
+	err = application.Unstage(name, org, uid)
+	if err != nil {
+		return singleInternalError(err, "failed delete previous pipeline runs")
 	}
 
 	pr := newPipelineRun(uid, app)
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -255,7 +255,7 @@ func Delete(kubeClient *kubernetes.Cluster, gitea GiteaInterface, org string, ap
 		return err
 	}
 
-	err = unstage(app, "")
+	err = Unstage(app.Name, app.Organization, "")
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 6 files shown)


internal/application/application.go
```diff
@@ -342,3 +365,42 @@ func (al ApplicationList) Swap(i, j int) {
...
+		LabelSelector: fmt.Sprintf("app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s",
+			app, org),
+	})
+	if err != nil {
+		return err
+	}
+
+	for _, pr := range l.Items {
+		if stageIdCurrent != "" && stageIdCurrent == pr.ObjectMeta.Name {
+			fmt.Printf("skip\n")
```
manno, line 395: This looks like a leftover from debugging?

internal/cli/clients/client.go
```diff
@@ -695,13 +695,32 @@ func (c *EpinioClient) AppShow(appName string) error {
...
+		WithTableRow("StageId", app.StageID).
 		WithTableRow("Routes", strings.Join(app.Routes, ", ")).
 		WithTableRow("Services", strings.Join(app.BoundServices, ", ")).
 		Msg("Details:")
 
 	return nil
 }
 
+// AppStageID returns the stage id of the named app, in the targeted org
+func (c *EpinioClient) AppStageId(appName string) (string, error) {
```
manno, line 707: `ID` should be capitalized (https://github.com/golang/go/wiki/CodeReviewComments#initialisms).

internal/api/v1/stage.go
```diff
@@ -120,6 +121,11 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		Instances: instances,
 	}
 
+	err = application.Unstage(name, org, uid)
```
manno, line 124: I can't remember does this also remove the workload? So would this interfere with zero-downtime?

manno (changes requested):

internal/application/application.go
andreas-kupries: Yes, d*t. Will remove.

internal/cli/clients/client.go
andreas-kupries: Right. Will fix.

internal/api/v1/stage.go
andreas-kupries: `Unstage` is new. It removes the PipelineRuns. The workload did not look affected in my tests, watching the pod stati.

internal/application/application.go
andreas-kupries: Done with 6e0d35f397412cbd7143acd3450e98600911dd05

internal/cli/clients/client.go
andreas-kupries: Done with 6e0d35f397412cbd7143acd3450e98600911dd05

manno (approved):

internal/api/v1/stage.go
jimmykarily: I was thinking that the old pipeline run (and staging pods) should be deleted only when the new staging succeeds. The reason is that, if the new staging process fails, the old version of the app will keep running and the user may decide to stick with that version for a while (e.g. until they fix the new version which fails). In that case, a user could still ask for the staging logs of the running version. This line would have deleted those and they won't be available anymore.

In other words, only get rid of old staging when you got a new successful one.

andreas-kupries: The new code should be movable to a place where we know that the new staging has succeeded or not.
That said, where (in the server) do we know when staging of an app has succeeded ?

jimmykarily: Right now I think it's on the client only. I created this spike to see if there is a way to achieve this on the server side: https://github.com/epinio/epinio/pull/443#discussion_r644510047

PR #445, opened 2021-06-01
Secure Epinio API ingress with TLS

### LINKED ISSUE #439: Epinio API over TLS
Follow up of #130 

Epinio CLI (or other consumers) should talk to the Epinio API over TLS. It could be either plain TLS (like apps have) or mTLS. In any case, this story is about encryption and not authentication. For the latter we have basic auth and that's good for now. 

## Acceptance Criteria

- The epinio API endpoint is secured with TLS
- The API is not accessible without TLS (disable http).

andreas-kupries:
Fixes #439


andreas-kupries:
Notes, issues:

  - cli does not actually work. It talks to the ingress, sure. Then fails with

    > error creating org: Post "https://epinio.192.168.192.2.omg.howdoi.website/api/v1/orgs": x509: certificate signed by unknown authority

    The basic created cert is self-signed, and the cli side cannot fully verify.
    I suspect that I have to store the ca cert in the config and make it available to the various clients.
    It should be ok for LE-signed certs.

  - Also broken, the `/` for the web dashboard. Here the error is

    ```
    2021/06/01 13:44:28 epinio/953 "level"=2 "msg"="received request"  "method"="GET" "uri"="/"
    2021/06/01 13:44:28 epinio/953 "level"=3 "msg"="request"  "body"="n/a"
    2021/06/01 13:44:28 http: panic serving 10.42.0.6:47672: no current org
    ```

    I suspect that this is just a consequence of the first issue.
    Because the first issue prevented the creation of the initial org.

andreas-kupries:
I can get install errors like 

> panic: Deployment epinio failed with error: Creating Epinio server cert failed: creation of local ssl certificate failed: Internal error occurred: failed calling webhook "webhook.cert-manager.io": Post "https://cert-manager-webhook.cert-manager.svc:443/mutate?timeout=10s": x509: certificate signed by unknown authority

despite the new code in epinio/server to wait for cert manager

andreas-kupries:
Confirmed with a hack disabling cert verification that this squashes the noted issues with cli and dashboard.

Based on https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/
I believe I have a way of going forward here:

  1. When a self-signed cert is made the CA is stored into the config.
  2. On reading the config, make the CA cert available, if present.

This means that for LE certs nothing should be in the config.
And we still have full verify for the self-signed, instead of disabling verification.

andreas-kupries:
Notes 
  - Manually running install of a dev deployment works. Script
    ```
    #!/bin/bash
    export EPINIO_DONT_WAIT_FOR_DEPLOYMENT=true
    epinio install --skip-default-org "$@"
    make patch-epinio-deployment
    sleep 30
    epinio org create workspace
    epinio target workspace
    ```

Adding tracing I see

```
Creating organization...
Name: workspace
2021/06/02 13:32:40 EpinioClient "level"=1 "msg"="POST https://epinio.172.21.0.2.omg.howdoi.website/api/v1/orgs"  
2021/06/02 13:32:40 EpinioClient "level"=2 "msg"="{ \"name\": \"workspace\" }"  
```

and the org is created, etc.

Doing the acceptance tests the creation of the org OTOH fails, even after the patched server is online. Page not found
```
Creating organization...
Name: workspace
2021/06/02 13:24:03 EpinioClient "level"=1 "msg"="POST https://epinio.172.20.0.3.omg.howdoi.website/api/v1/orgs"  
2021/06/02 13:24:03 EpinioClient "level"=2 "msg"="{ \"name\": \"workspace\" }"  
2021/06/02 13:24:03 EpinioClient/CreateOrg "level"=1 "msg"="return" "Organization"="workspace" 
error creating org: Not Found: 404 page not found
```

This was for a cluster created by the tests.

Creating a cluster manually and pointing the tests at it the org is created just fine.
It still fails, although later, the mini broker services are not coming up.

But for the base installation the cluster created by the tests seems to have differences to a cluster created manually which breaks access to the secured API.

I will now test minibroker setup for a manual cluster,/

andreas-kupries:
Extended the waiting period for the failing command to inspect the cluster a bit.
Results:
> % curl -k -u test-user:secure-testing http://epinio.172.20.0.3.omg.howdoi.website/api/v1/info ; echo
{"Version":"v0.0.14-9-g76544c9"}
% curl -k -u test-user:secure-testing https://epinio.172.20.0.3.omg.howdoi.website/api/v1/info ; echo
404 page not found

I.e. the server is accessible without TLS, and not with TLS.
The ingress looks ok to me.
[EAT_ingress.yaml.txt](https://github.com/epinio/epinio/files/6584867/EAT_ingress.yaml.txt)

(to come - same data for a manually created cluster)

andreas-kupries:
:bulb: I am doing my local runs using the traefik coming with k3d. That is v1.
The acceptance tests OTOH are disabling the traefik in k3d, forcing the use of traefik coming with epinio. That is v2.

So, my TLS work sort of works for Traefik V1 and does not at all for V2.
(Sort of, because the http is still open, bypassing security)

Note, given that my TLS work replicates the ingress setup done for apps, it stands to reason that
  1. apps are still accessible via http, not just https
  2. app are not accessible by https for V2.

That should be cross-checked.

andreas-kupries:
Confirmed with a manually created cluster and forced traefik v2 use, that is the cause for my 404 issues.
I now have to investigate traefik docs to see how v2 tls setup is done.

Oh! Found something different. Might only in part be a v1/v2 diff.

I used the http bypass to still push an app with a binary rebuilt for http access.
Allows me to compare the app ingress against my api ingress.
I.e. is my replication borken?

And curling the app it is indeed ok. Note, http blocked, https passed:
> % curl -k http://hello.172.27.0.2.omg.howdoi.website
404 page not found
% curl -k https://hello.172.27.0.2.omg.howdoi.website
[elided large html result]

Comparing the ingresses ... The app ingress has no load balancer either. So that seems to be a red herring.

Difference!
The app service uses port 8080.
The api service OTOH is on port 80.

I wonder if that triggers automagic behavior in traefik. I.e. keep 80 exposed (http ok), and in v2 the tls setup gets ignored ?!

andreas-kupries:
No. I missed a much clearer diff. Annotations.

app ingress has:
```
traefik.ingress.kubernetes.io/router.entrypoints: websecure
traefik.ingress.kubernetes.io/router.tls: "true"
```

Api does not. :facepalm:

jimmykarily:
> Confirmed with a hack disabling cert verification that this squashes the noted issues with cli and dashboard.
> 
> Based on https://forfuncsake.github.io/post/2017/08/trust-extra-ca-cert-in-go-app/
> I believe I have a way of going forward here:
> 
>     1. When a self-signed cert is made the CA is stored into the config.
> 
>     2. On reading the config, make the CA cert available, if present.
> 
> 
> This means that for LE certs nothing should be in the config.
> And we still have full verify for the self-signed, instead of disabling verification.

another option is to use "InsecureSkipVerify" when the domain is `omg.howdoi.website` when creating the client (https://golang.org/pkg/crypto/tls/#Client)

andreas-kupries:
I dislike the insecure skip verify as it would also disable any kind of verification done by the `epinio server` command also.

andreas-kupries:
Acceptance tests are ok for the http based operations, both cli and direct api testing.
The websocket based pieces fails.

... Websocket client side (dialer) does not use http transport. It comes with its own TLSClientConfig.

|Websock tls config|Result|Notes|
|---|---|---|
|default, none|:boom: | The error is: x509: certificate signed by unknown authority|
|skipverify| :heavy_check_mark: ||
|ca cert set|  :boom: | The error is: malformed HTTP response "\x00\x00\x18..."|

The failing cases have a common message as prefix, and different errors. The prefix is:
> :forbidden:failed to tail logs: Failed to connect to websockets endpoint (wss://epinio.172.31.0.2.omg.howdoi.website/api/v1/orgs/workspace/staging/36e7b7255c0e9178/logs?follow=true&stage_id=36e7b7255c0e9178). Response was = &lt;nil&gt;

andreas-kupries:
I might be running into https://github.com/gorilla/websocket/issues/601 here.

andreas-kupries:
Added a [comment](https://github.com/gorilla/websocket/issues/601#issuecomment-854541240) to that ticket.
I believe it is a true bug, although not in gorilla, but in the go http server, triggered by gorilla.
For now I am going with the `Force Http2 = no` workaround.
(I want my certs, not skip verify)

andreas-kupries:
The only :boom: left in the CI looks to be

```
should follow logs [It]
[4]       /home/carrier/actions-runner/_work/epinio/epinio/acceptance/api_v1_applications_test.go:589
[4] 
[4]       Unexpected error:
[4]           <*fmt.wrapError | 0xc000098380>: {
[4]               msg: "tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 192.168.128.1:52332->192.168.128.3:443: write: broken pipe",
[4]               err: {
[4]                   Op: "write",
[4]                   Net: "tcp",
[4]                   Source: {
[4]                       IP: [192, 168, 128, 1],
[4]                       Port: 52332,
[4]                       Zone: "",
[4]                   },
[4]                   Addr: {
[4]                       IP: [192, 168, 128, 3],
[4]                       Port: 443,
[4]                       Zone: "",
[4]                   },
[4]                   Err: {Syscall: "write", Err: 0x20},
[4]               },
[4]           }
[4]           tls: failed to send closeNotify alert (but connection was closed anyway): write tcp 192.168.128.1:52332->192.168.128.3:443: write: broken pipe
[4]       occurred
[4] 
[4]       /home/carrier/actions-runner/_work/epinio/epinio/acceptance/api_v1_applications_test.go:574
```

That is the `readLogs` helper, when it has completed reading the logs. The relevant lines are

```
				err := wsConn.Close()
```

Not sure if that is an actionable issue.

jimmykarily:
> I dislike the insecure skip verify as it would also disable any kind of verification done by the `epinio server` command also.

The epinio server and the epinio cli should be using different setttings for their clients. That should be possible with some refactoring (if not already).

andreas-kupries:
As I reminded myself, I should write the documentation for this now, before I am forgetting ...
Expect another commit sometime during the day with that.

Note also, from the ticket the following explanation:

@jimmykarily 
>  Instead of relying on the domain to figure out if certs are self-signed ...

> > This happens only for install (match for omg.howdoi.website).
The underlying mechanism however does not disable verification.
Install pulls the CA cert for the self-signed cert, and stuffs that into epinio's config file.
Then, whenever epinio is run it picks up that cert and adds it to the pool of CAs it verify against.

> > While I could have simply disabled verification for the self-signed cert this way of having the needed CA cert in the config means that the way is open for customers to create and use their own certs, self-signed or not, non-standard CA or not, and be able to work with all that. They simply have to stuff their CA cert(s) into the config.

andreas-kupries:
> >  I dislike the insecure skip verify as it would also disable any kind of verification done by the epinio server command also.

> The epinio server and the epinio cli should be using different setttings for their clients. That should be possible with some refactoring (if not already).

Everything currently uses the same http client and transport. The default transport of the net/http package.

rohitsakala (dismissed):

README.md
```diff
@@ -137,6 +138,24 @@ In case of trouble with Epinio's Traefik component or Ingress controllers, the
 [Traefik](docs/advanced.md#traefik) section in the
 [Advanced Topics](docs/advanced.md) document shall be your friend.
 
+#### A Quick Note On Security
+
+Epinio secures access to its API with TLS and basic authentication.
+
+None sign of that appeared in the above example because the installation process
```
andreas-kupries, line 145:
```suggestion
No sign of that appeared in the above example because the installation process
```

deployments/epinio.go
```diff
@@ -132,7 +132,27 @@ func (k Epinio) apply(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.I
...
 	}
 
-	message := "Creating Epinio server ingress"
+	// Wait for the cert manager to be present and active. It is required
+	for _, podname := range []string{
+		"webhook",
+		"cert-manager",
+		"cainjector",
+	} {
+		if err := c.WaitUntilPodBySelectorExist(ui, CertManagerDeploymentID, "app.kubernetes.io/name="+podname, k.Timeout); err != nil {
```
jimmykarily, line 141: Maybe `WaitForDeploymentCompleted` is better here? We should start talking about Deployments instead of Pods when it makes more sense.

internal/auth/certs.go
```diff
@@ -0,0 +1,133 @@
...
+	http.DefaultTransport.(*http.Transport).TLSClientConfig = config
+	websocket.DefaultDialer.TLSClientConfig = config
+
+	// See https://github.com/gorilla/websocket/issues/601 for
+	// what this is a work around for.
+	http.DefaultTransport.(*http.Transport).ForceAttemptHTTP2 = false
+}
+
+func CreateCertificate(ctx context.Context, cfg *rest.Config, name, namespace, systemDomain string) error {
+	// Create production certificate if it is provided by user
```
jimmykarily, line 49:
```suggestion
	// Create production certificate if systemDomain is provided by user
```

internal/cli/clients/install_client.go
```diff
@@ -159,6 +159,47 @@ func (c *InstallClient) Install(cmd *cobra.Command) error {
...
+
+	if strings.Contains(domain.Value.(string), "omg.howdoi.website") {
+		// Note 1:
+		// We are waiting here for the secret, instead of
+		// simply Get'ting it, because we cannot be sure that
+		// it was already created by cert-manager, from the
+		// cert request issued by the epinio deployment.  This
+		// is especially true when the epinio deployment does
+		// not wait for the server to be up. As it happens
+		// when a dev `ep install` is done (do't wait for
```
jimmykarily, line 178:
```suggestion
		// when a dev `ep install` is done (don't wait for
```

internal/cli/clients/install_client.go
```diff
@@ -159,6 +159,47 @@ func (c *InstallClient) Install(cmd *cobra.Command) error {
...
+		// deployment + skip default org).
+		//
+		// Note 2:
+		// The secret's name is (namespace + '-tls'), see the
+		// `auth.createCertificate` template for the created
```
jimmykarily, line 183: [nit] I think this part of the comment is obvious from the lines below. Remove the first part about the name of the secret?

README.md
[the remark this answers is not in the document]
andreas-kupries: Done by making a suggestion and committing that.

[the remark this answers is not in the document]
andreas-kupries: see above

internal/cli/clients/install_client.go
andreas-kupries: Ok.

deployments/epinio.go
andreas-kupries: Will try locally before committing. It certainly would be simpler. I.e. no replication of internal code from the other deployment.
I may have to re-read all the helpers to see what we have now. I had not realized that we had the ability to wait for a deployment already.

internal/cli/clients/install_client.go
andreas-kupries: Done, commit d9f2b91cd6fa548012900ff03d3b90a653c3ebb7

deployments/epinio.go
andreas-kupries: Hm. ... We might have a confusion here, because of names.
The `WaitForDeploymentCompleted` is for waiting on a kube `Deployment` resource.
This is not a `deployment` as we use it in `epinio install`.

That said, I will check if the helm chart deployed by the `cert-manager` deployment (sic) installs a kube `Deployment`.
And if not, maybe some other higher controller (`StateSet`, `ReplicaSet`, etc.) we could wait for.

andreas-kupries: Well, 3 kube deployments

```
% k get deployment : |grep cert
kubectl get deployment --all-namespaces
cert-manager      cert-manager-cainjector                      1/1     1            1           18m
cert-manager      cert-manager                                 1/1     1            1           18m
cert-manager      cert-manager-webhook                         1/1     1            1           18m
```

Now I am wondering if we should/could do waiting for an epinio deployment purely internally.
Because the goroutine for deployment knows when it is done.

Maybe can use the WaitGroup we already have waiting for all of them to complete ?

jimmykarily: I mainly suggested to wait for deployment because it's is a higher level structure in Kubernetes and its status better reflects the service's availability. In most cases we are not gaining a lot though. Using the wait groups will complicate that code even more I'm afraid. Feel free to leave this as it is.

jimmykarily (approved):

deployments/epinio.go
andreas-kupries: Resolving, defered to future cleanup.

PR #446, opened 2021-06-02
438 linkerd

### LINKED ISSUE #438: linkerd deployment
Follow up of #130 

As a first iteration we will add Linkerd to secure the "easy" components. Those are all the components that worked during the first investigation (see comments on #130).

## Acceptance Criteria

- `epinio install` should deploy linkderd control plane components (minimum set, no optional components)
- `epinio install` should add all the necessary annotations to the namespaces to allow linkerd to work
- There should be a `--skip-linkerd` flag that skips the installation of the linkderd control plane (e.g. in case the user has that already installed). This is similar to the `--skip-traefik` flag. The use of this flag should be documented.
- When `--skip-linkerd` is used, the annotations on namespaces and other objects are still applied.
- [optional] test linkderd installation in acceptance tests. E.g. create a test to run `linkderd tap` to check that namespaces where mTLSed or check for sidecars in pods.

jimmykarily:
Fixes #438

### DIFF
--- a/acceptance/epinio_installation_test.go
+++ b/acceptance/epinio_installation_test.go
@@ -1,0 +1,29 @@
+package acceptance_test
+
+import (
+	"strings"
+
+	"github.com/epinio/epinio/helpers"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Epinio Installation", func() {
+	It("has linkerd sidecars", func() {
+		kubectlCommand := "get pods -n epinio -l app.kubernetes.io/component=epinio-server -o=jsonpath='{.items[0].spec.containers[*].name}'"
+		out, err := helpers.Kubectl(kubectlCommand)
+		Expect(err).ToNot(HaveOccurred())
+		containers := strings.Split(out, " ")
+		Expect(containers).To(ContainElement("linkerd-proxy"))
+	})
+	It("has linkerd control plane components running", func() {
+		kubectlCommand := "get pods -n linkerd -o name"
+		out, err := helpers.Kubectl(kubectlCommand)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(out).To(MatchRegexp("linkerd-identity"))
+		Expect(out).To(MatchRegexp("linkerd-proxy-injector"))
+		Expect(out).To(MatchRegexp("linkerd-controller"))
+		Expect(out).To(MatchRegexp("linkerd-sp-validator"))
+		Expect(out).To(MatchRegexp("linkerd-destination"))
+	})
+})
--- a/assets/embedded-files/linkerd/install-job.yaml
+++ b/assets/embedded-files/linkerd/install-job.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: linkerd-install
+  namespace: linkerd
+spec:
+  template:
+    spec:
+      serviceAccountName: linkerd-installer
+      containers:
+      - name: linkerd-install
+        image: splatform/epinio-linkerd:2.10.2
+        imagePullPolicy: Always
+        command: ["/bin/sh","-c"]
+        args: 
+          - linkerd install | kubectl apply -f - && linkerd check
+      restartPolicy: Never
+  backoffLimit: 1
--- a/assets/embedded-files/linkerd/rbac.yaml
+++ b/assets/embedded-files/linkerd/rbac.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: linkerd-installer
+  namespace: linkerd
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: linkerd-installer
+subjects:
+  - kind: ServiceAccount
+    name: linkerd-installer
+    namespace: linkerd
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
--- a/assets/embedded-files/linkerd/uninstall-job.yaml
+++ b/assets/embedded-files/linkerd/uninstall-job.yaml
@@ -1,0 +1,19 @@
+---
+apiVersion: batch/v1
+kind: Job
+metadata:
+  name: linkerd-uninstall
+  namespace: linkerd
+spec:
+  template:
+    spec:
+      serviceAccountName: linkerd-installer
+      containers:
+      - name: linkerd-uninstall
+        image: splatform/epinio-linkerd:2.10.2
+        imagePullPolicy: Always
+        command: ["/bin/sh","-c"]
+        args: 
+          - linkerd uninstall --verbose | kubectl delete -f -
+      restartPolicy: Never
+  backoffLimit: 1
--- a/deployments/linkerd.go
+++ b/deployments/linkerd.go
@@ -1,0 +1,142 @@
+package deployments
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/epinio/epinio/helpers"
+	"github.com/epinio/epinio/helpers/kubernetes"
+	"github.com/epinio/epinio/helpers/termui"
+	"github.com/kyokomi/emoji"
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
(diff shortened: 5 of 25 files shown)


jimmykarily:
Opened for review. The only thing missing from the acceptance criteria in #438 is the optional test to check the installation with linkerd tap. I check if that's possible and add it to this PR as a separate commit or create a new PR for that after we merge this one.

assets/embedded-files/linkerd/install-job.yaml
```diff
@@ -0,0 +1,19 @@
...
+      serviceAccountName: linkerd-installer
+      containers:
+      - name: linkerd-install
+        image: splatform/epinio-linkerd:2.10.2
+        imagePullPolicy: Always
+        command: ["/bin/sh","-c"]
+        args: 
+          - linkerd install | kubectl apply -f - && linkerd check
+      restartPolicy: Never
+  backoffLimit: 1 # TODO: Is this ok?
```
jimmykarily, line 19: I don't think retrying this job would help in any way. What do you think?

assets/embedded-files/linkerd/uninstall-job.yaml
```diff
@@ -0,0 +1,19 @@
...
+      serviceAccountName: linkerd-installer
+      containers:
+      - name: linkerd-uninstall
+        image: splatform/epinio-linkerd:2.10.2
+        imagePullPolicy: Always
+        command: ["/bin/sh","-c"]
+        args: 
+          - linkerd uninstall --verbose | kubectl delete -f -
+      restartPolicy: Never
+  backoffLimit: 1 # TODO: Is this ok?
```
jimmykarily, line 19: same

deployments/traefik.go
```diff
@@ -136,15 +138,16 @@ func (k Traefik) GetVersion() string {
 
 func (k Traefik) Deploy(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
 
-	skipTraefik, err := options.GetBool("skip_traefik", TraefikDeploymentID)
+	skipTraefik, err := options.GetBool("skip-traefik", TraefikDeploymentID)
```
jimmykarily, line 141: not sure if this is the right place to replace all `_` in flags with `-`. We can do it in another PR. I think it's more common to use `-` in flag names.

deployments/traefik.go
```diff
@@ -136,15 +138,16 @@ func (k Traefik) GetVersion() string {
...
 	if err != nil {
-		return errors.Wrap(err, "Couldn't get skip_traefik option")
+		return errors.Wrap(err, "Couldn't get skip-traefik option")
 	}
 	if skipTraefik {
 		ui.Exclamation().Msg("Skipping Traefik Ingress deployment by user request")
 		return nil
 	}
 
+	// TODO: Don't do automatic detection!
```
jimmykarily, line 150: Accidentally committed (was a personal note). Will remove.

deployments/traefik.go
rohitsakala: okie

deployments/quarks.go
```diff
@@ -109,7 +109,9 @@ func (k Quarks) apply(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.I
 		action = "upgrade"
 	}
 
-	if err := c.CreateLabeledNamespace(QuarksDeploymentID); err != nil {
+	if err := c.CreateNamespace(QuarksDeploymentID, map[string]string{
+		kubernetes.EpinioDeploymentLabelKey: kubernetes.EpinioDeploymentLabelValue,
```
rohitsakala, line 113: Communication with `quarks` also happens through CRD's.

assets/embedded-files/linkerd/rbac.yaml
```diff
@@ -0,0 +1,19 @@
...
+metadata:
+  name: linkerd-installer
+subjects:
+  - kind: ServiceAccount
+    name: linkerd-installer
+    namespace: linkerd
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: cluster-admin
```
rohitsakala, line 19: Do we need `cluster-admin`?

acceptance/epinio_installation_test.go
```diff
@@ -0,0 +1,30 @@
...
+	It("has linkerd control plane components running", func() {
+		kubectlCommand := "get pods -n linkerd -o name"
+		out, err := helpers.Kubectl(kubectlCommand)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(out).To(MatchRegexp("linkerd-identity"))
+		Expect(out).To(MatchRegexp("linkerd-proxy-injector"))
+		Expect(out).To(MatchRegexp("linkerd-controller"))
+		Expect(out).To(MatchRegexp("linkerd-sp-validator"))
+		Expect(out).To(MatchRegexp("linkerd-destination"))
+		Expect(out).To(MatchRegexp("linkerd-destination"))
```
rohitsakala, line 28: Duplicated Check for `linkerd-destination` ?

deployments/linkerd.go
```diff
@@ -0,0 +1,144 @@
...
+	_, err := c.Kubectl.CoreV1().Namespaces().Get(
+		context.Background(),
+		LinkerdDeploymentID,
+		metav1.GetOptions{},
+	)
+	if err != nil {
+		return errors.New("Namespace " + LinkerdDeploymentID + " not present")
+	}
+
+	ui.Note().Msg("Upgrading Linkerd Ingress...")
```
rohitsakala, line 141: We can remove `Ingress` from note message I think

internal/organizations/organizations.go
```diff
@@ -67,6 +67,9 @@ func Create(ctx context.Context, kubeClient *kubernetes.Cluster, gitea GiteaInte
 					kubernetes.EpinioOrgLabelKey:        kubernetes.EpinioOrgLabelValue,
 					"quarks.cloudfoundry.org/monitored": "quarks-secret",
 				},
+				Annotations: map[string]string{
```
rohitsakala, line 70: Do we need linkerd sidecars in user application pods too ?

deployments/quarks.go
jimmykarily: ok removing then

assets/embedded-files/linkerd/rbac.yaml
jimmykarily: Providing fine-grained permissions would be better but we've taken a shortcut here as well: https://github.com/epinio/epinio/blob/main/assets/embedded-files/tekton/admin-role.yaml#L97

We also generally install Epinio as cluster-admins (unless the kubeconfig we use has different permissions).
I see two paths:
1. We assume the cluster "belongs" to Epinio. In this case it's ok to have cluster-admin permissions during Epinio installation (which is the case here and also when we run `epinio install`)
2. We play nice and we use roles with fine grained permissions for each and every action we take: `epinio install`, linkerd install job etc. This may mean we have to split the creation of those roles to a separate step and then use the created roles to actually install Epinio.

Option 2 is a lot more complicated than option 1 and doesn't offer much on clusters where the only tenant is Epinio. I would say let's stick to cluster-admin for installation until we have a use case where this is not ok.

That said, being cluster-admin when doing staging is probably not ok (see the linked file above).

acceptance/epinio_installation_test.go
jimmykarily: yes, removing.

deployments/linkerd.go
jimmykarily: copy-paste error. Fixing.

internal/organizations/organizations.go
jimmykarily: I think so. They will be talking to services created by the service-catalog. E.g. an application could be talking to another pod which is a Mariadb or other.

rohitsakala: Okie. Gotcha. Then we also need to inject `linkerd into `mariadb` right?

assets/embedded-files/linkerd/rbac.yaml
rohitsakala: okie. We can fix it later.

internal/organizations/organizations.go
jimmykarily: `mariadb` pods (or any other service), are created in the same org with the application. This organization is an annotated namespace and thus any pods created there will have a sidecar.

rohitsakala: Gotcha.

rohitsakala (approved):

PR #460, opened 2021-06-07
feat: Delete apps in an org parallelly

### LINKED ISSUE #370: Delete apps in an org parallelly
Ex - Suppose if there are 1000 apps, then deletion of apps in an org takes some time. We need to always keep in mind scalability and write the code.

Raullllll:
Fixes #370

I added a maximum number of concurrent app deletions. I thought it would be a good a idea just in case the system cannot handle too many requests at the same time. I set 100 for the max number of concurrent deletions, but please let me know if you think we should change this number or if this limit is not needed at all.
I didn't include unit tests as I didn't saw any, but please let me know if they need to be included.

Any feedback would be highly appreciated!

### DIFF
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -162,47 +162,45 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 		return err
 	}
 
+	// create a buffer, so there is no more than 'maxConcurrent' goroutines running at the same time
 	const maxConcurrent = 100
+	buffer := make(chan struct{}, maxConcurrent)
 	errChan := make(chan error, maxConcurrent)
-	success := make(chan bool)
-	stop := make(chan bool)
+	var wg sync.WaitGroup
+	var forLoopErr error
 
-	go func() {
-		var wg sync.WaitGroup
-		// create a buffer, so there is no more than 'maxConcurrent' goroutines running at the same time
-		buffer := make(chan struct{}, maxConcurrent)
-	loop:
-		for _, app := range apps {
-			buffer <- struct{}{}
-			wg.Add(1)
+loop:
+	for _, app := range apps {
+		buffer <- struct{}{}
+		wg.Add(1)
 
-			go func(app application.Application) {
-				defer wg.Done()
-				defer func() {
-					<-buffer
-				}()
-				err = application.Delete(cluster, gitea, org, app)
-				if err != nil {
-					errChan <- err
-				}
-			}(app)
-
-			select {
-			case <-stop:
-				break loop
-			default:
+		go func(app application.Application) {
+			defer wg.Done()
+			defer func() {
+				<-buffer
+			}()
+			err = application.Delete(cluster, gitea, org, app)
+			if err != nil {
+				errChan <- err
 			}
-		}
-		wg.Wait()
-		success <- true
-	}()
+		}(app)
 
-	//wait until all apps are deleted or we get an error
+		select {
+		case forLoopErr = <-errChan:
+			break loop
+		default:
+		}
+	}
+	wg.Wait()
+
+	if forLoopErr != nil {
+		return forLoopErr
+	}
+
 	select {
 	case err := <-errChan:
-		stop <- true
 		return err
-	case <-success:
+	default:
 		return nil
 	}
 
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -164,8 +164,8 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 
 	const maxConcurrent = 100
 	errChan := make(chan error, maxConcurrent)
-	success := make(chan struct{})
-	stop := make(chan struct{})
+	success := make(chan bool)
+	stop := make(chan bool)
 
 	go func() {
 		var wg sync.WaitGroup
@@ -174,10 +174,13 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 	loop:
 		for _, app := range apps {
 			buffer <- struct{}{}
+			wg.Add(1)
 
 			go func(app application.Application) {
 				defer wg.Done()
-				wg.Add(1)
+				defer func() {
+					<-buffer
+				}()
 				err = application.Delete(cluster, gitea, org, app)
 				if err != nil {
 					errChan <- err
@@ -192,13 +195,13 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 			}
 		}
 		wg.Wait()
-		success <- struct{}{}
+		success <- true
 	}()
 
 	//wait until all apps are deleted or we get an error
 	select {
 	case err := <-errChan:
-		stop <- struct{}{}
+		stop <- true
 		return err
 	case <-success:
 		return nil
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -185,7 +185,6 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 				if err != nil {
 					errChan <- err
 				}
-				<-buffer
 			}(app)
 
 			select {
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"net/http"
+	"sync"
 
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/internal/application"
@@ -123,16 +124,9 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
 		}
 	}
 
-	apps, err := application.List(cluster, org)
+	err = deleteApps(cluster, gitea, org)
 	if err != nil {
 		return APIErrors{NewAPIError(err.Error(), "", http.StatusInternalServerError)}
-	}
-
-	for _, app := range apps {
-		err = application.Delete(cluster, gitea, org, app)
-		if err != nil {
-			return APIErrors{NewAPIError(err.Error(), "", http.StatusInternalServerError)}
-		}
 	}
 
 	serviceList, err := services.List(cluster, org)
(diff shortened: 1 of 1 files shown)


internal/api/v1/organizations.go
```diff
@@ -161,3 +155,53 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
...
+		var wg sync.WaitGroup
+		// create a buffer, so there is no more than 'maxConcurrent' goroutines running at the same time
+		buffer := make(chan struct{}, maxConcurrent)
+	loop:
+		for _, app := range apps {
+			buffer <- struct{}{}
+
+			go func(app application.Application) {
+				defer wg.Done()
+				wg.Add(1)
```
jimmykarily, line 180: It could happen that `wg.Wait()` is executed before any of the `wg.Add(1)` calls have been made. It's better to call `wg.Add(1)` right before the go routing call to ensure it always executes before `wg.Wait()`.

internal/api/v1/organizations.go
```diff
@@ -161,3 +155,53 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
...
+				err = application.Delete(cluster, gitea, org, app)
+				if err != nil {
+					errChan <- err
+				}
+				<-buffer
```
jimmykarily, line 185: This should better be defered too to ensure that no-matter what happens during the go routine call, the buffer will be freed in the end.

internal/api/v1/organizations.go
```diff
@@ -161,3 +155,53 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
...
+func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) error {
+	apps, err := application.List(cluster, org)
+	if err != nil {
+		return err
+	}
+
+	const maxConcurrent = 100
+	errChan := make(chan error, maxConcurrent)
+	success := make(chan struct{})
+	stop := make(chan struct{})
```
jimmykarily, line 168: Since these structs have no fields, maybe using `bool` is better (cleaner code).

internal/api/v1/organizations.go
```diff
@@ -161,3 +155,53 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
...
+			}
+		}
+		wg.Wait()
+		success <- struct{}{}
+	}()
+
+	//wait until all apps are deleted or we get an error
+	select {
+	case err := <-errChan:
+		stop <- struct{}{}
```
jimmykarily, line 201: This will stop the loop so no more applications will be added for deletion (in the buffer) but we may have up to 99 applications currently being deleted. We should better wait for those go routines to finish. We can achieve that by moving the declaration of the wg variable (https://github.com/epinio/epinio/pull/460/files#diff-27a9ad008b6be81b374b2666592a4f2a279860da73cceaccf0268140cbf94686R171) ouside the main go routine (so that it's in scope here). Then we can simply defer the `wg.Wait()` call in the main thread (ideally right after the declaration of the variable).

internal/api/v1/organizations.go
```diff
@@ -161,3 +155,53 @@ func (oc OrganizationsController) Delete(w http.ResponseWriter, r *http.Request)
...
+				defer wg.Done()
+				wg.Add(1)
+				err = application.Delete(cluster, gitea, org, app)
+				if err != nil {
+					errChan <- err
+				}
+				<-buffer
+			}(app)
+
+			select {
```
jimmykarily, line 188: Instead of having a `select` outside the loop to signal the select inside the loop I think you can read the `errChan` here directly: "If there is an error, break the loop". With that change you probably don't need the main go routine anymore. After all it's only there to allow the execution to continue to the `select` right after. With that `select` gone, you can flatten the code to only be a for loop that spins up go routines while no error has happened.

This would change the code quite a bit so some of my other comments may not apply anymore.

jimmykarily (changes requested):
Hi Raul, thank you for this contribution!

I made some comments on the PR but I think you should first try my recommendation to get rid of the main go routine (because some of my other comments won't need addressing then). When working with go routines, I find myself going back and forth between 2 solutions because of edge cases I didn't think about. Since I haven't tried my suggestion, it could be the case here too :). If you think this won't work let me know and I'll have a look again.

internal/api/v1/organizations.go
Raullllll: yes I agree. Done in the next commit

internal/api/v1/organizations.go
Raullllll: done

internal/api/v1/organizations.go
Raullllll: done

internal/api/v1/organizations.go
Raullllll: All gouroutines in the buffer will finished before any code after wg.Wait() is executed, so current deletions will finish

internal/api/v1/organizations.go
Raullllll: This was my first approach, I tried without the main goroutine. But I found out this was not working. Read from the errChan will block until and error is sent to the channel that's what I used the select. If that is done there app will hang forever, unless I'm missing something :)

Raullllll:
@jimmykarily Thank you for your valuable feedback! I originally thought of doing it without the main goroutine, but I found out I needed it in order to wait for an error or a success if there was no errors. If I wait for the error channel to receive a message just outside the loop, app will stop there forever. Or maybe I misunderstood what you meant? Please have a look at the latest changes and let me know what you think.
Thanks!

internal/api/v1/organizations.go
jimmykarily: This `for` loop spins up a go routine and then checks to see if it should stop here (by checking the `stop` chan) or simply continue adding go routines. The only writer to the `stop` chan is the outer select, in the case of an error.

Starting go routines is relatively fast and the loop will probably have enough time to fill the buffer (100 go routines) before any of them has enough time to return an error. This mean that this stopping mechanism might only work when  there are more than 100 apps to delete and some app fails before we get to add the last application to the buffer. This might prevent the loop from adding more to the buffer but could still have 100 failing go routines. Since the errChan is buffered, that should be no problem.

Reading the errChan won't block as long as you keep the `default:` in the select (like you do now). That is not improving what I described above, just achieving the same with less code.

jimmykarily (changes requested):
>  If I wait for the error channel to receive a message just outside the loop, app will stop there forever.

What is meant is "check" the `errChan`, not "wait" for it. That is a `select` with a `default`, like you have it now outside the loop, but I meant you should do it inside the loop. So replace this check on the `stop` chan here: https://github.com/epinio/epinio/pull/460/files#diff-27a9ad008b6be81b374b2666592a4f2a279860da73cceaccf0268140cbf94686R191 with a check on the `errChan` (keeping the empty `default`). As far as I can tell this is equivalent code, but it doesn't need two channels and a main go routine.

Actually by doing it, you don't even need the success channel. You will already be in the main thread so there is noone else to signal.

I hope my comment makes sense. It's hard to describe go routine flow, without actually writing the code. If my comments are confusing maybe it's worth jumping in a pairing session, we should be able to short this out quickly. My timezone is EEST (GMT+3).

Again, thank you for your patience and your time.

Raullllll:
@jimmykarily  Thanks for your comments! I tried moving the errChan select inside the loop, but it fails in some edge cases. Let's say you have 100 apps, and the 99th app fails. The for loop finishes as soon as the 100 goroutines start executing. The select errChan goes to the default case, then for loop finishes. But the 99th goroutine is still running and fails, then send an error to errChan but this error is not received as the for loop has finished. That's why I added the main goroutine.
Do you see what I mean? If it is not clear, I would love to do a pairing session!
Thanks again for your feedback!

jimmykarily:
> @jimmykarily Thanks for your comments! I tried moving the errChan select inside the loop, but it fails in some edge cases. Let's say you have 100 apps, and the 99th app fails. The for loop finishes as soon as the 100 goroutines start executing. The select errChan goes to the default case, then for loop finishes. But the 99th goroutine is still running and fails, then send an error to errChan but this error is not received as the for loop has finished. That's why I added the main goroutine.
> Do you see what I mean? If it is not clear, I would love to do a pairing session!
> Thanks again for your feedback!

I get what you mean. When that happens, your main thread (outside the for loop) will wait until all the running routines are done (because right after the for loop you will have a `wg.Wait()` call). This means the failed go routine will have time to write its error to the errChan. All we need then is to collect the errors after the `wg.Wait()` with a `select` call, almost like the one you have now but with no need to send anything to the `stop` chan:

```
// Right after the for loop
wg.Wait() // Wait for all running go routines to finish (either successfully or with error)
select {
case err := <-errChan: // See if we have any errors 
  return err
default: // If there are no errors
  return nil
}
```

Raullllll:
ah ok! I see it now! But if we add this outside the loop, we will have to wait for all the apps to be deleted. Maybe add 2 selects, one inside the loop and another outside? so the inside select will also return err and stop the loop if an error is received
Thanks!

jimmykarily:
Yes I meant 2 `select`s  , but that won't work :facepalm:  (that's where writing the code would have helped). If the inner loop reads the error from the `errChan` (to be able to stop), then the error will no longer be available for the outer `select` to read it.

Given this method only returns one error and not the list of all failures, we could store the error read inside the for loop in a variable so it's not lost. E.g.

```
// before the for loop
var forLoopErr error

...
    // In the for loop
    select {
    case forLoopErr := <-errChan:
      break loop
    default:
    }
}

// Right after the for loop
wg.Wait() // Wait for all running go routines to finish (either successfully or with error)
if forLoopErr != nil { // If the loop was stopped because of an error
  return forLoopErr
}

select {
case err := <-errChan: // See if we have any errors (that didn't stop the for loop but did happen after the for loop was done)
  return err
default: // If there are no errors
  return nil
}
```

Raullllll:
@jimmykarily  you are right! This will work. I pushed a new commit with the changes, please have a look.
Thank you very much for your support with this!

andreas-kupries:
Heh. And just as I completed my reading of the code and writing up my understanding. :wink: 
Well, time to look at these changes and update my document and understanding.
When I am done I will post my own review. Will still take a while as I have a meeting in a moment.

jimmykarily:
@Raullllll yes, that's what I had in mind. It looks good now. I will wait for Andreas' review and also want to give it a hands-on try to see if we missed something before I approve. I hope to get it done tomorrow or worst case by the end of this week. Stay tuned!

andreas-kupries:
I approve.

Attached my write up and documentation of the current code: [dac.md](https://github.com/epinio/epinio/files/6623503/dac.md)

I would be pleased if this document could be added to the docs/ directory of the project, as reference for developers. Maybe make a reference to it in a code comment as well.

For posterity see also the [dac1.md](https://github.com/epinio/epinio/files/6623512/dac1.md). This document describes the state of the system at commit 528af5feead2d4c21ccc4ac80633507d7e0ffa21, before the dispatcher goroutine got removed, and the two issues I found with it while writing the document.

@jimmykarily :bell:

jimmykarily:
@andreas-kupries thanks for the review, good to have another pair of eyes. I don't think we should have a separate document to explain how a method works for 2 reasons:

- line numbers or even the way the method works could easily change without updating the document. One can't guess there is such a document when working on the code.
- one shouldn't have to look in other places to understand the code.

That said and given this method is a little complex I would be ok to have a (maybe shorter) version of your document at the top of the method. The part that describes what various lines do can be left out then, since it's easier to directly look at the code below.


I believe all is good. I just need to do some manual QA on it (just did a code review up to now) and if all is ok we can merge.

andreas-kupries:
> line numbers or even the way the method works could easily change without updating the document

Right. Could be handled by using unique markers in the code for the various blocks and referencing these in the document.

This kind of thing could be done with the code comments as well, i.e. have it reference markers in the code underneath the explanation.

jimmykarily (approved):

andreas-kupries:
@Raullllll Note that @jimmykarily has approved. That means you can hit `merge` now.

My last comments here regarding comments in code can be ignored.

The plan is now for me to add such comments summarizing my writeup after the merge is done.

Raullllll:
@andreas-kupries @jimmykarily  CI build pipeline failed. There was an error when trying to login into DockerHub. 
I don't think I can merge it since I don't have write access.
Thanks!

jimmykarily:
No worries, the one that fails is a copy of the one that succeeded (long story related to internal github action runners and secrets). I will merge now. Thanks @Raullllll, good job!

PR #465, opened 2021-06-08
Cleanup ingress left behind by app deletion

### LINKED ISSUE #461: Deleting an app leaves ingress behind
Reproduce:

- Delete the app
- `kubectl get ingress -n workspace` : the ingress is still there.

We should fix this and add a test to ensure this doesn't happen again.

NOTE: This may be solved with #347 (because of the ownership reference). We only have to add a test in that case.

andreas-kupries:
Fixes #461

### DIFF
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -51,6 +51,13 @@ func (a *Application) Delete(gitea GiteaInterface) error {
 		return pkgerrors.Wrap(err, "failed to delete application deployment")
 	}
 
+	err = a.kubeClient.Kubectl.ExtensionsV1beta1().Ingresses(a.Organization).
+		Delete(context.Background(), a.Name, metav1.DeleteOptions{})
+
+	if err != nil {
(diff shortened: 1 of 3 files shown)


rohitsakala (approved):

internal/application/application.go
```diff
@@ -51,6 +51,13 @@ func (a *Application) Delete(gitea GiteaInterface) error {
 		return pkgerrors.Wrap(err, "failed to delete application deployment")
 	}
 
+	err = a.kubeClient.Kubectl.ExtensionsV1beta1().Ingresses(a.Organization).
```
jimmykarily, line 54: Does this mean we need to delete the `Service` explicitly too?

andreas-kupries: Possibly yes. I will investigate.

andreas-kupries: Defered to separate ticket.

PR #466, opened 2021-06-08
Ignore git config files in app sources

### LINKED ISSUE #450: `epinio push` a git repository should not mess with Epinio's git
If you are in an application directory that is already a git repository and try to do `epinio push` while having uncommitted changes, this results in some strange results. In my case only the uncommitted files where available during staging.

We need to investigate and learn more on what is going on. One solution is to exclude the `.git` repository when creating the application archive. This won't create any problems even when we support `git push` deployments, we only have to make sure we don't do "git" twice.

thardeck:
to prevent issues with the remote storage in Gitea.

Fixes #450

### DIFF
--- a/internal/cli/clients/push.go
+++ b/internal/cli/clients/push.go
@@ -3,6 +3,7 @@ package clients
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"io/ioutil"
 	"path"
 	"time"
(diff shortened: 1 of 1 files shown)


internal/cli/clients/push.go
```diff
@@ -32,7 +32,11 @@ func collectSources(log logr.Logger, source string) (string, string, error) {
 		// application directory to get the proper paths to
 		// the files and directories to assemble in the
 		// tarball.
-
+		// Ignore git config files in the app sources to prevent conflicts with the gitea git repo
+		if f.Name() == ".git" || f.Name() == ".gitignore" || f.Name() == ".gitmodules" || f.Name() == ".gitconfig" || f.Name() == ".git-credentials" {
+			log.V(3).Info("Skipping upload of file/dir '" + f.Name() + "'.")
```
jimmykarily, line 37:
```suggestion
			log.V(3).Info(fmt.Sprintf("Skipping upload of file/dir '%s'.",  f.Name())
```

jimmykarily (dismissed):

rohitsakala (approved):

PR #472, opened 2021-06-09
API Error Handling and Contexts

manno:
This refers to #347 

It adds contexts to all network calls and makes API error handling less noisy.

### DIFF
--- a/acceptance/utilities_test.go
+++ b/acceptance/utilities_test.go
@@ -123,9 +123,11 @@ func makeAppWithDir(appName string, instances int, deployFromCurrentDir bool, ap
 
 	// And check presence
 
-	out, err := Epinio("app list", "")
-	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
-	ExpectWithOffset(1, out).To(MatchRegexp(fmt.Sprintf(`%s.*\|.*%d\/%d.*\|.*`, appName, instances, instances)))
+	EventuallyWithOffset(1, func() string {
+		out, err := Epinio("app list", "")
+		Expect(err).ToNot(HaveOccurred(), out)
+		return out
+	}, "5m").Should(MatchRegexp(fmt.Sprintf(`%s.*\|.*%d\/%d.*\|.*`, appName, instances, instances)))
 
 	return pushOutput
 }
--- a/deployments/google_services.go
+++ b/deployments/google_services.go
@@ -22,6 +22,8 @@ type GoogleServices struct {
 	Timeout time.Duration
 }
 
+var _ kubernetes.Deployment = &GoogleServices{}
+
 const (
 	GoogleServicesDeploymentID = "google-service-broker"
 	googleServicesVersion      = "0.1.0"
@@ -32,11 +34,11 @@ func (k *GoogleServices) ID() string {
 	return GoogleServicesDeploymentID
 }
 
-func (k *GoogleServices) Backup(c *kubernetes.Cluster, ui *termui.UI, d string) error {
+func (k *GoogleServices) Backup(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI, d string) error {
 	return nil
 }
 
-func (k *GoogleServices) Restore(c *kubernetes.Cluster, ui *termui.UI, d string) error {
+func (k *GoogleServices) Restore(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI, d string) error {
 	return nil
 }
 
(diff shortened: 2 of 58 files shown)


acceptance/utilities_test.go
```diff
@@ -123,9 +123,11 @@ func makeAppWithDir(appName string, instances int, deployFromCurrentDir bool, ap
 
 	// And check presence
 
-	out, err := Epinio("app list", "")
```
jimmykarily, line 126: oops :D !

internal/cli/clients/install_client.go
```diff
@@ -61,6 +61,8 @@ func (c *InstallClient) Install(cmd *cobra.Command) error {
 	defer log.Info("return")
 	details := log.V(1) // NOTE: Increment of level, not absolute.
 
+	ctx := context.Background()
```
jimmykarily, line 64: What is the purpose of passing a context everywhere, if that context has no way to be cancelled? As far as I understand a context exists as a way to signal various part of the code that what they are doing is no longer relevant (the went "out of context") and as such, they should stop doing it. This is the case with code that is triggered by http requests, when the request is no longer alive, or with code that it triggered from command line when the user does `ctrl+c`. Generally code that should run only as long as something else is still happening.

In this case, the only reason to cancel the context would be if the user stopped the installation (ctrl+c?) or if one of the deployments failed. We would need to create a "child" context with a CancelFunc in order to achieve this (as we do here for example: https://github.com/epinio/epinio/blob/main/internal/api/v1/applications.go#L236) and I don't see this here. Is this meant to happen in another PR?

internal/cli/delete.go
```diff
@@ -14,12 +14,12 @@ var CmdDeleteApp = &cobra.Command{
 	Short: "Deletes an application",
 	Args:  cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
-		client, err := clients.NewEpinioClient(cmd.Flags())
+		client, err := clients.NewEpinioClient(cmd.Context(), cmd.Flags())
```
jimmykarily, line 17: oh cmd has a Context already. I guess this handles the cancelling from the user's side. Maybe we need to handle the case when a deployment fails?

acceptance/utilities_test.go
manno: I had to move this into an Eventually block, it failed in Github actions with '1/2'. Worked locally, though.

deployments/google_services.go
```diff
@@ -186,15 +188,15 @@ func (k GoogleServices) Deploy(c *kubernetes.Cluster, ui *termui.UI, options kub
...
 		return err
 	}
 
 	return nil
 }
 
-func (k GoogleServices) Upgrade(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
+func (k GoogleServices) Upgrade(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
 	_, err := c.Kubectl.CoreV1().Namespaces().Get(
 		context.Background(),
```
jimmykarily, line 201: use ctx here?

deployments/minibroker.go
```diff
@@ -151,15 +153,15 @@ func (k Minibroker) Deploy(c *kubernetes.Cluster, ui *termui.UI, options kuberne
...
 		return err
 	}
 
 	return nil
 }
 
-func (k Minibroker) Upgrade(c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
+func (k Minibroker) Upgrade(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI, options kubernetes.InstallationOptions) error {
 	_, err := c.Kubectl.CoreV1().Namespaces().Get(
 		context.Background(),
```
jimmykarily, line 166: here too

jimmykarily (dismissed):

internal/cli/clients/install_client.go
manno: It's hard to implement a propagated context just in some part of the code base as the shared functions are changed to accept one.

I'm following convention here, by adding a context to all 'request' related funcs. Cobra also provides a top-level context  to commands, which we can propagate.

> A function that is never request-specific may use context.Background(), but err on the side of passing a Context even if you think you don't need to. The default case is to pass a Context; only use context.Background() directly if you have a good reason why the alternative is a mistake.
https://github.com/golang/go/wiki/CodeReviewComments#contexts

I like that this makes request-related funcs much more visible.

Improving cancel-ability we can do in another PR, as discussed.

deployments/google_services.go
manno: I fixed all the remaining places and removed TODO() and Background() contexts.

PR #473, opened 2021-06-09
Cleanup Makefile

### LINKED ISSUE #451: Cleanup Makefile
After discussing this in retro, we decided that the Makefile should have simpler flat targets that can be chained as desired. The targets should not depend on each other unless they are chains of simple targets.
E.g.

- `make build` should only build
- `make lint` should only lint
- `make release` could be a chain of `build lint vet` and other (flat and plain) targets.

This will allow us to mix and match targets both for local use (development), CI for testing and CI for releasing.

thardeck:
`make build` should only build
`make lint` should only lint
`make release` is a chain of build lint vet and other (flat and plain) targets

Fixes #451

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -26,7 +26,7 @@ jobs:
       - name: Build Epinio
         if: steps.branch.outputs.BRANCH_NAME == 'main'
         run: |
-          make build-all-small
+          make release
       - name: Login to Docker Hub
         uses: docker/login-action@v1
         with:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -26,7 +26,7 @@ jobs:
       - name: Build Epinio
```
rohitsakala, line 26: The name should be different since it is `make release`

thardeck: It still mainly builds the binaries, we just combined all the cleanup commands under release. @jimmykarily what do you think?
I do not mind calling it release but it is not completely true.

andreas-kupries (approved):
LGTM

rohitsakala (approved):

PR #475, opened 2021-06-10
Delete old pipelineruns only when new staging succeeds

### LINKED ISSUE #452: Only delete PipelineRuns when staging succeeds
This is a follow-up story to investigate is this is possible: https://github.com/epinio/epinio/pull/443#discussion_r644510047

andreas-kupries:
Was made to fix the issue #452 
The alternate pull request #483 looks to be better. It does the cleanup in the pipeline itself, as just another stage.

### DIFF
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -68,7 +68,7 @@ func (hc ApplicationsController) WaitUntilStaged(w http.ResponseWriter, r *http.
 
 	err = wait.PollImmediate(time.Second, duration.ToAppBuilt(),
 		func() (bool, error) {
-			l, err := client.List(context.TODO(),
+			l, err := client.List(ctx,
 				metav1.ListOptions{LabelSelector: models.EpinioStageIDLabel + "=" + stageId})
 			if err != nil {
 				return false, err
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -49,9 +49,8 @@ func (hc ApplicationsController) WaitUntilStaged(w http.ResponseWriter, r *http.
 
 	params := httprouter.ParamsFromContext(ctx)
 
-	// Not used right now. To come in a moment.
-	// org := params.ByName("org")
-	// name := params.ByName("app")
+	org := params.ByName("org")
+	name := params.ByName("app")
 	stageId := params.ByName("id")
 
 	log.Info("waiting until staging completes", "stageId", stageId)
@@ -96,6 +95,13 @@ func (hc ApplicationsController) WaitUntilStaged(w http.ResponseWriter, r *http.
 	resp := models.StageStatusResponse{}
 	if err != nil {
 		resp.ErrorMessage = err.Error()
+	} else {
+		// Sucessfully staged. We can now remove any old runs
+
+		err = application.Unstage(name, org, stageId)
+		if err != nil {
+			return singleInternalError(err, "failed delete previous pipeline runs")
+		}
 	}
 
 	err = jsonResponse(w, resp)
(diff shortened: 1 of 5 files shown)


internal/api/v1/stage.go
```diff
@@ -27,13 +31,87 @@ import (
...
+
+	cs, err := tekton.NewForConfig(cluster.RestConfig)
+	if err != nil {
+		return singleError(err, http.StatusInternalServerError)
+	}
+	client := cs.TektonV1beta1().PipelineRuns(deployments.TektonStagingNamespace)
+
+	err = wait.PollImmediate(time.Second, duration.ToAppBuilt(),
+		func() (bool, error) {
+			l, err := client.List(context.TODO(),
```
jimmykarily, line 71: I think you should pass the request context here instead

jimmykarily:
The only problem I see with this approach is, that it relies on someone checking the completion of staging with the new endpoint to do the cleanup. If another consumer of the API (the web dashboard for example), chooses to not check the status of staging, no cleanup would ever happen. This would be the case of "fire and forget" kind of staging.

Since the only consumer now is the cli and considering this approach fixes the problem for now, I would say it's ok. We can look for an alternative if this becomes insufficient at some point.

internal/api/v1/stage.go
andreas-kupries: Will investigate, hopefully fix.

andreas-kupries:
> The only problem I see with this approach is, that it relies on someone checking the completion of staging with the new endpoint to do the cleanup. If another consumer of the API (the web dashboard for example), chooses to not check the status of staging, no cleanup would ever happen. This would be the case of "fire and forget" kind of staging.

> Since the only consumer now is the cli and considering this approach fixes the problem for now, I would say it's ok. We can look for an alternative if this becomes insufficient at some point.

In response to the above I have managed to come up with 2 alternative ways of doing things.

  - Both have in common that the server does the polling and cleanup, however not triggered by the client, but by the `AppStage` itself.
  - First alternative would have any client do their own polling, as the cli does before this PR. Basically client polls for completion so that I can stop tailing the logs. Server polls for completion so that it can clean up pipeline runs. Easy to implement, nothing to sync between pieces, i.e. no coordination stuff required. However also loads the cluster API with more requests, with each client doing polling upping the load. 
  - Second alternative needs some global data structures in the server, and has more complexity. Server does the polling, client queries the server for the state of its polling. Either in a long-waiting request as the current code does, or by polling the server. The complexity comes in because the server has to store the polling result in memory for the (multiple) clients to find on the one hand, and not keep the information around for too long either, to keep memory usage down in the face of clients not asking for the state.

andreas-kupries:
If we go with these alternate solutions I would recommend we start with the first, where clients and server poll spearately. Only if this really gives us trouble because of load on the cluster should we consider the definitely more complex second alternative, as it requires much more coordination logic (global data structures, locking thereof, ...).

internal/api/v1/stage.go
andreas-kupries: Done, commit f25180ea1e97fd15e121ff23ab2390123ed23f0d

andreas-kupries:
Another possible solution, coming out of chatting about this.

See if we can extend the existing pipeline with a stage which would delete the old pipeline runs after it has started the app (posted the deployment).

- That would keep the client as is, polling for itself
- And the server would be pretty much unchanged as well (just the embedded pipeline definition would change)
- And we do not have to care about other clients.
- Clients would have no say when the deletion/cleanup happens, and do not have to care either.

Some research with tekton is required tough.

Onwards into that fray ...

andreas-kupries:
That alternative solution works.
See alternate PR #483

andreas-kupries:
Overcome by #483

PR #478, opened 2021-06-11
Internal docs explaining the logic of concurrently deleting apps

andreas-kupries:
A follow up to the recently merged #460 , adding documentation on the inner workings of that code.

### DIFF
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -162,7 +162,50 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 		return err
 	}
 
-	// create a buffer, so there is no more than 'maxConcurrent' goroutines running at the same time
+	// Operation of the concurrent code below:
+	//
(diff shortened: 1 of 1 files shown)


internal/api/v1/organizations.go
```diff
@@ -162,7 +162,50 @@ func deleteApps(cluster *kubernetes.Cluster, gitea *gitea.Client, org string) er
 		return err
 	}
 
-	// create a buffer, so there is no more than 'maxConcurrent' goroutines running at the same time
+	// Operation of the concurrent code below:
```
jimmykarily, line 165: I think any method that needs so much documentation is probably a method that should be re-written in a simpler way. On the other hand, I can't think of a simpler way right now so I'm ok to stick with the docs.

jimmykarily (approved):

internal/api/v1/organizations.go
andreas-kupries: I suspect that even "simple" concurrent code, like we have here, always will at least this much documentation.

Issue #479, opened 2021-06-11
Installation fails on MacOS


manno:
<details>
 <summary>Full Log</summary>

```
% ./epinio-darwin-amd64 install                                                                                                   
2021/06/11 12:07:34 InstallClient/Install "level"=1 "msg"="start"

🚢 Epinio installing...
2021/06/11 12:07:34 InstallClient/Install "level"=2 "msg"="process cli options"
2021/06/11 12:07:34 InstallClient/Install "level"=2 "msg"="fill defaults into options"
2021/06/11 12:07:34 InstallClient/Install "level"=2 "msg"="show option configuration"

Configuration...
  🧭  skip-traefik: false
  🧭  skip-linkerd: false
  🧭  system_domain:
  🧭  email_address: epinio@suse.com
  🧭  user: 6ca63b2809995443
  🧭  password: cf66b2a2baa2d4a2
2021/06/11 12:07:34 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="linkerd"

🚢 Deploying Linkerd...Executing  kubectl apply --filename /var/folders/0j/zmrkp2sj2xjc_63fpfg4skv00000gn/T/epinio865848608
Executing  kubectl apply --filename /var/folders/0j/zmrkp2sj2xjc_63fpfg4skv00000gn/T/epinio147322367

✔️  Linkerd deployed
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="traefik"

⚠️  Traefik Ingress already installed, skipping
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="ensure system-domain"
.
✔️  Created system_domain: 172.19.0.2.omg.howdoi.website
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="cert-manager"
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="epinio"
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="quarks"
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="epinio-registry"
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="tekton"
2021/06/11 12:08:55 InstallClient/Install "level"=2 "msg"="deploy"  "Deployment"="service-catalog"
[...]

PR #480, opened 2021-06-11
Add MacOS worker on Github Actions

manno:
refers to https://github.com/epinio/epinio/issues/479


manno:
No docker in the MacOS worker: 
> We have had some discussions with Docker about doing this but they have held firm on their request that using Docker on a service requires an enterprise license and docker enterprise is not at all supported on macOS.

> For Windows Microsoft holds the license for Docker Enterprise on Windows so we do not have any issues there and for Linux we use a specific build of Docker for Linux that is built and maintained by Azure.

jimmykarily:
We don't have a simple solution for this right now and it's not yet a priority. Closing.

PR #482, opened 2021-06-11
Make service params be a json string and validate it

### LINKED ISSUE #481: Allow passing helm chart parameters through minibroker
Although we allow passing parameters with:

```
epinio service create wordpressdb3 mariadb 10-3-22  auth.database wordpress
```

1. The parameters are written in the wrong place: 
  https://github.com/epinio/epinio/blob/main/internal/services/catalog_service.go#L401-L404 (They should be inside `spec`: https://github.com/kubernetes-sigs/service-catalog/blob/master/docs/parameters.md#design)
2. We only support top level keys. The example above needs to set an object `{ "auth": { "database": "wordpress"}}` but it sets `{ "auth.database": "wordpress"}` which doesn't have the [desired effect](https://github.com/bitnami/charts/tree/master/bitnami/mariadb#mariadb-common-parameters)

## Acceptance Criteria

- The user should be able to pass a full json instead of just key-value pairs
- There is a test that checks that it is possible to provision a mariadb database by defining the database in params. The database should automatically be created (the mariadb pod has mysql cli available so we can `kubectl exec` something to check if the database is there).

jimmykarily:
because the user may need to specify deeply nested params

Fixes #481

### DIFF
--- a/acceptance/catalog_services_test.go
+++ b/acceptance/catalog_services_test.go
@@ -1,6 +1,9 @@
 package acceptance_test
 
 import (
+	"fmt"
+
+	"github.com/epinio/epinio/helpers"
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
@@ -17,6 +20,17 @@ var _ = Describe("Catalog Services", func() {
 	Describe("service create", func() {
 		It("creates a catalog based service, with waiting", func() {
 			makeCatalogService(serviceName)
+		})
+
+		It("creates a catalog based service, with additional data", func() {
+			makeCatalogService(serviceName, `{ "db": { "name": "wordpress" }}`)
+			serviceInstanceName := fmt.Sprintf("service.org-%s.svc-%s", org, serviceName)
+
+			out, err := helpers.Kubectl(
+				fmt.Sprintf("get serviceinstance -n %s %s -o=jsonpath='{.status.externalProperties.parameters.db.name}'",
+					org, serviceInstanceName))
+			Expect(err).ToNot(HaveOccurred())
+			Expect(out).To(Equal("wordpress"))
 		})
 
 		It("creates a catalog based service, without waiting", func() {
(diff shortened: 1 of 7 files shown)


jimmykarily:
Example to create a mariadb database:

```
./dist/epinio-linux-amd64  service create wordpressdb6 mariadb 10-3-22  --data '{ "db": { "name": "wordpress" }}'
```

(the value we set is this: https://github.com/helm/charts/blob/master/stable/mariadb/values.yaml#L134 )

jimmykarily:
This PR misses a test

andreas-kupries:
Looks good to me, with the exception of missing tests, an issue you are already aware of.

jimmykarily:
@andreas-kupries I added a test: https://github.com/epinio/epinio/pull/482/files#diff-4b7ee420d22ec3d140ce703ec3716d7af436c3d5bffcb2e91915fa8038cb122dR25 . Can you have another look?

acceptance/catalog_services_test.go
```diff
@@ -19,6 +22,19 @@ var _ = Describe("Catalog Services", func() {
 			makeCatalogService(serviceName)
 		})
 
+		It("creates a catalog based service, with additional data", func() {
+			makeCatalogService(serviceName, `{ "db": { "name": "wordpress" }}`)
+			serviceInstanceName := fmt.Sprintf("service.org-%s.svc-%s", org, serviceName)
+			Eventually(func() string {
```
jimmykarily, line 28: Hmm, now that I think again, `Eventually` shouldn't be needed because `makeCatalogService` already waits for the service to appear. Also, if the service wasn't there the `Expect(err).ToNot(HaveOccurred())` would raise so the Eventually wouldn't help. I will try to remove the `Eventually` and test locally if it works.

jimmykarily: removed `Eventually` and it works fine.

andreas-kupries (approved):
LGTM.

PR #483, opened 2021-06-11
Move unstaging of old runs into the pipeline itself

### LINKED ISSUE #452: Only delete PipelineRuns when staging succeeds

andreas-kupries:
Fixes #452 

The competing #475 has been closed as overcome by this one.

### DIFF
--- a/assets/embedded-files/tekton/pipeline.yaml
+++ b/assets/embedded-files/tekton/pipeline.yaml
@@ -79,6 +79,21 @@ spec:
     workspaces:
     - name: source
       workspace: source
+  - name: clean
+    taskRef:
+      name: clean
+    params:
+      - name: APP_NAME
+        value: "$(params.APP_NAME)"
+      - name: ORG
+        value: "$(params.ORG)"
+      - name: STAGE_ID
+        value: "$(params.STAGE_ID)"
+    runAfter:
+    - stage
+    workspaces:
+    - name: source
+      workspace: source
 ---
 apiVersion: tekton.dev/v1beta1
 kind: Task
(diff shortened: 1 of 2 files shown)


assets/embedded-files/tekton/pipeline.yaml
```diff
@@ -79,6 +79,21 @@ spec:
...
+      name: clean
+    params:
+      - name: APP_NAME
+        value: "$(params.APP_NAME)"
+      - name: ORG
+        value: "$(params.ORG)"
+      - name: STAGE_ID
+        value: "$(params.STAGE_ID)"
+    runAfter:
+    - stage
```
jimmykarily, line 93: Maybe it's better to run the cleanup after `run` instead of `stage`. `run` is the task that creates/updates the Deployment Service and Ingress, so the staging process is really done when we finally run the provided image.

assets/embedded-files/tekton/pipeline.yaml
```diff
@@ -235,3 +250,41 @@ spec:
...
+      - -c
+      - |
+        # (kubectl get ..) Get all pipeline runs for the application specified by name and org.
+        # (jsonpath)       Just their names.
+        # (if test)        Exclude the current run (name == stage_id).
+        # (for/kube ...)   Delete the remainder.
+        #
+        echo "Cleaning up old runs of application '$(params.APP_NAME)' in organization '$(params.ORG)'"
+        echo "Current run to ignore is '$(params.STAGE_ID)'"
+        for run in $(kubectl get pipelinerun -n tekton-staging -l "app.kubernetes.io/name=$(params.APP_NAME),app.kubernetes.io/part-of=$(params.ORG)" -o "jsonpath={.items[*].metadata.name}")
```
jimmykarily, line 285: Like we discussed in person, if the user keeps pushing an application that fails to stage, all the failed staging resources will be left in the cluster until either a successful staging happens or the application is deleted.

I'm not sure if this is ok of if it would be better to:

- Delete the previous PipelineRuns if there is a success
- Delete the previous PipelineRuns even if the current one fails but there was no previous success (so that we only keep one failure around when all we got so far was failures). After all there is no running application to need to match to any staging logs.

If you also don't have a strong opinion on this, we can keep the current solution and revisit this is there is a different user need to cover in the future.

assets/embedded-files/tekton/pipeline.yaml
andreas-kupries: Can be done. Trivial one-line change. Just a moment.

assets/embedded-files/tekton/pipeline.yaml
andreas-kupries: > Delete the previous PipelineRuns even if the current one fails but there was no previous success

Even at first thought this logic would require quite some complex checks of the pipelineruns, and of the deployment, i.e. determine stage id of the running deployment, if any, and then take that into account. 

Actually this might have to be done at tekton level. Because right now the `cleanup` step runs only if the previous step (`stage` (or `run` in a moment) succeeds. Nothing happens on failure. And to handle your second case tekton would have to be told to run a step when a previous one fails! I.e. If `stage` fails run X, else (on success) run Y.

Definitely spiky again, IMHO.

assets/embedded-files/tekton/pipeline.yaml
andreas-kupries: Done, commit 3c392323b8a6f3030e6da07f00aab5e3cc526510

jimmykarily (approved):

Issue #489, opened 2021-06-15
Document the need for Traefik to be meshed with Linkerd in Epinio


andreas-kupries:
Document that

  - Traefik has to be meshed with Linkerd for fully secure communication from outside to epinio API, and applications.
  - This meshing is done automatically by the standard install commands (`install`, and `install-ingress` (See #491)).
  - A non-epinio Traefik (Custom user's, or system (`k3d`)) is __not__ meshed by default, and thus less secure.

See if we can document how to fix that issue for a non-epinio Traefik (configuration, and other steps to take).

### Old description

Epinio currently deploys Traefik after deploying Linkerd, and annotates Traefik for integration with the Linkerd mesh.

There are persistent doubts floating around about actually needing Traefik to be part of the mesh.

For example, communication with the Ingress Controller part of Traefik is indirect, by posting and removing Ingress kube resources.

On the other hand, Traefik seems to establish an `svclb` pod which might be a LoadBalancer. That would have to be part of the mesh to secure the communication from that LB to the service endpoints in Epinio and Application Pods.

On the third hand are some clues pointing to the svclb pod "only" setting up iptables rules for the necessary routing. In that case the pod may not be involved at all ?!

andreas-kupries:
@kkaempf In the discussion this morning, after standup, we wondered if Suse has experts around "Virtual Networking" as done by Linkerd or other tooling, which could help explain to us how this works at deeper levels. The name "Cilium" also floated around. Could you investigate/find such people ?

kkaempf:
Answered in Slack :wink:

andreas-kupries:
Answer: Yes, the meshing is needed, per various Slack threads.

Have to document the unmeshed cases (non-epinio Traefik).

andreas-kupries:
Renaming the issue now.

Issue #488, opened 2021-06-15
Epinio Production Install IP Chicken and Egg


andreas-kupries:
While trying to deploy Epinio on Digital Ocean a problem with the sequence in which things are being installed and provisioned was hit.

Up to now, Epinio installs its components in the following sequence:

- It installs Linkerd
- It installs Traefik (on an annotated namespace, so traefik control plane pods are meshed)
- It installs cert-manager
- It installs Epinio server (the Epinio API endpoint)
- It creates a certificate for the Epinio server using cert-manager CRDs

It actually does more, but these are the steps that reproduce the issue we have. The first bullet (linkerd) is probably also irrelevant to this discussion.

When a custom system domain is used (not the `omg.howdoi.website`), Epinio will ask the cert-manager to create a let's encrypt certificate for the Epinio server. This involves an HTTP challenge which is automatically solved by the cert-manager. Before that happens, cert-manager does a "self test" request, trying to access the http url which satisfies the http challenge. That url is in the form of "http://mycustomdomain.org/..."
Obviously for this to work, the domain used by the user (the system domain) should already resolve to the API address of the loadbalancer (in other words it should point to the cluster).

Long story short, the moment we ask the cert-manager to create a let's encrypt certificate for a custom domain, that domain must already point to the cluster. If that is not the case, the failed DNS query is cached and the cert-manager doesn't recover (maybe it will after DNS cache has expired, we didn't wait that long).

The chicken-egg problem in this case was that we don't know the loadbalancer IP address, before we deploy a service on the cluster that actually needs a load balancer IP. `epinio install` is the thing that triggers the IP address allocation but also the thing that will fail if the system domain doesn't resolve to that IP address. Obviously there is no time for the user to setup the DNS while `epinio install` is running (and before we hit the cert-manager problem).

We thought of 2 ways to solve this problem:

1. We implement an `epinio install traefik` command to allow the user to only install traefik to the cluster. This should trigger the IP address allocation. We will print the IP address when this command returns so the user can setup the system domain to point to that IP address. When that's done, the user can run `epinio install` to resume the epinio installation.

[...]

Issue #476, opened 2021-06-10
`--traefik-loadbalancer-ip` flag in `epinio installation`


jimmykarily:
## Why

Controlling epic #488 

Because before we can let cert-manager create any let's encrypt certs, the system-domain has to resolve to the load balancer IP address. Otherwise the self test will fail and the cert-manager won't be able to recover before certain time has passed.

We will split the traefik installation to a separate command (#477 ).
The `epinio install` command won't install traefik unless `--traefik-loadbalancer-ip` flag is used.  When it is set, it will be passed to the Epinio endpoint service in the LoadBalancerIP field. This setting is respected by most public cloud providers (but not by DigitalOcean).
If a public cloud provider doesn't respect this setting, the user will have to install the traefik using the separate command, connect the generated IP address with their system domain before they can proceed with `epinio install` (in which case they won't provide the `--traefik-loadbalancer-ip` flag).

## Flow

- The user buys a static IP address from the cloud provider (NOTE: This is not possible in DigitalOcean
but it's possible in most other public cloud providers).
- The user creates an A record to point their custom domain to that IP address
- The user runs `epinio install --traefik-system-domain mydomain.org --loadbalancer-ip 1.2.3.4`

The last command does a check to see if the system domain actually resolves to the given IP address.
If not, a warning is printed ([TBD]: exit then or continue?)

epinio install --with-traefik-loadbalancer-ip 1.2.3.4 --system-domain mydomain.org

manno:
What about the simple dev use case with omg.howdoi.website?

`epinio install --with-traefik`?

jimmykarily:
We can have the following flags:

```
--traefik-install # default false
--loadbalancer-ip # default not set
--system-domain # default 1.2.3.4.omg.howdoi.website
--production-certs # default false
```

Most combinations of these make sense except maybe for:

`--traefik-install false --loadbalancer-ip 1.2.3.4` because the ip is then not used anywhere (since we won't install traefik).

We should also make clear that the system-domain should already be pointing to the ip when both are set, otherwise we will hit the cert-manager issue. With the right validations and documentation we can get this right.

jimmykarily:
Chris pointed out that it may be hard to recover if the user releases the IP address after installing Epinio. E.g.

- The user gets an Elastic IP address (AWS) to be used as `--loadbalancer-ip` in `epinio install`
- The user sets up system domain with this IP address
- The user installs Epinio
- The user releases that Elastic IP address
- Boom!

[...]

PR #491, opened 2021-06-15
Separate Traefik Installation

### LINKED ISSUE #477: Split traefik installation to a separate command
## Why

Controlling epic #488 

Because when we install Epinio on a public cloud provider, before we let the cert-manager do the self test (for the http challenge in the case of custom system-domain), we need to be sure that the system domain resolves to the traefik load balancer IP address. Otherwise, the failed DNS response will be cached and even if the user fixes the DNS later, they cert-manager will keep failing. This will prevent it from creating a let's encrypt certificate.

For public cloud providers that support LoadBalancerIP a better option is the provide the --traefik-loadbalancer-ip flag in `epinio installation` (#476)

## How

The user may or may not want to install traefik. We already provide a way to skip it, but no
way to install it separately. We will replace the `--skip-traefik` flag of the `epinio install`
command, with a separate `epinio install traefik` command (name of command to be discussed).

This new command will install Traefik on the cluster and wait until it gets and IP address.
If no IP address appears until the timeout, the command will fail with an error.

We document that a System Traefik, or a user's custom Traefik will not be part of the Linkerd mesh.
Only the Traefik handled by epinio itself will be.

## Flow

1. The user installs Traefik with this new command
2. The user creates a DNS A record pointing their custom system-domain to that IP address
3. The user installs Epinio with `epinio install --system-domain mydomain.org`

## Notes

The quick start configuration may be affected, depending on the cluster foundation.

1. With `k3d` as foundation nothing changes in terms of user commands, nor internally, as k3d supplies its own Traefik, and Epinio foregoes installing its own. Now it simply never installs it automatically.
2. Effects are unknown for other foundations, like `minikube`, `KinD`, etc.

andreas-kupries:
Fixes #477 

User can install `linkerd` and `traefik` separately, via `epinio install-ingress`.
This separation then allows setup of DNS for custom domains.
I.e. point domain to the IP of the main traefik ingress.

See #488 for the main epic explaining the underlying issue.
See #476 for an competing/supporting change.
See #489 for needed documentation in this area (Traefik/Linkerd meshing)

### DIFF
--- a/deployments/traefik.go
+++ b/deployments/traefik.go
@@ -166,11 +166,35 @@ func (k Traefik) Deploy(ctx context.Context, c *kubernetes.Cluster, ui *termui.U
 	//  c  | no      | yes       | Something has claimed the namespace | Error
 	//  d  | no      | no        | Namespace is free for use           | Deploy
 
-	// Bug: The current order of checking (namespace first, then
-	// service), and associated actions, mishandles case A (Error
-	// instead of nothing).
+	log.Info("check presence, local service")
 
-	log.Info("check presence, namespace")
+	_, err = c.Kubectl.CoreV1().Services(TraefikDeploymentID).Get(
+		ctx,
+		"traefik",
+		metav1.GetOptions{},
+	)
+	if err == nil {
+		log.Info("service present")
+
+		ui.Exclamation().Msg("Traefik Ingress already installed, skipping")
+		return nil
+	}
+
+	log.Info("check presence, system service")
+
+	_, err = c.Kubectl.CoreV1().Services("kube-system").Get(
+		ctx,
+		"traefik",
+		metav1.GetOptions{},
+	)
+	if err == nil {
+		log.Info("service present")
+
+		ui.Exclamation().Msg("System Ingress present, skipping")
+		return nil
+	}
+
+	log.Info("check presence, traefik namespace")
 
 	_, err = c.Kubectl.CoreV1().Namespaces().Get(
 		ctx,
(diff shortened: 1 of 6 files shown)


deployments/traefik.go
```diff
@@ -166,24 +166,21 @@ func (k Traefik) Deploy(ctx context.Context, c *kubernetes.Cluster, ui *termui.U
...
-	log.Info("check presence, namespace")
-
-	_, err = c.Kubectl.CoreV1().Namespaces().Get(
+	_, err = c.Kubectl.CoreV1().Services(TraefikDeploymentID).Get(
 		ctx,
-		TraefikDeploymentID,
+		"traefik",
 		metav1.GetOptions{},
 	)
 	if err == nil {
```
jimmykarily, line 184: This is assuming that any error means the Service is missing. I think we should also check that if there is an error, it is about the service not being there, otherwise exit with an error. Something like:

```
if err == nil {
...
} else if !apierrors.IsNotFound(err)` {
  return err
}
```

jimmykarily: same thing below

andreas-kupries: Sure. Can do that.

jimmykarily (approved):

jimmykarily (approved):

deployments/traefik.go
andreas-kupries: Done. Commit 416156b7b1867e1d0570dcffc5c4b2dae3d3e1a7

PR #492, opened 2021-06-15
Cleanup some small things

manno:
Refers to #347 

* applied tekton.yaml message, was introduced for testing
* nil error in utilities, should be called after error checking
* quarks-secret debug message spam in logs every seconds
* models.Logs is an actor that uses a context, should go into application package
* kubeClient is a `Cluster`, not a client

### DIFF
--- a/acceptance/utilities_test.go
+++ b/acceptance/utilities_test.go
@@ -47,10 +47,10 @@ func Client() *http.Client {
 // func Curl is used to make requests against a server
 func Curl(method, uri string, requestBody *strings.Reader) (*http.Response, error) {
 	request, err := http.NewRequest(method, uri, requestBody)
-	request.SetBasicAuth(epinioUser, epinioPassword)
 	if err != nil {
 		return nil, err
 	}
+	request.SetBasicAuth(epinioUser, epinioPassword)
 	return Client().Do(request)
 }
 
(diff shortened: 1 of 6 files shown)


andreas-kupries (dismissed):
Looks good to me. I am quite ok with the `kubeClient`/`cluster` change. Should make things more readable, i.e. type behind the field. I know I had issues rememberin.

acceptance/utilities_test.go
```diff
@@ -47,10 +47,10 @@ func Client() *http.Client {
 // func Curl is used to make requests against a server
 func Curl(method, uri string, requestBody *strings.Reader) (*http.Response, error) {
 	request, err := http.NewRequest(method, uri, requestBody)
-	request.SetBasicAuth(epinioUser, epinioPassword)
 	if err != nil {
 		return nil, err
 	}
+	request.SetBasicAuth(epinioUser, epinioPassword)
 	return Client().Do(request)
```
andreas-kupries, line 54: D'oh (I was the person writing that bug). Thanks for fixing.

manno:
> Looks good to me. I am quite ok with the `kubeClient`/`cluster` change. Should make things more readable, i.e. type behind the field. I know I had issues rememberin.

Yes, but I didn't want to do another tree-wide edit... or should I? :)

andreas-kupries:
We can edit the other relevant files as we change them anyway for other PRs. Cleanup takes longer, also not a tree-wide thing.

andreas-kupries (approved):
Still ok

PR #493, opened 2021-06-15
Add AppCRD

manno:
refers to #347 - this will add (but it's not ready) creation and deletion of the app crd resource and make it the single source of truth for a limited set of features.

### comments

The fields of AppCRD are not a good match for us, I think. We also don't introduce the AppCRD controller, so no automatic status and ownerRef handling. AppCRD doesn't offer a typed client, which is a nuisance. Also it's a v1beta1 (not v1) CRD. We might need to think about building our own CRD (JSON schema + struct + generated typed client) later, once we know our requirements better? 

### new comments

https://github.com/epinio/epinio/issues/347#issuecomment-860521328


internal/api/v1/servicebindings.go
```diff
@@ -98,7 +100,7 @@ func (hc ServicebindingsController) Create(w http.ResponseWriter, r *http.Reques
 	}
 
 	for _, service := range theServices {
-		err = app.Bind(ctx, service)
+		err = wl.Bind(ctx, service)
```
jimmykarily, line 103: A service is bound to an application, not a specific workload. Even if the application didn't have a workload, we should still be able to bind the service to the application. As soon as a workload appears, it will have the service details mounted just because the Service was bound to the App. I don't mean we need to implement this functionality now but the change in the narrative is confusing.

internal/api/v1/applications.go
```diff
@@ -149,7 +198,8 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
 		return NewBadRequest("instances param should be integer equal or greater than zero")
 	}
 
-	err = app.Scale(r.Context(), updateRequest.Instances)
+	workload := application.NewWorkload(cluster, app)
+	err = workload.Scale(r.Context(), updateRequest.Instances)
```
jimmykarily, line 202: NOTE: We need to check if a `workload` actually exists before we try to scale it. E.g. 
- The user creates an application
- The user tries to scale it (before any staging has happened)
In this case we don't want to have a `500` error but rather let the user know they have to deploy first.

Issue #494, opened 2021-06-15
Windows: Cannot handle ANSI terminal control for output colorization


gunamata:
The characters are printed wrong on cmd.exe.

C:\>epinio install

←[34m🚢 Epinio installing...
←[0m
Configuration...
  🧭  skip-traefik: ←[35mfalse←[0m
  🧭  skip-linkerd: ←[35mfalse←[0m
  🧭  system_domain: ←[32m←[0m
  🧭  email_address: ←[32mepinio@suse.com←[0m
  🧭  user: ←[32ma7f36a3b8756de3f←[0m
  🧭  password: ←[32m558f1bbd2d8fbebc←[0m

←[34m🚢 Deploying Linkerd...←[0m

andreas-kupries:
Right. The full sequence are like `\033[0m` and the like. ANSI terminal control sequences.
On unix terminals they colorize the output. The inivisible `\033` is the escape character.
The example at the beginning means `reset everything to plain output`.

We have to disable that for Windows, looks like.
We use the go package `github.com/fatih/color`.

andreas-kupries:
Some investigation shows me that `fatih/color` should support Windows, by means of `mattn/go-colorable`.
Should be since 2017 after a first glance.

... for a first, upgrade from 1.90 to 1.12.0.
That brings in support for the environment variable `NO_COLOR`.
Simply setting that variable (any value) will disable colorized output.
This gives us a workaround when things go wrong.

andreas-kupries:
See PR #498 for a possible fix.

Even if not a fix, it should make a workaround available (Support for environment variable `NO_COLOR`. Setting it (to any value) should disable colorization.

andreas-kupries:
A question: What version of Windows is this ?

Issue #468, opened 2021-06-09
Extensions to the TLS/Cert handling for secure API


andreas-kupries:
Getting some experience with the new secured epinio API (TLS, Certs in config) we found some issues, mainly around easy switching between multiple epinio's on different clusters. Because the config contains epinio/cluster specific information, the aforementioned cert, and the API credentials (for a bit longer than that even).

After some discussion we have chosen to change the current state and extend it as follows:

  1. Keep the storage of cert and credentials in the config
  2. Create a new toplevel command `config` for operations on the configuration.
  3. Create a config subcommand `update` which reads cluster-specific configuration, i.e. cert and credentials from the currently targeted cluster, and saves this information into the config. The pulled data overwites the existing information in the config.
  4. Rework the install command. Remove the existing code for saving cert and credentials into the config, and run the new `config update` internally instead. In the same manner as is already done for `create-org workspace ; target workspace`.
  5. The `config` command can be extended in the future with an `add-cert` command for adding additional certs to the configuration. These certs should however be stored separately from the cluster cert, so that changing clusters will not clear them (See above the note about `update` overwriting data). That said, this extension is __not in scope for this ticket__
  6. Document the new features, under `Advanced` docs, or regular docs. Not in the `README` however, this has no bearing on the quick setup flow for a first user.

### Acceptance criteria

  1. [x] New `config` command.
  2. [x] New `config update` command.
  3. [x] Unchanged user-visible behaviour for `install` (The operations updating the config simply move internally)
  4. [ ] Stretch: Test cases for switching clusters ?
     Note that this would force the EAT to require two separate clusters, instead of the currently one cluster.
  5. [x] New test cases to confirm cert storage after `install` ?!
  6. [x] New test cases to confirm `config update` is working (Squash config, run update, confirm new config equals old).
  7. [x] documentation of the new command and features.

#### The data the discussion was based on

Arguments:

  1. Secure
  2. Easy switching between Clusters
  3. Support future custom CAs by customers
  4. Only small number of roundtrips from client to server

|     |Chosen       |Proposal/State   |Secure|Easy Switching Clusters|Future CAs|+1 Roundtrip|
|---  |---          |:---             |---   |---                    |---       |---         |
|     |             |Andreas (Current)|:ok:  |:boom:                 |:ok:      |:ok:        |
|     |             |Dimitry          |:boom:|:ok:                   |:boom:    |:ok:        |
[...]

PR #498, opened 2021-06-16
Upgrade to latest release of color package

andreas-kupries:
NO_COLOR support, plus bugfixes in deeper packages
(go-colorable f.ex is used for windows support)

Ref #494


andreas-kupries:
@gunamata Please try to build a window binary from this PR and check if color control is still broken.

agracey:
Would it be possible to make this controllable via ~/.config/epinio/config.yaml as well?

andreas-kupries:
I see no problems with doing that. And/or a global `--color` option. Which I had thought of.

With #468 in the works I will even have a place to hook a command managing the color setting into.

PR #505, opened 2021-06-18
Use app reference instead of app model

manno:
As it turns out, the functions dealing with k8s workloads, don't need the app model at all.
The struct member doesn't even need to be public.
(yes, we could just pass the appref as an arg to the functions, but we already have the struct)

Also cleaning up the workload struct's file, by moving unrelated functions into the 'application' file. 

Removes the app model from the workload struct.
It was there to maintain compatibility with existing code, but it turns
out the workload manager doesn't need access to the app model.

We pass just the reference to the app (org, name). The code is no longer
able to mutate the passed in app model.
The one function that did need to mutate the model (Complete()) always
returned it.

### DIFF
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -198,7 +198,7 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request) 
 		return NewBadRequest("instances param should be integer equal or greater than zero")
 	}
 
-	workload := application.NewWorkload(cluster, app)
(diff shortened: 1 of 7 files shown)


andreas-kupries (approved):
Looks good to me. Now running acceptance tests locally.

andreas-kupries:
local acceptance had only

```
[2] [Fail] Apps API Application Endpoints Staging when staging a new app [It] returns a success 
[2] /home/work/SUSE/dev/epinio/acceptance/api_v1_applications_test.go:405
```

failing (flake 6)

PR #519, opened 2021-06-22
feat: reworked the docs for dev deployment

andreas-kupries:
Doc only changes. No functional changes. CI troubles can be ignored.

Rephrased to impersonal voice, rearranged and explained the special commands needed, and added notes on the latest post-deployment actions of install.

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -104,7 +104,7 @@ running cluster.
 Another thing `epinio install` does after deploying all components is
 the creation and targeting of a standard organization, `workspace`.
 
-Wtih the failing server component these actions will fail, making the
+With the failing server component these actions will fail, making the
 installation fail. Using the option `--skip-default-org` instructs the
 command to forego these actions. Which in turn makes it necessary to
 run them manually to reach the standard state of a cluster. These are
--- a/docs/development.md
+++ b/docs/development.md
@@ -36,53 +36,95 @@ This is building Epinio for linux on amd64 architecture. If you are on a
 different OS or architecture you can use one of the available `build-*` targets.
 Look at the Makefile at the root of the project to see what is available.
 
-
 ### Installing Epinio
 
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -36,53 +36,95 @@ This is building Epinio for linux on amd64 architecture. If you are on a
...
 
-
 ### Installing Epinio
 
 You can have a look at [the dedicated document](/docs/install.md) for cluster
 specific instructions, but generally this should be sufficient to get you running:
 
 ```
-./dist/epinio-linux-amd64 install
+export EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1
```
thardeck, line 45: I suppose it is irrelevant if we use `true` or `1`?

docs/development.md
```diff
@@ -36,53 +36,95 @@ This is building Epinio for linux on amd64 architecture. If you are on a
...
+be adjusted accordingly.
+
+After making changes to the binary simply invoking `make
+patch-epinio-deployment` again will upload the changes into the
+running cluster.
+
+Another thing `epinio install` does after deploying all components is
+the creation and targeting of a standard organization, `workspace`.
+
+Wtih the failing server component these actions will fail, making the
```
thardeck, line 107: Typo `With`

thardeck (changes requested):

docs/development.md
andreas-kupries: Would have to check.

docs/development.md
andreas-kupries:
```suggestion
With the failing server component these actions will fail, making the
```

docs/development.md
andreas-kupries: Checked: Any non-empty string is sufficient.
https://github.com/epinio/epinio/blob/main/deployments/epinio.go#L195

thardeck (approved):

PR #525, opened 2021-06-22
Create application resource when pushing an app

### LINKED ISSUE #509: Pushing an app always creates an application resource
We previously introduced the app CRD and a separate endpoint. Pushing an app should also create an application resource, so that we always have one for workloads (=deployments).

## Acceptance Criteria

* epinio push first calls the new app create endpoint

manno:
Using `app push` will always call the AppCreate endpoint to create an application resource.

fixes #509

Some notes:

* APIError implemented the error interface, MultiError did not. Since both are interchangeable, adding `Error()`
* Not sure we ever use APIError as an error, though
* Adding the application resource test to an existing test case (golang app), to avoid another install/wait/delete cycle, could be it's own test, or I could merge it with even more tests (ingress/service test case is a good candidate)

### DIFF
--- a/internal/api/v1/errors.go
+++ b/internal/api/v1/errors.go
@@ -15,14 +15,13 @@ type ErrorResponse struct {
 	Errors []APIError `json:"errors"`
 }
 
-// APIErrors interface is used by all the handlers to return a single or
-// multiple errors
+// APIErrors interface is used by all handlers to return one or more errors
 type APIErrors interface {
 	Errors() []APIError
 	FirstStatus() int
 }
 
-// APIError fulfills the error and the APIErrors interface and contains a single error
+// APIError fulfills the error and APIErrors interfaces. It contains a single error.
 type APIError struct {
 	Status  int    `json:"status"`
 	Title   string `json:"title"`
@@ -54,7 +53,7 @@ func NewAPIError(title string, details string, status int) APIError {
 	}
 }
 
-// MultiError fulfills the APIErrors interface and contains multiple errors
+// MultiError fulfills the APIErrors interface. It contains multiple errors.
 type MultiError struct {
 	errors []APIError
 }
--- a/internal/api/v1/errors.go
+++ b/internal/api/v1/errors.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-// All our actions match this type. They can return a list of errors.
+// APIActionFunc is matched by all actions. Actions can return a list of errors.
 // The "Status" of the first error in the list becomes the response Status Code.
 type APIActionFunc func(http.ResponseWriter, *http.Request) APIErrors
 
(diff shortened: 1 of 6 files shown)


internal/api/v1/errors.go
```diff
@@ -15,25 +15,29 @@ type ErrorResponse struct {
 	Errors []APIError `json:"errors"`
 }
 
+// APIErrors interface is used by all the handlers to return a single or
+// multiple errors
```
andreas-kupries, line 19:
```suggestion
// APIErrors interface is used by all handlers to return one or more errors
```

internal/api/v1/errors.go
```diff
@@ -15,25 +15,29 @@ type ErrorResponse struct {
...
 type APIErrors interface {
 	Errors() []APIError
 	FirstStatus() int
 }
 
+// APIError fulfills the error and the APIErrors interface and contains a single error
```
andreas-kupries, line 25:
```suggestion
// APIError fulfills the error and APIErrors interfaces. It contains a single error.
```

internal/api/v1/errors.go
```diff
@@ -50,12 +54,20 @@ func NewAPIError(title string, details string, status int) APIError {
 	}
 }
 
-var _ APIErrors = MultiError{}
-
+// MultiError fulfills the APIErrors interface and contains multiple errors
```
andreas-kupries, line 57:
```suggestion
// MultiError fulfills the APIErrors interface. It contains multiple errors.
```

internal/api/v1/errors.go
```diff
@@ -15,25 +15,29 @@ type ErrorResponse struct {
 	Errors []APIError `json:"errors"`
 }
 
+// APIErrors interface is used by all the handlers to return a single or
```
andreas-kupries, line 18:
```suggestion
```

andreas-kupries (changes requested):
Some tweaks to various comments.

internal/cli/clients/client.go
```diff
@@ -417,12 +417,12 @@ func (c *EpinioClient) DeleteService(name string, unbind bool) error {
...
 			// names.
 
-			var apiError map[string][]api.APIError
+			var apiError api.ErrorResponse
 			if err := json.Unmarshal(bodyBytes, &apiError); err != nil {
 				return err
 			}
 
-			bound := strings.Split(apiError["errors"][0].Details, ",")
+			bound := strings.Split(apiError.Errors[0].Details, ",")
```
jimmykarily, line 425: Isn't there a case when `apiError.Errors` has no elements? Wouldn't this fail then? (the old code would have this issue too, I just wonder how we can be sure that `[0]` here exists)

andreas-kupries: The underlying structures are for one or more errors. (APIError, MultiError).
This can only happen if it is possible to generate a MultiError with zero elements.
@manno Is that possible ?

andreas-kupries (approved):

internal/cli/clients/client.go
manno: Sure, we could construct an ErrorResponse without an error in the API server. You would have to use MultiError in an action. 

While you could create such a struct, MultiError itself needs at least one error element.
Here is the only user of MultiError: https://github.com/epinio/epinio/blob/main/internal/api/v1/servicebindings.go#L85-L117

I don't think that code can return an empty one, see the check at the end.

PR #526, opened 2021-06-22
Ingress needs an annotation for traefik not a label

### LINKED ISSUE #520: `kubernetes.io/ingress.class` should be an annotation
The ingress Epinio creates for application has a *label*

```
  kubernetes.io/ingress.class: traefik
```

This is wrong. According to the documentation ([link1](https://cloud.google.com/community/tutorials/nginx-ingress-gke#configure_ingress_resource_to_use_nginx_ingress_controller), [link2](https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation)), that should be an annotation. 

The result is that when we deploy on gke, the application ingress gets its own public IP address and not the one of Traefik. This in turn means the application domain doesn't work (since it points to the traefik IP address).

Looking at the kubernetes docs, it seems that the annotation is deprecated, but looking at the gke docs it's probable still widely supported (see the 2 links above). We need to investigate what the proper solution is.

jimmykarily:
Fixes #520 (again)

### DIFF
--- a/assets/embedded-files/tekton/pipeline.yaml
+++ b/assets/embedded-files/tekton/pipeline.yaml
@@ -225,12 +225,12 @@ spec:
           annotations:
             traefik.ingress.kubernetes.io/router.entrypoints: websecure
             traefik.ingress.kubernetes.io/router.tls: "true"
+            kubernetes.io/ingress.class: traefik
           labels:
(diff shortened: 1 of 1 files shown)


andreas-kupries (approved):
:+1:

PR #527, opened 2021-06-22
Remove workloads before everything else during uninstall

### LINKED ISSUE #495: `epinio uninstall` with minibroker service pods
Reproduce:

- Install epinio
- enable incluster services
- create a mariadb service
- [optional?] push an app and bind the service to it
- Uninstall epinio with `epinio uninstall`

The linkerd deployment fails to uninstall with the following error:

```
| Please uninject the following pods before uninstalling the control-plane:                                                                                                                                                                  │
│     * mariadb-f4600bfb65c488165388-master-0                                                                                                                                                                                                │
│     * mariadb-f4600bfb65c488165388-slave-0                                                                                                                                                                                                 │
│ No resources found                                                                                                                                                                                                                         │
│ stream closed  
```

It seems that the mariadb pods where still around when linkerd tried to uninstall. The fix in this case would be to wait until the workloads namespaces are gone before we uninstall linkerd.

thardeck:
to prevent the linkerd uninstallation from failing if services were
created.
The services are removed as part of the workload namespace.

Fixes #495

### DIFF
--- a/internal/cli/clients/install_client.go
+++ b/internal/cli/clients/install_client.go
@@ -172,6 +172,10 @@ func (c *InstallClient) Uninstall(cmd *cobra.Command) error {
 
 	c.ui.Note().Msg("Epinio uninstalling...")
 
+	if err := c.DeleteWorkloads(ctx, c.ui); err != nil {
+		return err
+	}
+
 	wg := &sync.WaitGroup{}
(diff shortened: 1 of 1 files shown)


rohitsakala (approved):

jimmykarily:
Shouldn't we create a test to avoid regressions? I mean a test that reproduces #495 .

PR #501, opened 2021-06-17
Extended management of API certs and credentials in cli config

andreas-kupries:
Fixes #468

PR #528, opened 2021-06-22
Fix bugs introduced by the `config update-credentials` work.

andreas-kupries:
Fix bugs introduced by #501.

@thardeck Please check if that fixes your issues. If not you have to crank up `TRACE_LEVEL`.
__Attention__ With trace level up secret data, i.e. credentials and certs show up in the output.

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -1468,19 +1468,6 @@ func uniqueStrings(stringSlice []string) []string {
 }
 
 func getCredentials(log logr.Logger, ctx context.Context) (string, string, error) {
-	mainDomain, err := domain.MainDomain(ctx)
-	if err != nil {
-		return "", "", err
-	}
-
-	log.Info("got main domain", "domain", mainDomain)
-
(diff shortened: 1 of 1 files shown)


internal/cli/clients/client.go
```diff
@@ -1464,21 +1467,14 @@ func uniqueStrings(stringSlice []string) []string {
 	return list
 }
 
-func getCredentials(ctx context.Context) (string, string, error) {
-	mainDomain, err := domain.MainDomain(ctx)
-	if err != nil {
-		return "", "", err
-	}
-
-	if !strings.Contains(mainDomain, "omg.howdoi.website") {
```
jimmykarily, line 1473: I wonder what the reason for this check was. Anyway, this PR looks ok, unless I'm missing something.

andreas-kupries: Best guess based on who I remember writing the #501, a copy pasta error.

jimmykarily (approved):

thardeck (approved):

PR #535, opened 2021-06-23
Retry org creation when running into gateway and cert issues

### LINKED ISSUE #530: creating the default org could fail
```🚢 Creating organization...
Name: workspace
error creating org: Post "https://epinio.demo2.mycustomdomain.org /api/v1/orgs": x509: certificate is valid for 0034784f9d3f97cfba431b8a2c25d371.7b7a66337dbad9868232b2f833a75a16.traefik.default, not epinio.demo2.mycustomdomain.org
```

If you retry to creat the org with `epinio orgs create workspace` it will work. Obviously, sometimes, the certificate of the Epinio API is not ready immediately. We need to wait for something before we try to create the default org.

andreas-kupries:
Fixes #530 .

I do not see how we can test this in the EATs.

Even manual testing is a bit fraught, for a dev deployment.
I suspect it needs testing with a public cloud and the regular deployment flow (no `make patch-...`) to be sure.

### DIFF
--- a/deployments/epinio.go
+++ b/deployments/epinio.go
@@ -161,14 +161,15 @@ func (k Epinio) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI,
 		return auth.CreateCertificate(ctx, c.RestConfig, EpinioDeploymentID, EpinioDeploymentID, domain)
 	},
 		retry.RetryIf(func(err error) bool {
-			return strings.Contains(err.Error(), "failed calling webhook") ||
+			return strings.Contains(err.Error(), " x509: ") ||
+				strings.Contains(err.Error(), "failed calling webhook") ||
 				strings.Contains(err.Error(), "EOF")
 		}),
 		retry.OnRetry(func(n uint, err error) {
-			ui.Note().Msgf("Retrying creation of API cert via cert-manager (%d/10)", n)
+			ui.Note().Msgf("Retrying creation of API cert via cert-manager (%d/%d)", n, duration.RetryMax)
 		}),
 		retry.Delay(5*time.Second),
-		retry.Attempts(10),
+		retry.Attempts(duration.RetryMax),
 	)
 	if err != nil {
 		return errors.Wrap(err, "failed trying to create the epinio API server cert")
--- a/deployments/epinio.go
+++ b/deployments/epinio.go
@@ -165,7 +165,7 @@ func (k Epinio) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI,
 				strings.Contains(err.Error(), "EOF")
 		}),
 		retry.OnRetry(func(n uint, err error) {
-			ui.Note().Msgf("retrying to create the epinio cert using cert-manager")
+			ui.Note().Msgf("Retrying creation of API cert via cert-manager (%d/10)", n)
 		}),
 		retry.Delay(5*time.Second),
 		retry.Attempts(10),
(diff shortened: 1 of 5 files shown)


andreas-kupries:
:tada: Worked locally for dev deployment where I removed the `sleep 30` between `make patch-...` and org creation.
Output:

```
[...] deployment "epinio-server" successfully rolled out

🚢 Creating organization...
Name: workspace

🚢 Retrying (0/10) after Bad Gateway: Bad Gateway

✔️  Organization created.

🚢 Targeting organization...
Name: workspace

✔️  Organization targeted.
```

So, proper retry.

deployments/epinio.go
```diff
@@ -165,7 +165,7 @@ func (k Epinio) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI,
 				strings.Contains(err.Error(), "EOF")
 		}),
 		retry.OnRetry(func(n uint, err error) {
-			ui.Note().Msgf("retrying to create the epinio cert using cert-manager")
+			ui.Note().Msgf("Retrying creation of API cert via cert-manager (%d/10)", n)
 		}),
 		retry.Delay(5*time.Second),
 		retry.Attempts(10),
```
jimmykarily, line 171: refactor opportunity: make `10` a variable or even better, a constant. You can use it in the message above too.

deployments/epinio.go
```diff
@@ -165,7 +165,7 @@ func (k Epinio) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI,
 				strings.Contains(err.Error(), "EOF")
```
jimmykarily, line 165: There is one more error we should retry on here. Since you are at it maybe you can add it (or we can create an issue for it):

```
#1: creation of production ssl certificate failed: conversion webhook for cert-manager.io/v1alpha2, Kind=Certificate failed: Post "https://cert-manager-webhook.cert-manager.svc:443/convert?timeout=30s": x509: certificate signed by unknown authority
```

It happens on a public cloud provider the first time you install Epinio. If you uninstall and retry it usually works. Must have something to do with the fact that the first time, images have to be downloaded making things slower, thus we hit this error. I don't want to hijack this PR though, I can move it to another issue.

deployments/epinio.go
andreas-kupries: Right, of course. Will be done.

deployments/epinio.go
andreas-kupries: Ah, that is the epinio deployment. Touched only because disliked the message ... I will add a `x509` pattern as I have in my retry.

andreas-kupries: Actually, might be better to replace `failed calling webhook` with the `x509`. That last should call anything certs, regardless of webhook or not.

deployments/epinio.go
andreas-kupries: Done with commits b9ad829867534bab1224e007655c7c8b61b8040b, f4ade8929dbf718b249526880f2374efbb77eb50, and c772e169b2c6779124cc7279ab13c274b76139b4.

deployments/epinio.go
andreas-kupries: Done with commits b9ad829867534bab1224e007655c7c8b61b8040b, f4ade8929dbf718b249526880f2374efbb77eb50, and c772e169b2c6779124cc7279ab13c274b76139b4.

jimmykarily (approved):

PR #539, opened 2021-06-24
fix: Drop general silencing of the usage display on cli command line errors

### LINKED ISSUE #537: Running a command without the required parameters should show `--help`
If I run an epinio command and forget that there are additional parameters required, I have to rerun the command with `--help`.

Instead if required parameters are missing `--help` should be shown instead of just an error message.

At the moment it looks like this:
```
epinio services list-plans         
accepts 1 arg(s), received 0
```

And it should look like this:
```
epinio services list-plans         
accepts 1 arg(s), received 0

Lists all plans provided by the named service class

Usage:
  epinio service list-plans CLASS [flags]

Flags:
  -h, --help   help for list-plans

Global Flags:
      --config-file string       (EPINIO_CONFIG) set path of configuration file (default "/home/thardeck/.config/epinio/config.yaml")
      --trace-level int          (TRACE_LEVEL) Only print trace messages at or above this level (0 to 5, default 0, print nothing)
```

## Acceptance criteria
* Missing required parameters result in the output of `--help` of this command
* There is still an error message shown what is missing above or below

andreas-kupries:
Fixes #537

:notes: Dynamically silence usage display for errors unrelated to the commandline, i.e. from the backend and such. In most places this simply means silencing the moment we are in the command function. In very few places this silence has to be lifted again for specific issues.

The ensemble command specs are unchanged.

### DIFF
--- a/internal/cli/push.go
+++ b/internal/cli/push.go
@@ -75,6 +75,8 @@ var CmdPush = &cobra.Command{
 	Short: "Push an application from the specified directory, or the current working directory",
 	Args:  cobra.RangeArgs(1, 2),
 	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.SilenceUsage = true
+
 		client, err := clients.NewEpinioClient(cmd.Context(), cmd.Flags())
 		if err != nil {
 			return errors.Wrap(err, "error initializing cli")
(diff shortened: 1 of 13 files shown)


internal/cli/push.go
```diff
@@ -84,27 +86,29 @@ var CmdPush = &cobra.Command{
...
 			path = args[1]
 		}
 
 		if _, err := os.Stat(path); err != nil {
+			// Path issue is user error. Show usage
+			cmd.SilenceUsage = false
 			return errors.Wrap(err, "path not accessible")
 		}
 
 		i, err := instances(cmd)
```
jimmykarily, line 101: This is also a user error. You can un-silence the usage in the `instances()` function too.

jimmykarily: or do that inside the block below (like you did above too). There is another call to `instances()` here: https://github.com/epinio/epinio/pull/539/files#diff-3bc90f8ea54bc289631b44849d75d342573625452a6fde0144e7e2d98be80725R159

if you don't un-silence inside the `instances` method then you need to handle that there explicitly too.

jimmykarily (approved):

internal/cli/push.go
andreas-kupries: Done in `instances()`, commit 0a3700f9b1bf969e7270d4930d862e73a7a146da

jimmykarily (approved):

PR #541, opened 2021-06-24
Replace sample app with 12factor app

### LINKED ISSUE #523: 12factor app doesn't work (gke, digitalocean)
First we had #521 . I tried to disable load balancing on gke (check the network settings when creating a cluster in the web ui) to see if that allows Epinio to work but after successfully pushing an application, the application route still gives a `502
Bad Gateway`. There are some errors that might be relevant in the traefik pod:

```
...
│ traefik time="2021-06-22T12:28:58Z" level=error msg="Cannot create service: subset not found" ingress=factor12 namespace=workspace providerName=kubernetes serviceName=factor12 servicePort=8080                                           │
...
│ linkerd-proxy [  1002.635758s]  WARN ThreadId(01) server{orig_dst=10.4.0.12:8080}: linkerd_app_core::errors: Failed to proxy request: error trying to connect: Connection refused (os error 111) client.addr=10.4.0.5:55472                │
```
After fixing #521 maybe no longer need to disable load balancing and things just work. We need to test this.

rohitsakala:
fixes #523


agracey:
I don't think this should be "replaces", just in addition

agracey:
More broadly, the sample apps should be in their own repos.

rohitsakala:
Got it @agracey

PR #542, opened 2021-06-24
Add owner ref to the application resource

### LINKED ISSUE #511: When removing the application resource, all owned resources are removed
Adding ownerRefs to all workload parts (deployments, secrets, ingresses, ...) so that the application resource owns them and a cascading delete removes them.

relies on #509 

## Acceptance Criteria

* ownerRefs are added
* delete app code is simplified to rely on ownerRefs

manno:
Fixes #511

Test failures disappeared with a fresh cluster. This PR modifies the tekton pipeline.
Not adding new tests, removal of ingress and service is already tested by existing tests.

## Note about OwnerRefs

* Owner references are namespaced, an owner reference on a resource can only refer to an owner in the same namespace, or a cluster resource.
* Owner references refer to a specific UID, not just app and name of a resource.

The application resource can't easily own the `PipelineRun` resource, as that's in another namespace.

We have two secrets (`registry-creds`, `ca-cert`) in the org namespace, but they are owned by the org (=namespace), not the application.

### DIFF
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -140,17 +140,18 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		}
 	}
 
+	owner := metav1.OwnerReference{
+		APIVersion: app.GetAPIVersion(),
+		Kind:       app.GetKind(),
+		Name:       app.GetName(),
+		UID:        app.GetUID(),
+	}
 	params := stageParam{
 		AppRef:    req.App,
 		Git:       req.Git,
 		Route:     req.Route,
 		Instances: instances,
-		Owner: metav1.OwnerReference{
-			APIVersion: app.GetAPIVersion(),
-			Kind:       app.GetKind(),
-			Name:       app.GetName(),
-			UID:        app.GetUID(),
-		},
+		Owner:     owner,
 	}
 
 	mainDomain, err := domain.MainDomain(ctx)
@@ -171,7 +172,7 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		return InternalError(err, fmt.Sprintf("failed to create pipeline run: %#v", o))
 	}
 
-	err = auth.CreateCertificate(ctx, cluster.RestConfig, params.Name, params.Org, mainDomain)
+	err = auth.CreateCertificate(ctx, cluster, params.Name, params.Org, mainDomain, &owner)
 	if err != nil {
 		return InternalError(err)
 	}
--- a/internal/api/v1/stage.go
+++ b/internal/api/v1/stage.go
@@ -40,6 +40,7 @@ type stageParam struct {
 	Route     string
 	Stage     models.StageRef
 	Instances int32
+	Owner     metav1.OwnerReference
 }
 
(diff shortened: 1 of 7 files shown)


internal/api/v1/stage.go
```diff
@@ -157,18 +165,18 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
...
 
-	pr := newPipelineRun(uid, app, mainDomain, registryURL, deploymentImageURL)
+	pr := newPipelineRun(uid, params, mainDomain, registryURL, deploymentImageURL)
 	o, err := client.Create(ctx, pr, metav1.CreateOptions{})
 	if err != nil {
 		return InternalError(err, fmt.Sprintf("failed to create pipeline run: %#v", o))
 	}
 
-	err = auth.CreateCertificate(ctx, cluster.RestConfig, app.Name, app.Org, mainDomain)
+	err = auth.CreateCertificate(ctx, cluster.RestConfig, params.Name, params.Org, mainDomain)
```
jimmykarily, line 174: Shouldn't this certificate have an owner too?

internal/application/workload.go
```diff
@@ -35,24 +35,6 @@ func (a *Workload) Delete(ctx context.Context, gitea GiteaInterface) error {
 		return pkgerrors.Wrap(err, "failed to delete repository")
 	}
```
jimmykarily, line 36: If a workload is a specific `Deployment`, then deleting the git repo is not part of deleting the workload. In any case, all that's left now is the repo deletion which makes the method name even less appropriate. Anyhow, leave it for the upcoming refactoring.

jimmykarily (approved):

internal/application/workload.go
manno: Yes, I'll change that with my current story. I have to change the delete code anyhow.

internal/api/v1/stage.go
manno: Oh yes!

manno:
I rebased on main and added the owner reference to the certificate, will have to investigate on Monday why this PR fails tests.

internal/auth/certs.go
```diff
@@ -60,14 +59,31 @@ func CreateCertificate(ctx context.Context, cfg *rest.Config, name, namespace, s
...
 		return errors.Wrap(err, fmt.Sprintf("creation of %s ssl certificate failed", origin))
 	}
+
+	client, err := cluster.ClientCertificate()
+	if err != nil {
+		return err
+	}
+
+	if owner != nil {
+		obj.SetOwnerReferences([]metav1.OwnerReference{*owner})
```
jimmykarily, line 73: Would it be possible to pass the owner to the `createCertificate` function and create the resource with the desired owner in one step?

manno: I wanted to separate functions that use a client to create objects from the funcs that construct the object being created. That way you keep construction and creation (?) separate. You don't need to pass too many arguments around that way and you keep the error handling, client, context (maybe logging) in one top level function.

Ah, I saw your comment on Slack. `SetOwnerReferences` doesn't reach out to the API server.

jimmykarily: Oh true, my mistake. All good.

jimmykarily (approved):

PR #543, opened 2021-06-24
feat: Report already bound services as notes, not errors

### LINKED ISSUE #533: Do not error when trying to bind a bound service on push
@jimmykarily reports that

```
🚢 Binding Services
error pushing app to server: Conflict: {"errors":[{"status":409,"title":"Service 'wordpress' already bound","details":""}]}
```

appears when pushing an app with `-b` and the service is already bound.
He proposes to convert this into a notification instead of failing.

Notes from myself
  - This can happen when pushing an update to an app
  - Agreed to make a non-failure
  - Will require extending the bind response to carry such notifications.
  - API generally can only report successes and errors.

andreas-kupries:
Fixes #533

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -342,9 +342,24 @@ func (c *EpinioClient) BindService(serviceName, appName string) error {
 		return err
 	}
 
-	_, err = c.post(api.Routes.Path("ServiceBindingCreate", c.Config.Org, appName), string(js))
+	b, err := c.post(api.Routes.Path("ServiceBindingCreate", c.Config.Org, appName), string(js))
 	if err != nil {
 		return err
+	}
+
+	br := &models.BindResponse{}
+	if err := json.Unmarshal(b, br); err != nil {
+		return err
+	}
+
+	if len(br.WasBound) > 0 {
+		c.ui.Success().
+			WithStringValue("Service", serviceName).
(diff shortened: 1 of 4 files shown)


andreas-kupries:
TODO: update the tests

internal/cli/clients/client.go
```diff
@@ -1233,12 +1248,28 @@ func (c *EpinioClient) Push(ctx context.Context, name string, source string, par
...
-		c.ui.Note().Msg("Done")
+		br := &models.BindResponse{}
+		if err := json.Unmarshal(b, br); err != nil {
+			return err
+		}
+
+		msg := c.ui.Note()
+		text := "Done"
+		if len(br.WasBound) > 0 {
+			text = text + ", With Already Bound Services"
```
thardeck, line 1264: Should we write something like `With Already Bound Service[s]` or make it dependent on the number of services, because this can be confusing, if we mention serviceS and the user wonders if there are more then one.

Just a nitpick, I am also fine with this solution.

thardeck (approved):

internal/cli/clients/client.go
andreas-kupries: Ok. I will fix the nit in some other PR, that way we do not have to through CI again.
Thanks for reviewing.

PR #554, opened 2021-06-28
500 print traefik ip address

### LINKED ISSUE #500: `epinio install-ingress` should print the IP address of Traefik
The reason we split the Traefik installation to a separate command was that we wanted to be able to setup DNS before we install epinio.

To do that, we need to know the load balancer IP address of the Traefik service. Epinio has no command to give that information to the user which means the only way for the user to get that is to use `kubectl`.
Epinio installation should not rely on the user knowing how to use kubectl to get the IP address, after all that's what Epinio does by definition, abstracts and simplifies Kubernetes operations.

We could either have a separate command `epinio ingress-ip` or just let the `epinio install-ingress` wait until and IP is assigned (or a timeout occurs) and the print it to the standard output. Being able to find that IP address at any time may be useful so the separate command may be a good idea.

jimmykarily:
Fixes #500

### DIFF
--- a/internal/cli/clients/install_client.go
+++ b/internal/cli/clients/install_client.go
@@ -2,6 +2,7 @@ package clients
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"os"
 	"strings"
@@ -153,10 +154,16 @@ func (c *InstallClient) Install(cmd *cobra.Command) error {
 
 	installationWg.Wait()
 
+	traefikServiceIngressInfo, err := c.traefikServiceIngressInfo(ctx)
+	if err != nil {
+		return err
+	}
+
 	c.ui.Success().
 		WithStringValue("System domain", domain.Value.(string)).
 		WithStringValue("API User", c.config.User).
 		WithStringValue("API Password", c.config.Password).
+		WithStringValue("Traefik Ingress info", traefikServiceIngressInfo).
 		Msg("Epinio installed.")
 
 	return nil
@@ -273,7 +280,14 @@ func (c *InstallClient) InstallIngress(cmd *cobra.Command) error {
 		return err
 	}
 
-	c.ui.Success().Msg("Epinio Ingress done.")
+	traefikServiceIngressInfo, err := c.traefikServiceIngressInfo(ctx)
+	if err != nil {
+		return err
+	}
+
+	c.ui.Success().
+		WithStringValue("Traefik Ingress info", traefikServiceIngressInfo).
+		Msg("Epinio Ingress done.")
 
 	return nil
 }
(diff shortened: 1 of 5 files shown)


internal/cli/clients/install_client.go
```diff
@@ -376,3 +390,22 @@ func (c *InstallClient) fetchIP(ctx context.Context, ip *string) error {
...
+		LabelSelector: "app.kubernetes.io/name=traefik",
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(serviceList.Items) == 0 {
+		return "not found", nil
+	}
+
+	traefikServiceIngressInfo, err := json.Marshal(serviceList.Items[0].Status.LoadBalancer.Ingress)
```
jimmykarily, line 405: In some cases, the `ingress` field may have an `ip`, [in other cases](https://stackoverflow.com/questions/35179410/how-to-wait-until-kubernetes-assigned-an-external-ip-to-a-loadbalancer-service#comment94349664_49989421) it may have `hostname`. I decided to just print the whole thing.

internal/cli/clients/install_client.go
```diff
@@ -376,3 +390,22 @@ func (c *InstallClient) fetchIP(ctx context.Context, ip *string) error {
...
+
+func (c *InstallClient) traefikServiceIngressInfo(ctx context.Context) (string, error) {
+	serviceList, err := c.kubeClient.Kubectl.CoreV1().Services("").List(ctx, metav1.ListOptions{
+		LabelSelector: "app.kubernetes.io/name=traefik",
+	})
+	if err != nil {
+		return "", err
+	}
+	if len(serviceList.Items) == 0 {
+		return "not found", nil
```
jimmykarily, line 402: I made no effort to try to detect a Traefik which is not deployed by Epinio. I check the labels of the one installed by `k3d` and thy don't match the label above. I'm not sure if it's worth trying to detect previously installed Traefik.

andreas-kupries (commented):
Works for local install. I believe should be tested with GKE also.

jimmykarily:
Works on GKE:

```
~/workspace/suse/epinio (500-print-traefik-ip-address)*$ ./dist/epinio-linux-amd64 install-ingress

🚢 Epinio installing its Ingress (Traefik)...

Configuration...
  🧭  skip-linkerd: false

🚢 Deploying Linkerd...
✔️  Linkerd deployed

🚢 Deploying Traefik Ingress..................................................
✔️  Traefik Ingress deployed

✔️  Epinio Ingress done.
Traefik Ingress info: [{"ip":"34.116.245.142"}]

```

andreas-kupries (approved):

andreas-kupries (approved):

PR #558, opened 2021-06-29
fix: attempt to retrieve stage id of an app without workload

### LINKED ISSUE #536: Scaling an app to zero prevents epinio from working
If you scale down an app with `epinio app update --instances 0` to zero, epinio is not able to fix it afterwards because it can not connect anymore.

This can be fixed manually by using k9s to increase the replicas.

It should be possible to scale to zero to basically pause an app, and it is possible when done directly in k9s.

## Acceptance criteria
* Scaling down an app to 0 is possible
* The app can be scaled up again with epinio

andreas-kupries:
Fixes #536.

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -337,6 +337,28 @@ var _ = Describe("Apps", func() {
 				return out
 			}, "1m").Should(MatchRegexp(`Status .*\|.* 1\/1`))
 		})
+
+		Describe("no instances", func() {
+			BeforeEach(func() {
+				out, err := Epinio(fmt.Sprintf("app update %s --instances 0", appName), "")
+				Expect(err).ToNot(HaveOccurred(), out)
+			})
+			It("lists apps without instances", func() {
+				Eventually(func() string {
+					out, err := Epinio("app list", "")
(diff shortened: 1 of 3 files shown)


andreas-kupries:
Ok. The CI failures look to be an actual issue in my PR.
It seems that my changes somehow disabled the `--staging` option of `app logs`.
I get the same logs with and without it in my local trial.

Investigating ... :construction: Back to `in progress` ...

andreas-kupries:
Right, the system needs the `stageId`, and is not getting the correct value, but an empty string ... And that points directly to my changes in the retrieval of same!

andreas-kupries:
The original claim that the stageID can be pulled out of the deployment is incorrect. There is no such label on the deployment.
The pod does have it, `epinio.suse.org/stage-id: e4e5b5cb34515edf` for the app I am looking at.

Huh ... Closer look ... Found it ... `spec.template.metadata.labels`. Of course. The pod spec carried by the deployment.
Ok. Lets read it from that field.

andreas-kupries:
Local CI focused on the failing test cases is now :heavy_check_mark:

rohitsakala (approved):

Issue #559, opened 2021-06-29
Extract acceptance cluster setup from test suite


jimmykarily:
We should be able to use any cluster setup we want to run the acceptance tests. Right now, the test suite is spinning up a k3d cluster and does all the setup.

We should move the setup of the cluster to a separate script (with a make target for it) and simply pass the KUBECONFIG file to the test suite. This will allow us to point tests to other clusters and test different setups.

If we also allow passing the `system-domain` to be used, this will even allow us to test using real DNS (not the `omg.howdoi.website` reflector). This should provide more stable tests too.

PR #560, opened 2021-06-29
Configurable magic domain

manno:
Make the magic domain configurable.

Hopefully helps to fix current CI problems (#559).

### DIFF
--- a/acceptance/acceptance_suite_test.go
+++ b/acceptance/acceptance_suite_test.go
@@ -28,17 +28,21 @@ func TestAcceptance(t *testing.T) {
 	RunSpecs(t, "Acceptance Suite")
 }
 
-var nodeSuffix, nodeTmpDir string
+var (
+	nodeSuffix, nodeTmpDir string
 
-// serverURL is the URL of the epinio API server
-var serverURL, websocketURL string
-var registryMirrorName = "epinio-acceptance-registry-mirror"
+	// serverURL is the URL of the epinio API server
+	serverURL, websocketURL string
+	registryMirrorName      = "epinio-acceptance-registry-mirror"
+	epinioMagicDomain       = "omg.howdoi.website"
+)
 
 const (
-	networkName         = "epinio-acceptance"
-	registryMirrorEnv   = "EPINIO_REGISTRY_CONFIG"
-	registryUsernameEnv = "REGISTRY_USERNAME"
-	registryPasswordEnv = "REGISTRY_PASSWORD"
+	networkName          = "epinio-acceptance"
+	registryMirrorEnv    = "EPINIO_REGISTRY_CONFIG"
+	registryUsernameEnv  = "REGISTRY_USERNAME"
+	registryPasswordEnv  = "REGISTRY_PASSWORD"
+	epinioMagicDomainEnv = "EPINIO_MAGIC_DOMAIN"
 
 	// skipCleanupPath is the path (relative to the test
 	// directory) of a file which, when present causes the system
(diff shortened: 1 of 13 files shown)


jimmykarily:
This is useful if someone wants to switch from `omg.howdoi.website` to `nip.io` or other. We still use the magic domain as a switch to decide whether to:
- Generate let's encrypt certs or not
- Create a node port service or not for the registry

For the first one, we do that because we assume a domain like 'omg.howdoi.website" is not publicly accessible thus let's encrypt won't be able to access our http challenge url. This is not necessarily true, since the IP address to which the magic domain will resolve can perfectly be a public IP address. So the way we generate certificates should be based on whether the cluster is publicly accessible and not on the domain used. When we have more options (e.g. custom signers, use your own cert etc) the decision on what certificate to create should be a flag provided by the user.

For the second one, I think the logic is the same. We can't use TLS for the registry when the registry is not publicly accessible because of the http challenge. Then we just to the assumption that `omg` domain means the cluster is not publicly accessible so we setup the registry differently.

I'm not sure if let's encrypt would refuse to generate a cert for a magic dns in general but in any case, the conditions on which we decide things in code should be directly related to the thing being decided. If there is an indirect conditions (e.g. assume access to the cluster based on the domain), this logic should be in one place.

This PR is fine as a first step. After all we can switch to another magic domain if `omg` misbehaves.

manno:
Ok, I'll try to fix it up and get it through CI.

I think in general wildcard domains with public IPs are fine with Letsencrypt, but they might require some special handling?
The [sslip.io server](https://github.com/cunnie/sslip.io/blob/master/bosh-release/src/sslip.io-dns-server/xip/xip.go#L386) has extra code for that, while xip and nip don't.

acceptance/acceptance_suite_test.go
```diff
@@ -28,17 +28,21 @@ func TestAcceptance(t *testing.T) {
...
+	epinioMagicDomain       = "omg.howdoi.website"
+)
 
 const (
-	networkName         = "epinio-acceptance"
-	registryMirrorEnv   = "EPINIO_REGISTRY_CONFIG"
-	registryUsernameEnv = "REGISTRY_USERNAME"
-	registryPasswordEnv = "REGISTRY_PASSWORD"
+	networkName          = "epinio-acceptance"
+	registryMirrorEnv    = "EPINIO_REGISTRY_CONFIG"
```
jimmykarily, line 42:
```suggestion
```

jimmykarily: Not needed after: https://github.com/epinio/epinio/commit/edc14ab

internal/cli/clients/install_client.go
```diff
@@ -366,6 +366,7 @@ func (c *InstallClient) fillInMissingSystemDomain(ctx context.Context, domain *k
 
 		if ip != "" {
 			domain.Value = fmt.Sprintf("%s.omg.howdoi.website", ip)
```
jimmykarily, line 368:
```suggestion

```

internal/cli/root.go
```diff
@@ -59,6 +59,10 @@ func Execute() {
 	viper.BindPFlag("no-colors", pf.Lookup("no-colors"))
 	argToEnv["colors"] = "EPINIO_COLORS"
 
+	pf.StringP("magic-domain", "", "omg.howdoi.website", "Magic domain to use self-signed certs with")
+	viper.BindPFlag("magic-domain", pf.Lookup("magic-domain"))
+	argToEnv["magic-domain"] = "EPINIO_MAGIC_DOMAIN"
```
jimmykarily, line 64: This makes `--magic-domain` a global flag. If you installed Epinio with `--magic-domain=nip.io` and you started the epinio server with no value (in which case the default `omg.howdoi.website` will be used, this line will have a problem: https://github.com/epinio/epinio/pull/560/files#diff-3ea7fa44784ec45a11489f5d0b1d94d2ce86ea59f9e971eb44944eed3be1b080R161

jimmykarily: A fix would be to not allow the user to specify the magic domain except only during installation. This value will have to be stored somewhere in order to be used in post-installation code.

This becomes a bit complicated and I start to think that we should better spend time working on stories that solve the real issue (https://github.com/epinio/epinio/pull/560#issuecomment-871121956). No domain should be special in the code and no decisions should be made based on the domain. Our defaults just have work together (e.g. default domain when one is not set: `1.2.3.4.omg.howdoi.website`, default issuer when one is not defined: `CA issuer`).

manno: Yes, I agree, there should be no special domain. Instead we should allow the user to configure features.

It was interesting to see how far omg.whowi.website spread across the code.
CI and DNS is working again, we no longer need this workaround, so I'd close this issue.

Issue #63, opened 2021-01-21
Bring your own Cert for Applications


jimmykarily:
Depends on: https://github.com/SUSE/carrier/issues/62

We should implement the automated TLS setup for applications but we should also allow the usage of existing certificates. Some users may already be paying for their certificates and they may want to use them.

After we fix #62 we should find a way to allow users to use existing certificates instead of letting Carrier create new ones for them.

jimmykarily:
@agracey suggested that we could implement "bring your own cert" as "bring your own issuer". In other words, allow the users to specify another issues (their own issuer) that serves their own certificates for various domains (at some point we will support multiple domains). If the domain in use is not known to the user's issuer it could fallback to let's encrypt.

We can have a spike to investigate if that is possible.

agracey:
Is this done with https://github.com/epinio/epinio/issues/564? Or is it separate?

jimmykarily:
No this is separate. This is mentioned briefly in #384 but maybe deserves another card (or is this one that card?). What we want it the user to be able to use any value for the `--tls-issuer` flag of the install command. This way, they can create an issuer with a custom CA they own. Optionally, epinio will provide a command to help with that.

jimmykarily:
Specifying an arbitrary string as the issuer name is already supported (@andreas-kupries ?). What's left is:

- Verify that arbitrary issuer names are supported
- Document how to create an issuer with an existing ca cert
- When installing epinio, validate that the given issuer exists before trying to install. If for example the user makes a typo, this shouldn't result in a broken installation (should just print an error and exit before anything is installed).
- [optional] auto completion for existing issuers on the `epinio install` command.

manno:
Epinio now supports setting a custom cert-manager issuers when installing. 

https://github.com/epinio/epinio/blob/d9562397e69a4c0f47351740aff0b0154ba60df6/docs/user/howtos/certificate_issuers.md


We can open another issue, if we want to revisit custom certificates (not created by a CA/letsencrypt, but e.g. bought online) for  pushed apps.

PR #562, opened 2021-06-30
559 extract acceptance cluster setup

### LINKED ISSUE #559: Extract acceptance cluster setup from test suite

jimmykarily:
Fixes #559 

This should allow us to use other types of clusters (other than k3d) by simply setting the KUBECONFIG environment variable.

Some of the changes also allow us to use real DNS (not the magic one) like this in the future:

```bash
make acceptance-cluster-delete # Delete the old cluster
EXPOSE_ACCEPTANCE_CLUSTER_PORTS=1 make acceptance-cluster-setup # Create a new one and bind ports 80 and 443 on the host
export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig # Point to the new cluster
EPINIO_SYSTEM_DOMAIN=some.real.domain.org make test-acceptance # Use a domain that points to the runner VM's IP address
```

We can't predict the IP address of the k3d container but the IP address of the VM doesn't change so we can use a real domain pointing to that IP address (even if it's not publicly accessible). This way we avoid problems related to `omg.howdoi.website` DNS server. Each worker would have to infer the `EPINIO_SYSTEM_DOMAIN` to be used base on it's hostname or something like that (each runner would have a different system domain).

The only reason this can't work yet is that whenever we see a system-domain that is not `omg.howdoi.website`, we assume it's a production deployment and we try to create let's encrypt certificates with HTTP challenges. Since our runners are not publicly accessible, http challenges are not possible. There are stories in the backlog that could solve this (e.g. #63) but another way would be to implement a flag that explicitly enables the let's encrypt signer during installation (or after?) and avoid deducting it from the system-domain. See also comments here: https://github.com/epinio/epinio/pull/560#issuecomment-871121956

For now, with this PR we can at least try tests on different types of clusters.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -53,4 +53,8 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
         run: |
-          GINKGO_NODES=8 make test-acceptance
+          export GINKGO_NODES=8
+          make acceptance-cluster-delete
+          make acceptance-cluster-setup
+          export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
+          make test-acceptance
--- a/.github/workflows/outside-pr.yml
+++ b/.github/workflows/outside-pr.yml
@@ -42,4 +42,8 @@ jobs:
           REGISTRY_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
           REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
         run: |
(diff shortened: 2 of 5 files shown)


jimmykarily:
CI is busted (on DNS issues). Tests pass locally.

jimmykarily:
Create this story to be able to use real DNS in CI: https://github.com/epinio/epinio/issues/564 (among other things)

manno (approved):

PR #563, opened 2021-06-30
Split test suite

manno:
So we can run on  multiple workers.

This splits API tests and extracts common functionality into packages.

Should help with #559


jimmykarily (dismissed):

andreas-kupries (approved):
lgtm

Issue #377, opened 2021-05-11
Epinio documentation


jimmykarily:
Before we call Epinio [SLC](https://medium.com/@nimay/shifting-prototyping-mindsets-from-mvp-to-slc-eab5bc42b6b2) we need documentation that is:
- up-to-date
- usable (searchable?)
- just enough (not necessarily diving into the internals of Epinio but enough for our users to use all Epinio features).

We can have documentation in any format that satisfies the above but a separate site might be the only option for "usable" if we want to support searching.

kkaempf:
Focus on correctness and completeness for the initial open-source release. Leave extras (like searching) for the 'beta' release.

jimmykarily:
That makes README.md and `/docs/*.md` files a viable solution.

jimmykarily:
Note: document this: https://github.com/epinio/epinio/issues/380

rohitsakala:
Action Items

- [ ]  Update README.md
- [ ] Update /docs/*.md

PR #565, opened 2021-06-30
Deploy from GIT url

### LINKED ISSUE #514: Deploy from Git URL
We can create an app with a Git repo (creds maybe) and push it without uploading a tarball first.
The Git info is stored in or referenced from the application resource.

might rely on #513

andreas-kupries:
Fixes #514

### DIFF
--- a/internal/cli/push.go
+++ b/internal/cli/push.go
@@ -14,42 +14,49 @@ import (
 var ()
 
 func init() {
+	CmdPush.Flags().Int32P("instances", "i", v1.DefaultInstances,
+		"The number of desired instances for the application, default only applies to new deployments")
+	CmdPush.Flags().String("git", "", "git revision of sources. PATH becomes repository location")
 	CmdPush.Flags().StringSliceP("bind", "b", []string{}, "services to bind immediately")
-	CmdPush.RegisterFlagCompletionFunc("bind", func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
-		// `cmd`, `args` are ignored.
-		// `toComplete` is the option value entered so far.
-		// This is a StringSlice option.
-		// This means that the option value is a comma-separated string of values.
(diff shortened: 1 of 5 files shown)


internal/cli/push.go
```diff
@@ -83,8 +90,18 @@ var CmdPush = &cobra.Command{
...
 
+		gitRevision, err := cmd.Flags().GetString("git")
+		if err != nil {
+			return errors.Wrap(err, "could not read option --git")
+		}
+
 		var path string
 		if len(args) == 1 {
+			if gitRevision != "" {
+				cmd.SilenceUsage = false
```
jimmykarily, line 101: I think you need to unsilence in this case too:

```
./dist/epinio-linux-amd64 push --git main https://github.com/epinio/example-12factor.git
```
This is missing the app name but it prints no error, just exits.

andreas-kupries: The code here should be unsilencing things. Double Negation: silence = false => do not silence = unsilence.
Will investigate.

andreas-kupries: Hm. removing the statement makes no difference. No output ... disabling the `silence = true` at the top ... no difference ... I am beginning to hate cobra error handling.

andreas-kupries: :bulb: I am using errors.Wrap, except, the `err` is likely nil :scream:

jimmykarily:
Can we add docs for this? Or should we do that here: https://github.com/epinio/epinio/issues/377 ? I don't want us to start delegating documentation to that story but given it's already in progress, maybe we will get unnecessary conflicts or even loose docs on rebases and such. I'm ok if you just add a commit to the wip branch of #377 to document this.

internal/cli/push.go
andreas-kupries: Fixed with commit b526fc85e7baa8906a9444a9b828f1d36ff67625

andreas-kupries:
I will do the doc in #377

jimmykarily (approved):

PR #573, opened 2021-07-02
Add release-drafter to generate release notes

manno:


manno:
release-drafter only supports PRs: https://github.com/release-drafter/release-drafter/issues/501

PR #580, opened 2021-07-05
CI: Concurrent jobs to run the acceptance test api, cli suites

### LINKED ISSUE #577: Spike: Run split test suite in parallel on multiple workers
(Converted from icebox entry)

Can the testsuite runtime be improved by running tests in parallel on multiple workers ?

andreas-kupries:
Fixes #577

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,6 +16,32 @@ on:
     - cron:  '0 0 * * *'
 
 jobs:
+  linter:
+    runs-on: self-hosted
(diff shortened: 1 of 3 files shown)


manno:
I was wondering if we can move the unit tests into the linter job, but if I understand correctly having the unit tests with the acceptance tests requires less setup?

andreas-kupries:
Yes. If the unit tests go into linter it will need all the build setup also.
It should not matter in terms of time needed for execution.
Only that we would have the same setup three times.

manno (approved):

Issue #508, opened 2021-06-21
CLI (except install,update creds) doesn't need a kubernetes connection


manno:
Thinking of Epinio as differenct components, we would speak of:

* installer
* cli
* web server
* api server

The cli parts that deal with applications, shouldn't need a separate kubernetes connection (`KUBECONFIG`).

## Acceptance

* the remaining code parts that use kubectl use the api server instead
* implements Match* as API endpoint (some could be part of `Index`)

jimmykarily:
Superseded by #620 (was a duplicate but kept the newer one because it had a more detailed description).

PR #587, opened 2021-07-07
Support App environment Variables

### LINKED ISSUE #375: Set environment variables
Depends on #347 

Various applications may need to be configured through environment variables. For example, a rails application needs to have the `RAILS_MASTER_KEY` environment variable set to a specific secret (https://edgeguides.rubyonrails.org/security.html#custom-credentials).

## Acceptance Criteria

The user can specify env vars for a created app (i.e. when creating the app). See https://github.com/epinio/epinio/issues/375#issuecomment-868301329

When pushing the app, staging will create a deployment. The deployment will mount the secret as env vars.

mm: I'd suggest to postpone `--staging` env vars for a separate follow-up story.

## Previous Notes

Also buildpacks are moving away from `buildpack.yaml` configuration towards env vars (https://github.com/paketo-buildpacks/nginx/blob/main/rfcs/0001-buildpack-yml-to-env-var.md).

We need a way to allow the user to set environment variables that will be available to the application at run time.

We should also investigate if the same solution would work for staging with buildpacks (see the link above).

## Previous Questions

- Where are the variables stored? Is it on the deployment itself?
- What happens when we push the application again?
- How does the user delete a variable?
- Can the solution look similar to the concept of services? E.g. each environment variable being a secret behind the scenes that can be mounted to any application, kinda like a service, so it has its own lifecycle.

### LINKED ISSUE #510: Listing apps, should show application resources instead of deployments
Now listing apps, we see all the deployments. With the introduction of an application resource, that might exist before the deployment (workload), exists, we want the user to see it in the CLI output.

**Note**: The same might be true for other commands, like `app show`. We might need to create follow up stories.

relies on #509 

## Acceptance Criteria

* epinio show app includes apps without deployments

andreas-kupries:
Fix #375 
Fix #510

### Todo

  - More tests

### DIFF
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -131,12 +131,26 @@ func (hc ApplicationsController) Show(w http.ResponseWriter, r *http.Request) AP
 		return OrgIsNotKnown(org)
 	}
 
+	exists, err = application.Exists(ctx, cluster, models.NewAppRef(appName, org))
+	if err != nil {
+		return InternalError(err)
+	}
+
+	if !exists {
+		return AppIsNotKnown(appName)
+	}
+
+	// Application exists. It may not have a workload however.
+
 	app, err := application.Lookup(ctx, cluster, org, appName)
 	if err != nil {
 		return InternalError(err)
 	}
 	if app == nil {
-		return AppIsNotKnown(appName)
+		// While the app exists, it has no workload.
+		// Return something barebones.
+		app = models.NewApp(appName, org)
+		app.Status = `Inactive, without workload. Launch via "epinio app push"`
 	}
 
 	js, err := json.Marshal(app)
@@ -173,13 +187,26 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request) 
 		return OrgIsNotKnown(org)
 	}
 
+	exists, err = application.Exists(ctx, cluster, models.NewAppRef(appName, org))
+	if err != nil {
+		return InternalError(err)
+	}
+
+	if !exists {
+		return AppIsNotKnown(appName)
+	}
+
+	// Application exists. It may not have a workload however.
+
 	app, err := application.Lookup(ctx, cluster, org, appName)
 	if err != nil {
 		return InternalError(err)
 	}
 
 	if app == nil {
-		return AppIsNotKnown(appName)
+		// App without workload cannot be scaled at the moment.
+		// TODO: Extend to stash the request in the app or attached resource
+		return NewAPIError("Unable to scale application without workload", "", http.StatusBadRequest)
 	}
 
 	defer r.Body.Close()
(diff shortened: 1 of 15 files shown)


thardeck (dismissed):

internal/api/v1/applications.go
```diff
@@ -222,24 +249,41 @@ func (hc ApplicationsController) Logs(w http.ResponseWriter, r *http.Request) {
...
 		if app == nil {
-			jsonErrorResponse(w, AppIsNotKnown(appName))
+			// While app exists it has no workload
+			jsonErrorResponse(w, NewAPIError("No logs available for application without workload", "", http.StatusBadRequest))
+			return
 		}
 	}
 
 	if appName == "" && stageID == "" {
 		jsonErrorResponse(w, BadRequest(errors.New("You need to speficy either the stage id or the app")))
```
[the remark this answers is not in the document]
andreas-kupries: will do

andreas-kupries, line 285:
```suggestion
		jsonErrorResponse(w, BadRequest(errors.New("You need to specify either the stage id or the app")))
```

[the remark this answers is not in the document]
andreas-kupries: And done, via github suggestion

thardeck (dismissed):

internal/cli/env.go
```diff
@@ -0,0 +1,188 @@
...
+		}
+
+		return nil
+	},
+	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
+		if len(args) != 0 {
+			return nil, cobra.ShellCompDirectiveNoFileComp
+		}
+
+		// TODO: arg 1: EV match
```
manno, line 177: I don't understand this TODO

internal/cli/clients/client.go
```diff
@@ -177,6 +319,10 @@ func (c *EpinioClient) ServicePlanMatching(ctx context.Context, serviceClassName
 
 	result := []string{}
 
+	// TODO Create and use server endpoints. Maybe use existing
```
manno, line 322: Ah nice, these TODOs are for
https://github.com/epinio/epinio/issues/508

internal/cli/clients/client.go
```diff
@@ -90,6 +90,148 @@ func NewEpinioClient(ctx context.Context, flags *pflag.FlagSet) (*EpinioClient,
 	return epinioClient, nil
 }
 
+// EnvList displays a table of all environment variables and their
+// values for the named application.
+func (c *EpinioClient) EnvList(ctx context.Context, appName string) error {
+	log := c.Log.WithName("EnvList")
+	log.Info("start")
+	defer log.Info("return")
+	// details := log.V(1) // NOTE: Increment of level, not absolute.
```
manno, line 99:
```suggestion
```

internal/application/workload.go
```diff
@@ -72,6 +134,10 @@ func (a *Workload) Scale(ctx context.Context, instances int32) error {
 
 // UnbindAll dissolves all bindings from the application.
 func (a *Workload) UnbindAll(ctx context.Context, cluster *kubernetes.Cluster, svcs []string) error {
+
+	// TODO: Optimize this action to perform a single restart for
```
manno, line 138: Maybe add an issue for this TODO and link it here. Though, 'optimize' issues don't make it in the backlog often...

internal/cli/env.go
andreas-kupries: The command line completion should should complete the second argument (index 1) based on the existing environment variables of the app.

internal/cli/clients/client.go
andreas-kupries: Yes

internal/application/workload.go
andreas-kupries: Done, #598

PR #590, opened 2021-07-07
Replace omg.howdoi.website detection and use provided issuer instead

manno:
* add arg `--tls-issuer` to specify the name of the cert manager issuer to use
* add args to CLI and server to toggle the creation and use of the node
  port workaround for the internal registry
  (`--enable-internal-registry-node-port`,
  `--use-internal-registry-node-port`)
* remove decision on tekton trust, always add certificate to tekton and
  add TODO to remove workaround once we can use the mesh
* epinio install will use current TRACE_LEVEL for server

No tests were added since we don't have tests for 'production'/letsencrypt installations. Really, we have no tests for installation at all.

fixes #590


manno:
no longer green after rebase on main

internal/application/application.go
```diff
@@ -236,6 +236,11 @@ func Delete(ctx context.Context, cluster *kubernetes.Cluster, gitea GiteaInterfa
 		return err
 	}
 
+	err = cluster.Kubectl.CoreV1().Secrets(appRef.Org).Delete(ctx, fmt.Sprintf("%s-tls", appRef.Name), metav1.DeleteOptions{})
```
jimmykarily, line 239: Maybe we should do this instead? https://cert-manager.io/docs/usage/certificate/#cleaning-up-secrets-when-certificates-are-deleted

internal/cli/server.go
```diff
@@ -31,6 +32,14 @@ func init() {
 	flags.Int("port", 0, "(PORT) The port to listen on. Leave empty to auto-assign a random port")
 	viper.BindPFlag("port", flags.Lookup("port"))
 	viper.BindEnv("port", "PORT")
+
+	flags.String("tls-issuer", deployments.SelfSignedIssuer, "(TLS_ISSUER) The cluster issuer to use for workload certificates")
+	viper.BindPFlag("tls-issuer", flags.Lookup("tls-issuer"))
+	viper.BindEnv("tls-issuer", "TLS_ISSUER")
+
+	flags.Bool("use-internal-registry-node-port", true, "(USE_INTERNAL_REGISTRY_NODE_PORT) Use the internal registry via a node port")
```
jimmykarily, line 40: There is no point int setting this to true if `enable-internal-registry-node-port` was set to `false` at installation time right?
It seems to me, the only reason we need this flag is because we have no place to store the value of the `enable-internal-registry-node-port` flag that was used. Is there a way to detect what value was used? For example, check the registry service to see what type it is?

jimmykarily:
I verified that I can use another domain and still use self-signed certificates with:

```
/dist/epinio-linux-amd64 install --skip-default-org --verbosity=1 --system-domain 10.86.4.38.nip.io
```

jimmykarily:
Since there is no auto-detection of "production" setup anymore, we need to document the recommended way to deploy in production. That should prominent in the README.md (even if it's just a link to another page).

internal/cli/server.go
jimmykarily: How does the user define this for the server? Epinio sets up the server command automatically inside the server deployment, the user doesn't have a way to provide flags. The install command is the one that creates the server command and that one has access to the `--enable-internal-registry-node-port` flag. Maybe we should use that to define the `--use-internal-registry-node-port`?

jimmykarily:
I face this issue:

1. I create a gke cluster and get a kubeconfig to talk to it
2. I run this command to setup ingress: `./dist/epinio-linux-amd64 install-ingress`
3. I run this command to install Epinio:
```
TRACE_LEVEL=100 ./dist/epinio-linux-amd64 install --verbosity=1  --enable-internal-registry-node-port=false --tls-issuer letsencrypt-production  --system-domain pr-590.karakasilis.me
```

I get this error:

```

🕞  Starting app.kubernetes.io/name=epinio-server in epinio .

🕞    Starting pod epinio-server-5cb687c87f-ktqw7 in epinio ...........................
✔️  Registry deployed
2021/07/12 16:32:13 InstallClient/Install "level"=2 "msg"="return"  
...

✔️  Gitea deployed
2021/07/12 16:32:14 InstallClient/Install "level"=2 "msg"="return"  
....................................................................................................................................................

🕞   Applying tekton staging resources .
2021/07/12 16:33:28 InstallClient/Install "level"=2 "msg"="return"  
panic: Deployment tekton failed with error: Applying tekton staging resources failed:
: Failed to get registry CA from tekton-staging namespace: failed to decode certificate
failed find PEM data


goroutine 29 [running]:
github.com/epinio/epinio/internal/cli/clients.(*InstallClient).Install.func1(0xc0007ef620, 0x21012e0, 0xc0000c0000, 0x2117f40, 0xc000276820, 0x21235e0, 0xc0000d4400, 0xc000040e74)
	/home/dimitris/workspace/suse/epinio/internal/cli/clients/install_client.go:148 +0x263
created by github.com/epinio/epinio/internal/cli/clients.(*InstallClient).Install
	/home/dimitris/workspace/suse/epinio/internal/cli/clients/install_client.go:145 +0xe1b
```

internal/application/application.go
manno: I agree, but this will be done for all the cert generated by cert manager. Should be ok for us.

I'll see if it works , otherwise it'll have to go in a separate PR.

internal/cli/server.go
manno: Yes, that was the plan. Let me check if I forgot to do this...

manno: Ah here it is, some awful indirection, but all variable names match the pattern (snake case, kebap case, ...)

https://github.com/epinio/epinio/pull/590/files#diff-892d335f47fd05c8157a81b96c028b3d6c1a88492ca015135d24fe86a1ede046R308-R309

So the client puts this as an env var into the epinio deployment yaml via our ##substitution##. (I feel we should invest in real templates soon;)

I thought 'use' was more approriate for the server flag, but maybe it should be the same name as the client flag 'enable' ?

internal/application/application.go
manno: I think it worked.

internal/cli/server.go
jimmykarily: maybe `use` should be used in all cases? I mean, they use it in a different way but they both use it somehow. Otoh, I'm still thinking we could get rid of the server flag but inspecting the registry service. It's only used in one place (the staging api) in order to decide what the registry URL should look like.

jimmykarily (approved):

PR #602, opened 2021-07-09
Try to push docker img directl

### LINKED ISSUE #574: Spike: Stage an existing Docker image
We can speed up CI, by removing the buildpack tekton job from a large number of our tests. They would just stage an existing Docker image.

In the future, users can stage Docker images via `epinio push`.

See #485 also

manno:
not to be commited, not clean and still not working due to wrong yaml 

Fixes #574


rohitsakala:
Closing in favour of #629

PR #605, opened 2021-07-12
Fix minikube installation script

### LINKED ISSUE #599: Remove the broken minikube start instruction from the docs

rohitsakala:
Fixes #599 

* Remove private minikube iso image
* Add minikube tool script for automatic installation
* Add working bash variables

### DIFF
--- a/scripts/minikube-start.sh
+++ b/scripts/minikube-start.sh
@@ -6,9 +6,8 @@ require_tools minikube
 : "${VM_CPUS:=4}"
 : "${VM_MEMORY:=16384}"
 : "${VM_DISK_SIZE:=120g}"
-
-: "${MINIKUBE_ISO_VERSION:=0.1.8}"
-: "${MINIKUBE_ISO_URL:=https://github.com/f0rmiga/opensuse-minikube-image/releases/download/v${MINIKUBE_ISO_VERSION}/minikube-openSUSE.x86_64-${MINIKUBE_ISO_VERSION}.iso}"
+: "${VM_DRIVER:=virtualbox}"
+: "${K8S_VERSION:=1.19.9}"
 
 if ! minikube status > /dev/null; then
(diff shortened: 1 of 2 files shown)


scripts/minikube-start.sh
```diff
@@ -6,9 +6,8 @@ require_tools minikube
 : "${VM_CPUS:=4}"
 : "${VM_MEMORY:=16384}"
 : "${VM_DISK_SIZE:=120g}"
-
-: "${MINIKUBE_ISO_VERSION:=0.1.8}"
-: "${MINIKUBE_ISO_URL:=https://github.com/f0rmiga/opensuse-minikube-image/releases/download/v${MINIKUBE_ISO_VERSION}/minikube-openSUSE.x86_64-${MINIKUBE_ISO_VERSION}.iso}"
+: "${VM_DRIVER:=virtualbox}"
```
thardeck, line 9: Why does Minikube need a vm environment and what happens if Virtualbox is not installed?

thardeck (changes requested):

scripts/minikube-start.sh
rohitsakala: These are the default configuration on minikube which works. Users should install something like Virtualbox or docker or hyperkit whichever they prefer.

thardeck: Minikube recommends Docker on there site as preferred option (https://minikube.sigs.k8s.io/docs/drivers/), but if this is a special use case, I do not mind.

thardeck (approved):

scripts/minikube-start.sh
kkaempf: I use `minikube start --docker-opt storage-driver=overlay2` to run minikube on Docker. The `docker-opt` addresses a performance warning you get during startup.

PR #613, opened 2021-07-14
Create private epinio CA

andreas-kupries:
fix #603 

Merge should wait for #590 to be merged, and then rebased.

PR #614, opened 2021-07-14
Update docs for tls-issuer arg

### LINKED ISSUE #564: Use a CA issuer instead of self-signed certificates
Creating a CA issuer is very simple: https://cert-manager.io/docs/configuration/selfsigned/#bootstrapping-ca-issuers

and it has the benefit that the user's need only trust the CA once and it is going to work for all the generated certificates. With self-signed certs, one needs to create an exception for each and every cert generated by Epinio (e.g. the different apps).

See also the discussion here: https://github.com/epinio/epinio/issues/476#issuecomment-862468253

Right now we rely on the system-domain in order to decide whether to use let's encrypt certificates or not. This is wrong (see discussion here: https://github.com/epinio/epinio/pull/560#issuecomment-871121956). Ideally the user should be able to change to another issuer even after installing Epinio but this is complicated. For now, it is ok to let the user specify the type of issuer they want to use at installation time. E.g.

```bash
epinio install --tls-private-ca # the default
```

or

```
epinio install --tls-lets-encrypt
```

Being able to use a private CA no matter what the domain is not only more flexible and better UX but it also unblocks us here: https://github.com/epinio/epinio/pull/562#issue-680740923. In other words, a user may want to have a custom domain and still not use let's encrypt for certs.

This story could simply replace self-signed certs with a CA signer or it could also implement the explicit flags above and decouple the code from the `omg.howdoi.website` domain. Let's decide if we want to split in 2 stories or not.

## Acceptance Criteria

The epinio installer CLI accepts a new argument: `--tls-issuer=string`
For now we support `letsencrypt-production` and `selfsigned-issuer`.

The string is a reference to the cert-manager clusterissuer name. A few places in the code will behave differently depending on the type.

* For `letsencrypt-production` the installer sets up Epinio to use letsencrypt. This is the previous 'production' setup.
* For `selfsigned-issuer` the installer sets up Epinio to generate a private CA. This is old the `omg.howdoi.website` setup.

We keep the old defaults, `systemdomain=omg.howdoi.website`
 and `tls=selfsigned-issuer`.

To use the letsencrypt issuer, you must use a public IP and set up DNS for the system domain.
 
There should be a separate flag to control the node port creation for the internal registry (https://github.com/epinio/epinio/blob/main/assets/container-registry/chart/container-registry/templates/container-registry.yaml#L56-L82), something like `--use-node-port-workaround-for-internal-registry`

### Notes

* https://github.com/epinio/epinio/pull/560/files lists all the switches.
* this also needs to add the new args to epinio server deployment.
* we still have the cluster signed certificate from quarks secret for the internal registry.
* with this issue solved, specifying a custom issuer is possible. A follow-up story to make the letsencrypt issuer configurable, so that one can switch to the `dnschallenge`, is no longer needed.
* another follow-up could add support for `external-dns`, to automate the DNS zone management

manno:
Fixes #564


docs/user/explanations/configuration.md
```diff
@@ -28,17 +28,19 @@ Epinio's configuration contains
...
 User name and password are used by the client to authenticate itself
 when talking to Epinio's API server. The `epinio install` command
 saves the initial information to the configuration.
 
-The certificate information is set only when installing epinio for
-development. This uses an `omg.howdoi.website` domain with a
-self-signed certificate. `epinio install` saves the associated CA
+The certificate information is set only when installing epinio with
+`--tls-issuer=selfsigned-issuer`.  The installation uses a the wildcard domain
+`omg.howdoi.website` and the self-signed issuer by default.
```
andreas-kupries, line 41: Hasn't this changed with #590 to always save the cert ?
Also `a the wild...`.

docs/user/howtos/certificate_issuers.md
```diff
@@ -0,0 +1,64 @@
+# Using a Certificate Issuer
+
+Epinio comes with two [cert-manager issuers](https://cert-manager.io/docs/configuration/) for creating certificates:
+
+* letsencrypt-production
+* selfsigned-issuer
```
andreas-kupries, line 6: Plus the new `epinio-ca` from #613, also the new default

docs/user/tutorials/install_epinio_on_gke.md
```diff
@@ -41,9 +41,9 @@ The output of the command will print the `loadbalancer-ip`.
 #### Install Epinio in cluster
```
andreas-kupries, line 41:
```suggestion
#### Install Epinio In Cluster
```

Like most other changes to titles ?

andreas-kupries (changes requested):
Just small stuff

docs/user/explanations/configuration.md
manno: I might not have understood that correctly. I think we always save the cert to the config, but this is about being able to specify some field on the cert?

I might need to look that up in code.

docs/user/tutorials/install_epinio_on_gke.md
manno: Yes, I thought we could just keep level 4 headlines as sentences, because the font will have a similar size ;)
Better to be consistent though.

docs/user/explanations/configuration.md
manno: Yes, I misunderstood and saving is explained in the next paragraph

andreas-kupries (approved):

PR #619, opened 2021-07-15
Allow mounting of the SA for linkerd sidecars in app pods

### LINKED ISSUE #532: Workloads should be meshed
Reproduce:

- Install epinio

Although the namespace is annotated for linkerd, the application pod doesn't get a sidecar. This is because we create the application with `automountServiceAccountToken: false` (see here: https://github.com/linkerd/linkerd2/issues/3183). We do this because we don't want the code inside the application pods to be able to talk to kubernetes (which is what the servicetoken is allowing). Linkerd on the other side needs this.

We need to find a way to have a sidecar without allowing application code to do nasty things on the cluster.

andreas-kupries:
Fix #532 

Note that this expects the SA to be suitably limited so that the application cannot talk to kubernetes.
An __unmodified__ default SA for the namespace is good for this.

### DIFF
--- a/assets/embedded-files/tekton/pipeline.yaml
+++ b/assets/embedded-files/tekton/pipeline.yaml
@@ -209,7 +209,7 @@ spec:
                 app.kubernetes.io/name: "$(params.APP_NAME)"
             spec:
               serviceAccountName: "$(params.ORG)"
-              automountServiceAccountToken: false
+              automountServiceAccountToken: true
               containers:
               - name: "$(params.APP_NAME)"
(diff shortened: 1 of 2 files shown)


jimmykarily (approved):

jimmykarily:
I still see this in the events of the application pods:

```
Normal  InjectionSkipped   2m26s (x4 over 95m)  linkerd-proxy-injector  Linkerd sidecar proxy injection skipped: automountServiceAccountToken set to "false"                                                                             │
```
(and no sidecar exists)

PR #623, opened 2021-07-15
Add builder image arg to push

### LINKED ISSUE #455: Allow the user to use latest buildpacks and stack (change the builder version)
The user may want to use a different builder image for the following reasons:

- Use a newer version (security updates, new features etc)
- Use an older version (compatibility issues, apps not ready to upgrade etc)
- Use a custom version (with custom buildpacks they maintain, smaller image etc)

There is only one place where the url to the builder image is used and that is in the tekton pipeline. Being able to change that is only a matter or propagating the user's preference down to the pipeline. Maybe persisting that preference in the Epinio config is also good UX (in which case there should be a way to inspect the currently used image, store the builder image used for staging on the apps themselves and allow the user to change the persisted setting)

manno:
fixes #455

### DIFF
--- a/deployments/tekton.go
+++ b/deployments/tekton.go
@@ -27,12 +27,9 @@ import (
 	"github.com/pkg/errors"
 	"github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"github.com/tektoncd/pipeline/pkg/client/clientset/versioned"
-	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	yaml2 "sigs.k8s.io/yaml"
-
-	typedbatchv1 "k8s.io/client-go/kubernetes/typed/batch/v1"
 )
 
 type Tekton struct {
@@ -287,13 +284,6 @@ func (k Tekton) Deploy(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI
 		return err
 	}
 
-	s := ui.Progress("Warming up cluster with builder image")
-	err = k.warmupBuilder(ctx, c)
-	if err != nil {
-		return err
-	}
-	s.Stop()
-
 	return nil
 }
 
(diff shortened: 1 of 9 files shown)


deployments/tekton.go
```diff
@@ -484,6 +484,7 @@ func (k Tekton) createClusterRegistryCredsSecret(ctx context.Context, c *kuberne
 // in order to avoid pulling it the first time we an application is staged.
 // TODO: This doesn't work in a multi-node cluster because it will only pull
 // the image on one node. Maybe we could use a dummy daemonset for that.
+// This does not help with custom builder images.
```
jimmykarily, line 487: On a follow-up story, we should make it possible to change the default builder image permanently. This setting should either be persisted in the local epinio config file (thus it could be different for each epinio user), or globally (somewhere in the cluster).

In any case, I think we should completely remove this warm-up now, because:

- It only works on 1-node clusters (it's a pod not a daemonset).
- It only helps when using the default builder image.
- It's only needed on the first push.
- If needed, it can be handled outside `epinio install`. E.g. in our tests we could run a make target to warm up the cluster with this image. The users could do something similar on their clusters if they find the first push very important to be fast.

internal/cli/push.go
```diff
@@ -155,7 +164,7 @@ var CmdPush = &cobra.Command{
 		}
 		params.Services = services
 
-		err = client.Push(cmd.Context(), args[0], path, params)
+		err = client.Push(cmd.Context(), params)
```
jimmykarily, line 167: Can we report the builder image in the this output?

```

🚢 About to push an application with given name and sources into the specified organization
Name: sample
Sources: assets/sample-app/
Organization: workspace
```

rohitsakala: Done.

deployments/tekton.go
rohitsakala: Removed the job. Done. No need of make target since tests are passing.

internal/cli/clients/client.go
```diff
@@ -1458,6 +1465,7 @@ func (c *EpinioClient) Push(ctx context.Context, name, source string, params Pus
 		WithStringValue("Name", appRef.Name).
 		WithStringValue("Organization", appRef.Org).
 		WithStringValue("Route", fmt.Sprintf("https://%s", route)).
+		WithStringValue("Builder Image", params.BuilderImage).
```
jimmykarily, line 1468: can we print the same under line 1291 above? (the text before the push)

jimmykarily (approved):

PR #627, opened 2021-07-19
Remove check of (non) dependencies

### LINKED ISSUE #622: All commands fail on powershell
When attempting to use Epinio via powershell, I get the following error:
image

According to @jimmykarily, this error should only be shown when calling `epinio server`.

jimmykarily:
`git` and `sh` are only needed when pushing the application code from
the epinio server to gitea. This check was happening on the client where
those tools are not needed.

Fixes #622

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,14 @@ FLAKE_ATTEMPTS ?= 2
 acceptance-cluster-delete:
 	k3d cluster delete epinio-acceptance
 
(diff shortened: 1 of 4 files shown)


rohitsakala (approved):

rohitsakala:
I didn't run tests locally. but LGTM

rohitsakala (approved):

PR #629, opened 2021-07-19
Add --docker-image-url flag feature

### LINKED ISSUE #574: Spike: Stage an existing Docker image

rohitsakala:
fixes #574 

Reduced 

CLI Tests from 32 min to 20 min
API Tests from 22 min to 10 min


internal/cli/clients/client.go
```diff
@@ -1346,59 +1353,54 @@ func (c *EpinioClient) Push(ctx context.Context, name, rev, source string, param
...
+	c.ui.Normal().Msg("Deploying application ...")
+	deployRequest := models.DeployRequest{
+		App:            appRef,
+		Instances:      params.Instances,
+		Route:          route,
+		Git:            gitRef,
+		CustomImageURL: params.Docker,
+	}
+	if stageID != "" {
+		deployRequest.Stage = models.StageRef{ID: stageID}
```
jimmykarily, line 1395: Should the deploy step care how the image was created? Can't it simply accept an `ImageURL` (replacing the `CustomImageURL`) every time? The internal registry is nothing but a convenience, maybe we don't need to handle it differently. We can keep the StageID around though. It is useful information for the Deployment and the rest of the resources.

internal/cli/push.go
```diff
@@ -95,10 +96,20 @@ var CmdPush = &cobra.Command{
 			return errors.Wrap(err, "could not read option --git")
 		}
 
+		dockerImageURL, err := cmd.Flags().GetString("docker-image-url")
+		if err != nil {
+			return errors.Wrap(err, "could not read option --docker")
```
jimmykarily, line 101:
```suggestion
			return errors.Wrap(err, "could not read option --docker-image-url")
```

acceptance/api/v1/applications_test.go
```diff
@@ -117,6 +119,46 @@ var _ = Describe("Apps API Application Endpoints", func() {
 		return env.Curl("POST", url, strings.NewReader(body))
 	}
 
+	waitForPipeline := func(stageID string) {
+		Eventually(func() string {
+			out, err := helpers.Kubectl(fmt.Sprintf("-n tekton-staging get pipelinerun %s  -o jsonpath='{.status.conditions[0].status}'", stageID))
```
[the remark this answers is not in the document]
rohitsakala: Done.

acceptance/api/v1/suite_test.go
```diff
@@ -63,21 +65,21 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	// Eventually is used to retry in case the rollout of the patched deployment
 	// is not completely done yet.
 	fmt.Println("Ensure default workspace exists")
-	testenv.EnsureDefaultWorkspace()
+	testenv.EnsureDefaultWorkspace(epinioBinary)
```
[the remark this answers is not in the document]
rohitsakala: Yeah makes sense. Leaving this for now for sometime later

deployments/registry.go
```diff
@@ -217,7 +217,8 @@ func (k Registry) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.U
 			return strings.Contains(err.Error(), "x509: certificate signed by unknown authority") ||
 				strings.Contains(err.Error(), "connection refused") ||
 				strings.Contains(err.Error(), "no endpoints available") ||
-				strings.Contains(err.Error(), "EOF")
+				strings.Contains(err.Error(), "EOF") ||
+				strings.Contains(err.Error(), "i/o timeout")
```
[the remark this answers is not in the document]
rohitsakala: I saw this error on my laptop on K3d. I saw it only once and I couldn't reproduce it anymore. My laptop has good resources. So, I added it and it is not due to slow system. 😕

assets/embedded-files/epinio/server.yaml
```diff
@@ -45,6 +45,9 @@ rules:
   resources:
```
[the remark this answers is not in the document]
rohitsakala: Before, we had a tekton Task which created the deployment, service and ingress. now, we do it in through epinio server.

internal/cli/push.go
rohitsakala: Done.

internal/cli/clients/client.go
rohitsakala: Right, I refactored it into using two variables ImageURL and RegistryURL. The imageURL is sent to the deploy endpoint when docker image is specified by the user else in all other cases, registryURL is sent to the deploy endpoint since we don't know ImageURL (the complete one). Savy?

rohitsakala: Done.

internal/api/v1/stage.go
```diff
@@ -41,8 +46,8 @@ func (app *stageParam) GitURL(server string) string {
 // ImageURL returns the URL of the image, using the ImageID. The ImageURL is
 // later used in app.yml.  Since the final commit is not known when the app.yml
```
jimmykarily, line 47: Since we are at it, this comment about the final commit seems well out of date (from when we were adding a commit on top).

internal/api/v1/stage.go
```diff
@@ -153,7 +158,13 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 
 	log.Info("staged app", "org", org, "app", params.AppRef, "uid", uid)
 
-	resp := models.StageResponse{Stage: models.NewStage(uid)}
+	if viper.GetBool("use-internal-registry-node-port") {
```
jimmykarily, line 161: Can we add a comment here? E.g. 

```
The ImageURL in the response should be the one accessible by kubernetes. In stageParam above, the registry is passed with the "internal" url, since it's where tekton will push from within the cluster.
```

internal/cli/clients/client.go
```diff
@@ -39,6 +39,10 @@ import (
 	"k8s.io/apimachinery/pkg/util/validation"
 )
 
+const (
+	LocalRegistry = "127.0.0.1:30500/apps"
```
jimmykarily, line 43: not used anymore?

rohitsakala: Done.

internal/api/v1/stage.go
rohitsakala: Done.

internal/api/v1/stage.go
rohitsakala: Done.

jimmykarily (approved):

jimmykarily (approved):

PR #642, opened 2021-07-23
Add epinio installation intial test

### LINKED ISSUE #606: `epinio install` should print credentials only after updating the config file
I (re)-installed Epinio and I got this: 

```
API User: de7d7d0014503aaa
API Password: db76cb32bccdbdae
Traefik Ingress info: [{"ip":"192.168.48.4"}]
🚢 Updating the stored credentials from the current cluster
✔️  Ok
~/workspace/epinio (564-custom-issuer)*$ vim ~/.config/epinio/config.yaml 
...
 23 pass: 990614b81b7fe49b                                                          
 24 user: 804b43fc5b71fc7f
...
```

The credentials printed in the stdout don't match the (correct) ones in the config file. I think this happens because we print the config values before we update the config. We should print the current credentials in every case. Maybe we should update the config before we print.

rohitsakala:
fixes #606


helpers/epinio/epinio.go
```diff
@@ -0,0 +1,36 @@
+package epinio
```
jimmykarily, line 1: isn't this helper only used in tests? Maybe we should put it under `acceptance/` to make it clear.

.github/workflows/main.yml
```diff
@@ -129,3 +129,46 @@ jobs:
...
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Install Tools
+        run: make tools-install
+      - name: Add Tools to PATH
+        run: |
+          echo "`pwd`/output/bin" >> $GITHUB_PATH
+      - name: API Acceptance Tests
```
jimmykarily, line 165:
```suggestion
      - name: Installation Acceptance Tests
```

acceptance/install/config_test.go
```diff
@@ -0,0 +1,56 @@
...
+	"github.com/epinio/epinio/helpers/epinio"
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Epinio Installation", func() {
+	var flags string
+	epinioBinary := fmt.Sprintf("../../dist/epinio-%s-%s", runtime.GOOS, runtime.GOARCH)
+
+	configFile := "./assets/tests/config.yaml"
```
jimmykarily, line 17: This test is altering the file which means subsequent tests will find it altered. If a similar test is added in the future, the order in which the tests are run could make this test fails. I think it's better if we copy the file to a temporary location and use it from there. The temporary file should be deleted when the tests is done.

helpers/epinio/epinio.go
jimmykarily: Also, if we are introducing a helper, shouldn't it be re-used here: https://github.com/epinio/epinio/blob/main/acceptance/testenv/epinio.go#L100 ?

.github/workflows/main.yml
rohitsakala: Done.

acceptance/install/config_test.go
rohitsakala: Done.

helpers/epinio/epinio.go
rohitsakala: Done. Moved to the acceptance folder. Initially, I thought of re-using the testenv for these tests. But I decided not to since testenv is explicitly used only for creating testing environments and this package I created is used inside the tests.

acceptance/install/config_test.go
```diff
@@ -0,0 +1,72 @@
...
+	epinioUser := "epinio"
+	epinioPassword := "password"
+
+	BeforeEach(func() {
+		input, err := ioutil.ReadFile("../../assets/tests/config.yaml")
+		Expect(err).NotTo(HaveOccurred())
+		f, err := ioutil.TempFile("", "config")
+		Expect(err).NotTo(HaveOccurred())
+
+		configFile = f.Name()
```
jimmykarily, line 31: NOTE: next time we do the same (on another test), we should move it to a helper function.

jimmykarily (approved):

PR #648, opened 2021-07-27
Make serviceaccounttoken true for all apps

### LINKED ISSUE #532: Workloads should be meshed

rohitsakala:
fixes #532


andreas-kupries (approved):
Looks good. Passed local test (focus on the changed test only)

PR #653, opened 2021-07-28
Add --ingress-service-ip flag to epinio cli

### LINKED ISSUE #476: `--traefik-loadbalancer-ip` flag in `epinio installation`

rohitsakala:
fixes #476

### DIFF
--- a/deployments/traefik.go
+++ b/deployments/traefik.go
@@ -116,6 +116,7 @@ func (k Traefik) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI
 	if err != nil {
 		return err
 	}
+	loadBalancerIP := options.GetStringNG("ingress-service-ip")
 
 	// Setup Traefik helm values
 	var helmArgs []string
@@ -125,6 +126,7 @@ func (k Traefik) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI
 	// Overwrite globalArguments until https://github.com/traefik/traefik-helm-chart/issues/357 is fixed
 	helmArgs = append(helmArgs, `--set "globalArguments="`)
 	helmArgs = append(helmArgs, `--set-string deployment.podAnnotations."linkerd\.io/inject"=enabled`)
+	helmArgs = append(helmArgs, fmt.Sprintf("--set-string service.spec.loadBalancerIP=%s", loadBalancerIP))
 
 	helmCmd := fmt.Sprintf("helm %s traefik --namespace %s %s %s", action, TraefikDeploymentID, traefikChartURL, strings.Join(helmArgs, " "))
 	if out, err := helpers.RunProc(helmCmd, currentdir, k.Debug); err != nil {
--- a/deployments/traefik.go
+++ b/deployments/traefik.go
@@ -213,7 +213,7 @@ func (k Traefik) Deploy(ctx context.Context, c *kubernetes.Cluster, ui *termui.U
 	if err == nil {
 		log.Info("service present")
 
-		ui.Exclamation().Msg("System Ingress present, skipping")
+		ui.Exclamation().Msg("System Traefik Ingress present, skipping traefik installation and related flags `--ingress-service-ip`  will be ignored")
 		return nil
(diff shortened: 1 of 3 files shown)


deployments/traefik.go
```diff
@@ -116,6 +116,7 @@ func (k Traefik) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI
 	if err != nil {
 		return err
 	}
+	loadBalancerIP := options.GetStringNG("ingress-service-ip")
```
jimmykarily, line 119: When traefik installation is skipped, either explicitly (--skip-traefik) or implicitly (because it's found), this option will be ignored. I think we should show a warning to the user to ensure they know their setting wasn't respected. E.g.

```
Skipping Traefik Ingress deployment by user request. The `ingress-service-ip` setting will be ignored.
```

jimmykarily: Oh, you already do that here: https://github.com/epinio/epinio/pull/653/files#diff-c0a23dc7003156100c2e86d23fcfe738c2037b8e32c5d95c956baa7f120a0ec3R156
I think we are still missing the implicit skip though.

internal/cli/install.go
```diff
@@ -119,29 +142,26 @@ var CmdInstall = &cobra.Command{
...
+		return errors.Wrap(err, "could not read option --skip-traefik")
+	}
+
+	ingressIP, err := cmd.Flags().GetString("ingress-service-ip")
+	if err != nil {
+		return errors.Wrap(err, "could not read option --ingress-service-ip")
+	}
+
+	if ingressIP != "" && skipTraefik {
+		return errors.New("cannot have --skip-traeifk and --ingress-service-ip together")
```
andreas-kupries, line 160:
```suggestion
		return errors.New("cannot have --skip-traefik and --ingress-service-ip together")
```

Typo fix.

andreas-kupries (dismissed):

internal/cli/install.go
rohitsakala: Solved.

deployments/traefik.go
rohitsakala: Done.

internal/cli/install.go
```diff
@@ -99,10 +99,33 @@ var TraefikOptions = kubernetes.InstallationOptions{
 	},
 }
 
+var CommonInstallOptions = kubernetes.InstallationOptions{
+	{
+		Name:        "ingress-service-ip",
+		Description: "IP address to be assgined to ingress loadbalancer service",
```
andreas-kupries, line 105:
```suggestion
		Description: "IP address to be assigned to ingress loadbalancer service",
```

andreas-kupries (changes requested):
Just another typo.

andreas-kupries (approved):

Issue #600, opened 2021-07-09
Write test plan for installing on supported platforms. Automate that plan.


andreas-kupries:
Like AWS.

manno:
Also maybe a separeate test suite for "installer tests". That tests letsencrypt with a predefined domain? And minikube as well as AWS CI runners for the installer suite?

jimmykarily:
Maybe extend the work started here: #642

manno:
We have a https://github.com/epinio/epinio/tree/main/acceptance/install
and https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)

PR #664, opened 2021-08-06
More installer tests

manno:
Adding the scenarios from https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)#test-scenarios

The scenarios collapse tests for different cloud providers, client platforms, install options, push options and buildpacks.
The list is much shorter than a full matrix.

TODO in another PR:
* complete  the tests
* Macos and windows github runners cannot easily install docker/k3d. at least it's possible on windows.
* we have no code to setup/teardown/clean external clusters, yet.
* adding more push test to the install suite. So we have a stronger assurance, that the install works and we don't need to run the full acceptance suite on all the providers/platforms.

See #600 for initial work on macos/windows runners.

We discussed briefly during planning and may be confident enough to remove complicated push tests from the test scenario list, in favour of a more simple push to show the installation worked. 

refers to #600


acceptance/testenv/download.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"io/ioutil"
+	"os"
+	"path"
+
+	"github.com/epinio/epinio/helpers"
+	"github.com/pkg/errors"
+)
+
+// CompressionType is used by DownloadDescriptor, it's the compression of the downloaded
```
andreas-kupries, line 12: downloaded what ? (archive ? file ? ...)

andreas-kupries (dismissed):
My question is just a nit.

andreas-kupries (approved):

Issue #667, opened 2021-08-09
Epinio API uses forward-auth authentication


jimmykarily:
Follow up of #486 
Depends on #666 

## Acceptance Criteria

- Basic auth flow is completely removed
- Epinio is using forward-auth with dex and a static user for authentication
- The Epinio cli can still authenticate with the API
- The documentation is update and describes how forward auth works in Epinio (any images are updated too)

## Notes

Epinio cli should be able to authenticate and use the Epinio API. This can happens either completely automatically (by using the username/password to complete the oauth2 flow) or semi-automatically (by letting the user complete the flow and writing the final token in the config). If we automate the oauth2 flow in the epinio cli for the default static user, we should make sure we **don't** do that for other idps (e.g. GitHub). The whole purpose of using oauth2 is for the user to not have to give you their third party credentials. The static user credentials belong to Epinio so that's ok in that case.

jimmykarily:
This is how some people have solved the programmatic authentication with dex: https://github.com/kubeflow/kfctl/issues/140#issuecomment-718512435

It's either this, or we can simply have an endpoint in epinio that shows the right epinio config section for the current token. So the user runs something like `epinio login` which gives url to the user to visit. Something like: 


```
$ epinio login
Visit: https://epinio.mydomain.org/cli-token
```
That will lead the user to the start of the oauth flow (the endpoint will be protected by the traefik-forward-auth middleware). When authentication is successul, the served page will say something like:

```
Add this section to your epinio config to allow the epinio cli to talk to Epinio:

token:
  ... things here ...

```

As an improvement, the cli could start a local server (like other clis do, e.g. `fly`). Then the original request will redirect to that local server passing the token in params. The cli will automatically write the config file then.

jimmykarily:
> As an improvement, the cli could start a local server (like other clis do, e.g. fly). Then the original request will redirect to that local server passing the token in params. The cli will automatically write the config file then.

like what is described here for example: https://developer.okta.com/blog/2018/07/16/oauth-2-command-line#the-trick-to-oauth-20-on-the-command-line

jimmykarily:
There is a problem with logging out users. Traefik forward auth, doesn't require the user to pass the Bearer token every time. Instead it creates a cookie which allows authorization until it's expired:

- https://github.com/thomseddon/traefik-forward-auth/blob/c4317b7503fb0528d002eb1e5ee43c4a37f055d0/README.md#logging-out
[...]

PR #668, opened 2021-08-09
666 setup traefik forward auth

### LINKED ISSUE #666: Epinio sets up traefik forward auth with dex
Follow up of #486 

PoC branch: https://github.com/epinio/epinio/tree/486-multiple-users-dex

Epinio authentication should happen with traefik-forward-auth (https://doc.traefik.io/traefik/v2.0/middlewares/forwardauth/) that redirects to dex (https://dexidp.io/docs/kubernetes/). By default, Epinio will create a "static" dex user.

To be able to split the implementation in smaller stories, this feature won't be enabled by default yet. The basic-auth mechanism will still be in-place and enabled. The forward-auth middleware wont' be use anywhere yet. It can still be enabled on an ingress by changing an annotation, so the follow up stories can use this.

In future stories we will add ways for the user to set up and enable other (upstream) identity providers.

## Acceptance Criteria

- Epinio install sets up the forward-auth middleware with dex and a static user (installs traefik-forward-auth, dex etc)
- The base auth flow is still enabled and default.

jimmykarily:
Fixes #666


jimmykarily:
To try this branch, you have to edit the epinio ingress and change the middleware from basic auth to this:

```
traefik.ingress.kubernetes.io/router.middlewares: traefik-forward-auth-forward-auth@kubernetescrd    
```

internal/cli/clients/client.go
```diff
@@ -1699,6 +1699,7 @@ func uniqueStrings(stringSlice []string) []string {
 	return list
 }
 
+// TODO: Fix this. We need to complete the oauth2 process and get a token
```
jimmykarily, line 1702: Moved to #667

internal/web/dex_controller.go
```diff
@@ -0,0 +1,257 @@
+package web
```
jimmykarily, line 1: Remove this controller. Traefik forward auth is taking care of these endpoints now.

deployments/traefik_forward_auth.go
```diff
@@ -0,0 +1,480 @@
...
+	if err := k.createService(ctx, c); err != nil {
+		return errors.Wrap(err, "create Traefik Forward Auth service")
+	}
+
+	if err := k.createCertificate(ctx, c, options, ui); err != nil {
+		return errors.Wrap(err, "create Traefik Forward Auth service")
+	}
+
+	// Wait until the cert is there before we create the Ingress
+	// TODO: Fix all the ingresses to wait before they are created
```
jimmykarily, line 114: Do this in this PR or another?

internal/web/dex_controller.go
jimmykarily: removed

internal/cli/clients/client.go
jimmykarily: changed the comment

deployments/traefik_forward_auth.go
jimmykarily: Did that

deployments/dex.go
```diff
@@ -0,0 +1,289 @@
...
+	DexDeploymentID = "dex"
+	dexVersion      = "2.30.0"
+	dexChartFile    = "dex-0.6.0.tgz"
+)
+
+func (k *Dex) ID() string {
+	return DexDeploymentID
+}
+
+func (k *Dex) Backup(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI, d string) error {
```
manno, line 37: Backup/Restore are no longer part of the interface in main, instead PreInstallCheck is needed

deployments/dex.go
```diff
@@ -0,0 +1,289 @@
...
+	if err := c.WaitForNamespace(ctx, ui, TraefikForwardAuthDeploymentID, duration.ToDeploymentNamespaceCreated()); err != nil {
+		return errors.Wrapf(err, "waiting for %s deployment", TraefikForwardAuthDeploymentID)
+	}
+	traefikAuthClientSecretName := fmt.Sprintf("%s-client", TraefikForwardAuthDeploymentID)
+	traefikAuthClientSecret, err := c.WaitForSecret(ctx, TraefikForwardAuthDeploymentID, traefikAuthClientSecretName, duration.ToAuthClientSecretCreated())
+	if err != nil {
+		return errors.Wrapf(err, "waiting for %s secret", traefikAuthClientSecretName)
+	}
+
+	apiUserOpt, err := options.GetOpt("user", "")
```
manno, line 149: You can use `GetString(string, string) (string, error)` to avoid the cast or `GetStringNG(string) string` to just get the value.

deployments/epinio.go
```diff
@@ -371,7 +377,7 @@ func (k *Epinio) createIngress(ctx context.Context, c *kubernetes.Cluster, subdo
 										}}}}}}},
 				TLS: []networkingv1.IngressTLS{{
 					Hosts:      []string{subdomain},
-					SecretName: "epinio-tls",
+					SecretName: EpinioDeploymentID + "-tls",
```
manno, line 380: Are you preparing Epinio for multiple deployments? Maybe/Eventually we should invest in a function "names.TLSSecretSomething" instead of constructing the name every time.

deployments/registry.go
```diff
@@ -251,36 +298,9 @@ func (k Registry) apply(ctx context.Context, c *kubernetes.Cluster, ui *termui.U
...
 	}
 
-	ui.Success().Msg("Registry deployed")
-
-	log.Info("apply done")
-
-	return nil
-}
-
-func (k Registry) GetVersion() string {
```
manno, line 261: I just wish all the deployments had the same order for these funcs :)

internal/duration/duration.go
```diff
@@ -10,18 +10,21 @@ import (
...
-	orgDeletion         = 5 * time.Minute
-	serviceSecret       = 5 * time.Minute
-	serviceProvision    = 5 * time.Minute
-	serviceLoadBalancer = 5 * time.Minute
-	podReady            = 5 * time.Minute
-	appBuilt            = 10 * time.Minute
-	warmupJobReady      = 30 * time.Minute
-	certManagerReady    = 5 * time.Minute
-	kubedReady          = 5 * time.Minute
+	systemDomain             = 2 * time.Minute
```
manno, line 13: This seems weird to me. Maybe we should have `const slow = 2 * time.Minute`, `const medium`, `const long` instead of all these numbers?

Anyhow, not important.

deployments/traefik_forward_auth.go
```diff
@@ -0,0 +1,485 @@
...
+								Secret: &corev1.SecretVolumeSource{
+									SecretName: "epinio-ca-root",
+								},
+							},
+						},
+					},
+					Containers: []corev1.Container{
+						{
+							Name:  "traefik-forward-auth",
+							Image: "docker.io/thomseddon/traefik-forward-auth:" + TraefikForwardAuthVersion,
```
manno, line 370: Aha! ;)
I'm missing some documentation/note pointing me to https://github.com/thomseddon/traefik-forward-auth
Maybe in a later PR

deployments/cert_manager.go
```diff
@@ -260,6 +261,32 @@ func (cm CertManager) apply(ctx context.Context, c *kubernetes.Cluster, ui *term
...
+	// as per the docs:
+	// https://cert-manager.io/docs/faq/kubed/#syncing-arbitrary-secrets-across-namespaces-using-kubed
+	secretName := "epinio-ca-root"
+
+	emptySecret := v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      secretName,
+			Namespace: CertManagerDeploymentID,
+			Annotations: map[string]string{
+				"kubed.appscode.com/sync": fmt.Sprintf("kubed-source-namespace=%s", CertManagerDeploymentID),
```
manno, line 275: Ok, so this signals to me the source namespace of the secret, when I see it on some namespace.
That's good.
You could shorten this, as the kubed connection is already clear from the key name:
```suggestion
				"kubed.appscode.com/sync": fmt.Sprintf("source-namespace=%s", CertManagerDeploymentID),
```

manno (changes requested):
I think this needs to be rebased on main?

deployments/cert_manager.go
jimmykarily: On the namespace resource, the key is simply `kubed-source-namespace` and `kubed.appscode.com/sync` doesn't exist.

So this annotation is saying "when you find a namespace with a label key `kubed-source-namespace` and a label value of `CertManagerDeploymentID`, copy this secret over"

deployments/dex.go
jimmykarily: Will remove (I think you mean `PreDeployCheck`, which [is already there](https://github.com/epinio/epinio/pull/668/files#diff-e77a4e78f949937daf2ca120d63abb5af20e28cb27c1ad16427b3f8277b39667R255))

deployments/dex.go
jimmykarily: Oh, right! I was copying from [the epinio deployment](https://github.com/epinio/epinio/blob/main/deployments/epinio.go#L112-L125). I will fix it there too.

deployments/epinio.go
jimmykarily: I just remember that in some places we construct the name from the deployment name and here we had that hardcoded. I changed it just to make sure, if we ever rename the deployment, that this secret name will also be changed.

Maybe it deserves a bigger refactoring like the one you describe. Not sure if we want to do it here.

deployments/registry.go
jimmykarily: True but the files are so big that we rarely scroll. Search and code navigation does the trick most of the time. If I start moving things it will be hard to review the actual changes. I think this should happen (if at all) in a PR dedicated to moving code (no changes at all) to make reviewing easier.

internal/duration/duration.go
jimmykarily: It feels weird to me too. We just gave names to magic numbers and now we have magic constants! In general, these are upper limits on the time we expect it to take. When that proves to be small, we just bump it until it's good. The reason for having multiple of them here is to avoid bumping the timeout for irrelevant operations when a bump is needed.

So the rule is, re-use the same timeout when the operation is similar. In reality though, it's hard to tell which operations are "similar". For example, if I'm waiting for cert manager to create a secret and I do it 5 minutes before I install cert-manager, that obviously is going to take 5+ minutes. If I do the same operation (wait for the secret) but after cert-manager is already installed, it may take only some seconds. Is that the same operation then?

deployments/epinio.go
manno: Nah, let's do it later

deployments/registry.go
manno: Yes, if we do it, it should be separate

manno (approved):

deployments/traefik_forward_auth.go
jimmykarily: I added a bullet in the Acceptance Criteria in #667

Issue #503, opened 2021-06-17
Issues getting started with epinio on a windows machine


gunamata:
Hi, 

I am trying to get epinio sample application push working on my windows laptop, and ran into issues. Here are are steps I followed.

1. Start with a fresh Rancher Desktop installation. Install epinio prerequisites (kubectl, helm). 

```
c:\2021\Guna\SUSE\epinio\assets>kubectl get nodes
NAME              STATUS   ROLES                          AGE   VERSION
desktop-hiq5d5u   Ready    builder,control-plane,master   68m   v1.21.0+k3s1

c:\2021\Guna\SUSE\epinio\assets>kubectl get ns
NAME              STATUS   AGE
kube-system       Active   70m
default           Active   70m
kube-public       Active   70m
kube-node-lease   Active   70m
kube-image        Active   69m

```

2. Install epinio using `epinio install --verbosity=1`

```
c:\2021\Guna\SUSE\epinio\assets>epinio install --verbosity=1

←[34m🚢 Epinio installing...
←[0m
Configuration...
  🧭  skip-traefik: ←[35mfalse←[0m
  🧭  skip-linkerd: ←[35mfalse←[0m
  🧭  system_domain: ←[32m←[0m
  🧭  email_address: ←[32mepinio@suse.com←[0m
  🧭  user: ←[32m80eab6de208b75a2←[0m
  🧭  password: ←[32m5d53fd58eaf9f17a←[0m

←[34m🚢 Deploying Linkerd...
←[0m
←[32m✔️  Linkerd deployed
←[0m
←[33m⚠️  Traefik Ingress already installed, skipping
←[0m
🕞  Waiting for LoadBalancer IP on traefik service. ←[35m.←[0m

←[32m✔️  Created system_domain: 192.168.46.233.omg.howdoi.website
←[0m
←[34m🚢 Deploying Epinio...
←[0m

←[34m🚢 Deploying Quarks...
←[0m
←[34m🚢 Deploying Tekton...
←[0m←[34m🚢 Deploying ServiceCatalog...
←[0m
←[34m🚢 Deploying CertManager...
←[0m
←[34m🚢 Deploying Gitea...
←[0m
←[34m🚢 Deploying Registry...
←[0mI0617 11:28:37.158803   13904 request.go:655] Throttling request took 1.0333661s, request: POST:https://192.168.46.233:6443/api/v1/namespaces

[...]

PR #671, opened 2021-08-10
Cluster patching for Windows. Plus documentation.

### LINKED ISSUE #660: Windows binary should default to "--no-colors"
As Windows cannot [display emojies](503) or handle [color output](659) in a reasonable way, the `--no-colors` flag should default to `true` in Windows.

### LINKED ISSUE #672: Windows install fails on dex deployment
* epinio (built from source) @ c7afb19

* Windows 10 Home, Kubernetes via Rancher Desktop

* Running `epinio install` in PowerShell

Installation fails with
```
panic: Deployment dex failed with error: Failed to extract embedded file:  - file does not exist


goroutine 52 [running]:
github.com/epinio/epinio/internal/cli/clients.(*InstallClient).Install.func1(0xc000254150, 0x24ca9b0, 0xc000034050, 0x24d5670, 0xc00044b400, 0x24d9df8, 0xc00020d0b0, 0xc00020d094)
        /abuild/projects/rancher/epinio/epinio/internal/cli/clients/install_client.go:168 +0x23f
created by github.com/epinio/epinio/internal/cli/clients.(*InstallClient).Install
        /abuild/projects/rancher/epinio/epinio/internal/cli/clients/install_client.go:165 +0x10e5
```

### LINKED ISSUE #673: '--no-colors' option has no effect on Windows
Passing the `--no-colors` command line option to Epinio on Windows has no visual effect.

andreas-kupries:
References #503 
Fixes #673 
Fixes #672
Fixes #660

### Notes

- `kubectl cp` syntax specifies a pod file as `pod`:`path`. Note the `:`. That is a special character on Windows and breaks the command. This got replaced by a series of `kubectl exec` commands using a tar command client- and pod-side to copy the file in, move it into the proper place, and ensure correct permissions.
- Note that in the above we cannot use absolute paths, like `/epinio` either, they are breaking the command as well. Because on Windows these are __drive-relative__ paths and get expanded by the kubectl binary before trying to use them. Luckily for us relative paths, i.e `epinio` do work. Because the CWD for `kubectl exec` is the root directory of the pod, making relative and absolute paths equivalent.
- Newly added to the patch script is the `EPINIO_BINARY_TAG` variable. It contains the tag for the image the binary will be placed in. From within a git checkout the default, provided by a `git describe ...` command is good enough. Outside of a checkout use `make tag` in the checkout to determine the tag, and then set the environment variable explicitly.

### DIFF
--- a/scripts/patch-epinio-deployment.sh
+++ b/scripts/patch-epinio-deployment.sh
@@ -16,8 +16,14 @@
 # Patching the deployment forces the pod to restart with a now existing image
 # and the correct binary is in place.
 
-export EPINIO_BINARY_PATH=${EPINIO_BINARY_PATH:-'dist/epinio-linux-amd64'}
+export EPINIO_BINARY_PATH="${EPINIO_BINARY_PATH:-dist/epinio-linux-amd64}"
+export EPINIO_BINARY_TAG="${EPINIO_BINARY_TAG:-$(git describe --tags --abbrev=0)}"
 
+echo
+echo Configuration
+echo "  - Binary: ${EPINIO_BINARY_PATH}"
+echo "  - Tag:    ${EPINIO_BINARY_TAG}"
+echo
 echo "Creating the PVC"
 cat <<EOF | kubectl apply -f -
 ---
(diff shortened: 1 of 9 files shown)


andreas-kupries:
Confirmed working :tada: :fireworks: :green_heart:

manno (approved):

scripts/patch-epinio-deployment.sh
```diff
@@ -60,7 +66,19 @@ echo "Waiting for dummy pod to be ready"
 kubectl wait --for=condition=ready pod -n epinio epinio-copier
 
 echo "Copying the binary on the PVC"
-kubectl cp ${EPINIO_BINARY_PATH} epinio/epinio-copier:/epinio/epinio
+# Notes
+# 1. kubectl cp breaks because of the colon in `pod:path`. Thus the more complex tar construction.
```
jimmykarily, line 70: Will probably be fixed at some point: https://github.com/kubernetes/kubernetes/pull/102047

PR #677, opened 2021-08-13
Rename --ingress-service-ip to --loadbalancer-ip

manno:
It's the IP of the external loadbalancer.

refers to #600


andreas-kupries (approved):
:+1:

PR #678, opened 2021-08-13
Increase default deployment installation timeout

### LINKED ISSUE #654: Azure AKS: Linkerd installation times out on fresh AKS clusters (incl. fresh resources)
I installed epinio-0.0.19 to a newly created AKS cluster (kube versions tested v1.19.11 and v1.20.7, 2 Standard_D3_v2 nodes).

Tested `epinio install-ingress` and plain `epinio install`.

On any initial epinio installation, the linkerd deployment times out:

```
> ./epinio-0.0.19 install

🚢 Epinio installing...

Configuration...
  🧭  skip-traefik: false
  🧭  skip-linkerd: false
  🧭  system_domain: 
  🧭  email_address: epinio@suse.com
  🧭  user: 050020026671ca82
  🧭  password: 5254901c87555a3a
  🧭  tls-issuer: epinio-ca
  🧭  use-internal-registry-node-port: true

🚢 Deploying Linkerd...error installing Epinio: failed waiting Linkerd install job to complete: timed out waiting for the condition
```

Observations:

I think the timeout for linkerd to become ready is to short for Azure AKS, because

- the second try (epinio install-uninstall-install) will always succeed, and epinio will work afterwards
- if I delete the AKS cluster, and recreate the cluster using the same AKS resource names, the first epinio installation
will then succeed as well - I know from experience, that AKS will reuse resources (e.g. kube service, image cache, aso.), if 
a cluster with the same resource names will be recreated within a specific timeframe.
- if I create an AKS cluster with resource naming that's not been used before, the first epinio install will time out for sure

So beside image caching and download times, it could also be about kubernetes configurations, which take too long to create, but aren't cleaned up on uninstall.

Related issues:

- https://github.com/epinio/epinio/issues/649
- https://github.com/epinio/epinio/issues/647

rohitsakala:
fixes #654


deployments/linkerd.go
```diff
@@ -178,3 +187,49 @@ func (k Linkerd) Upgrade(ctx context.Context, c *kubernetes.Cluster, ui *termui.
...
+		},
+		Spec: batchv1.JobSpec{
+			Template: corev1.PodTemplateSpec{
+				Spec: corev1.PodSpec{
+					ServiceAccountName: serviceAccountName,
+					Containers: []corev1.Container{
+						{
+							Name:            jobName,
+							Image:           imageName,
+							ImagePullPolicy: "Always",
```
manno, line 212: This will download the image every time. I don't think that's wanted?

manno (changes requested):
Note: Title of PR doesn't match content.

manno (approved):

deployments/linkerd.go
rohitsakala: Changed it

rohitsakala: Removed it.

Issue #620, opened 2021-07-15
Epinio cli should work without kube access


jimmykarily:
There are some commands that will always need access to the kube api. E.g.
- epinio install
- epinio config update-credentials

These are considered "admin" commands. Regular epinio users should not have to run those commands. This means, all the commands that are not "admin" commands should be possible to run with no access to the kubernetes cluster. That is, with no kubeconfig in place.

Right now Epinio is still talking to the kube api for various other reasons. E.g. for shell autocompletion (find available services/apps/etc) or in order to discover the Epinio API by looking up for the ingress.

## Acceptance Criteria

- All the code that fetches information from the cluster by talking to kube is migrated to API endpoints
- No API endpoint autodiscovery happens. There should be a command to target an instance of Epinio (e.g. `epinio login`). This command should update the epinio config file if run successfully. This command is an alternative to running `epinio config update-credentials` as an admin. It can be used when no kubernetes cluster access is possible.

agracey:
This seems like a good change to make as part of the work for multiple users

malcolmlewis:
Hi
This would also enable the rpm completions package to build without needing kubectl as part of build requires. There is a bit of a disconnect with kubectl on openSUSE/SLE (different repos in use) except Tumbleweed.

agracey:
I think this is a dup of #508, right?

jimmykarily:
Yes. Although #508 is older, I will keep this one because the description feels a bit more detailed.

manno:
> No API endpoint autodiscovery happens. There should be a command to target an instance of Epinio (e.g. epinio login). This command should update the epinio config file if run successfully. This command is an alternative to running epinio config update-credentials as an admin. It can be used when no kubernetes cluster access is possible.

Not sure I understand that acceptance criteria. Is that about the API endpoint and handled by https://github.com/epinio/epinio/issues/681?
Then that new `login` command would allow the user to set url/user/pass in the epinio.yaml config.

jimmykarily:
> > No API endpoint autodiscovery happens. There should be a command to target an instance of Epinio (e.g. epinio login). This command should update the epinio config file if run successfully. This command is an alternative to running epinio config update-credentials as an admin. It can be used when no kubernetes cluster access is possible.
> 
> Not sure I understand that acceptance criteria. Is that about the API endpoint and handled by #681?
> Then that new `login` command would allow the user to set url/user/pass in the epinio.yaml config.

[...]

PR #692, opened 2021-08-16
Caching api/wss urls in the configuration

### LINKED ISSUE #681: Cache the epinio API url
to avoid asking the kube api what the ingress is with every command.

Rename `epinio config update-credentials` to `epinio config update` which will update:

- the CA in the config
- the basic auth credentials in the config
- the epinio url in the config

This command should contact the kube api, discover the epinio server ingress and update the cached value.

In the future we could implement a command to manually set the value. E.g. move the current `epinio target` to `epinio orgs target` (which makes more sense) and let `epinio target` set the epinio url. Or even `epinio login https://...`.

andreas-kupries:
Fixes #681

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -245,7 +245,7 @@ func (c *EpinioClient) ConfigUpdate(ctx context.Context) error {
 
 	details.Info("retrieving credentials")
 
-	user, password, err := getCredentials(details, ctx)
+	user, password, err := getCredentials(ctx, details)
 	if err != nil {
 		c.ui.Exclamation().Msg(err.Error())
 		return nil
@@ -254,7 +254,7 @@ func (c *EpinioClient) ConfigUpdate(ctx context.Context) error {
 	details.Info("retrieved credentials", "user", user, "password", password)
 	details.Info("retrieving server locations")
 
-	api, wss, err := getAPI(details, ctx)
+	api, wss, err := getAPI(ctx, details)
 	if err != nil {
 		c.ui.Exclamation().Msg(err.Error())
 		return nil
@@ -263,7 +263,7 @@ func (c *EpinioClient) ConfigUpdate(ctx context.Context) error {
 	details.Info("retrieved server locations", "api", api, "wss", wss)
 	details.Info("retrieving certs")
 
-	certs, err := getCerts(details, ctx)
+	certs, err := getCerts(ctx, details)
 	if err != nil {
 		c.ui.Exclamation().Msg(err.Error())
 		return nil
@@ -1724,7 +1724,7 @@ func uniqueStrings(stringSlice []string) []string {
 	return list
 }
 
-func getAPI(log logr.Logger, ctx context.Context) (string, string, error) {
+func getAPI(ctx context.Context, log logr.Logger) (string, string, error) {
 	cluster, err := kubernetes.GetCluster(ctx)
 	if err != nil {
 		return "", "", err
@@ -1741,7 +1741,7 @@ func getAPI(log logr.Logger, ctx context.Context) (string, string, error) {
 }
 
 // TODO: https://github.com/epinio/epinio/issues/667
-func getCredentials(log logr.Logger, ctx context.Context) (string, string, error) {
+func getCredentials(ctx context.Context, log logr.Logger) (string, string, error) {
 	cluster, err := kubernetes.GetCluster(ctx)
 	if err != nil {
 		return "", "", err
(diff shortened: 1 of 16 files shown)


internal/cli/clients/epinio_api.go
```diff
@@ -24,10 +25,32 @@ type EpinioAPIClient struct {
 var epinioClientMemo *EpinioAPIClient
 
 func GetEpinioAPIClient(ctx context.Context) (*EpinioAPIClient, error) {
+	// Check for information cached in memory, and return if such is found
 	if epinioClientMemo != nil {
 		return epinioClientMemo, nil
 	}
 
+	// Check for information cached in the peinio configuration,
```
jimmykarily, line 33:
```suggestion
	// Check for information cached in the Epinio configuration,
```

jimmykarily:
Unfortunately we still make a request to the kube API for every `epinio` command: https://github.com/epinio/epinio/blob/main/internal/cli/clients/client.go#L69

This is another issue though. Let's fix it on another PR.

andreas-kupries:
Ah. I see. Yes. Although I am not sure if the creation of the cluster structure actually talks to the kubernetes API.
Regaardless, this could be moved behind an in-memory cache (like we do for a few things already, and if `GetCluster` is not doing so already), and creation then deferred until the first actual need.

jimmykarily:
yes maybe the easiest solution would be to set the `Cluster` field to a method instead of a concrete value and memoize that.

jimmykarily (approved):

jimmykarily:
> Unfortunately we still make a request to the kube API for every `epinio` command: https://github.com/epinio/epinio/blob/main/internal/cli/clients/client.go#L69
> 
> This is another issue though. Let's fix it on another PR.

We can fix this on #620

internal/cli/clients/epinio_api.go
```diff
@@ -24,10 +25,32 @@ type EpinioAPIClient struct {
...
 
 func GetEpinioAPIClient(ctx context.Context) (*EpinioAPIClient, error) {
+	// Check for information cached in memory, and return if such is found
 	if epinioClientMemo != nil {
 		return epinioClientMemo, nil
 	}
 
+	// Check for information cached in the Epinio configuration,
+	// and return if such is found. Cache into memory as well.
+	configConfig, err := config.Load()
```
manno, line 35: nit: Why not just `cfg`?

> Variable names in Go should be short rather than long. This is especially true for local variables with limited scope. Prefer c to lineCount. Prefer i to sliceIndex.

https://github.com/golang/go/wiki/CodeReviewComments#variable-names

internal/cli/enable.go
```diff
@@ -14,9 +16,13 @@ var CmdEnable = &cobra.Command{
 	Use:           "enable",
 	Short:         "enable Epinio features",
 	Long:          `enable Epinio features that are not enabled by default`,
-	Args:          cobra.ExactArgs(0),
 	SilenceErrors: true,
 	SilenceUsage:  true,
+	Args:          cobra.MinimumNArgs(1),
+	RunE: func(cmd *cobra.Command, args []string) error {
+		cmd.Usage()
+		return errors.New(fmt.Sprintf(`Unknown method "%s"`, args[0]))
```
manno, line 24: nit: there is also https://pkg.go.dev/fmt#Errorf and https://pkg.go.dev/github.com/pkg/errors?utm_source=godoc#Errorf

manno (dismissed):

internal/cli/clients/client.go
```diff
@@ -1737,7 +1778,7 @@ func getCredentials(log logr.Logger, ctx context.Context) (string, string, error
 	return user, pass, nil
 }
 
-func getCerts(ctx context.Context, log logr.Logger) (string, error) {
+func getCerts(log logr.Logger, ctx context.Context) (string, error) {
```
manno, line 1781: only mentioning this, since it's a private func no need to change it:

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx: 

https://pkg.go.dev/context#pkg-overview

andreas-kupries: Heh. And here my change brought this one function into sync with the others in the file.

andreas-kupries: Fixed with commit c7c989546ecc858c5df015650247b70c76d1be0c

internal/cli/enable.go
andreas-kupries: Fixed with commit d96c92939a49edd006e3e7119f6d4157e96c9025

internal/cli/clients/epinio_api.go
andreas-kupries: Fixed with commit 53b826e55dfc618ab4406227517b4deb030a2bd5

manno (approved):

PR #694, opened 2021-08-17
No longer copy epinio config/binary to ginkgo node tmp dir

manno:
We only have one cluster, which is shared by all ginkgo nodes.

### DIFF
--- a/acceptance/api/v1/suite_test.go
+++ b/acceptance/api/v1/suite_test.go
@@ -94,16 +94,12 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	Expect(os.Getenv("KUBECONFIG")).ToNot(BeEmpty(), "KUBECONFIG environment variable should not be empty")
 
-	env = testenv.New(nodeTmpDir, testenv.Root())
-
-	out, err := env.CopyEpinio()
+	// Get config from the installation (API credentials)
+	out, err := testenv.CopyEpinioConfig(nodeTmpDir)
 	Expect(err).ToNot(HaveOccurred(), out)
-
 	os.Setenv("EPINIO_CONFIG", nodeTmpDir+"/epinio.yaml")
 
-	// Get config from the installation (API credentials)
-	out, err = proc.Run("", false, "cp", testenv.EpinioYAML(), nodeTmpDir+"/epinio.yaml")
-	Expect(err).ToNot(HaveOccurred(), out)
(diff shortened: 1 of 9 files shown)


acceptance/api/v1/suite_test.go
```diff
@@ -93,19 +93,11 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 
 	Expect(os.Getenv("KUBECONFIG")).ToNot(BeEmpty(), "KUBECONFIG environment variable should not be empty")
+	Expect(os.Getenv("EPINIO_CONFIG")).ToNot(BeEmpty(), "EPINIO_CONFIG environment variable should not be empty")
 
 	env = testenv.New(nodeTmpDir, testenv.Root())
 
-	out, err := env.CopyEpinio()
-	Expect(err).ToNot(HaveOccurred(), out)
-
-	os.Setenv("EPINIO_CONFIG", nodeTmpDir+"/epinio.yaml")
```
andreas-kupries, line 102: Question: Where does the setup of this EV happen now, with this removed ?

manno: It's done in SetupEnv (https://github.com/epinio/epinio/blob/0ea808eacf7524bc1c0dca48c7d61969a742d29a/acceptance/api/v1/suite_test.go#L41), I verified it's still present in the ginkgo node parallel worker

andreas-kupries: Ok.

andreas-kupries: Except CI goes :boom: with

>[3]   EPINIO_CONFIG environment variable should not be empty
[3]   Expected
[3]       <string>: 
[3]   not to be empty

indicating trouble around that EV

manno: Yes, because the `env.Epinio` switches directories in some tests... latest commit should help. However, I'm also seeing 401 errors in tests locally :?

andreas-kupries: I do not believe so. The error above is from the check at the beginning of every suite that the EV is actually set.
I.e. this is a failure in the global suite setup.
Because of that error no tests are actually run.

andreas-kupries: The install test has already failed in the same manner for the latest commit

manno: Haha, makes sense. My local environment is different.

andreas-kupries (approved):

PR #700, opened 2021-08-18
Prefix app service and ingress names to ensure validity as a kube resource name

### LINKED ISSUE #690: Application names should map to "safe" URLs
Right now application names like `12factor` can't be used because they result in a route like `12factor.omg.howdoi.website` which is not a valid url (because it starts with a number?).

We need some small library that generates url-safe routes for an application given the application name.
Ideally the route should still have a part of it that looks like the application name so it's easy to tell which application this route belongs too.

andreas-kupries:
Fixes #690

### DIFF
--- a/helpers/kubernetes/cluster.go
+++ b/helpers/kubernetes/cluster.go
@@ -17,6 +17,7 @@ import (
 	kind "github.com/epinio/epinio/helpers/kubernetes/platform/kind"
 	minikube "github.com/epinio/epinio/helpers/kubernetes/platform/minikube"
 	"github.com/epinio/epinio/helpers/termui"
+	"github.com/epinio/epinio/internal/names"
 
 	appsv1 "k8s.io/api/apps/v1"
 	apibatchv1 "k8s.io/api/batch/v1"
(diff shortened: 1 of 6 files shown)


helpers/kubernetes/cluster.go
```diff
@@ -610,7 +611,8 @@ func (c *Cluster) GetVersion() (string, error) {
 
 // ListIngressRoutes returns a list of all routes for ingresses in `namespace` with the given selector
 func (c *Cluster) ListIngressRoutes(ctx context.Context, namespace, name string) ([]string, error) {
-	ingress, err := c.Kubectl.NetworkingV1().Ingresses(namespace).Get(ctx, name, metav1.GetOptions{})
+	ingress, err := c.Kubectl.NetworkingV1().Ingresses(namespace).Get(
+		ctx, names.GenerateResourceName("i-"+name), metav1.GetOptions{})
```
jimmykarily, line 615: I think it should be the caller that worries about passing the right ingress name. This method can stay generic by expecting the correct name to be passed.

andreas-kupries: Done with 89db30b8515d48c9fa4fa9207b7cb352090912ac

jimmykarily (approved):

PR #701, opened 2021-08-18
688 cache image

### LINKED ISSUE #688: build cache for staging
The buildpack tekton task supports the CACHE_IMAGE and "cache workspace" settings to cache builds artifacts so that subsequent pushes are faster. Right now, pushing Rails or Wordpress on Epinio is equally slow every time because we don't use that. We should try to use one of the settings to speed things up.

## Acceptance Criteria

- a second push of the same app, uses some caching mechanism from the tekton pipeline
- [optional] there is a way to push without caches and maybe delete the cache altogether
- the cache should be deleted at least when the application is deleted.

jimmykarily:
Fixes #688

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -194,6 +194,44 @@ var _ = Describe("Apps", func() {
 		})
 	})
 
+	Describe("build cache", func() {
+		push := func(arg ...string) (string, error) {
+			appDir := "../assets/sample-app"
+			return env.Epinio(appDir, "app", append([]string{"push", appName}, arg...)...)
+		}
+		BeforeEach(func() {
+			out, err := push()
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+
+		When("pushing for the second time", func() {
+			AfterEach(func() {
+				env.DeleteApp(appName)
+			})
+
+			It("is using the cache PVC", func() {
+				out, err := helpers.Kubectl("get", "pvc", "-n", deployments.TektonStagingNamespace, fmt.Sprintf("%s-%s", org, appName))
+				Expect(err).ToNot(HaveOccurred(), out)
+
+				out, err = push()
+				Expect(err).ToNot(HaveOccurred(), out)
+
+				Expect(out).To(MatchRegexp("Reusing cache layer"))
+			})
+		})
+		When("deleting the app", func() {
+			It("deletes the cache PVC too", func() {
+				out, err := helpers.Kubectl("get", "pvc", "-n", deployments.TektonStagingNamespace, fmt.Sprintf("%s-%s", org, appName))
+				Expect(err).ToNot(HaveOccurred(), out)
+				env.DeleteApp(appName)
+
+				out, err = helpers.Kubectl("get", "pvc", "-n", deployments.TektonStagingNamespace, fmt.Sprintf("%s-%s", org, appName))
+				Expect(err).To(HaveOccurred(), out)
+				Expect(out).To(MatchRegexp(fmt.Sprintf(`persistentvolumeclaims "%s-%s" not found`, org, appName)))
+			})
+		})
+	})
+
 	Describe("push and delete", func() {
 		It("shows the staging logs", func() {
 			By("pushing the app")
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/epinio/epinio/deployments"
 	"github.com/epinio/epinio/helpers"
 	v1 "github.com/epinio/epinio/internal/api/v1"
+	"github.com/epinio/epinio/internal/names"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
@@ -210,7 +211,8 @@ var _ = Describe("Apps", func() {
 			})
 
 			It("is using the cache PVC", func() {
-				out, err := helpers.Kubectl("get", "pvc", "-n", deployments.TektonStagingNamespace, fmt.Sprintf("%s-%s", org, appName))
+				out, err := helpers.Kubectl("get", "pvc", "-n",
+					deployments.TektonStagingNamespace, names.GenerateResourceName(org, appName))
 				Expect(err).ToNot(HaveOccurred(), out)
 
 				out, err = push()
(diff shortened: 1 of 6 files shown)


assets/embedded-files/tekton/stage-pipeline.yaml
```diff
@@ -57,6 +67,29 @@ spec:
...
+kind: Task
+metadata:
+  name: cleanup
+  namespace: tekton-staging
+spec:
+  workspaces:
+  - name: source
+  steps:
+  - name: cleanup
+    image: lachlanevenson/k8s-kubectl
```
jimmykarily, line 84: I'm just re-using what the other task uses. Anything with `rm` available would do.

internal/api/v1/stage.go
```diff
@@ -148,6 +178,12 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		BuilderImage: req.BuilderImage,
 	}
 
+	// TODO: Delete the PVC when the application is deleted
```
jimmykarily, line 181: This is done.

internal/api/v1/models/app.go
```diff
@@ -62,6 +64,11 @@ func (ar *AppRef) EnvSecret() string {
 	return ar.Name + "-env"
 }
 
+func (ar *AppRef) PVCName() string {
+	// TODO: Can this exceed the allowed limit in characters?
+	return fmt.Sprintf("%s-%s", ar.Org, ar.Name)
```
jimmykarily, line 69: Irrelevant but since I remembered it here: The application image name doesn't have the org in it. This means the same application on different orgs could result in a conflict on the image name? We append more things to the image name so maybe that is impossible to happen, but I will check.

andreas-kupries: IIRC
Application routes are already global.
Which means that you cannot have two applications with the same name, even in different orgs.

jimmykarily: true. I'm not sure if we have any friendly validation for this though. It probably allows the user to push with a non-unique name and fails miserably later, not sure.

andreas-kupries: Note however, this apparently generates the name of a kube resource, i.e. of a `PVC`.
Might need a prefix. Will, if an org name like `12monkeys` is allowed.
Certainly needs `names.GenerateResourceName` from #700 to ensure the length limit.

internal/api/v1/models/app.go
```diff
@@ -62,6 +64,10 @@ func (ar *AppRef) EnvSecret() string {
 	return ar.Name + "-env"
 }
 
+func (ar *AppRef) PVCName() string {
+	return fmt.Sprintf("%s-%s", ar.Org, ar.Name)
```
andreas-kupries, line 68: This apparently generates the name of a kube resource, i.e. of a `PVC`.
Might need a prefix. Will, if an org name like `12monkeys` is allowed.
See #690 for analogous issues.
Certainly needs `names.GenerateResourceName` from #700 to ensure the length limit.
Note, the test's `k get pvc ...` has to be modified to match.

internal/api/v1/stage.go
```diff
@@ -51,6 +52,35 @@ func (app *stageParam) ImageURL(registryURL string) string {
...
+func (c ApplicationsController) ensurePVC(ctx context.Context, cluster *kubernetes.Cluster, pvcName string) error {
+	_, err := cluster.Kubectl.CoreV1().PersistentVolumeClaims(deployments.TektonStagingNamespace).Get(ctx, pvcName, metav1.GetOptions{})
+	if err != nil && !apierrors.IsNotFound(err) { // Unknown error, irrelevant to non-existence
+		return err
+	}
+	if err == nil { // pvc already exists
+		return nil
+	}
+
+	// From here on, only if the PVC is missing
```
andreas-kupries, line 64:
```suggestion
	// This is reached only when the expected PVC is missing
```

internal/api/v1/stage.go
```diff
@@ -148,6 +178,11 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 		BuilderImage: req.BuilderImage,
 	}
 
+	err = hc.ensurePVC(ctx, cluster, req.App.PVCName())
+	if err != nil {
+		return InternalError(err, "failed to ensure a PersistenVolumeClaim for the application source and cache")
```
andreas-kupries, line 183:
```suggestion
		return InternalError(err, "failed to ensure a PersistentVolumeClaim for the application source and cache")
```

andreas-kupries (changes requested):
Changes around the pvc resource name, guided by #690 and #700.
Plus typo and phrasing suggestions.

internal/api/v1/models/app.go
jimmykarily: `12monkeys` namespace (org) is allowed and so does the `12monkeys.sample` PVC. I used `names.GenerateResourceName` helper (https://github.com/epinio/epinio/pull/701/commits/4d8e8b0bb5bf2db53ffa6e6d15bc493e652bad63)

andreas-kupries: Ok

andreas-kupries (approved):

acceptance/apps_test.go
```diff
@@ -221,11 +223,13 @@ var _ = Describe("Apps", func() {
...
+				out, err := helpers.Kubectl("get", "pvc", "-n",
+					deployments.TektonStagingNamespace, names.GenerateResourceName(org, appName))
 				Expect(err).ToNot(HaveOccurred(), out)
 				env.DeleteApp(appName)
 
-				out, err = helpers.Kubectl("get", "pvc", "-n", deployments.TektonStagingNamespace, fmt.Sprintf("%s-%s", org, appName))
+				out, err = helpers.Kubectl("get", "pvc", "-n",
+					deployments.TektonStagingNamespace, names.GenerateResourceName(org, appName))
 				Expect(err).To(HaveOccurred(), out)
 				Expect(out).To(MatchRegexp(fmt.Sprintf(`persistentvolumeclaims "%s-%s" not found`, org, appName)))
```
andreas-kupries, line 234: Here is the `-` which should be a `.`

andreas-kupries (approved):

PR #703, opened 2021-08-19
Redirect all http traffic to https using traffic

### LINKED ISSUE #661: Configure http -> https redirection for all Traefik ingresses
Right now, all our ingresses are configured with TLS, but when you try to connect using plain http, you get a 404. It seems, Traefik can be configured to redirect to https automatically with something like:

```
--entrypoints.web.http.redirections.entryPoint.to=websecure
--entrypoints.web.http.redirections.entryPoint.scheme=https
--entrypoints.web.http.redirections.entrypoint.permanent=true
```

We should do that for all the ingresses in Epinio (epinio api, applications etc).

Links:

- https://github.com/traefik/traefik/tree/v2.4/pkg/provider/kubernetes/ingress/fixtures
- https://doc.traefik.io/traefik/routing/routers/#tls

jimmykarily:
Fixes #661

This will only work for the Traefik Epinio installs. If the user decides to use their previously deployed traefik (with `epinio install --skip-traefik`), we have no control on how that is run so we can't setup automatic redirection.

### DIFF
--- a/acceptance/epinio_installation_test.go
+++ b/acceptance/epinio_installation_test.go
@@ -1,6 +1,8 @@
 package acceptance_test
 
 import (
+	"fmt"
+	"net/http"
 	"strings"
 
 	"github.com/epinio/epinio/helpers"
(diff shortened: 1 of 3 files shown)


jimmykarily:
Maybe it's possible to add a test for this.

kkaempf:
> This will only work for the Traefik Epinio installs. If the user decides to use their previously deployed traefik (with `epinio install --skip-traefik`), we have no control on how that is run so we can't setup automatic redirection.

This PR should be accompanied with a doc fix describing this (and how to fix it in an already installed traefik)

jimmykarily:
> > This will only work for the Traefik Epinio installs. If the user decides to use their previously deployed traefik (with `epinio install --skip-traefik`), we have no control on how that is run so we can't setup automatic redirection.
> 
> This PR should be accompanied with a doc fix describing this (and how to fix it in an already installed traefik)

done

acceptance/epinio_installation_test.go
```diff
@@ -28,4 +30,28 @@ var _ = Describe("Epinio Installation", func() {
...
+		client.CheckRedirect = func(req *http.Request, via []*http.Request) error {
+			return http.ErrUseLastResponse
+		}
+
+		r, err := client.Do(request)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(r.StatusCode).To(Equal(http.StatusMovedPermanently))
+		location, err := r.Location()
+		Expect(err).ToNot(HaveOccurred())
+		Expect(location.String()).To(Equal(fmt.Sprintf("%s/info", serverURL)))
```
andreas-kupries, line 54: The `fmt.Sprintf("%s/info", serverURL)` is used twice (here and in line 37).
Maybe make it a variable at the top and share?

andreas-kupries (changes requested):
Just a nit about DRY.

acceptance/epinio_installation_test.go
jimmykarily: done (amended)

andreas-kupries: I do not see the variable. Only that the PR was rebased to newer main.

andreas-kupries (changes requested):

andreas-kupries (approved):

PR #708, opened 2021-08-20
Remove direct kube access from client commands.

### LINKED ISSUE #620: Epinio cli should work without kube access

andreas-kupries:
Fixes #620

### DIFF
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -168,6 +168,62 @@ func (hc ApplicationsController) Show(w http.ResponseWriter, r *http.Request) AP
 	return nil
 }
 
+func (hc ApplicationsController) ServiceApps(w http.ResponseWriter, r *http.Request) APIErrors {
+	ctx := r.Context()
+	params := httprouter.ParamsFromContext(ctx)
+	org := params.ByName("org")
+
+	cluster, err := kubernetes.GetCluster(ctx)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	exists, err := organizations.Exists(ctx, cluster, org)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	if !exists {
+		return OrgIsNotKnown(org)
+	}
+
+	var appsOf = map[string]models.AppList{}
+
+	apps, err := application.List(ctx, cluster, org)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	for _, app := range apps {
+		w := application.NewWorkload(cluster, app.AppRef())
+		bound, err := w.Services(ctx)
+		if err != nil {
+			return InternalError(err)
+		}
+		for _, bonded := range bound {
+			bname := bonded.Name()
+			if theapps, found := appsOf[bname]; found {
+				appsOf[bname] = append(theapps, app)
+			} else {
+				appsOf[bname] = models.AppList{app}
+			}
+		}
+	}
+
+	js, err := json.Marshal(appsOf)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	_, err = w.Write(js)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	return nil
+}
+
 func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request) APIErrors {
 	ctx := r.Context()
 	params := httprouter.ParamsFromContext(ctx)
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -14,6 +14,7 @@ import (
 	"github.com/epinio/epinio/internal/api/v1/models"
 	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/cli/clients/gitea"
+	"github.com/epinio/epinio/internal/duration"
 	"github.com/epinio/epinio/internal/organizations"
 	"github.com/gorilla/websocket"
 	"github.com/julienschmidt/httprouter"
@@ -233,6 +234,54 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request) 
 
 	return nil
 }
+
+func (hc ApplicationsController) Running(w http.ResponseWriter, r *http.Request) APIErrors {
+	ctx := r.Context()
+	params := httprouter.ParamsFromContext(ctx)
+	org := params.ByName("org")
+	appName := params.ByName("app")
+
+	cluster, err := kubernetes.GetCluster(ctx)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	exists, err := organizations.Exists(ctx, cluster, org)
+	if err != nil {
+		return InternalError(err)
+	}
+
+	if !exists {
+		return OrgIsNotKnown(org)
+	}
+
+	exists, err = application.Exists(ctx, cluster, models.NewAppRef(appName, org))
+	if err != nil {
+		return InternalError(err)
+	}
+
+	if !exists {
+		return AppIsNotKnown(appName)
+	}
+
+	app, err := application.Lookup(ctx, cluster, org, appName)
+	if err != nil {
+		return InternalError(err)
+	}
+	if app == nil {
+		// While app exists it has no workload
+		return NewAPIError("No status available for application without workload", "", http.StatusBadRequest)
+	}
+
+	err = cluster.WaitForDeploymentCompleted(
+		ctx, nil, org, appName, duration.ToAppBuilt())
+	if err != nil {
+		return InternalError(err)
+	}
+
+	return nil
+}
+
 func (hc ApplicationsController) Logs(w http.ResponseWriter, r *http.Request) {
 	ctx := r.Context()
 	params := httprouter.ParamsFromContext(ctx)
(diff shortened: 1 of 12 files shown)


internal/cli/clients/push.go
```diff
@@ -131,53 +125,25 @@ func (c *EpinioClient) deployCode(req models.DeployRequest) ([]byte, error) {
...
-				}
-				// it worked
-				if pr.Status.CompletionTime != nil {
-					return true, nil
-				}
-			}
-			// pr exists, but still running
-			return false, nil
-		})
+	_, err := c.get(api.Routes.Path("StagingComplete", app.Org, id))
```
jimmykarily, line 140: I'm wondering if the request timeout is long enough for the server side timeout (ToAppBuilt) to be effective. I mean, we don't set any timeout here but that doesn't mean the cli won't timeout before the server does. If that is true, we no longer control how long the cli waits for staging to complete. Maybe we need to pass something to the http client to achieve that.

internal/cli/clients/push.go
```diff
@@ -131,53 +125,25 @@ func (c *EpinioClient) deployCode(req models.DeployRequest) ([]byte, error) {
...
 
-	err := c.Cluster.WaitForDeploymentCompleted(
-		ctx,
-		c.ui, app.Org, app.Name, duration.ToAppBuilt())
-	if err != nil {
-		return errors.Wrap(err, "waiting for app to come online failed")
-	}
-
-	return nil
+	_, err := c.get(api.Routes.Path("AppRunning", app.Org, app.Name))
```
jimmykarily, line 147: same here

internal/domain/domain.go
```diff
@@ -17,7 +17,7 @@ func MainDomain(ctx context.Context) (string, error) {
 		return mainDomain, nil
 	}
 
-	cluster, err := kubernetes.GetCluster(ctx)
+	cluster, err := kubernetes.GetCluster(ctx) // Push -> MainDomain -> here (client-side! FIXIT?)
```
jimmykarily, line 20: fixed already?

andreas-kupries: Yes. Should remove that.

internal/cli/clients/push.go
andreas-kupries: Hm. Will check. If true will put in some more thought.

andreas-kupries: The server side timeout are handled at

- https://github.com/epinio/epinio/blob/fc07dc09a1603908ac1351925e54a83bc62145b6/internal/api/v1/stage.go#L260
- https://github.com/epinio/epinio/blob/fc07dc09a1603908ac1351925e54a83bc62145b6/internal/api/v1/applications.go#L333

It is (default) 10 minutes.

Where is the client timeout handled ... Oh, wait, you are saying above that the client does not have a timeout here ...
While I could see a browser timing out, why do you believe the client would despite not having set one ?
AFAICT from the net/http docs the default timeout is `no timeout`.

internal/domain/domain.go
andreas-kupries: Addressed by commit a867f2312964f497cb812b873512d16b475c3c4a

internal/api/v1/router.go
```diff
@@ -67,16 +67,18 @@ func patch(path string, h http.HandlerFunc) routes.Route {
...
+	"AppShow":         get("/orgs/:org/applications/:app", errorHandler(ApplicationsController{}.Show)),
+	"AppLogs":         get("/orgs/:org/applications/:app/logs", ApplicationsController{}.Logs),
+	"StagingLogs":     get("/orgs/:org/staging/:stage_id/logs", ApplicationsController{}.Logs),
+	"StagingComplete": get("/orgs/:org/staging/:stage_id/complete", errorHandler(ApplicationsController{}.Staged)), // See stage.go
+	"AppDelete":       delete("/orgs/:org/applications/:app", errorHandler(ApplicationsController{}.Delete)),
+	"AppUpload":       post("/orgs/:org/applications/:app/store", errorHandler(ApplicationsController{}.Upload)), // See upload.go
+	"AppStage":        post("/orgs/:org/applications/:app/stage", errorHandler(ApplicationsController{}.Stage)),  // See stage.go
+	"AppDeploy":       post("/orgs/:org/applications/:app/deploy", errorHandler(ApplicationsController{}.Deploy)),
+	"AppUpdate":       patch("/orgs/:org/applications/:app", errorHandler(ApplicationsController{}.Update)),
+	"AppRunning":      get("/orgs/:org/applications/:app/running", errorHandler(ApplicationsController{}.Running)),
```
jimmykarily, line 81: We can add a test for this new endpoint

internal/api/v1/applications.go
```diff
@@ -167,6 +168,62 @@ func (hc ApplicationsController) Show(w http.ResponseWriter, r *http.Request) AP
 	return nil
 }
 
+func (hc ApplicationsController) ServiceApps(w http.ResponseWriter, r *http.Request) APIErrors {
```
jimmykarily, line 171: Can you add a comment that describes what this endpoint is about?

internal/api/v1/router.go
andreas-kupries: Will do.

internal/api/v1/applications.go
andreas-kupries: Sure.

internal/api/v1/applications.go
```diff
@@ -233,6 +290,54 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
 
 	return nil
 }
+
+func (hc ApplicationsController) Running(w http.ResponseWriter, r *http.Request) APIErrors {
```
jimmykarily, line 294: Can you add a comment on what this endpoint is all about? The response is not interesting so it should be used only to wait until and application is in running state. Let's document that a bit in comments.

internal/api/v1/deploy.go
```diff
@@ -361,3 +365,11 @@ func existingReplica(ctx context.Context, client *k8s.Clientset, app models.AppR
 	}
 	return *result.Spec.Replicas, nil
 }
+
+func appDefaultRoute(ctx context.Context, name string) (string, error) {
```
jimmykarily, line 369: This can be moved to the domain package I think.

internal/cli/clients/client.go
```diff
@@ -1725,6 +1700,10 @@ func uniqueStrings(stringSlice []string) []string {
 }
 
 func getAPI(ctx context.Context, log logr.Logger) (string, string, error) {
+	// This is called only by the admin command `config update`
+	// which has to talk to the cluster to retrieve the
+	// information. This is allowed.
```
jimmykarily, line 1705: maybe it would be safer to have a different "admin client" class that we can use when kube access might be needed. This way it will be harder to accidentally introduce commands that access the kube api when they shouldn't. I think we can rely on code reviews to avoid that for now but if we fail we should think about extracting the functionality to another client.

internal/api/v1/applications.go
andreas-kupries: Done. Commit 1c96529a8810d7ed6c73dac70ab1d8498f212db6

internal/api/v1/applications.go
andreas-kupries: Done. Commit 1c96529a8810d7ed6c73dac70ab1d8498f212db6

internal/cli/clients/client.go
andreas-kupries: Agreed, while also out of scope for now. Will make a ticket for this tech debt.

internal/cli/clients/push.go
jimmykarily: ok then

internal/api/v1/deploy.go
andreas-kupries: will do

jimmykarily (approved):

internal/cli/clients/client.go
andreas-kupries: See #719

internal/api/v1/deploy.go
andreas-kupries: Done. Commit b2a3b244de2d247b927ef74d1041a301a88498c3

internal/api/v1/router.go
andreas-kupries: Commit 61c5559eb239444f454779ec09b5535a44f6bcd2 extends an existing test (app deploy) to not just deploy, but also invoke the new endpoint.

jimmykarily (approved):

PR #711, opened 2021-08-20
Add github action for AWS EKS

### LINKED ISSUE #687: Install tests: setup persistent AWS cluster for tests.
(tests will epinio uninstall to remove resources)

In order to create the tests for `load-balancer-ip` and in order to test on AWS (https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)) we need to have access to a permanent cluster.

# Acceptance Criteria

- There is a way (script or docs) to create the cluster on AWS
- One the install tests (new or existing) on that cluster
- The cluster is cleaned up and re-used every time
- There is no way to run 2 jobs at the same time on the same cluster (each one messing the other job's results).

Consider these solutions for the first item above:

- https://github.com/suse/catapult
- other? (@kkaempf)
- custom script

rohitsakala:
fixes #687

This github action will create and delete an eks cluster at 6 am every day.

A successful run of this build can be found here - https://github.com/epinio/epinio/runs/3380893247

### DIFF
--- a/.github/workflows/eks.yml
+++ b/.github/workflows/eks.yml
@@ -1,0 +1,46 @@
+name: EKS-CI
+
+on:
+  schedule:
+    - cron:  '0 6 * * *'
+
+env:
+  SETUP_GO_VERSION: '^1.13.7'
+
+jobs:
+  acceptance-cli:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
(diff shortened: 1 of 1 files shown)


jimmykarily:
Since this is not running any tests let's not merge it. We can continue on the same branch and merge when we implement a test case from here: https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)

PR #718, opened 2021-08-23
feat: provide user information to the api server

### LINKED ISSUE #706: Epinio knows who the user is (basic auth)
Followup of #486 and #686

As a first step multi user (with basic auth), we want to let the Epinio server know, who the current user is. This is possible by forwarding some headers from the Traefik basic auth middleware.

https://doc.traefik.io/traefik/middlewares/http/basicauth/#headerfield

## Acceptance Criteria

- There is a header that lets epinio server know who the user is
- The epinio server prints every request and the user that made it in the default log lever (TRACE_LEVEL=0) without verbosity being on. This acts as a log of user actions.
- It's documented how to add a User to Epinio (by updating the secret).

andreas-kupries:
Fixes #706 

- traefik config extended to insert user as request header seen by the server
- server extended to demo arrival and presence of user information in requests.
- documented process of adding a new user able to access the api


internal/cli/server.go
```diff
@@ -121,7 +121,8 @@ func logRequestHandler(h http.Handler, logger logr.Logger) http.Handler {
 func logRequest(r *http.Request, log logr.Logger) {
 	uri := r.URL.String()
 	method := r.Method
-	log.V(1).Info("received request", "method", method, "uri", uri)
+	log.V(1).Info("received request",
```
jimmykarily, line 124: iirc the value here is not the absolute target TRACE_LEVEL but increases the current trace level. This results in logs appearing only when `TRACE_LEVEL` is set to `2`. This makes `TRACE_LEVEL=1` useless and boring since nothing happens there. I think we should even show the requests in `TRACE_LEVEL=0`, if not then at least in `TRACE_LEVEL=1`.

There is a problem with the liveness probe of the server pod though, which is set to the `api/v1/info` endpoint. This results in request logs like this appearing in regular intervals:

```
2021/08/24 09:54:42 epinio/5089 "level"=2 "msg"="received request"  "method"="GET" "uri"="/api/v1/info" "user"=""  
```

This is annoying and confusing to the user and we should find a way to filter it. Since that endpoint is useful in general, we don't want to completely filter it out in logs. We should better create a special endpoint for liveness/readiness probes that can be filtered out in logs. Then we can show requests in `TRACE_LEVEL=0` I think.

Feel free to do that here or we can create a separate issue for that.

jimmykarily: According [to the docs](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#when-should-you-use-a-liveness-probe) we don't even need a liveness probe. Epinio server knows how to die when something goes wrong. 
We still need the readiness probe though.

assets/embedded-files/epinio/basicauth.yaml
```diff
@@ -7,3 +7,4 @@ metadata:
 spec:
   basicAuth:
     secret: epinio-api-auth-secret
+    headerField: X-WebAuth-User
```
jimmykarily, line 10: It seems that if we ever want to support another middleware (e.g. forwardauth) we would need to check for some other header instead: https://doc.traefik.io/traefik/v2.0/middlewares/forwardauth/#authresponseheaders

Anyway, let's think about it then.

docs/developer/howtos/README.md
```diff
@@ -0,0 +1,16 @@
...
+  - [How to work with Epinio during Development](development.md)
+  - [How to add a user for API access](new-api-user.md)
+  - [What happens inside when deploying a development binary](behind-the-curtains.md)
+
+[comment1]: # Comment is done by abusing link label syntax (without a link referencing it)
+[comment2]: # .
+[comment3]: # One of the things is not like the others. IOW misclassified, I believe.
+[comment4]: # See also issue https://github.com/epinio/epinio/issues/698#issuecomment-900893447
+[comment5]: # where it is argued that the behind the curtains text should be folded into the
+[comment6]: # development document.
```
jimmykarily, line 16: I don't understand this section. What is this about? Also it doesn't render well (I clicked on the three-dots menu -> view file).

internal/cli/server.go
andreas-kupries: Will make a new endpoint, and then reduce level.

assets/embedded-files/epinio/basicauth.yaml
andreas-kupries: Agreed.

docs/developer/howtos/README.md
andreas-kupries: It is an inline markdown comment. As a comment it should not render at all.
That said, I better remove it and make a ticket out of it.

andreas-kupries: Ticket is #721 . Removal will be committed when the new probe endpoint is done.

internal/cli/server.go
andreas-kupries: Commits
- 8458f6aab0224a3d9d31f8754336e245a3d2fea5
- 67eee30fbb4dd49456df78acfee9d0ccbd444003

andreas-kupries: Requests are now logged for TRACE_LEVEL=1 and higher.

jimmykarily (approved):

jimmykarily (approved):

PR #553, opened 2021-06-28
Add create app to CLI and fix delete

manno:
With the new apps create command, it was also necessary to fix apps delete, so it can delete apps that don't have a workload.

fixes #515

PR #720, opened 2021-08-23
Delete stale serviceBindings while deleting services

### LINKED ISSUE #630: Services can't be deleted if previously bound to now deleted app
While working on a test for Rails (https://github.com/epinio/epinio/pull/628), I created an application to which I bound a service. In the `AfterEach` block, I first deleted the application and then the service. The test failed because the service was not deleted (https://github.com/epinio/epinio/pull/628/checks?check_run_id=3112534175)

This is happening because when we delete an application, we don't unbind the services, we just delete the application (somewhere around here: https://github.com/epinio/epinio/blob/main/internal/api/v1/applications.go#L434). Then, when we delete the Service, if there are no bound apps (which is true since we've now deleted the app), we don't delete any `ServiceBindings` . (which is normally happening here: https://github.com/epinio/epinio/blob/main/internal/api/v1/services.go#L365).

`ServiceBindings` don't have an owener reference and they are not automatically deleted when we delete a ServiceInstance. When we try to delete the epinio service, the ServiceInstance is not deleted and fails as this:

```
~/workspace/epinio (441-create-rails-acceptance-test)*$ kubectl get serviceinstances -A
NAMESPACE        NAME                                               CLASS                         PLAN      STATUS                                    AGE
orgs-625899451   service.org-orgs-625899451.svc-service-310176721   ClusterServiceClass/mariadb   10-3-22   DeprovisionBlockedByExistingCredentials   16m
```
The reason is that there is still a ServiceBinding around:

```
~/workspace/epinio (441-create-rails-acceptance-test)*$ kubectl get servicebindings -A
NAMESPACE        NAME                                                                  SERVICE-INSTANCE                                   SECRET-NAME                                                           STATUS   AGE
orgs-625899451   service.org-orgs-625899451.svc-service-310176721.app-apps-625897000   service.org-orgs-625899451.svc-service-310176721   service.org-orgs-625899451.svc-service-310176721.app-apps-625897000   Ready    6m14s
```

rohitsakala:
fixes #630 

TODO: Add tests


internal/api/v1/services.go
```diff
@@ -369,7 +369,12 @@ func (sc ServicesController) Delete(w http.ResponseWriter, r *http.Request) APIE
 		}
 	}
 
-	// Everything looks to be ok. Delete.
+	// Delete all service bindings. There can be service
+	// bindings whose apps are already deleted.
+	err = service.DeleteBindings(ctx)
```
andreas-kupries, line 374: What happens if the app is __not__ deleted ? This looks as if it pulls the rug from under such an application.

internal/application/application.go
```diff
@@ -209,6 +210,25 @@ func Delete(ctx context.Context, cluster *kubernetes.Cluster, gitea GiteaInterfa
 		return err
 	}
 
+	// get bounded services
+	app, err := Lookup(ctx, cluster, appRef.Org, appRef.Name)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return err
+	}
+
+	// delete all service bindings
```
andreas-kupries, line 219: Note that we have `Bind` and `Unbind` functions for adding and removing bindings between apps and services.
See `internal/application/workload.go`.
There is even `UnbindAll` which dissolves all bindings from an app. and it is not used, there is just the definition ?!

And I thought that an `app delete` would remove the bindings already, i.e. invoke `UnbindAll` ...
But that is the reported error, right ?

I believe I will check the history for

  1. Where UnbindAll` was added, and if it was used at the time.
  2. If it was used when it was added, where that use was removed.

@jimmykarily Ping also about this.

andreas-kupries: See 7d16af874edcd56dec66bd5ea9ba87a78147dfe0 where UnbindAll was introduced, and used.
See the following commit df9feadb287ba34417a9923986b54ff1699e728b where it got dropped.

I suspect that part of the df9fea commit has to be reverted to fix our problem.

@manno Ping.

andreas-kupries (changes requested):
Checking the history I suspect that most of the current work is superflous in the face of the existing unused `UnbindAll` function which does the removal of service bindings on app deletion and was dropped at some point.

internal/application/application.go
andreas-kupries: Relevant PR holding these commits is #553

andreas-kupries: And we have a test for this situation: `It("unbinds bound services when deleting an app", func() {`. (See apps_test.go)

Why did that keep passing ? ... Oh, it checks only the delete output for reporting on the service, not that the binding is properly gone. And the reporting of what services are bound to the deleted app, that was not affected

internal/api/v1/services.go
rohitsakala: Removed this part of the code.

internal/application/application.go
rohitsakala: You are right. Changed code to use it.

acceptance/apps_test.go
```diff
@@ -384,11 +384,11 @@ var _ = Describe("Apps", func() {
 			})
 		})
 
-		It("unbinds bound services when deleting an app", func() {
+		It("unbinds bound services when deleting an app and then delete's service", func() {
```
andreas-kupries, line 387:
```suggestion
		It("unbinds bound services when deleting an app, and then deletes the service", func() {
```

acceptance/apps_test.go
```diff
@@ -404,6 +404,14 @@ var _ = Describe("Apps", func() {
 				Expect(err).ToNot(HaveOccurred(), out)
 				return out
 			}, "1m").ShouldNot(MatchRegexp(`.*%s.*`, appName))
+
+			By("checking if servicebinding is deleted")
```
andreas-kupries, line 408:
```suggestion
			By("checking if the servicebinding is deleted")
```

acceptance/apps_test.go
```diff
@@ -404,6 +404,14 @@ var _ = Describe("Apps", func() {
...
+			serviceBindings, err := helpers.Kubectl("get", "servicebindings",
+				"--namespace", org)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(serviceBindings).NotTo(ContainSubstring(fmt.Sprintf("svc.org-%s.svc-%s.app-%s", org, serviceName, appName)))
```
andreas-kupries, line 412: The tests have a `helper/catalog/names.go` file (`catalog` package). Might be sensible to put the `Sprintf` into a function there to hide the details of the construction here.

internal/application/application.go
```diff
@@ -209,6 +209,21 @@ func Delete(ctx context.Context, cluster *kubernetes.Cluster, gitea GiteaInterfa
 		return err
 	}
 
+	// get bounded services
```
andreas-kupries, line 212:
```suggestion
	// get application workload, if any, and its bounded services, if any
```

andreas-kupries (changes requested):
Only a few small things

acceptance/apps_test.go
rohitsakala: Done.

andreas-kupries (approved):

PR #726, opened 2021-08-26
chore: extend the code documentation

andreas-kupries:
Extend the documentation we have for our code, in the code's comments.
Look into generating documentation from these comments.

### DIFF
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -235,7 +235,7 @@ func (hc ApplicationsController) ServiceApps(w http.ResponseWriter, r *http.Requ
 	return nil
 }
 
-// Update handles the API endpoint /orgs/:org/applications/:app (POST)
+// Update handles the API endpoint /orgs/:org/applications/:app (PATCH)
 // It modifies the specified application. Currently this is only the
 // number of instances to run.
 func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request) APIErrors {
--- a/internal/api/v1/applications.go
+++ b/internal/api/v1/applications.go
@@ -23,11 +23,13 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+// ApplicationsController represents all functionality of the API related to applications
 type ApplicationsController struct {
 	conn *websocket.Conn
 }
 
-// Create creates a new and empty application. I.e. without a workload.
+// Create handles the API endpoint /orgs/:org/applications (POST)
+// It creates a new and empty application. I.e. without a workload.
 func (hc ApplicationsController) Create(w http.ResponseWriter, r *http.Request) APIErrors {
 	ctx := r.Context()
 	params := httprouter.ParamsFromContext(ctx)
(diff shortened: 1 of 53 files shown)


internal/api/v1/applications.go
```diff
@@ -230,8 +235,9 @@ func (hc ApplicationsController) ServiceApps(w http.ResponseWriter, r *http.Requ
 	return nil
 }
 
-// Update modifies the specified application. currently this is only
-// the number of instances to run.
+// Update handles the API endpoint /orgs/:org/applications/:app (POST)
```
rohitsakala, line 238: Is Update a PATCH operation ?

internal/api/v1/applications.go
```diff
@@ -75,7 +77,8 @@ func (hc ApplicationsController) Create(w http.ResponseWriter, r *http.Request)
 	return nil
 }
 
-// Index lists all the known applications, with and without workload.
```
rohitsakala, line 78: Some comments have operations like GET POST mentioned but some are not.

internal/api/v1/models/app.go
```diff
@@ -30,18 +30,25 @@ func (a *App) AppRef() AppRef {
 	return NewAppRef(a.Name, a.Organization)
 }
 
+// AppList is just a collection of app references
```
rohitsakala, line 33: AppList is a collection is better ?

internal/auth/certs.go
```diff
@@ -1,5 +1,3 @@
-// Package auth collects structures and functions around the
```
rohitsakala, line 1: Is this not required ?

rohitsakala (changes requested):

internal/api/v1/applications.go
```diff
@@ -23,11 +23,13 @@ import (
...
 )
 
+// ApplicationsController represents all functionality of the API related to applications
 type ApplicationsController struct {
 	conn *websocket.Conn
 }
 
-// Create creates a new and empty application. I.e. without a workload.
+// Create handles the API endpoint /orgs/:org/applications (POST)
+// It creates a new and empty application. I.e. without a workload.
```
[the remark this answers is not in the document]
andreas-kupries: I started a new sentence.

internal/api/v1/applications.go
andreas-kupries: Huh. Yes.
```suggestion
// Update handles the API endpoint /orgs/:org/applications/:app (PATCH)
```

internal/api/v1/applications.go
andreas-kupries: Yes. Added the information only when I noticed same url, different http methods. Can be consolidated to always have the method.

internal/api/v1/models/app.go
andreas-kupries:
```suggestion
// AppList is a collection of app references
```

internal/auth/certs.go
andreas-kupries: No. The description is at the top of `auth.go`. Repeating it here causes the docs generator to repeat it as well in the output.

PR #729, opened 2021-08-27
Add test scenario for AKS cluster

### LINKED ISSUE #702: Install tests: when we have external clusters, set up windows and macos test runners

manno:
Adds AKS cluster creation and the test scenario from https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)#test-scenarios

* AKS loadbalancer status returns an IP, not a hostname like on EKS
* AKS docs suggested split `install-ingress` is also needed, it's not part of the table though and I'm unsure if it's required.

fixes #702


.github/workflows/aks.yml
```diff
@@ -36,38 +45,46 @@ jobs:
           go-version: ${{ env.SETUP_GO_VERSION }}
       - name: Setup Ginkgo Test Framework
         run: go install github.com/onsi/ginkgo/ginkgo@v1.16.2
-      - name: Install Dependencies
-        run: |
-          choco install kubernetes-cli eksctl
 
-      - name: Configure AWS credentials
+      - name: Configure AWS credentials for Route53
```
andreas-kupries, line 49: Question: So while this is an MS/Azure Cluster, the DNS is done via Amazon ?!

.github/workflows/aks.yml
```diff
@@ -36,38 +45,46 @@ jobs:
...
-          export KUBECONFIG=$PWD/kubeconfig-epinio-ci
+          export KUBECONFIG=$PWD/kubeconfig
           make test-acceptance-install
 
-      - name: Delete EKS cluster
-        if: ${{ always() }}
-        env:
-          KUBECONFIG: kubeconfig-epinio-ci
+      - name: Delete AKS cluster
+        if: ${{ always() && !github.event.inputs.keep_cluster }}
```
andreas-kupries, line 86: What is the `always()` needed for ? IIUC correctly it is always true, and thus superfluous in the condition ?!

andreas-kupries (changes requested):
Just questions. Well, it implies a request to document things around the questions a bit more. ;)

.github/workflows/aks.yml
svollath: Guess that's fair for the test env - no need to test az-cli for DNS resources, when we implemented route53 in the first place. We don't provide code or even (provider-)specific docs on DNS setup - so thers's also nothing which could fail for us with Azure DNS. We do only need the "System-DOMAIN" and "Traefik-Ext-IP" - if there's something wrong with that, it would also fail with route53.

manno: The Route53 code was already there and we have our zones on AWS. Zone updates/DNS setup are not part of the test scenario, so I use the same zone.

.github/workflows/aks.yml
manno: I'm not sure if that's needed, but didn't invest another thirty minutes to find out. 

You need the `always()`, when you have no other condition, so that the job runs even for failed tests, i.e. for cleanup. I think it's save to drop it, as we still evaluate the `github.event.inputs.keep_cluster` expression. But expressions in Github Actions surprised me multiple times, so I feel we should test that change, because we don't want to keep clusters running.

`github.event.inputs.keep_cluster` is null for a scheduled run, and by default the empty string for a manually scheduled run. For a string null and empty string should both evaluate to false, according to the docs.

.github/workflows/aks.yml
andreas-kupries: Please note that my question is not to imply that I am against this. I was simply confused to see Amazon things in a flow for MS/Azure. A small note in a comment that this mix is intentional would be enough to unconfuse future me reading it again. And hopefully other maintainers.

.github/workflows/aks.yml
andreas-kupries: I am ok with keeping it. Just confused. Maybe a quick note summarizing your explanation in the flow itself ?

.github/workflows/aks.yml
manno: This https://github.com/epinio/epinio/pull/729/files#diff-efbe2e406ecf70a5533ff28e4d18451538484832f80130758de3a649d8d03dbbR35 was not enough I guess ;)

Ok, will comment the workflow.

.github/workflows/aks.yml
manno: This one is needed: https://github.com/epinio/epinio/blob/main/.github/workflows/eks.yml#L66

.github/workflows/aks.yml
andreas-kupries: Different location I only became aware of later, when reaching it in the PR. I prefer local explanations. Also ok are references to other relevant places and note of explanation there, although in second place after local explanations.

.github/workflows/aks.yml
andreas-kupries: Agreed, based on your explanation above.

andreas-kupries (approved):

PR #731, opened 2021-08-30
Folded `behind the curtains` text into the single howto referencing it

### LINKED ISSUE #721: Behind the Curtains Document is not a How To
Document should be moved to explanation at least. Alternatively note that https://github.com/epinio/epinio/issues/698#issuecomment-900893447 argues that its content should be folded into the development howto.

andreas-kupries:
Fixes #721


kkaempf (approved):
lgtm

Issue #724, opened 2021-08-25
Staging logs don't always show up


agracey:
I'm seeing an intermittent issue where the staging logs don't always show up. It seems to be pretty random and I'm betting it's a network issue or something. 

The next time I see it, I'll see if they can be retrieved afterwards with `epinio logs --staging` and update this ticket

andreas-kupries:
Could it be a race also ?
A quickly started app with not much of logs, faster than the client trying to retrieve them ?
What application was used ?

Do you have captures of the output with the logs missing ?
(Extended TRACE_LEVEL ? (TRACE_LEVEL=6 will capture all tracing we currently have))

Given your suspicion of network issues ... Cluster with epinio is not on the same box as the client ?

andreas-kupries:
Looking a bit a logs for a run showing the staging logs I noticed

```
2021/08/30 11:16:14 epinio/631 "level"=1 "msg"="received request"  "method"="GET" "uri"="/api/v1/orgs/workspace/staging/b0e789281a9e2dc5/logs?follow=true\u0026stage_id=b0e789281a9e2dc5" "user"="a42607e9bc02c85b"
2021/08/30 11:16:14 epinio/631 "level"=2 "msg"="request"  "body"="n/a"

2021/08/30 11:16:21 log-tracing "level"=1 "msg"="starting the tail for pod b0e789281a9e2dc5-cleanup-cdpbq-pod-fvs5v"  
```

Note the 7 seconds between receiving the request and the first action which may generate user-visible data!
(The server log does not show the log entries streamed over the web socket)

I will dig a bit more into this area, see if I can determine what is done in that time.

andreas-kupries:
Moving back into the icebox until we have a repro with extended tracing

agracey:
I don't believe I've seen this since 0.1.1. I don't know if it's load related but going to close until I see it again and can repro.

PR #734, opened 2021-08-30
Mostly more tracing the log streaming facilities, plus bug fix for install

andreas-kupries:
Refers #724 . Not a fix, code for research.

### DIFF
--- a/helpers/kubernetes/tailer/setup.go
+++ b/helpers/kubernetes/tailer/setup.go
@@ -1,3 +1,5 @@
+// tailer manages objects which tail the logs of a collection of pods specified by a label selector.
+// This is similar to what the cli tool `stern` does.
 package tailer
 
 import (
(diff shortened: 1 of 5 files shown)


andreas-kupries:
@agracey ... Should not have pinged you as reviewer, only here. Hopefully this will help you getting better information out of your systems when it happens again. It requires TRACE_LEVEL=6 for client and server. This is also how I got the info about the kube watcher events having these large gaps between them. No idea how kube implements the watcher. Part of the problem is we watch for all pods with a particular label in __all__ namespaces. This may be something kube is not good at.

(Note: The server's trace level is copied from the trace_level `epinio install` runs under. That makes it easy to get a server with tracing active, just run `TRACE_LEVEL=nnn epinio install`.

kkaempf:
I can confirm that the fix in `internal/cli/clients/client.go` makes my minikube install succeed.

manno (approved):

PR #739, opened 2021-08-31
Org becomes Namespace

### LINKED ISSUE #544: We should rename orgs
The name organization likely gives the impression of harder multi-tenancy than we will want to offer (at least in the near- to mid-term).

I know we stayed away from calling them "spaces" because that can get confusing with the underlying "namespace" from k8s but that might be ok?

andreas-kupries:
fix #544

chore: rename org(s) to namespace(s)
  - Renamed cli command name
  - Renamed skip option
  - Updated API (as public as the cli command name)
  - Updated tests
  - Various error messages
  - Documentation

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -157,5 +157,5 @@ minikube-delete:
 	@./scripts/minikube-delete.sh
 
 install: build
-	EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1 ./dist/epinio-linux-amd64 install --skip-default-org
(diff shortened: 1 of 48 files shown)


manno (changes requested):
This serialized struct is also part of two API requests: https://github.com/epinio/epinio/blob/main/internal/api/v1/models/app.go#L59

andreas-kupries:
Fixed.

manno (approved):

PR #753, opened 2021-09-03
Improve user facing error messages

### LINKED ISSUE #586: Polish: Nicer looking message when an org is missing / not found
Epinio currently prints

> error listing apps: Not Found: {"errors":[{"status":404,"title":"Organization 'workspace' does not exist","details":""}]}

when a command needing an org run into a missing org (example `app list`).

This looks complicated, messy, is difficult to read, and conveys the feeling of an internal error, bug in epinio.
Should be something like

> Unable to XXX: Targeted organization "workspace" does not exist.

Likely have a lot of more issues reported in this bad manner. This is the one just seen.

thardeck:
Fixes #586

### DIFF
--- a/internal/api/v1/errors.go
+++ b/internal/api/v1/errors.go
@@ -114,7 +114,7 @@ func UserNotFound() APIError {
 // OrgIsNotKnown constructs an API error for when the desired org does not exist
 func OrgIsNotKnown(org string) APIError {
 	return NewAPIError(
-		fmt.Sprintf("Namespace '%s' does not exist", org),
+		fmt.Sprintf("Targeted namespace '%s' does not exist", org),
 		"",
 		http.StatusNotFound)
 }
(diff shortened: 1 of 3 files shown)


andreas-kupries (approved):

internal/api/v1/errors.go
```diff
@@ -114,7 +114,7 @@ func UserNotFound() APIError {
 // OrgIsNotKnown constructs an API error for when the desired org does not exist
 func OrgIsNotKnown(org string) APIError {
 	return NewAPIError(
-		fmt.Sprintf("Namespace '%s' does not exist", org),
+		fmt.Sprintf("Targeted namespace '%s' does not exist", org),
```
andreas-kupries, line 117: This requires matching updates in the testsuite. CI :boom: because of that.

andreas-kupries (changes requested):

andreas-kupries (approved):

PR #754, opened 2021-09-03
Fix parameter config for changelog generator action

andreas-kupries:


kkaempf (approved):
Give it a try, it can't get worse ;-)

PR #755, opened 2021-09-03
Fix second parameter which really should take a sequence and doesn't.

andreas-kupries:


kkaempf (approved):
github, you're drunk, go home. ;-)

Issue #757, opened 2021-09-03
Make CI (all Badges) green - and stable


svollath:
We released e.g. epinio 0.0.24, while most badges in the README are red. While general CI might have been green for the release, others were red.

In addition, I today failed with installing epinio (v0.0.23 and v0.0.24) on e.g. EKS.

manno:
CI reports this error: 

```
🚢 Create Custom Service
  Name: service-23503500
  Namespace: 
  | PARAMETER |  VALUE  |
  |-----------|---------|
  | mariadb   | 10-3-22 |
  error creating service: invalid character 'M' looking for beginning of value
```

By slightly modifying the error reporting, I can see the actual error: 

```
error creating service: cannot parse JSON response: Method Not Allowed
: invalid character 'M' looking for beginning of value
```

Trace logging shows the outgoing request from the CLI: `EpinioClient "level"=1 "msg"="POST https://epinio.aks.example.com/api/v1/namespaces//custom-services"`.

However the server's logs show: `"msg"="received request"  "method"="GET" "uri"="/api/v1/namespaces/custom-services"`.

How did the POST end up as a GET? We renamed the 'orgs' endpoints to 'namespace' and ended up with these rules:

```
	"OrgDelete": delete("/namespaces/:org", errorHandler(OrganizationsController{}.Delete)),
	"ServiceCreateCustom": post("/namespaces/:org/custom-services", errorHandler(ServicesController{}.CreateCustom)),
```

Now, the request has a double slash in `namespaces//custom-services`, because the test setup did not target a namespace. Our router probably matches the URL to `delete("/namespaces/custom-services")` and detects the method does not match.

manno:
I'll have to wait for the scheduled workflows.

Issue #758, opened 2021-09-04
Issue installing epinio with external ClusterIssuer certificates & custom dns - attempts to create workspaces with magicDNS


JamesTryand:
**Describe the bug**
for this i've changed my email to ```myemailaddress@gmail.com``` and the base domain to ```by.example.com``` from my actual one, and naturally i'm not including the secrets

I am running epinio within a non-externally facing system. I am using a Digital Ocean DNS01 lets encrypt ClusterIssuer 

However when I install epinio it fails when creating the workspace as it is attempting to use the omg.howdoi.website rather than the custom domain provided. 

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Install Epinio Certmanager with 
```epinio install-cert-manager --email-address myemailaddress@gmail.com```
2. Install Cluster Issuer  
```
kubectl apply -f do-secret.yaml
kubectl apply -f do-dns-live.yaml
```
with a do-dns-live.yaml that looks similar to : 
```
apiVersion: cert-manager.io/v1
kind: ClusterIssuer
metadata:
  name: do-dns-live
  namespace: cert-manager
spec:
  acme:
    email: myemailaddress@gmail.com
    server: https://acme-v02.api.letsencrypt.org/directory
    preferredChain: "ISRG Root X1"
    privateKeySecretRef:
      name: do-issuer-account-key-live
    solvers:
    - dns01:
        cnameStrategy: Follow
        digitalocean:
          tokenSecretRef:
            name: digitalocean-dns
            key: access-token
      selector:
        dnsZones:
          - 'by.example.com'
```
3. Then install the rest of epinio with : 
```
epinio install \
        --skip-cert-manager \
        --tls-issuer=do-dns-live \
        --system-domain=by.example.com \
        --email-address myemailaddress@gmail.com \
        --skip-ssl-verification \
        --trace-level 5 \
        --verbosity 1
```
4. At this point it _mostly_ works, however it reports an error at the creating workspace, and then although the ingress is available, it returns an error and blank page when i visit ```https://epinio.by.example.com``` 

**Expected behavior**
I would expect to be able to browse to a working ```https://epinio.by.example.com``` 

**Logs**
When installing I get the following log : 

<details>
  <summary>Click to expand<summary>  

```
2021/09/04 13:27:50 InstallClient/Install "level"=1 "msg"="start"

🚢 Epinio v0.0.24 installing...
2021/09/04 13:27:50 InstallClient/Install "level"=2 "msg"="process cli options"
[...]

PR #762, opened 2021-09-06
Fix AKS/EKS Install Tests

manno:
refers to #757

The tests didn't not target the namespace. This somehow (?) worked, until we renamed 'org' to 'namespace', which made the routes so similiar, that a "Method not supported" error was reported by the httprouter.

### DIFF
--- a/.github/workflows/aks.yml
+++ b/.github/workflows/aks.yml
@@ -7,12 +7,16 @@ on:
     - cron:  '0 5 * * *'
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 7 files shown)


andreas-kupries:
`Method not reported` is also an issue in #758. It uses 0.0.24, which has the rename.

andreas-kupries (approved):
It looks ok. I like the formatError changes reporting both json issues better, and reporting all the collected errors (when it is a multi-error)

PR #764, opened 2021-09-06
Fix linter warnings

manno:
This fixes some of the linter warnings. 

I create another PR to create a seperate workflow for the scheduled lint,  so it can drop e.g. errcheck to avoid warnings for old code. The other workflow, for PRs, will have more checks, for new code.

refers to #757

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -115,7 +115,7 @@ func (c *EpinioClient) EnvList(ctx context.Context, appName string) error {
 	return nil
 }
 
-// EnvUnset adds or modifies the specified environment variable in the
+// EnvSet adds or modifies the specified environment variable in the
 // named application, with the given value. A workload is restarted.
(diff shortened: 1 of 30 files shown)


internal/cli/clients/client.go
```diff
@@ -827,7 +824,6 @@ func (c *EpinioClient) Info() error {
 	c.ui.Success().
 		WithStringValue("Platform", platform).
 		WithStringValue("Kubernetes Version", kubeVersion).
-		WithStringValue("Gitea Version", giteaVersion).
```
andreas-kupries, line 830: This requires a matching change in the testsuite (before test setup - This part still looks for this string - CI :boom: because of that)

andreas-kupries (changes requested):
So many issues in our code which needed fixing.
Cannot approve yet as CI does not yet pass.
The issue is CI looking for the eliminated Gitea version information as part of suite setup.

andreas-kupries (approved):

PR #765, opened 2021-09-07
Fix golangci-lint workflow

manno:
The workflow didn't work previously, because the generated files were
missing.

This PR adds detection for schedule vs. pull_request events.
In schedule mode, we use a relaxed config, to avoid to many warnings.
In PR mode the config is stricter, so new code lints.
The PR mode config is also in the default config location and should be
picked up by IDEs.

refers to #757

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -14,14 +14,28 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+
+      - name: Generate Golang
+        run: |
+          export PATH=$PATH:/home/runner/go/bin/
+          make embed_files
+
+      - id: configfile
+        run: |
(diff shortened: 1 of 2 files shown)


.github/workflows/golangci-lint.yml
```diff
@@ -14,14 +14,28 @@ jobs:
...
+          else
+            echo "::set-output name=file::.golangci-schedule.yml"
+          fi
+
       - name: golangci-lint
         uses: golangci/golangci-lint-action@v2
         with:
           # Required: the version of golangci-lint is required and must be specified without patch version: we always use the latest patch version.
-          version: v1.39
+          version: v1.42.1
```
andreas-kupries, line 35: Does this specify the patch version (trailing `.1`) ? Contrary to the comment ?

manno: It does :(

jimmykarily (approved):

PR #767, opened 2021-09-07
Fix to use namespace key for default namespace

manno:
Leftover from previous rename


andreas-kupries (approved):
Have the same fix in a local branch.

Issue #768, opened 2021-09-07
Wrong API endpoint of existing default config file is kept


svollath:
**Describe the bug**
Using Epinio-0.0.24 on linux. If there's a config file from a previous install already, e.g. `~/.config/epinio/config.yaml` - `epinio install` will read that file, and use the previous configured API endpoint, even if the system domain will change.

**To Reproduce**
Have a pre existing `~/.config/epinio/config.yaml` config file - e.g. from "install=>uninstall" procedure.
Make sure the `Traefik Ingress info` (traefik public IP) will differ for the next installation (if you always get the same IP, you could change the former one in the config file, to see the effect later).

1. Install Epinio with `epinio install` (e.g. getting `Traefik Ingress info: [{"ip":"20.75.240.33"}]`)
2. Uninstall Epinio with `epinio uninstall`
3. Install Epinio with `epinio install` (e.g. getting `Traefik Ingress info: [{"ip":"20.72.177.217"}]`)
4. See Epinio picking up a new external IP (e.g. `Traefik Ingress info: [{"ip":"20.72.177.217"}]`), and later fail with
e.g. `🚢 Retrying (0/10) after Post "https://epinio.20.75.240.33.omg.howdoi.website/api/v1/namespaces": dial tcp 20.75.240.33:443: i/o timeout`, which is the old API endpoint.

**Expected behavior**
Since finally Epinio will rewrite the default config file with the current API endpoint information, I'd expected that it will not read, or keep the existing value from the default config file, when running a plain `epinio install`.

**Logs**
existing "~/.config/epinio/config.yaml":
```
api: https://epinio.20.75.240.33.omg.howdoi.website
certs: |
  -----BEGIN CERTIFICATE-----
********
  -----END CERTIFICATE-----
colors: true
epinio_protocol: http
epinio_workloads_namespace: epinio-workloads
epinio_ws_protocol: ws
gitea_namespace: gitea
gitea_protocol: http
namespace: ""
org: workspace
pass: ********
user: ********
wss: wss://epinio.20.75.240.33.omg.howdoi.website
```

Output from `epinio install`:
```
> ./epinio install                                          

🚢 Epinio v0.0.24 installing...

Configuration...
  🧭  skip-traefik: false
  🧭  skip-linkerd: false
  🧭  skip-cert-manager: false
  🧭  system_domain: 
  🧭  email_address: epinio@suse.com
  🧭  user: 082338fdc859745b
  🧭  password: eb21266ce034e7b9
[...]

PR #772, opened 2021-09-07
Installer reloads config so create org uses new API endpoint

manno:
I also moved the ConfigUpdate code into a separate file, as it is not a client of the API server. It uses only the cluster (via kubeconfig) to find out config values and write them to disk.

Added a re-initialization of the Epinio API client in the final install stages, so an outdated Epinio config doesn't break the installation. 

fixes https://github.com/epinio/epinio/issues/768

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -19,8 +19,6 @@ import (
 	"time"
 
 	"github.com/avast/retry-go"
-	"github.com/epinio/epinio/deployments"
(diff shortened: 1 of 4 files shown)


andreas-kupries (approved):
lgtm. Certainly easier/simpler than what I had in mind.

@JamesTyrand You may wish to check if this solves #758, under the assumption that you had a config file around which contained an old api url. Note my comment on how to make a local client which has the change and deploys without having to patch the server pod.

andreas-kupries (approved):

PR #779, opened 2021-09-08
Safety: Report empty namespace.

andreas-kupries:
This is an internal error.
It should not happen due to the default namespace set by the config package.

fixes https://github.com/epinio/epinio/issues/783

### DIFF
--- a/internal/cli/clients/client.go
+++ b/internal/cli/clients/client.go
@@ -99,6 +99,10 @@ func (c *EpinioClient) EnvList(ctx context.Context, appName string) error {
 		WithStringValue("Application", appName).
 		Msg("Show Application Environment")
 
+	if err := c.TargetOk(); err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


internal/cli/clients/client.go
```diff
@@ -1712,6 +1780,13 @@ func (c *EpinioClient) curlWithCustomErrorHandling(endpoint, method, requestBody
 	return bodyBytes, nil
 }
 
+func (c *EpinioClient) TargetOk() error {
+	if c.Config.Org == "" {
+		return errors.New("Internal Error: No targeted")
```
manno, line 1785:
```suggestion
		return errors.New("Internal Error: No namespace targeted")
```

manno: This is an internal error, because we have the default in the config normally, right?

manno (changes requested):

internal/cli/clients/client.go
andreas-kupries: Yes.

PR #788, opened 2021-09-09
699 remove gitea

### LINKED ISSUE #699: Replace gitea with S3 compatible storage
Right now, in Epinio v0.0.21, the process of uploading and staging the code is as follows:

- The cli uploads the code as a tarball.
- The server endpoint extracts the tarball and pushes it to Gitea
- When the cli asks the code to be staged, the tekton pipeline is creating a temporary PVC to be used as the `source` tekton workspace (https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#volumeclaimtemplate). That PVC is deleted when the PipelineRun is deleted (which happens when the next staging process succeeds).

In order to implement build cache (https://github.com/epinio/epinio/issues/688) we will need to make the PVC for the sources a permanent PVC which will not be automatically deleted as above, but we will rather need to manage its lifecycle in our code (e.g. delete it when the app is deleted etc). See the linked issue for more details.

Having a permanent PVC that has a copy of the application’s code makes Gitea feel a bit redundant. Instead of storing the code on Gitea when the cli uploads it, we could directly store it on that permanent application PVC. We can then make the sources available to the tekton pipeline using the `sources` workspace (https://github.com/tektoncd/pipeline/blob/main/docs/workspaces.md#using-persistentvolumeclaims-as-volumesource) without any need for a “clone” Task.

Such a solution allows us to get rid of Gitea which is currently only acting as a storage for code. There are some things we need to think about though, namely:

- What do we replace the gitref with? Possibly a UUID or anything that won’t create conflicts should do.
- Is staging faster this way? Probably yes, because we don’t need to clone anything and the code for staging is living on a PVC (like it did before, after cloning)
- Is the “upload” endpoint faster?
- Is there any limit on how many PVCs we can have? Is one PVC per app prohibiting? Even if there is a limit, we would face that problem anyway, since the tekton pipeline right now, creates a PVC for each PipelineRun (so minimum one PVC per pushed app).
- Where does a PVC live in kubernetes? Do we need any “placement” rules to make sure source code transactions are as fast as possible?
- How will this work in the future, when/if we implement rollbacks for apps?
- Does the new solution take up more disk space in total than what we currently have?

Regarding the last item, here is a comparison:

Gitea:

- Gitea deployment size - flat size, doesn’t increase with the number of apps
- One git tree per pushed application - depending on the types of changes on the source code, this can be from a little larger than a “flat” checkout of the code to seriously larger (e.g. if big blobs have been removed at some time in history)
- One PVC per PipelineRun - Full clone of the same git repo (or a shallow copy?)

So worst case scenario: 2 full git clones + gitea deployment size

PVCs:

- One “shallow” copy of the source code per pushed app - multiplied by the number of previous versions we want to keep around (e.g. when we implement rollbacks). While a new push is happening we will need 2 copies of the code for some time (before we delete the old copy upon successful staging).

One of the problems with the suggested solution is, how do we get the uploaded tarball (the "upload" request) written in the application PVC. The code runs inside the epinio server pod and that pod doesn't have that PVC bound. We probably need some temporary pod with the PVC mounted to which we will send the file over. This should be fast and secure. It should also handle the cases when something goes wrong and not leave partly copied code around that will break some next staging.

manno:
Fixes #699


jimmykarily:
This PR is missing a test to check that old objects are deleted from S3 when a new staging is successful. Also has some TODOs in code that need fixing.

The functionality is there already though. Epinio deploys and uses Minio if no external S3 settings are provided. It is using the external S3 storage otherwise.

docs/user/explanations/advanced.md
```diff
@@ -19,6 +19,9 @@ The quick way of pushing an application, as explained at
 [Quickstart: Push an application](../tutorials/quickstart.md#push-an-application), uses a local
 directory containing a checkout of the application's sources.
 
+
+TODO: Update this (no gitea)
```
jimmykarily, line 23: We also need a section about S3 settings and how to configure an external S3

internal/api/v1/upload.go
```diff
@@ -59,26 +54,38 @@ func (hc ApplicationsController) Upload(w http.ResponseWriter, r *http.Request)
...
 
-	log.Info("uploaded app", "org", org, "app", name)
+	username, err := GetUsername(r)
+	if err != nil {
+		return UserNotFound()
+	}
+	// TODO: When staging, check the metadata to ensure the blob that is requested
+	// actually belongs to the app
+	// TODO: We need the metadata to be able to tell if the blob belongs to an existing
+	// app or not (leftovers?).
```
jimmykarily, line 78: Let's also check if these metadata are really part of the S3 protocol or just a minio thing. In other words, can we be sure those data will always be there, regardless of the chose S3 solution?

jimmykarily: Moved this to a different issue. Needs discussion. 

https://github.com/epinio/epinio/issues/801

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,203 @@
...
+	return nil
+}
+
+// New returns an instance of an s3 manager
+func New(connectionDetails ConnectionDetails) (*Manager, error) {
+	useSSL := connectionDetails.UseSSL
+
+	// TODO add support for skip verify and private certs
+	// t, _ := minio.DefaultTransport(useSSL)
+	// t.TLSClientConfig.InsecureSkipVerify = true
```
jimmykarily, line 83: This should only be needed if someone uses and external, yet insecure S3. This should be a different feature, if anyone ever asks for it. I'll remove these for now.

jimmykarily:
I still want to fix this: https://github.com/epinio/epinio/pull/788/files#diff-44a311773090a2abd70d10d9f026c5d98e1a8f88f37381e50f8b4a30fe784854R316 (not fail when trying to delete an already deleted object from S3) and then update the documentation.

I want to get the code review done first before I write the documentation to avoid having to update docs (and images) again after code review. If this PR gets an approval I can move the documentation update to another PR.

I will open for review now, I will work on the last issue above in the meantime.

internal/application/application.go
```diff
@@ -317,6 +311,12 @@ func Unstage(ctx context.Context, cluster *kubernetes.Cluster, appRef models.App
 		if err != nil {
 			return err
 		}
+
+		// TODO: write a test for this?
+		// TODO: Don't fail if the file is already gone?
```
jimmykarily, line 316: Hmm, I'm not sure we need a test for this. The underlying library will give a 204 if the object is missing (so no error): https://github.com/minio/minio-go/blob/master/api-remove.go#L186-L189

Should we skip this test?

acceptance/api/v1/applications_test.go
```diff
@@ -131,16 +135,58 @@ var _ = Describe("Apps API Application Endpoints", func() {
...
+		secretKey, err := base64.StdEncoding.DecodeString(string(out))
+		Expect(err).ToNot(HaveOccurred(), string(out))
+
+		rand, err := randstr.Hex16()
+		Expect(err).ToNot(HaveOccurred(), out)
+		out, err = helpers.Kubectl("run", "-it",
+			"--restart=Never", "miniocli"+rand, "--rm",
+			"--image=minio/mc", "--command", "--",
+			"/bin/bash", "-c",
+			fmt.Sprintf("mc alias set minio http://minio.minio-epinio.svc.cluster.local %s %s 2>&1 > /dev/null && mc --quiet ls minio/epinio", string(accessKey), string(secretKey)))
```
andreas-kupries, line 177: This might need some in-code docs. It took a while to understand that the alias command creates a shorthand name for the server which is subsequently used by the following `ls`. It did not help that the `ls` was nearly out of sight at the right border of the display.

jimmykarily:
I tested the PR once again with AWS S3 to see if metadata work there too:

image


All good!

acceptance/api/v1/applications_test.go
```diff
@@ -361,30 +439,6 @@ var _ = Describe("Apps API Application Endpoints", func() {
 			Expect(err).ToNot(HaveOccurred())
 		})
 
-		When("uploading a broken tarball", func() {
```
andreas-kupries, line 364: How has the test become superfluous ?

jimmykarily: We no longer unpack the tarball on the "Upload" endpoint. We don't even do so in the "Stage" endpoint. We unpack at the very last minute, inside the pipeline as a separate Task. This check became a bit more complicated to implement now. We would need to upload a broken tarball, stage the app and look at the failed pipeline logs to see the error.

Maybe there is a some command we can run on the "Upload" endpoint to verify the validity of the tarball and return an error early? Untaring just for that is probably a bad idea since it's an expensive operation.

Otoh, this is only testing that the library we use to create the tarball, actually works. If it didn't we should be able to tell immediately (all tests would fail).

For these reasons, we decided to simply remove this test. If you think it's worth the effort, I could try one of the above (check pipeline output or find some way to validate the tarball without extracting).

andreas-kupries: No, I was just wondering. I am ok now.

acceptance/api/v1/applications_test.go
jimmykarily: Done with this: https://github.com/epinio/epinio/pull/788/commits/110337c636e9b56d912f787ad46244a773c73a85 ?

internal/api/v1/importgit.go
```diff
@@ -0,0 +1,101 @@
...
+	"github.com/epinio/epinio/helpers/tracelog"
+	"github.com/epinio/epinio/internal/api/v1/models"
+	"github.com/epinio/epinio/internal/s3manager"
+	git "github.com/go-git/go-git/v5"
+	"github.com/go-git/go-git/v5/plumbing"
+	"github.com/julienschmidt/httprouter"
+)
+
+// ImportGit handles the API endpoint /namespaces/:org/applications/:app/import-git.
+// It receives a Git repo url, it clones that (shallow clone), creates a tarball
```
andreas-kupries, line 21:
```suggestion
// It receives a Git repo url and revision, clones that (shallow clone), creates a tarball
```

internal/application/application.go
manno: I vote skip :)

internal/cli/info.go
```diff
@@ -12,7 +12,7 @@ var ()
 var CmdInfo = &cobra.Command{
 	Use:   "info",
 	Short: "Shows information about the Epinio environment",
-	Long:  `Shows status and version for Kubernetes, Gitea, Tekton and Eirini.`,
+	Long:  `Shows status and version for Kubernetes, Tekton etc.`,
```
andreas-kupries, line 15: I thought that we had already changed this to simply use a more generic `server components`.
```suggestion
	Long:  `Shows status and versions for epinio's server-side components.`,
```

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+type Manager struct {
+	minioClient       *minio.Client
+	connectionDetails ConnectionDetails
+}
+
+type ConnectionDetails struct {
+	Endpoint string
+	// UseSSL Toggles the use of SSL for the s3 connection. If we deploy
```
andreas-kupries, line 27:
```suggestion
	// UseSSL toggles the use of SSL for the s3 connection. If we deploy
```

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,198 @@
...
+	// our own thing, linkderd will take care of this and we set useSSL to
```
andreas-kupries, line 28:
```suggestion
	// our own thing, linkerd will take care of this and we set UseSSL to
```

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,198 @@
...
+		Endpoint:        endpoint,
+		UseSSL:          useSSL,
+		AccessKeyID:     key,
+		SecretAccessKey: secret,
+		Bucket:          bucket,
+		Location:        location,
+	}
+}
+
+// Validate makes sure the S3 settings provided are valid
```
andreas-kupries, line 49:
```suggestion
// Validate makes sure the provided S3 settings are valid
```

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,198 @@
...
+// The user should either provide all the mandatory settings
```
andreas-kupries, line 50: either ... or what if she doesn't ?
To make sure, the sentence is an incomplete `either/or` construct with the `or` branch missing.

internal/s3manager/s3manager.go
```diff
@@ -0,0 +1,198 @@
...
+	allMandatoryEmpty := details.Endpoint == "" &&
+		details.AccessKeyID == "" &&
+		details.SecretAccessKey == "" &&
+		details.Bucket == ""
+	optionalSet := details.Location != ""
+
+	// If mandatory fields are partly set
+	partlyMandatory := !(allMandatorySet || allMandatoryEmpty)
+	if partlyMandatory {
+		return errors.New("when specifying an external s3 server, you  must set all  mandatory S3 options")
```
andreas-kupries, line 65:
```suggestion
		return errors.New("when specifying an external s3 server, you must set all mandatory S3 options")
```

andreas-kupries (changes requested):
Just comment nits.

docs/user/explanations/advanced.md
jimmykarily: Moved documentation here: https://github.com/epinio/epinio/issues/804

internal/application/application.go
jimmykarily: Skipping.

internal/s3manager/s3manager.go
jimmykarily: Fixed with: ```The user should provide all the mandatory settings or no settings at all```

andreas-kupries (approved):

andreas-kupries (approved):
all nits addressed

andreas-kupries (approved):

PR #791, opened 2021-09-10
feat: Ensure tests run with custom epinio config file

### LINKED ISSUE #696: Can't run tests against existing epinio installation
Environment:
- Azure AKS cluster (v1.20.7), or
- Amazon EKS cluster (v1.20.4)
- Epinio pre-installed (v0.0.21, two step install with custom domain)

How to reproduce:
- checkout epinio git repo (tag v0.0.21)
- prepare buildenv (add go to PATH, `make tools-install`, `go get -u github.com/onsi/ginkgo/ginkgo`)
- copy epinios config.yaml to \<epinio-repo\>/tmp/epinio.yaml
- set env vars for KUBECONFIG=\<kubeconfig\>, EPINIO_CONFIG=\<epinio-repo\>/tmp/epinio.yaml, EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1, EPINIO_SKIP_PATCH=1
- run e.g. `make test-acceptance-api`

Expected output:
- tests skip epinio installation
- framework will execute tests with correct kube auth

Output:
- epinio installation is skipped (but a v0.0.21 binary will be built for ginkgo nodes)
- framework is able to e.g. "Setup cluster services" and "Setup google", but later fails to connect to kubernetes for any test:

```
[1] Running Suite: APIv1 Suite
[1] ==========================
[1] Random Seed: 1629216531 - Will randomize all specs
[1] Parallel test node 1/2.
[1] 
[1] I'm running on runner = skylab
[1] REGISTRY_USERNAME or REGISTRY_PASSWORD environment variables are empty. Pulling from dockerhub will be subject to rate limiting.
[2] Running Suite: APIv1 Suite
[2] ==========================
[2] Random Seed: 1629216531 - Will randomize all specs
[2] Parallel test node 2/2.
[2] 
[1] Compiling Epinio on node 1
[1] Ensure default workspace exists
[1] Setup cluster services
[1] Setup google
[1] SynchronizedBeforeSuite is done, checking Epinio info endpoint
[1] STEP: creating an org
[1] STEP: targeting an org
[1] STEP: creating an org
[1] STEP: targeting an org
[1] STEP: creating application resource first
[1] • Failure [12.354 seconds]
[1] Apps API Application Endpoints
[1] /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:29
[1]   Deploying
[1]   /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:415
[1]     with non-staging using custom docker image
[1]     /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:495
[1]       when deploying a new app
[1]       /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:510
[1]         returns a success [It]
[1]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:517
[1] 
[1]         401 Unauthorized
[1]                                                                                                                                                                                            
[1]         Expected                                                                                                                                                                           
[1]             <int>: 401                                                                                                                                                                     
[1]         to equal                                                                                                                                                                           
[1]             <int>: 200                                                                                                                                                                     
[1] 
[1]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:525
[1] ------------------------------
[2] STEP: creating an org
[2] STEP: targeting an org
[2] STEP: creating an org
[2] STEP: targeting an org
[2] STEP: creating application resource first
[2] • Failure [12.326 seconds]
[2] Apps API Application Endpoints
[2] /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:29
[2]   Deploying
[2]   /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:415
[2]     with staging
[2]     /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:437
[2]       when deploying a new app
[2]       /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:438
[2]         returns a success [It]
[2]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:439
[2] 
[2]         Unexpected error:
[2]             <*json.SyntaxError | 0xc00065ef20>: {                                                                                                                                          
[2]                 msg: "invalid character 'U' after top-level value",                                                                                                                        
[2]                 Offset: 5,                                                                                                                                                                 
[2]             }                                                                                                                                                                              
[2]             invalid character 'U' after top-level value                                                                                                                                    
[2]         occurred                                                                                                                                                                           
[2] 
[2]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:451
[2] ------------------------------
[1] STEP: creating an org
[1] STEP: targeting an org
[1] STEP: creating an org
[1] STEP: targeting an org
[1] STEP: creating application resource first
[1] • Failure [12.425 seconds]
[1] Apps API Application Endpoints
[1] /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:29
[1]   Deploying
[1]   /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:415
[1]     with non-staging using custom docker image
[1]     /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:495
[1]       when deploying a new app
[1]       /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:510
[1]         returns a success [It]
[1]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:517
[1] 
[1]         401 Unauthorized
[1]                                                                                                                                                                                            
[1]         Expected                                                                                                                                                                           
[1]             <int>: 401                                                                                                                                                                     
[1]         to equal                                                                                                                                                                           
[1]             <int>: 200                                                                                                                                                                     
[1] 
[1]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:525
[1] ------------------------------
[2] STEP: creating an org
[2] STEP: targeting an org
[2] STEP: creating an org
[2] STEP: targeting an org
[2] STEP: creating application resource first
[2] • Failure [12.422 seconds]
[2] Apps API Application Endpoints
[2] /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:29
[2]   Deploying
[2]   /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:415
[2]     with staging
[2]     /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:437
[2]       when deploying a new app
[2]       /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:438
[2]         returns a success [It]
[2]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:439
[2] 
[2]         Unexpected error:
[2]             <*json.SyntaxError | 0xc000111600>: {                                                                                                                                          
[2]                 msg: "invalid character 'U' after top-level value",                                                                                                                        
[2]                 Offset: 5,                                                                                                                                                                 
[2]             }                                                                                                                                                                              
[2]             invalid character 'U' after top-level value                                                                                                                                    
[2]         occurred                                                                                                                                                                           
[2] 
[2]         /suse/svollath/github/epinio/acceptance/api/v1/applications_test.go:451
[2] ------------------------------
[1] STEP: creating an org
[1] STEP: targeting an org
[1] • Failure [5.335 seconds]
[1] Services API Application Endpoints, Mutations
[1] /suse/svollath/github/epinio/acceptance/api/v1/services_mutation_test.go:16
[1]   POST api/v1/orgs/:org/applications/:arg/servicebindings/
[1]   /suse/svollath/github/epinio/acceptance/api/v1/services_mutation_test.go:609
[1]     returns a 'not found' when the application does not exist [It]
[1]     /suse/svollath/github/epinio/acceptance/api/v1/services_mutation_test.go:685
[1] 
[1]     401 Unauthorized
[1]                                                                                                                                                                                            
[1]     Expected                                                                                                                                                                               
[1]         <int>: 401                                                                                                                                                                         
[1]     to equal                                                                                                                                                                               
[1]         <int>: 404
...
```

rohitsakala:
fixes #696

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,7 +85,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
-          make install
+          make prepare_ci_environment_k3d
           make test-acceptance-cli
 
   acceptance-api:
@@ -130,7 +130,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
-          make install
+          make prepare_ci_environment_k3d
           make test-acceptance-api
 
   acceptance-apps:
@@ -175,7 +175,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
-          make install
+          make prepare_ci_environment_k3d
           make test-acceptance-apps
 
   acceptance-install:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -85,6 +85,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
+          make install
           make test-acceptance-cli
 
   acceptance-api:
@@ -129,6 +130,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
+          make install
           make test-acceptance-api
 
   acceptance-apps:
@@ -173,6 +175,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
+          make install
           make test-acceptance-apps
 
   acceptance-install:
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,5 @@ minikube-start:
 minikube-delete:
 	@./scripts/minikube-delete.sh
 
-install: embed_files build-linux-amd64
-	EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1 ./dist/epinio-linux-amd64 install --skip-default-namespace
-	@./scripts/patch-epinio-deployment.sh
+prepare_ci_environment_k3d: embed_files build-linux-amd64
+	@./scripts/prepare-ci-environment-k3d.sh
--- a/Makefile
+++ b/Makefile
@@ -156,6 +156,6 @@ minikube-start:
 minikube-delete:
 	@./scripts/minikube-delete.sh
 
-install: build
+install: embed_files build-linux-amd64
 	EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1 ./dist/epinio-linux-amd64 install --skip-default-namespace
 	@./scripts/patch-epinio-deployment.sh
--- a/acceptance/api/v1/suite_test.go
+++ b/acceptance/api/v1/suite_test.go
@@ -11,6 +11,8 @@ import (
 	"github.com/epinio/epinio/acceptance/helpers/proc"
 	"github.com/epinio/epinio/acceptance/testenv"
 	"github.com/epinio/epinio/helpers"
+	epinioConfig "github.com/epinio/epinio/internal/cli/config"
+
 	"github.com/onsi/ginkgo/config"
 
 	. "github.com/onsi/ginkgo"
@@ -59,13 +61,14 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	// install to do it's thing without needing the patch script to run first.
 	// Eventually is used to retry in case the rollout of the patched deployment
 	// is not completely done yet.
+	epinioBinary := testenv.EpinioBinaryPath()
 	fmt.Println("Ensure default workspace exists")
 	testenv.EnsureDefaultWorkspace(epinioBinary)
 
 	fmt.Println("Setup cluster services")
 	testenv.SetupInClusterServices(epinioBinary)
 
-	out, err = helpers.Kubectl("get", "pods", "--namespace", "minibroker", "--selector=app=minibroker-minibroker")
+	out, err := helpers.Kubectl("get", "pods", "--namespace", "minibroker", "--selector=app=minibroker-minibroker")
 	Expect(err).ToNot(HaveOccurred(), out)
 	Expect(out).To(MatchRegexp(`minibroker.*2/2.*Running`))
 
@@ -95,7 +98,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	Expect(err).ToNot(HaveOccurred(), out)
 	os.Setenv("EPINIO_CONFIG", nodeTmpDir+"/epinio.yaml")
 
-	env = testenv.New(nodeTmpDir, testenv.Root())
+	config, err := epinioConfig.LoadFrom(nodeTmpDir + "/epinio.yaml")
+	Expect(err).NotTo(HaveOccurred())
+	env = testenv.New(nodeTmpDir, testenv.Root(), config.User, config.Password)
 
 	out, err = env.Epinio(nodeTmpDir, "target", "workspace")
 	Expect(err).ToNot(HaveOccurred(), out)
--- a/acceptance/api/v1/suite_test.go
+++ b/acceptance/api/v1/suite_test.go
@@ -49,12 +49,8 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 
 	testenv.CreateRegistrySecret()
 
-	epinioBinary := testenv.EpinioBinaryPath()
-	err := testenv.EnsureEpinio(epinioBinary)
-	Expect(err).ToNot(HaveOccurred(), "installing epinio")
-
-	out, err := testenv.PatchEpinio()
-	Expect(err).ToNot(HaveOccurred(), out)
+	err := testenv.CheckEpinio()
+	Expect(err).ToNot(HaveOccurred(), "checking epinio")
 
 	// Now create the default org which we skipped because it would fail before
 	// patching.
--- a/acceptance/api/v1/suite_test.go
+++ b/acceptance/api/v1/suite_test.go
@@ -4,13 +4,10 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"strconv"
 	"testing"
-	"time"
 
 	"github.com/epinio/epinio/acceptance/helpers/proc"
 	"github.com/epinio/epinio/acceptance/testenv"
-	"github.com/epinio/epinio/helpers"
 	epinioConfig "github.com/epinio/epinio/internal/cli/config"
 
 	"github.com/onsi/ginkgo/config"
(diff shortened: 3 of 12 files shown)


Makefile
```diff
@@ -156,6 +156,6 @@ minikube-start:
 minikube-delete:
 	@./scripts/minikube-delete.sh
 
-install: build
-	EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1 ./dist/epinio-linux-amd64 install --skip-default-namespace
+install: embed_files build-$(shell uname | tr "[:upper:]" "[:lower:]")-amd64
```
manno, line 159: I wonder if that name is too generic? It's installing for CI, not useful for users. More like a CI/dev Makefile target?

Also, when this builds the platform binary, do tests still need to build it again (https://github.com/epinio/epinio/blob/main/acceptance/testenv/epinio.go#L125-L136), or is the server binary enough?
Hm, probably doesn't matter, as rebuilding is fast.

rohitsakala: Removed BuildEpinio from tests and also removed generic bash call.

thardeck (dismissed):

Makefile
andreas-kupries: Maybe `install` :arrow_right:  `ci_setup` ?

andreas-kupries (commented):
lgtm

andreas-kupries (approved):
lgtm

andreas-kupries (dismissed):
lgtm again

acceptance/api/v1/suite_test.go
```diff
@@ -45,16 +47,11 @@ var _ = SynchronizedBeforeSuite(func() []byte {
...
 	fmt.Printf("Compiling Epinio on node %d\n", config.GinkgoConfig.ParallelNode)
-	testenv.BuildEpinio()
 
 	testenv.CreateRegistrySecret()
 
-	epinioBinary := testenv.EpinioBinaryPath()
-	err := testenv.EnsureEpinio(epinioBinary)
-	Expect(err).ToNot(HaveOccurred(), "installing epinio")
-
-	out, err := testenv.PatchEpinio()
```
jimmykarily, line 56: Search for "PatchEpinio" in this diff. You remove it here but you keep it in other cases. Shouldn't the Make target take care of everything, including patching? Maybe we can also remove this function, unless the installation suite still needs it.

rohitsakala: Right. Removed now in other places

Makefile
rohitsakala: Done

manno (changes requested):
This removes `BuildEpinio`, so no platform binary is created. The Makefile also hardcodes the `dist/epinio-linux-amd64` binary. While the server binary is always Linux, the tests would have previously worked on Mac, both binaries were built automatically.

I think it's sensitive for `ci_install` to actually mean `install-epinio-on-k3d-on-linux-for-ci`, as a shortcut.
However, now `make test-acceptance-cli` will fail on a Mac, where it previously worked. Not only is there no install target, but the binary for end to end testing is not built for Darwin.

rohitsakala:
> This removes `BuildEpinio`, so no platform binary is created. The Makefile also hardcodes the `dist/epinio-linux-amd64` binary. While the server binary is always Linux, the tests would have previously worked on Mac, both binaries were built automatically.
> 
> I think it's sensitive for `ci_install` to actually mean `install-epinio-on-k3d-on-linux-for-ci`, as a shortcut.
> However, now `make test-acceptance-cli` will fail on a Mac, where it previously worked. Not only is there no install target, but the binary for end to end testing is not built for Darwin.

Yup. I reverted the change back to the generic make target. But I am not sure why you felt generic was bad in your previous comment. I initially thought we need to have a generic make target for install which @jimmykarily wanted. Fortunately we didn't require a new make target for CI since the generic one works for CI too.

acceptance/api/v1/suite_test.go
```diff
@@ -63,13 +60,14 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	// install to do it's thing without needing the patch script to run first.
 	// Eventually is used to retry in case the rollout of the patched deployment
 	// is not completely done yet.
+	epinioBinary := testenv.EpinioBinaryPath()
 	fmt.Println("Ensure default workspace exists")
 	testenv.EnsureDefaultWorkspace(epinioBinary)
```
jimmykarily, line 65: Having a default namespace is part of installing Epinio. If we take care of patching outside the pipeline, I think we should take care of creating the missing namespace there too.

Now that I look at it again, we are mixing things a bit. For example, we don't install Epinio and we don't patch the epinio server pod but we do compile the client binary and we try to create the missing namespace here. What if the cluster we are talking to, is running a different version of the Epinio server or missing components (because it's an "old" cluster for example)?

In my opinion, we should either do everything in the pipeline (like we did) or nothing at all (which is what this PR should be all about). Doing nothing means:

- Compile/Install/Patch Epinio outside the pipeline. Also create the missing default namespace if needed.
- Pass a kubeconfig, an epinio config and an epinio binary path to the pipeline
- The pipeline runs the tests using the above three, assuming they all match the epinio version which is tested.

With this plan, consistency is a responsibility of the developer or the pipeline and not the test suite.

Another option would be, to still have a helper do the first step in go code and call it conditionally in the before suite block, if the information of step 2 is not passed to the pipeline. I think this is what we tried to do initially but living next to the test code made it too easy to spread. After a certain point, the setup of the cluster was all over the place. That's why I believe the setup should better be handled outside the acceptance test suite.

Of course the installation tests are a different matter.

rohitsakala: Agreed.

manno:
>  The generic one will work on Mac too. I just tested it. Let me know if you face any trouble.

If it builds the server binary as well as the clients platform binary that's great. Definitely get it back ;)

When I wrote "generic", I only meant the name of the "install" target. As Makefiles are somewhat user facing, I was wondering if just "install" could be misunderstood.

acceptance/install/suite_test.go
```diff
@@ -35,7 +35,7 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	testenv.CreateRegistrySecret()
 
 	return []byte(strconv.Itoa(int(time.Now().Unix())))
```
jimmykarily, line 37:
```suggestion
	return []byte{}
```

scripts/prepare-ci-environment-k3d.sh
```diff
@@ -0,0 +1,70 @@
+#!/bin/bash
+
+set -e
+
+# UNAME should be darwin or linux
+UNAME="$(uname | tr "[:upper:]" "[:lower:]")"
+
+# EPINIO_BINARY is used to execute the installation commands
+EPINIO_BINARY="./dist/epinio-"${UNAME}"-amd64"
```
jimmykarily, line 9: Maybe use:

```
SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )"
```
and use that to refer to the binary? This way it doesn't matter where you invoke this script from.

scripts/prepare-ci-environment-k3d.sh
```diff
@@ -0,0 +1,70 @@
...
+check_dependency kubectl helm
+# Create docker registry image pull secret
+create_docker_pull_secret
+
+# Install Epinio
+EPINIO_DONT_WAIT_FOR_DEPLOYMENT=1 "${EPINIO_BINARY}" install --skip-default-namespace
+
+# Patch Epinio
+./scripts/patch-epinio-deployment.sh
+sleep 10
```
jimmykarily, line 42: the script is already waiting for the copying to happen. I don't think you need this sleep.

.github/workflows/main.yml
```diff
@@ -85,6 +85,7 @@ jobs:
           make acceptance-cluster-delete
           make acceptance-cluster-setup
           export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig
+          make prepare_ci_environment_k3d
```
jimmykarily, line 88: Now this page needs updating: https://github.com/epinio/epinio/blob/main/docs/developer/howtos/acceptance_tests.md

I wonder if you should explicitly set the 3 needed environment variables here, just to make it obvious that this is the way to configure the tests. Right now `EPINIO_CONFIG`, `EPINIO_BINARY` and `KUBECONFIG` are all set in 3 different places. It's not obvious that these are the 3 things needed in order to point the tests to an Epinio installation.

Maybe if we document it properly in the linked page, it will be enough.

jimmykarily (changes requested):

scripts/prepare-ci-environment-k3d.sh
manno: Apparently we already do that in other scripts. I would be content with just calling the scripts from the top level dir only, it's not user facing.

jimmykarily: fair

scripts/prepare-ci-environment-k3d.sh
rohitsakala: Removed.

.github/workflows/main.yml
rohitsakala: Done. Updated the docs

Issue #785, opened 2021-09-08
Manifest file to store options


agracey:
**Is your feature request related to a problem? Please describe.**

Currently, each push of a new app requires you to specify configuration options in the push command. It would be nice to offer a way to use something similar to the CF manifest.yml. 

**Describe the solution you'd like**

I would like to have all of the configuration options stored in a file (likely manifest.yml or epinio.yml) that gets read when pushing an app. 

Options specified on the command line would override options in the manifest.

The configuration options needed:
- name
- [deployment template name](https://github.com/epinio/epinio/issues/361)
- services to bind
- environment variables
- routes (including path and ingress class when needed)

We also need to include the ability to generate the manifest from a running application.

**Describe alternatives you've considered**

**Additional context**

andreas-kupries:
@thardeck Please read. I believe my thoughts, especially in the code section are important to your work on #643 also.

## Thinking .... (Always a good idea :wink:)

> The configuration options needed:
> - name
> - [deployment template name](https://github.com/epinio/epinio/issues/361)
> - services to bind
> - environment variables
> - routes (including path and ingress class when needed)

For the initial PR I would say only that `name`, `services`, and `env variables` are relevant.
The latter dependent on having #643 done. Templating and routes depend on things not on our schedule, IIRC. When they are added and implemented adding them to the manifest would be part of their acceptance criteria.

> I would like to have all of the configuration options stored in a file (likely manifest.yml or epinio.yml) that gets read when pushing an app. 

My prefered name is `epinio.yml`. Makes the association clear. Should also prevent mis-identification of CF, kubecf, etc. manifests by epinio as for its own use. Any support for such (and their different yaml spec) can be done as separate enhancements, if actually necessary.

> Options specified on the command line would override options in the manifest.

Sensible.
  - Trivially done for app `name`. Although it makes the `name` argument of `push` optional.
  - Services (and EVs): Data are lists. The easy specification for override here is to go for total replacement. If there are any options for services (or EVs) the relevant part of the manifest is dropped entirely. No attempt at merging the two.

> We also need to include the ability to generate the manifest from a running application.

Doable.
  - Name of application is always available, whether the app is active or at rest (i.e. no workload).
  - EVs are also always available Stored separately, and the `app env` commands can attach EVs to apps at rest.
  - Services ... Can be bound currently only to active apps. We do not support pre-binding of services to apps at rest.

[...]

Issue #643, opened 2021-07-26
Set environment variables on push


agracey:
Similar to how we can set services with the `-b` flag, we should have a `-e` flag to set environment variables when pushing

PR #803, opened 2021-09-15
Support binding services to apps at rest

### LINKED ISSUE #802: Support pre-binding of services to applications at rest.

See #785 where this came up. Having the feature will make things much easier on that front.
Further, we already support app EVs for apps at rest.
Not having this for services is a limitation I see no real reason, except that apps at rest were introduced late, and not all parts of epinio got immediate support. This looks to be such a part.

  - The `service bind/unbind` commands operate properly on apps without workload.
  - The `app create` command supports the `--bind` option of `app push`.
  - Internally, on `push` the import of the services into the deployment is done by the server component, and not the client anymore. 

None.

andreas-kupries:
Fix #802. This is a step towards #785. See also #643.


andreas-kupries:
The branch https://github.com/epinio/epinio/tree/802-services-for-apps-at-rest-stashed holds the code written before restarting from scratch. This is kept for now in case we can reuse bits and pieces in the coming work.

andreas-kupries:
First testable change ... Integrate service binding resources into the app delete cascade, via owner references.
This removes the need for the API server to explicitly unbind services from the deployment.
Note that custom services do not have app-specific resources when bound and are not taking part in this.
Never did.

andreas-kupries:
Collapsed the majority of the PR history in preparation for resolving conflicts with current main.
Easier to manage a single resolution instead of doing it over and over again is git works through the chain of changes.

docs/developer/explanations/service-management-internals.md
```diff
@@ -0,0 +1,267 @@
...
+         ((stage sources))
+         ((deploy image (stage result)))
+                server side -- per S: ((bind sequencing))
+
+user <-- client :report ok fail
+```
+
+### Semantics: `app delete A`
+
+Deletes the named application `A`.
```
jimmykarily, line 232: and the services secret I guess (if one exists)?

internal/api/v1/applications.go
```diff
@@ -344,12 +373,106 @@ func (hc ApplicationsController) Update(w http.ResponseWriter, r *http.Request)
...
 
-	workload := application.NewWorkload(cluster, app.AppRef())
-	err = workload.Scale(r.Context(), updateRequest.Instances)
+	app, err := application.Lookup(ctx, cluster, org, appName)
 	if err != nil {
 		return InternalError(err)
 	}
 
+	// TODO: Can we optimize to perform a single restart regardless of what changed ?!
+	// TODO: Should we ?
```
jimmykarily, line 382: Interesting, I think we should but it can be tricky in some cases. I'm not sure if all the information is kept on one object (the `Deployment`). We can keep mutating the receiver of those methods (the "app") until we are ready and then apply the thing once, but if there are more than one kubernetes objects involved, the applying method can't fail midway and leave things in an inconsistent state.

internal/api/v1/models/app.go
```diff
@@ -83,6 +94,20 @@ func (ar *AppRef) EnvSecret() string {
 	return names.GenerateResourceName(ar.Name + "-env")
 }
 
+// SvcSecret returns the name of the kube secret holding the
+// bound services of the referenced application
+func (ar *AppRef) SvcSecret() string {
```
jimmykarily, line 99: This is a generator, not a getter (shouldn't be used to find the name of an existing secret). Maybe the name of the method should make that clear? (e.g. `GenSvcSecretName`)

docs/developer/explanations/service-management-internals.md
andreas-kupries: Yes. That is one of the earliest changes in the (now collapsed) history - Make the service binding resources owned by the app, and auto-deleted. This is turn deletes the secrets.

internal/api/v1/servicebindings.go
```diff
@@ -155,6 +186,22 @@ func (hc ServicebindingsController) Delete(w http.ResponseWriter, r *http.Reques
 		return OrgIsNotKnown(org)
 	}
 
+	apiErr := DeleteBinding(ctx, cluster, org, appName, serviceName, username)
+	if apiErr != nil {
+		return apiErr
+	}
+
+	w.Header().Set("Content-Type", "application/json")
+	_, err = w.Write([]byte{})
```
jimmykarily, line 195: Consider using `jsonResponse` from this PR (now merged): https://github.com/epinio/epinio/pull/811/files#diff-168b4a07f391b22def39fd870b0373811deb114cd513793f9cf5a3182f8452ceR81

internal/api/v1/applications.go
andreas-kupries: Yep. Hence the more questing tone of the comment. Not sure if possible, and if yes, if it would be worth it. The currently 3 changes (at most) might even cause at kube level only a single restart anyway, if the deployment controller sees them quickly enough.

internal/api/v1/models/app.go
andreas-kupries: Could be done. There are several methods like this in the file now (First was for environment, `EnvSecret` nearby).

internal/api/v1/servicebindings.go
andreas-kupries: Yes, that is part of what I have to do as part of the rebase / conflict resolution.

internal/api/v1/models/app.go
andreas-kupries: Done, commit 4b318c1c1d975ef6571608271baf99f66951d126

docs/developer/explanations/service-management-internals.md
andreas-kupries: See next sentence, which implies this.

internal/application/application.go
```diff
@@ -377,3 +318,50 @@ func Logs(ctx context.Context, logChan chan tailer.ContainerLogLine, wg *sync.Wa
 	logger.Info("fetch")
 	return tailer.FetchLogs(ctx, logChan, wg, config, cluster)
 }
+
+// fetch is a common helper for Lookup and List. It fetches all
+// information about an application from the cluster.
+func fetch(ctx context.Context, cluster *kubernetes.Cluster, app *models.App) error {
```
jimmykarily, line 324: I feel this should have the `models.App` as a receiver (iow, be a method of), after all, all it does it mutating it.

internal/application/scale.go
```diff
@@ -0,0 +1,130 @@
...
+
+	return int32(result), nil
+}
+
+// ScalingSet sets the desired number of instances for the named application.
+// When the function returns the number is saved.
+func ScalingSet(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef, instances int32) error {
+	return scaleUpdate(ctx, cluster, appRef, func(scaleSecret *v1.Secret) {
+		scaleSecret.Data[instanceKey] = []byte(strconv.Itoa(int(instances)))
+	})
```
jimmykarily, line 45: This is the only consumer of `scaleUpdate` thus the only possible implementation of the `modifyScaling` parameter is this one here. Is there a reason not to collapse these 2 methods in one that does this and only this kind of update? I don't see any other way to use `scaleUpdate` (with a different `modifyScaling`).

internal/application/application.go
andreas-kupries: I will look into that. About to push the rebase.

internal/application/scale.go
andreas-kupries: Heh. I can do the inline (after the rebase is in). I copied the file from environment and did not bother with the inline when the number of users shrunk to one. ... And now I am wondering if this update function could be shared among env, scale, and services. ...

internal/application/application.go
andreas-kupries: Ah. `App` is a non-local type. Making the change implies moving `fetch` into the `models` package.

andreas-kupries: And if we do that we will get a circular dependency because the function uses methods from `application`, and is used in `application`. Verdict: Declined.

internal/application/scale.go
andreas-kupries: That goes into the rabbit hole that the associated load functions can be generalized also.

I am declining the inline request, and will make a ticket to rework this part of the system, i.e. generalize the load/update across these state-holding secrets.

andreas-kupries: Ticket is https://github.com/epinio/epinio/issues/832

internal/application/services.go
```diff
@@ -0,0 +1,176 @@
...
+			},
+		}
+		err = cluster.CreateSecret(ctx, appRef.Org, *svcSecret)
+
+		if err != nil {
+			return nil, err
+		}
+
+		err = cluster.LabelSecret(ctx, appRef.Org, secretName, map[string]string{
+			"app.kubernetes.io/name":       appRef.Name,
```
jimmykarily, line 164: same

internal/application/workload.go
```diff
@@ -32,6 +36,166 @@ func NewWorkload(cluster *kubernetes.Cluster, app models.AppRef) *Workload {
...
+		// iteration, over the new services now, adds all which are not in old, i.e. actually new.
+
+		newVolumes := []corev1.Volume{}
+		newMounts := []corev1.VolumeMount{}
+
+		for _, volume := range deployment.Spec.Template.Spec.Volumes {
+			_, hasold := oldServices[volume.Name]
+			_, hasnew := new[volume.Name]
+
+			// Note that volumes which are in not in old are passed and kept. These are the volumes
```
jimmykarily, line 139:
```suggestion
			// Note that volumes which are not in old are passed and kept. These are the volumes
```

jimmykarily (changes requested):

docs/developer/explanations/service-management-internals.md
```diff
@@ -0,0 +1,267 @@
...
+__Note__: The service binding resources and associated secrets of
+__catalog-based__ services are owned by the app resource, as they are
+tied to the app (Materialization of the n:m relation between
+applications and services). This makes removal on app deletion easier,
+as it will happen automatically as part of the cascade taking down
+everything associated with an application.
+
+The bindings of custom services are not treated this way, as these are
+shared between applications, and thus cannot be owned by a single
+one. Unbinding them is not deleting anything either also.
```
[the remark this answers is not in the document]
andreas-kupries: Definitely.

internal/api/v1/applications.go
```diff
@@ -73,16 +72,71 @@ func (hc ApplicationsController) Create(w http.ResponseWriter, r *http.Request)
...
 	err = application.Create(ctx, cluster, appRef, username)
 	if err != nil {
 		return InternalError(err)
 	}
 
-	err = jsonResponse(w, models.ResponseOK)
+	// [INSTANCES CODING]
+	// Save instance information - Force default if not specified
+	// ATTENTION: == 0 -- Use default number of instances
+	//          :  > 0 -- Desired + 1, reduce by one to get actual value.
```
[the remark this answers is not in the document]
andreas-kupries: - The `nil` value of integers is `0`.
  - I did not really wanted to open the can of worms of allowing even a single negative number and its effect on validation

Using `*integer` instead of `integer` for the field ... Ok, while it felt like a hack I am now unsure why I thought it would be more of a hack than the coding I am using now.

So, we could go and make the `Instances` field a pointer to integer. Then we can use the `nil` as the signal, and a regular int value.

I will work on that now.

internal/application/scale.go
```diff
@@ -0,0 +1,130 @@
...
+				},
+			},
+		}
+		err = cluster.CreateSecret(ctx, appRef.Org, *scaleSecret)
+
+		if err != nil {
+			return nil, err
+		}
+
+		err = cluster.LabelSecret(ctx, appRef.Org, secretName, map[string]string{
```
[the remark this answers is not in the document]
andreas-kupries: Yeah, we have this separation of creating and labeling quit a bit over the place, not just here. I do not remember why it was done in the past. Maybe we had not figured out how to set labels as part of creation ?

I would prefer to declare the proposed change as out of scope for this ticket and create a new ticket to investigate fix this tech debt.

internal/application/services.go
andreas-kupries: s.a.

docs/developer/explanations/service-management-internals.md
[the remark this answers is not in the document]
andreas-kupries: I also have to remove the `planning.md` document, as it was purely tracking work in the repo while I was working.

internal/application/scale.go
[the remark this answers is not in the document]
jimmykarily: sounds good

internal/api/v1/applications.go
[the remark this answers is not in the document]
andreas-kupries: Done, commit 45ac3302f6d0dd381f5a2004da18101a3df2cb99

internal/application/scale.go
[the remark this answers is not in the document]
andreas-kupries: As the redo of the instance field was done so quickly I had a quick look ...

When I saw that we had a `CreateLabeledSecret` (`CLS`) I thought the change would be easy.
Then I noticed that the signature is different.

The `CreateSecret` (`CS`) takes a pre-filled v1.Secret structure.

`CLS` on the other takes a name and creates just a bare-bones secret with labels.
Oh, and it uses `LabelSecret` (`LS`) internally, so even that is a two-step dance of create, then label.

That said, fixing up the `CLS` signature first, to take a secret (or writing a variant doing so), and then converting all places using `CS` and `LS` to the new `CLS` or its variant, and then fixing `CLS` (and its variant) to not use `LS` internally looks to be the way to go.

Still looking to be better as separate ticket and PR.

Ticket is #843

jimmykarily (approved):

PR #809, opened 2021-09-15
adding gitjob instructions

### LINKED ISSUE #738: Write docs for "How to use GitJob"
Improve this:
https://gist.github.com/agracey/33a748f1fe9931357199d25907a553ea

agracey:
Closes #738

### DIFF
--- a/docs/user/howtos/gitjob_push.md
+++ b/docs/user/howtos/gitjob_push.md
@@ -1,0 +1,104 @@
+One of the more interesting features of Rio was that it would let you set up a deployment that rebuilds and republishes when your code stored in Git is changed. 
+
+We can recreate this functionality using the GitJob CRD that's a part of [Rancher Fleet](https://fleet.rancher.io/).
+
+NOTE: We will improve this experience in the future!
+
+## Setup
+
+### Install GitJob
+
+If you don't have Rancher (or standalone Fleet) installed, we need to install the GitJob operator by following the isntructions found at https://github.com/rancher/gitjob#running.
+
+
+Then we need to setup the Service Account to run our Jobs with (since we don't need to do anything directly with the kube api, we don't need to add any role bindings to it):
+
+```
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: epinio-push
+```
+
+### Upload Epinio Config
+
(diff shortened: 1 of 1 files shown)


agracey:
I punted on GitJob installation itself as my cluster already has it installed via Fleet.

docs/user/howtos/gitjob_push.md
```diff
@@ -0,0 +1,104 @@
...
+
+### Upload Epinio Config
+
+So the GitJob can authenticate and push correctly, we can upload our Epinio config file to the cluster with:
+
+```
+kubectl create secret generic epinio-creds --from-file=$HOME/.config/epinio/config.yaml
+```
+
+This will create a secret containing the config.yaml that was created locally when you do `epinio install` or `epinio config update`
```
jimmykarily, line 31: Alternatively, we can give permissions to the ServiceAccount to read the Epinio credentials secret and run `epinio config update` from the gitjob itself. This will mean, one less step in this setup. Otoh, it will need us to create a role with that permission and explain why, which probably makes the guide more complicated.

I think both ways are equal, complexity-wise

docs/user/howtos/gitjob_push.md
```diff
@@ -0,0 +1,104 @@
...
+apiVersion: gitjob.cattle.io/v1
+kind: GitJob
+metadata:
+  # The name of the GitJob, doesn't need to match the project.
+  name: samplepush
+spec:
+  syncInterval: 15
+  git:
+    # The git repo and branch to track. 
+    repo: https://github.com/scf-samples/12factor
```
jimmykarily, line 49: Can we use this instead: https://github.com/epinio/example-12factor ?

jimmykarily:
This is quick and easy, good job! I wonder if there should be some link to this page in some other page. We don't have a "contents" section somewhere, do we?

docs/user/howtos/gitjob_push.md
agracey: 👍 done

docs/user/howtos/gitjob_push.md
agracey: Yeah, and I think something that is done client side is easier to document than something magic inside our platform 😄

jimmykarily (approved):

PR #811, opened 2021-09-16
API client and server logging

manno:
* the cli/clients package is now the cli-client for Epinio,
  cli/clients/epinioapi is the API client library for the Epinio API
* cli/client uses epinioapi package for talking to the endpoints
* server always uses jsonResponse when answering requests
* every endpoint returns a json body
* adds more request/response structs
* adds generic response struct as an intermediate, until endpoints
  answer with the affected resource itself

refers to #775 - preparing for common request/response logging

### DIFF
--- a/internal/api/v1/organizations.go
+++ b/internal/api/v1/organizations.go
@@ -57,12 +57,7 @@ func (oc NamespacesController) Match(w http.ResponseWriter, r *http.Request) API
 
 	log.Info("deliver matches", "found", matches)
 
-	js, err := json.Marshal(matches)
-	if err != nil {
-		return InternalError(err)
-	}
-	w.Header().Set("Content-Type", "application/json")
-	_, err = w.Write(js)
+	err = jsonResponse(w, models.NamespacesMatchResponse{Names: matches})
 	if err != nil {
 		return InternalError(err)
 	}
@@ -115,12 +110,7 @@ func (oc NamespacesController) Index(w http.ResponseWriter, r *http.Request) API
 		})
 	}
 
-	js, err := json.Marshal(namespaces)
-	if err != nil {
-		return InternalError(err)
-	}
-	w.Header().Set("Content-Type", "application/json")
-	_, err = w.Write(js)
+	err = jsonResponse(w, namespaces)
 	if err != nil {
 		return InternalError(err)
 	}
@@ -170,8 +160,10 @@ func (oc NamespacesController) Create(w http.ResponseWriter, r *http.Request) AP
 		return InternalError(err)
 	}
 
+	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusCreated)
-	_, err = w.Write([]byte{})
+
+	err = jsonResponse(w, models.ResponseOK)
 	if err != nil {
 		return InternalError(err)
 	}
@@ -225,7 +217,8 @@ func (oc NamespacesController) Delete(w http.ResponseWriter, r *http.Request) AP
 
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusOK)
-	_, err = w.Write([]byte{})
+
+	err = jsonResponse(w, models.ResponseOK)
 	if err != nil {
 		return InternalError(err)
 	}
--- a/internal/api/v1/services.go
+++ b/internal/api/v1/services.go
@@ -67,12 +67,7 @@ func (sc ServicesController) Show(w http.ResponseWriter, r *http.Request) APIErr
 		responseData[key] = value
 	}
 
-	js, err := json.Marshal(responseData)
-	if err != nil {
-		return InternalError(err)
-	}
-	w.Header().Set("Content-Type", "application/json")
-	_, err = w.Write(js)
+	err = jsonResponse(w, models.ServiceShowResponse{Details: responseData})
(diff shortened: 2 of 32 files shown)


internal/api/v1/organizations.go
```diff
@@ -170,8 +160,10 @@ func (oc NamespacesController) Create(w http.ResponseWriter, r *http.Request) AP
 		return InternalError(err)
 	}
 
+	w.Header().Set("Content-Type", "application/json")
```
jimmykarily, line 163: isn't `jsonResponse` setting this, 2 lines below?

internal/api/v1/organizations.go
```diff
@@ -225,7 +217,8 @@ func (oc NamespacesController) Delete(w http.ResponseWriter, r *http.Request) AP
 
 	w.Header().Set("Content-Type", "application/json")
```
jimmykarily, line 218: same

internal/api/v1/services.go
```diff
@@ -312,8 +304,10 @@ func (sc ServicesController) Create(w http.ResponseWriter, r *http.Request) APIE
 		}
 	}
 
+	w.Header().Set("Content-Type", "application/json")
```
jimmykarily, line 307: same

internal/cli/clients/push.go
```diff
@@ -1,96 +1,240 @@
...
+		return fmt.Errorf("%s: %s", "app name incorrect", strings.Join(errorMsgs, "\n"))
 	}
 
-	defer response.Body.Close()
-	bodyBytes, err := ioutil.ReadAll(response.Body)
+	c.ui.Normal().Msg("Create the application resource ...")
+
+	request := models.ApplicationCreateRequest{Name: appRef.Name}
+
+	_, err := c.API.AppCreate(
```
jimmykarily, line 92: Why does this method have an error func while others (e.g. `AppUpload` below) don't? Shouldn't we be consistent?

internal/cli/server.go
```diff
@@ -111,38 +110,47 @@ func ReadyRouter() *httprouter.Router {
...
 
 		// log the request first, then ...
 		logRequest(r, log)
 
 		// ... call the original http.Handler
 		h.ServeHTTP(w, r)
-	}
 
-	return http.HandlerFunc(fn)
+		if log.V(15).Enabled() {
```
jimmykarily, line 134: not for this PR but, we need to put names to these values. What is level 15? What are we supposed to write there? Should I put my message there, lower or higher?

jimmykarily (changes requested):

internal/api/v1/organizations.go
manno: Haha, great that you found that.
Yes, but this method is one of the few exceptations, that want to return their own header.  In the next line:

>  `w.WriteHeader(http.StatusCreated)`

So, I read the API docs for the http funcs and I think this way we return both, the content type and the 201 status code. The `SetHeader` in `jsonResponse` will be ineffective and the call to `r.Write` will not send another header.

I would suggest to add the status code to `jsonResponse` (or it's replacement) in the future.

internal/api/v1/organizations.go
manno: Yes, again bypassing the SetHeader in jsonResponse, because I didn't want to make this PR even bigger.

internal/cli/clients/push.go
manno: Yes, I think there is something wrong with our http client helpers, they encapsulate the response struct. So the only way to see the real status code is to pass in a "custom error handler". Some API endpoints return a status code that's different from 200 and need to do this, because everything != 200 is considered an error by our helpers. 

I want to change that, maybe by exposing the response struct, but not in this PR.
Not a 100% sure how the result will look like.

internal/cli/server.go
manno: So, I made my peace with those numbers. I thought we picked [logr](https://github.com/go-logr/logr), because it doesn't have log levels.

> Verbosity-levels on info logs. This gives developers a chance to indicate arbitrary grades of importance for info logs, without assigning names with semantic meaning such as "warning", "trace", and "debug." Superficially this may feel very similar, but the primary difference is the lack of semantics. Because verbosity is a numerical value, it's safe to assume that an app running with higher verbosity means more (and less important) logs will be generated.

I'm still pro log-levels, but this is also fine. No need to overthink it I guess...

internal/api/v1/organizations.go
jimmykarily: This is hard to follow. So, any `w.Header().Set()` that comes after an explicit `w.WriteHeader` will also have no effect? Then I wonder why we can still write the body (inside the `jsonResponse` method).

In any case, I think you are right, we should pass the status code to `jsonResponse`. I think the response object should have one more field for that.

internal/cli/server.go
jimmykarily: We, as developers, randomly choose a verbosity "number". But logs are not fireworks. The user doesn't simply want more. The user is probably looking for some piece of information and I doubt they will start increasing the verbosity until they find the thing they want. Me at least, I simply increase it to "100" or similar to be sure I get what I need on the first try.

Let's discuss this outside this PR though.

jimmykarily (approved):

PR #829, opened 2021-09-21
Fix error when CRD doesn't have a conditions field yet

jimmykarily:
We got this error randomly in CI:

```
error: .status.conditions accessor error: <nil> is of the type <nil>, expected []interface{}
```

possibly originating here:

https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/unstructured/helpers.go#L137

This commit is re-using existing code to wait for CRDs. We no longer
check conditions but it works as is in existing code. If this ever
fails, we may need to extend the WaitForCRD function to also check the
conditions field for establishment.

### DIFF
--- a/deployments/minio.go
+++ b/deployments/minio.go
@@ -3,7 +3,6 @@ package deployments
 import (
 	"context"
 	"fmt"
-	"strconv"
 	"time"
 
 	"github.com/epinio/epinio/helpers"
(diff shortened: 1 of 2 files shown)


manno (approved):
Great, so there was an existing method ;)
I did search for it, but couldn't find it.

jimmykarily:
> Great, so there was an existing method ;)
> I did search for it, but couldn't find it.

I just stumbled on it when I was about to write it again :D

PR #831, opened 2021-09-22
Wait until CRDs are also established

### LINKED ISSUE #830: Wait for CRDs to be established
We had 2 ways to wait for CRDs and we switched to our implementation here: https://github.com/epinio/epinio/pull/829#issue-1002451675

the other method was waiting for the condition "established" (shelling out to `kubectl`) but failed when the CRD wasn't there yet (or some weird situation like that, we are not sure).

The method we use now, polls until the CRD is there but doesn't wait for the "established" condition. Now we get this error:

```
 🚢 Deploying Minio..................
⚠️  Deployment minio-operator failed with error: Installing minio/tenant.yaml failed:
error: unable to recognize "/tmp/epinio3364112752": no matches for kind "Tenant" in version "minio.min.io/v2"
: exit status 1
```

(from here: https://github.com/epinio/epinio/runs/3664330284?check_suite_focus=true#step:11:101)

A possible solution is the extend the `WaitForCRD` method to also check for the `established` condition before it returns.

jimmykarily:
Fixes #830

### DIFF
--- a/helpers/kubernetes/cluster.go
+++ b/helpers/kubernetes/cluster.go
@@ -22,6 +22,7 @@ import (
 	apibatchv1 "k8s.io/api/batch/v1"
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
 	apiextensions "k8s.io/apiextensions-apiserver/pkg/client/clientset/clientset"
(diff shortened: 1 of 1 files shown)


jimmykarily:
Another option would be re re-use this code: https://github.com/kubernetes/kubectl/blob/d11f42ab701cb66eec24aa24b3f19939d6409e32/pkg/cmd/wait/wait.go#L121-L123

but it looked more compicated.

manno (approved):

PR #837, opened 2021-09-24
Remove service-catalog, minibroker and google broker

jimmykarily:
and all the relevant Epinio commands

Fixes #837


jimmykarily:
TODO:

- Remove ServiceCatalog
- Remove Minibroker
- Remove Google broker
- Fix tests
- [TBD] Stop refering to custom services as "custom". They are the only services now. Can be done on another PR.

andreas-kupries:
> [TBD] Stop refering to custom services as "custom". They are the only services now. Can be done on another PR.

Ok if done as part of the beta2 cycle. Do not wish to publish beta 2 without this being clean

acceptance/apps/rails_test.go
```diff
@@ -90,8 +90,24 @@ var _ = Describe("RubyOnRails", func() {
...
 
 		// Create a database for Rails
+		out, err = proc.Run("", false, "helm", "repo", "add", "bitnami", "https://charts.bitnami.com/bitnami")
+		Expect(err).ToNot(HaveOccurred(), out)
+
 		serviceName = catalog.NewServiceName()
-		out, err = env.Epinio("", "service", "create", serviceName, "mariadb", "10-3-22")
+		out, err = proc.Run("", false,
+			"helm", "install", serviceName, "bitnami/mariadb", "-n", rails.Org,
+			"--set", "auth.database=\"mydb\"",
```
andreas-kupries, line 99:
```suggestion
			"--set", `auth.database="mydb"`,
```

Less issue with quoted characters. See also following lines.

internal/services/services.go
```diff
@@ -21,31 +21,12 @@ func Lookup(ctx context.Context, kubeClient *kubernetes.Cluster, org, service st
...
-		return nil, err
-	}
-
-	catalogServices, err := CatalogServiceList(ctx, kubeClient, org)
-	if err != nil {
-		return nil, err
-	}
-
-	return append(customServices, catalogServices...), nil
+	return CustomServiceList(ctx, kubeClient, org)
```
andreas-kupries, line 29: Tech debt, to be resolved in separate PR:
  - `Merge services.go`, `custom_service.go`, inline custom specific functions into the generic functionality.
  - Drop the service interface ?!

andreas-kupries (changes requested):

andreas-kupries:
Another general comment: There is no true conflict with anything I wrote for the services bound to apps at rest.

The only thing which can be affected, but is not a conflict is that I extended the service interface, specifically `GetBinding` got an additional argument for ownership information. This argument was only ever used by the catalog services. It is irrelvant to custom services and ignored. Because ustom services do not have per-app binding structures to delete when the app goes away.

This one change of mine can be removed, and IMHO should be.

acceptance/apps/rails_test.go
andreas-kupries: Oh, I believe I see your issue and error for the test suite.

You have bash quoting here. The `"` around `mydb`, etc. are wrong.
Proper fix may be:
```suggestion
			"--set", "auth.database=mydb",
```
etc.

andreas-kupries: The very fact that your commands are identical here and in the terminal is the issue I believe, and tripping you up.

internal/services/services.go
jimmykarily: I'm wondering if the interface will be useful soon, if we decide to try some other kind of Service (e.g. Crossplane). Usually though, this kind of thinking leads to left-overs in code which nobody ever cleans up. I think I prefer that we removed the interface. Maybe on another PR?

acceptance/apps/rails_test.go
andreas-kupries: Out of band: Various ways of quoting, including the proposed, did not work

internal/services/services.go
andreas-kupries: See first line of my original comment, another PR is fine.

jimmykarily: Moved here: #844

jimmykarily:
> Another general comment: There is no true conflict with anything I wrote for the services bound to apps at rest.
> 
> The only thing which can be affected, but is not a conflict is that I extended the service interface, specifically `GetBinding` got an additional argument for ownership information. This argument was only ever used by the catalog services. It is irrelvant to custom services and ignored. Because ustom services do not have per-app binding structures to delete when the app goes away.
> 
> This one change of mine can be removed, and IMHO should be.

Moved here: #844

jimmykarily:
I've been debugging the rails test failure for days now. I tried a number of things. Here are the findings:

I deployed mariadb with helm : `helm install mariadb-test bitnami/mariadb --set auth.database=production,auth.username=myuser,auth.password=mypassword,volumePermissions.enabled=true,auth.rootPassword=mypassword`). This gave me:

```
NAME            NAMESPACE       REVISION        UPDATED                                 STATUS          CHART           APP VERSION
mariadb-test    default         1               2021-09-30 07:17:08.356173474 +0000 UTC deployed        mariadb-9.6.0   10.5.12 
```

I was able to:

- connect to that database from rails running on the host
- connect to that database using mysql cli from withing a pod running on the same k3d cluster

I wasn't able to connect to that database from rails running inside the pod. I got the error:

```
TypeError: no implicit conversion of Integer into String
```

The underlying reason is this:

```
cnb@rails-5bbcdf4bf5-lckf2:/$ /cnb/lifecycle/launcher bin/rails c

irb(main):008:0> Mysql2::Client.new(host: "mariadb-test.default", username: "myuser", password: "mypassword", database: "production").query("select database()").to_a
=> [{"database()"=>0}]
```
(that `0` should have been `production`)

Our verdict is that mysql2 gem does not play nice with the mysql libraries included in the buildpack generated image. That gem hasn't got any updates for 2 years. Given that we can connect to the database from some other pod using mysql cli (tried both, linkerd messed and not), it means it's not a networking issue. Maybe an encoding incompatibility?

We also tried to talk to a DigitalOcean mysql and after we changed the password to not use the `caching_sha2_password` plugin (which wasn't available), it worked fine.

Regarding why it worked with minibroker, I think it didn't. Epinio has a bug that it reports the app as running even when the pod is still coming up. The test probably reported success too early and the app failed behind the scenes

The current plan is to use some other database for the rails test (postgresql?) and if that works, update the example-rails repository too.

jimmykarily:
I tried postgresql (deployed with: `helm install postgres bitnami/postgresql --set volumePermissions.enabled=true,postgresqlUsername=myuser,postgresqlPassword=mypassword,postgresqlDatabase=production`)

and it works like a charm. I will adapt the test and if that works I will also change the README of the example-rails repo.

jimmykarily:
> The current plan is to use some other database for the rails test (postgresql?) and if that works, update the example-rails repository too.

Done: https://github.com/epinio/example-rails/pull/2

Adapted the test, passes locally. Let's see what CI has to say.

andreas-kupries:
One CI failure

> Apps
[1] /home/sles/actions-runner/_work/epinio/epinio/acceptance/apps_test.go:26
[1]   list across namespaces
[1]   /home/sles/actions-runner/_work/epinio/epinio/acceptance/apps_test.go:562
[1]     lists all applications belonging to all namespaces [It]
[1]     /home/sles/actions-runner/_work/epinio/epinio/acceptance/apps_test.go:588

It does not look like something transient. More some issue in the test setup coming home to roost.

andreas-kupries:
Looked a bit more. The failure looks to be independent of the broker removal.
More like a concurrency issue.

The test in question targets multiple namespaces, and while that looks ok maybe other tests have changed things underneath the command.

That said, given that the command is `list --all` it should not validate the target namespace IMHO. Because in that mode it operates across all anyway, not just the target.

So. This failure is IMHO not a blocker against merging.

jimmykarily:
> Looked a bit more. The failure looks to be independent of the broker removal. More like a concurrency issue.
> 
> The test in question targets multiple namespaces, and while that looks ok maybe other tests have changed things underneath the command.
> 
> That said, given that the command is `list --all` it should not validate the target namespace IMHO. Because in that mode it operates across all anyway, not just the target.
> 
> So. This failure is IMHO not a blocker against merging.

yes and the test passes locally. I will open a new issue to fix the (not required) validation

andreas-kupries (approved):

jimmykarily:
Moved here: https://github.com/epinio/epinio/issues/856

PR #845, opened 2021-09-29
fix: Ensure tests can use EPINIO_CONFIG env variable

rohitsakala:


andreas-kupries (commented):
That looks ok. I will trial that locally and then give my final :+1:  :-1:

andreas-kupries (approved):
Testing went fine. :green_heart: go.

Issue #816, opened 2021-09-17
Static site generated docs


jimmykarily:
Something like this one:

https://docs.kubewarden.io/

or the one from quarks

kkaempf:
https://docs.epinio.io is live, based on https://github.com/epinio/docs

PR #846, opened 2021-09-29
Drop user docs from source repo (follow up from #816)

kkaempf:
Moved to https://github.com/epinio/docs

Signed-off-by: Klaus Kämpf <kkaempf@suse.de>


andreas-kupries (approved):
The technical part is trivial and ok. My understanding of the discussion in the daily is that it can go ahead.

thardeck (approved):

PR #848, opened 2021-09-30
Show path information for service parameters

### LINKED ISSUE #821: Output of service show has the parameters
# Acceptance Criteria

It should be clear from the `service show` output how to access the service secrets, i.e. where they are mounted, from their application.

andreas-kupries:
fix #821 

  - Shows access paths for service parameters
  - Shows scope warning, i.e. that the paths are limited in visibility, to the app container

Modified not just `service show`, but also `create-custom`.
:warning: Did not change `create`, as catalog services are getting removed.

### DIFF
--- a/internal/cli/usercmd/client.go
+++ b/internal/cli/usercmd/client.go
@@ -480,12 +480,14 @@ func (c *EpinioClient) ServiceDetails(name string) error {
 	}
 	serviceDetails := resp.Details
 
-	msg := c.ui.Success().
-		WithTable("Parameter", "Value", "Access Path").
-		WithTableRow("Epinio User", resp.Username, "")
+	c.ui.Note().
+		WithStringValue("User", resp.Username).
+		Msg("")
+
+	msg := c.ui.Success()
 
 	if len(serviceDetails) > 0 {
-		msg = msg.WithTableRow("", "", "")
+		msg = msg.WithTable("Parameter", "Value", "Access Path")
 
 		keys := make([]string, 0, len(serviceDetails))
 		for k := range serviceDetails {
@@ -496,9 +498,11 @@ func (c *EpinioClient) ServiceDetails(name string) error {
 			msg = msg.WithTableRow(k, serviceDetails[k],
 				fmt.Sprintf("/services/%s/%s", name, k))
 		}
-	}
 
-	msg.Msg("")
+		msg.Msg("")
+	} else {
+		msg.Msg("No parameters")
+	}
 
 	c.ui.Exclamation().
 		Msg("Beware, the shown access paths are only available in the application's container")
--- a/internal/cli/usercmd/client.go
+++ b/internal/cli/usercmd/client.go
@@ -480,21 +480,22 @@ func (c *EpinioClient) ServiceDetails(name string) error {
 	}
 	serviceDetails := resp.Details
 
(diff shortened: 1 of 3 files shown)


internal/cli/usercmd/client.go
```diff
@@ -476,17 +480,27 @@ func (c *EpinioClient) ServiceDetails(name string) error {
...
 	keys := make([]string, 0, len(serviceDetails))
 	for k := range serviceDetails {
 		keys = append(keys, k)
 	}
 	sort.Strings(keys)
 	for _, k := range keys {
-		msg = msg.WithTableRow(k, serviceDetails[k])
+		path := ""
+		switch k {
+		case "Status":
```
jimmykarily, line 492: I think "Status" is no more?

internal/cli/usercmd/client.go
```diff
@@ -476,17 +480,27 @@ func (c *EpinioClient) ServiceDetails(name string) error {
...
+		case "Username":
```
jimmykarily, line 493: Slight problem, when service has a "Username" key, then the epinio username is not printed at all:

```
~/workspace/epinio (821-service-parameter-paths)*$ epinio service create myservice2 Username dimitris

🚢 Create Service
Name: myservice2
Namespace: workspace
| PARAMETER |  VALUE   |          ACCESS PATH          |
|-----------|----------|-------------------------------|
| Username  | dimitris | /services/myservice2/Username |

⚠️  Beware, the shown access paths are only available in the application's container

✔️  Service Saved.
Name: myservice2
Namespace: workspace
~/workspace/epinio (821-service-parameter-paths)*$ epinio service show myservice2

🚢 Service Details
Name: myservice2
Namespace: workspace
✔️  
|          |          | ACCESS PATH |
|----------|----------|-------------|
| Username | dimitris |             |

⚠️  Beware, the shown access paths are only available in the application's container

```

Also the table has no headers. First, I think we should call it "Epinio user" to make it clear that it's not a key from the service itself.
Also, since it's not a service key, maybe it shouldn't be in this looped object in the first place? It's service meta information, maybe it belongs to some other object/attribute.

internal/cli/usercmd/client.go
andreas-kupries: Right, with custom as service this is gone. Will fix.

internal/cli/usercmd/client.go
andreas-kupries: True, should move the information out of the general map. Requires changes to the response structure.
That will be a slightly larger change. Will do after the planning.

internal/cli/usercmd/client.go
andreas-kupries: Done, commit 37c434ba7258d68c130d1b7bba04eecc3b17ed3d

internal/cli/usercmd/client.go
andreas-kupries: Done, commit 37c434ba7258d68c130d1b7bba04eecc3b17ed3d

Note: Using CI for testing in the hope that it works better. Here I am running into pull errors for the linkerd images.
That does not bode well for local testing.

andreas-kupries:
Ok. Easy change.

andreas-kupries:
Tweaked, commit 65e33e37e0d7074ff8324733207e4b03757f7416

jimmykarily (approved):

PR #861, opened 2021-10-04
Implement --env parameter for push and apps create/update

### LINKED ISSUE #643: Set environment variables on push

thardeck:
Fixes #643

### DIFF
--- a/internal/cli/options.go
+++ b/internal/cli/options.go
@@ -62,12 +62,20 @@ func bindOption(cmd *cobra.Command) {
 		})
 }
 
+// envOption initializes the --env/-e option for the provided command
+func envOption(cmd *cobra.Command) {
+	cmd.Flags().StringSliceP("env", "e", []string{}, "environment variables to be used")
+	// nolint:errcheck // Unable to handle error in init block this will be called from
+}
+
 // appConfiguration processes the `--bind` and `--instances` options of
 // the command into a proper application configuration.
-// TODO 803/643 EV option
 func appConfiguration(cmd *cobra.Command) (models.ApplicationUpdateRequest, error) {
 	result := models.ApplicationUpdateRequest{}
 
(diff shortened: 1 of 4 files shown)


internal/cli/options.go
```diff
@@ -90,6 +95,29 @@ func appConfiguration(cmd *cobra.Command) (models.ApplicationUpdateRequest, erro
...
+		}
+		result.Environment = append(result.Environment, models.EnvVariable{
+			Name:  pieces[0],
+			Value: pieces[1],
+		})
+	}
+
+	// nil --> Default / No change
+	// - AppCreate API will replace it with `v1.DefaultInstances`
+	// - AppUpdate API will treat it as no op, i.e. keep current instances.
```
andreas-kupries, line 116: The three preceding comment lines are about the instances, not services, nor EV. Should be moved up to where the instances are handled in this function

andreas-kupries (changes requested):

internal/cli/options.go
```diff
@@ -83,13 +91,26 @@ func appConfiguration(cmd *cobra.Command) (models.ApplicationUpdateRequest, erro
...
 
+	assignments, err := cmd.Flags().GetStringSlice("env")
+	if err != nil {
+		return result, errors.Wrap(err, "failed to read option --env")
+	}
+
+	for _, assignment := range assignments {
+		pieces := strings.Split(assignment, "=")
+		if len(pieces) != 2 {
+			return result, errors.New("Bad --env assignment, expected `name=value` as value")
```
kkaempf, line 106: Maybe add the `assignment` value to the error message to show which `--env` assignment is meant here ?!

andreas-kupries (approved):

Issue #361, opened 2021-05-05
Allow deployment template to be customized


agracey:
It would be very helpful to give an escape hatch to deploy custom yaml as part of the workload being deployed

Some usecases:
- Adding NetworkPolicies to app
- Adding HPA to app
- Adding custom sidecars 
- Deploying built container to knative
- Class of service differences in deployments

This feature is a foot-gun but if we don't expose it through the CLI and require manual editing of a ConfigMap, then it's the Operators responsibility to write good templates. We can even say that if the template is changed, we no longer guarantee the stability of workloads.

- [x] #1224

agracey:
The assumption being that the user has permissions to create the objects being requested

agracey:
This could be doable with "class of service" templates to make safer

agracey:
Formalizing the request slightly:

I would like to be able to customize the yaml that gets deployed when I push an app. It would also be very valuable to developers to allow a different template selection when pushed (but has a default). 

This is likely best done by having a CRD that defines a template to be filled in.

jimmykarily:
We had a short discussion about this and about use cases. The use case as we understand it, is this:

Some operator, familiar with kubernetes, has a strong opinion on how an application deployment should look like. Maybe also, that operator is not satisfied with how Epinio deploys apps or maybe what we do is not enough and doesn't cover their needs.
That operator still wants to have the "epinio experience" with `epinio push` and all. The cluster's users will be developers and they don't have to know how the application is deployed.

Now a fact:

Epinio application deployment is just a Deployment exposed through a Service and an Ingress. Also uses some auto-generated certificates for TLS.

From all the above, the only things that change (unless I'm missing something) are, the Deployment's container image (which is the staging artifact), the application route (for the Ingress) and some labels on the various resources (the name of the app, the org etc). The rest are really just a template and they are similar in all apps.

The application resources described above used to be templates in the Tekton Pipeline, then we moved them to go code. In any case, conceptually, they are still templates with some values interpolated.

Now, here comes an idea we came up with:

[...]

PR #862, opened 2021-10-04
Improvements from Helm Deployment Spike

manno:
refs #361

### DIFF
--- a/internal/api/v1/router.go
+++ b/internal/api/v1/router.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 
 	"github.com/epinio/epinio/helpers/routes"
+	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/julienschmidt/httprouter"
 )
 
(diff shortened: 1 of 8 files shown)


internal/api/v1/router.go
```diff
@@ -43,6 +44,8 @@ func jsonErrorResponse(w http.ResponseWriter, responseErrors APIErrors) {
 func errorHandler(action APIActionFunc) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
 		if errors := action(w, r); errors != nil {
+			log := tracelog.Logger(r.Context())
+			log.V(1).Info("responding with json error response", "action", action, "errors", errors)
```
andreas-kupries, line 48: Merge the two lines, drop the temp variable ?

andreas-kupries (approved):

internal/api/v1/stage.go
```diff
@@ -142,11 +142,11 @@ func (hc ApplicationsController) Stage(w http.ResponseWriter, r *http.Request) A
 
 	log.Info("staging app", "org", org, "app", req)
 
-	cs, err := tekton.NewForConfig(cluster.RestConfig)
+	tc, err := cluster.ClientTekton()
```
jimmykarily, line 145: I implemented a function for the "is it staging?" check here: https://github.com/epinio/epinio/pull/860/files#diff-44a311773090a2abd70d10d9f026c5d98e1a8f88f37381e50f8b4a30fe784854R100

I can fix on rebase though, no worries.

internal/api/v1/router.go
manno: Will do in another PR

PR #863, opened 2021-10-05
Add retry to API client

manno:
* add status code to custom error
* retry on all status codes > 399
* implemented only for NamespaceCreate, AppRunning, StagingComplete,
  which already retried on certain strings

refs #361


andreas-kupries (approved):

manno:
I fixed a bug in the doWithCustomErrorHandling func and switched AppCreate to 'do()'.

PR #872, opened 2021-10-07
Manifest support, continued

### LINKED ISSUE #785: Manifest file to store options

andreas-kupries:
Fixes #785 

@jimmykarily Note that the manifest (yaml) as described is already similar to the `AppManifest` structure of #867 , just a bit more. And that may also be a nice starting structure for the new app CRD you are looking into. The separate secrets for env, instances, services, etc. could be merged into that, i.e. removed.

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -7,6 +7,7 @@ import (
 	"net/http"
 	"os"
 	"path"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


andreas-kupries:
The companion documentation PR is https://github.com/epinio/docs/pull/10

PR #885, opened 2021-10-13
Use k3d v5.0.0 in CI (and development)

jimmykarily:


jimmykarily:
> LGTM, I've locally tested `acceptance-cluster-setup.sh` without any issue.
> 
> Just a question for my curiosity: where is `k3d.sh` used?

Through some very-hard-to-follow bash, it's called when you run `make tools-install`. This part of the code needs to be simplified.

Issue #878, opened 2021-10-11
Spike on gin framework


jimmykarily:
Follow up of #818

Let's try to introduce [gin](https://github.com/gin-gonic/gin)

On this first iteration, let's try to replace our router with the gin one. There are a number of other features in gin which we could use (e.g. validation, model binding). We can do it in separate stories if they are complicated. Look at the #818 for more about what gin supports.

PR #886, opened 2021-10-13
Spike, use gin(-gonic) as router replacement

andreas-kupries:
Ref #878

Current state:

- api endpoints disconnected :warning: This breaks CI
- non-api endpoints (dashboard, assets, ready) handled via gin :heavy_check_mark: 

For the api endpoints the error handling will be interesting (what with our APIError interface)


kkaempf:
Hmm, looks like the introduction of `gin` is rather straightforward.
How would you assess the complexity of this port, @andreas-kupries ?

kkaempf:
Thanks !

And no worries, the current spike is not about getting done but about learning a lot (about gin) within a week. And it looks like you're well underway :wink:

Issue #891, opened 2021-10-14
Zero instance option doesn't work with epinio push


juadk:
**Describe the bug**
I tried to deploy an app with zero instance as it's asked in our test scenario 2 ([GKE](https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)) 

Ref #893, this ticket depends on resolution of this bug somewhat.

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Install Epinio on the platform you prefer, I used k3s `v1.21.5+k3s2`:
`epinio install`

2. Push an app:
`epinio push sample0 sample-app -i 0`

3. Look at cluster:
```
kubectl get pods -n workspace
NAME                       READY   STATUS    RESTARTS   AGE
sample0-5bf4575cb9-b8zvw   2/2     Running   0          15m
```

No error message

**Expected behavior**
We shouldn't see an instance as we used the `-i 0` option.

**Logs**
<details>
  <summary>Push logs</summary>
  <pre>
epinio push sample0 sample-app -i 0 --trace-level 5 --verbosity 1
2021/10/14 11:28:18 Config-0xc000112360 "level"=4 "msg"="Loading"  "from"="/root/.config/epinio/config.yaml"
2021/10/14 11:28:18 Config-0xc000112360 "level"=4 "msg"="Loaded"  "value"="namespace=(workspace), user=(197aa71f9db5254a), pass=(099d51a556365888), api=(https://epinio.192.168.121.19.omg.howdoi.website), wss=(wss://epinio.192.168.121.19.omg.howdoi.website), color=(true), @(/root/.config/epinio/config.yaml)"
2021/10/14 11:28:18 EpinioApiClient "level"=4 "msg"="query configuration"  
2021/10/14 11:28:18 Config-0xc000112000 "level"=4 "msg"="Loading"  "from"="/root/.config/epinio/config.yaml"
2021/10/14 11:28:18 Config-0xc000112000 "level"=4 "msg"="Loaded"  "value"="namespace=(workspace), user=(197aa71f9db5254a), pass=(099d51a556365888), api=(https://epinio.192.168.121.19.omg.howdoi.website), wss=(wss://epinio.192.168.121.19.omg.howdoi.website), color=(true), @(/root/.config/epinio/config.yaml)"
2021/10/14 11:28:18 EpinioApiClient "level"=4 "msg"="cached in config"  
[...]

Issue #879, opened 2021-10-11
Implement CI scenario num.2


jimmykarily:
https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)

PR #893, opened 2021-10-15
Add CI scenario 2 - GKE

juadk:
Ref #879 

This PR add a new automated test for GKE in the CI:

- Client: Linux
- Install tests: Split Install for Traefik and Letsencrypt Prod
- Install options: Custom domain
- Push options: zero instances

Current state:

Two things will need to be changed later:
1. The `--tls-issuer=letsencrypt-production` doesn't work, I'm still trying to debug it but the option could be added later. And we have exactly the same issue on the AKS CI.
This issue will be fixed by https://github.com/epinio/epinio/pull/915, so I can add the flag `--tls-issuer=letsencrypt-production` back now and wait the merge before re-triggering the CI.

2. Due to #891 , I can't check that replica is set to 0 [at the end of the test ](https://github.com/epinio/epinio/pull/893/files#diff-f34e1b5f65ea939bed362830c889062d74c107d3401dc4f179efb522eeac77fcR107). Until the bug is fixed, I check that the replica  is equal to 1.

CI results [here](https://github.com/epinio/epinio/actions/workflows/gke.yaml)

### DIFF
--- a/.github/workflows/gke.yaml
+++ b/.github/workflows/gke.yaml
@@ -1,0 +1,97 @@
+name: GKE-CI
+
+on:
+  schedule:
+    - cron:  '0 5 * * *'
+  workflow_dispatch:
+    inputs:
+      keep_cluster:
+        description: "Keep the cluster afterwards? (empty/yes)"
+        required: false
+        default: ""
+      gke_domain:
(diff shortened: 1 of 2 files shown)


juadk:
Like the Letsencrypt issue is identified and is about to be fixed, I can add back the flag `--tls-issuer=letsencrypt-production`.

PR #914, opened 2021-10-20
App health support

### LINKED ISSUE #900: Health checks for workloads
Epinio creates deployments for the pushed apps.

Pods should have a health/ready/(startup) check.
Is the k8s default good enough?

What's the minimum we can do for apps. Should this be customizable, e.g. when an app implements a separate health check endpoint?
Some apps might need a startup probe, to make them wait before they start.

https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

Example: https://github.com/workadventure-suse/workadventure-kustomize/blob/master/base/front-deployment.yaml#L88-L98

andreas-kupries:
Fix #900 

This is the implementation PR.

https://github.com/epinio/docs/pull/18 is the companion PR for the documentation.

### DIFF
--- a/internal/api/v1/application/create.go
+++ b/internal/api/v1/application/create.go
@@ -85,6 +85,7 @@ func (hc Controller) Create(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err)
 	}
 
+	// Instances
 	desired := DefaultInstances
 	if createRequest.Configuration.Instances != nil {
 		desired = *createRequest.Configuration.Instances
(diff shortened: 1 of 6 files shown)


jimmykarily:
@andreas-kupries does #1224 deprecate this one? I mean, if someone want to define readiness/health probes on applications, they can simply provide a custom helm chart, right?

andreas-kupries:
That looks to be a possibility, yes.

jimmykarily:
ok let's close this then. @agracey suggested that we allow the user to define custom values for the app helm chart (e.g. the health probe port etc) but that's another issue

agracey:
New issue: https://github.com/epinio/epinio/issues/1252

PR #915, opened 2021-10-20
Make sure there is an ingressclass for Traefik

jimmykarily:
otherwise the Ingresses created by cert-manager for the challenges,
won't work. That's because they get the new "ingressClass" field which
requires an Ingress Class to exist.

More here:
- https://github.com/traefik/traefik/blob/4235cef1b239bdffdb2fe7c62ebcac83df55781f/docs/content/providers/kubernetes-ingress.md#ingressclass
- https://github.com/traefik/traefik-helm-chart/blob/master/traefik/values.yaml#L64
- https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#specifying-the-class-of-an-ingress
- https://cert-manager.io/docs/usage/ingress/

Probably cert-manager started adding that field here:

https://github.com/jetstack/cert-manager/commit/57e9e57fbf015783da258e40fe8f932f89f4fdac


juadk:
Just tested on GKE and it works perfectly!
image
Deploying the sample-app worked too.

PR #941, opened 2021-10-27
Cache users instead of querying the cluster on each request.

### LINKED ISSUE #927: Implement a cache in authMiddleware for the list of available Users
This is a followup of #871 to fix this comment: https://github.com/epinio/epinio/pull/913#discussion_r733683122

andreas-kupries:
Fix #927

First, a basic load generator for the server, collecting response statistics. Also an initial script invoking for the generator for a matrix of counts and delays.

See https://github.com/epinio/epinio/issues/927#issuecomment-952624065 for the baseline performance of the unmodified server, without any kind of caching.


andreas-kupries:
Closing as rejected. We do not want this kind of debug tool directly in our client. It should be run only by us. Further, the `ab` tool can give us the same kind of numbers without work from us.

Issue #594, opened 2021-07-08
Path based routing


agracey:
I'd like to have multiple applications share a hostname with path based routing. 

The obvious use case here is splitting `/api` from `/static`  so UI code is in a different project (built then hosted statically) from the API code. The reason they need to be able to be on the same host is for CORS simplicity.

jimmykarily:
To simplify the command line syntax, let's switch to this:

```
epinio app update sample --route someother.10.86.4.38.omg.howdoi.website/api --route someotherdomain.org
```

Epinio should figure out which part of the `route` is the domain and which is the path.

PR #951, opened 2021-11-02
Rename "domains" to "routes"

jimmykarily:
as a preparation for https://github.com/epinio/epinio/issues/594

This is a dummy renaming. I'm just doing it on a separate PR to make reviewing easier both for this one and the one that actually implements path based routing.

Made the relevant changes here as well: https://github.com/epinio/application/pull/1/files


thardeck (dismissed):
There seems to be one failed test but looks good to me besides that.

thardeck (approved):

PR #954, opened 2021-11-03
594 path based routing

### LINKED ISSUE #594: Path based routing

jimmykarily:
Fixes #594

### DIFF
--- a/acceptance/helpers/machine/machine.go
+++ b/acceptance/helpers/machine/machine.go
@@ -69,9 +69,9 @@ func (m *Machine) DeleteNamespace(namespace string) {
 	out, err := m.Epinio("", "namespace", "delete", "-f", namespace)
 	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
 
-	namespaces, err := m.Epinio("", "namespace", "list")
-	ExpectWithOffset(1, err).ToNot(HaveOccurred())
-	ExpectWithOffset(1, namespaces).ToNot(MatchRegexp(namespace))
+	out, err = m.Epinio("", "namespace", "show", namespace)
+	ExpectWithOffset(1, err).To(HaveOccurred())
+	ExpectWithOffset(1, out).To(MatchRegexp(".*Not Found: Targeted namespace.*does not exist.*"))
 }
 
 func (m *Machine) VerifyOrgNotExist(org string) {
--- a/acceptance/helpers/machine/machine.go
+++ b/acceptance/helpers/machine/machine.go
@@ -63,6 +63,17 @@ func (m *Machine) TargetOrg(org string) {
 	ExpectWithOffset(1, out).To(MatchRegexp("Currently targeted namespace: " + org))
 }
 
+func (m *Machine) DeleteNamespace(namespace string) {
+	By("deleting a namespace")
+
+	out, err := m.Epinio("", "namespace", "delete", "-f", namespace)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
+
+	namespaces, err := m.Epinio("", "namespace", "list")
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, namespaces).ToNot(MatchRegexp(namespace))
+}
+
 func (m *Machine) VerifyOrgNotExist(org string) {
 	out, err := m.Epinio("", "namespace", "list")
 	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
--- a/internal/application/ingresses.go
+++ b/internal/application/ingresses.go
@@ -8,6 +8,7 @@ import (
 	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/epinio/epinio/internal/auth"
 	"github.com/epinio/epinio/internal/names"
+	"github.com/epinio/epinio/internal/routes"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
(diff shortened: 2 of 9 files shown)


internal/application/ingresses.go
```diff
@@ -141,65 +154,41 @@ func SyncIngresses(ctx context.Context, cluster *kubernetes.Cluster, appRef mode
...
+				Port: networkingv1.ServiceBackendPort{
+					Number: 8080,
+				}}}
+
+	ingress.Spec.TLS = []networkingv1.IngressTLS{{
+		Hosts: []string{ingress.Spec.Rules[0].Host},
+		// TODO:
+		// Should we re-use the same certificate if one exists for that domain?
+		// (e.g. other app on a different path). Who is the owner then?
+		// Is it ok to have 2 differnet certificates for the same domain?
```
jimmykarily, line 187: I think the only implication is that in the case of letsencrypt, we will issue more requests than we need to. But sharing certificates between ingresses when the host is similar requires more complicated handling. I'm not sure it's worth it.

acceptance/helpers/machine/machine.go
```diff
@@ -63,6 +63,17 @@ func (m *Machine) TargetOrg(org string) {
...
 
+func (m *Machine) DeleteNamespace(namespace string) {
+	By("deleting a namespace")
+
+	out, err := m.Epinio("", "namespace", "delete", "-f", namespace)
+	ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
+
+	namespaces, err := m.Epinio("", "namespace", "list")
+	ExpectWithOffset(1, err).ToNot(HaveOccurred())
+	ExpectWithOffset(1, namespaces).ToNot(MatchRegexp(namespace))
```
andreas-kupries, line 74: Alternative, should we wish to avoid `list` ... `show`, and expect error.

internal/application/ingresses.go
```diff
@@ -84,7 +94,10 @@ func SyncIngresses(ctx context.Context, cluster *kubernetes.Cluster, appRef mode
 		}
 		log.Info("creating app ingress", "org", appRef.Org, "app", appRef.Name, "", desiredRoute)
 
-		ing := newAppIngress(appRef, desiredRoute, username)
+		route := routes.FromString(desiredRoute)
+		ingressName := names.IngressName(fmt.Sprintf("%s-%s", appRef.Name, route))
+		ing := route.ToIngress(ingressName)
```
andreas-kupries, line 99:
```suggestion
		ingress := route.ToIngress(ingressName)
```

I.e. nicer name ? Would need updates where used, of course.

internal/application/ingresses.go
```diff
@@ -141,65 +154,41 @@ func SyncIngresses(ctx context.Context, cluster *kubernetes.Cluster, appRef mode
...
-			TLS: []networkingv1.IngressTLS{
-				{
-					Hosts: []string{
-						route,
-					},
-					SecretName: name + "-tls",
-				},
-			},
-		},
+// completeIngress that takes an Ingress as created by the routes#ToIngress
```
andreas-kupries, line 157:
```suggestion
// completeIngress takes an Ingress as created by the routes#ToIngress
```

internal/names/names.go
```diff
@@ -20,8 +20,11 @@ func ServiceName(base string) string {
 // IngressName returns the name of a kube ingress derived from the
 // base string. It ensures that things like leading digits are
 // sufficiently hidden to prevent kube from erroring out on the name.
+// It also replaces "/" characters with "-" to produce a valid resource
+// name from a route (which may have "/" in it).
```
andreas-kupries, line 24:
```suggestion
// name from a route (which may contain "/" characters).
```

internal/routes/routes.go
```diff
@@ -0,0 +1,78 @@
...
+	if len(splitRoute) > 1 {
+		path = "/" + splitRoute[1]
+	} else {
+		path = "/"
+	}
+	return Route{Domain: domain, Path: path}
+}
+
+// FromIngress returns a Route resource matching the given Ingress
+// NOTE: Epinio doesn't create Ingresses with multiple rules. For the reason,
```
andreas-kupries, line 66:
```suggestion
// NOTE: Epinio doesn't create Ingresses with multiple rules. For that reason,
```

andreas-kupries (changes requested):
Some small things

acceptance/helpers/machine/machine.go
jimmykarily: 07024aa1f3842da96c806101142960a3b20e4c24

internal/application/ingresses.go
jimmykarily: 07024aa1f3842da96c806101142960a3b20e4c24

internal/application/ingresses.go
jimmykarily: 07024aa1f3842da96c806101142960a3b20e4c24

internal/routes/routes.go
jimmykarily: 07024aa1f3842da96c806101142960a3b20e4c24

internal/names/names.go
jimmykarily: 07024aa1f3842da96c806101142960a3b20e4c24

internal/application/ingresses.go
jimmykarily: changed this comment to simply explain the situation

andreas-kupries (approved):

PR #956, opened 2021-11-03
Installer waits for minio tenant to be ready

manno:
Directly after installing Epinio users might fail to push apps, because the tenant is
not ready to store uploads.


andreas-kupries (dismissed):

andreas-kupries:
The API failure looks to be an insufficient timeout (i.e. 4 seconds not enough in some cases for things change/settle).

andreas-kupries (approved):
lgtm

jimmykarily:
CI failure is irrelevant (changes are only on the installer). Merging.

PR #959, opened 2021-11-04
Don't list orgs when targeting a specific one

### LINKED ISSUE #958: Namespace#Index shouldn't fail when a namespace is deleted mid-flight
The "Index" endpoint, after finding all the namespaces, tries to populate them with more data (applications, services, etc). If a namespace is removed during that time, the action fails. It should instead remove that namespace from the internal list and continue with the rest.

This bug is causing acceptance tests to fail randomly because some tests are listing namespaces while other ginkgo nodes are deleting their namespaces.

jimmykarily:
because tests fail when an org is deleted while "list" endpoint is still
running.

Fixes #958

### DIFF
--- a/internal/api/v1/namespace/index.go
+++ b/internal/api/v1/namespace/index.go
@@ -6,6 +6,7 @@ import (
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/internal/api/v1/response"
 	"github.com/epinio/epinio/internal/application"
+	epinioerrors "github.com/epinio/epinio/internal/errors"
 	"github.com/epinio/epinio/internal/organizations"
 	"github.com/epinio/epinio/internal/services"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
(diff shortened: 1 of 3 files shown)


andreas-kupries (dismissed):

internal/api/v1/namespace/index.go
```diff
@@ -34,12 +35,16 @@ func (oc Controller) Index(c *gin.Context) apierror.APIErrors {
 	for _, org := range orgList {
 		appNames, err := namespaceApps(ctx, cluster, org.Name)
 		if err != nil {
-			return err
+			return apierror.InternalError(err)
```
andreas-kupries, line 38: Shouldn't this also check for a missing namespace ?
I mean, you check for missing namespace after getting the services.
IMHO this check should also be done after getting the apps, above.

andreas-kupries (changes requested):
Looks to be a partial fix.

internal/api/v1/namespace/index.go
jimmykarily: namespaceApps won't generate an error if a namespace is missing. That's because it never runs `organization.Exists` like the `namespaceServices` does (through `services.List`). It's inconsistent but as far as the fix is concerned, it can't break anymore.

andreas-kupries: Oh, I see now. It uses the `ListAppRefs` and simply gets an empty slice ... Ok then.

andreas-kupries (approved):

internal/api/v1/namespace/index.go
jimmykarily: We could achieve consistency by making the `services.List` a bit more resilient, if it didn't check for the existence of the namespace. It could simply set it in the selector and return an empty list if the namespace is gone. But since this is a hot fix, let's not change more than we have to.

PR #962, opened 2021-11-04
Enable cross CI to release helm chart

juadk:
This PR adds a repository dispatch action to send an API call to the [epinio/epinio-helm-chart ](https://github.com/epinio/epinio-helm-chart)repo.
The latest Epinio tag is sent and it is used for automatically update the appVersion chart field.
See also https://github.com/epinio/epinio-helm-chart/pull/1

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -59,3 +59,17 @@ jobs:
           prerelease: "true"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # Allow to release Epinio Helm chart automatically when we release Epinio.
+      # The latest tag is sent to the Helm chart repo.
+      - name: Get the last tag
+        id: last-tag
+        run: |
+          tag=$(git -P tag | tail -1)
+          echo '::set-output name=TAG::'$tag
+      - name: Repository Dispatch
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -59,3 +59,17 @@ jobs:
           prerelease: "true"
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      # Allow to release Epinio Helm chart automatically when we release Epinio.
+      # The latest tag is sent to the Helm chart repo.
+      - name: Get the last tag
+        id: last-tag
+        run: |
+          tag=$(git -P tag | tail -1)
```
manno, line 67: I wonder if there is an edge case where this behaves different from the "git describe --tags" used by the Makefile.

manno (approved):

.github/workflows/release.yml
juadk: If I'm not wrong, `git describe --tags`  gives also the commit SHA.
Something like:
```
git describe --tags
epinio-0.1.1-3-g87b21da
```

manno: Only if you're not directly on the tag:

```
% git checkout v0.1.4
% git describe --tags
v0.1.4
```

PR #963, opened 2021-11-04
Add a purge step in CI tests when route53 is used

### LINKED ISSUE #939: Purge route53 DNS entries
The DNS entries added for CI tests in route53 are never deleted, we should add a delete command in the POST of the GH actions to fix this.

ldevulder:
Some tests use route53 to create/update a DNS entry but this newly created route was not purge at the end.

This fixes #939.

### DIFF
--- a/acceptance/helpers/route53/change.go
+++ b/acceptance/helpers/route53/change.go
@@ -42,11 +42,11 @@ type DNSAnswer struct {
 	Protocol     string   `json:"Protocol"`
 }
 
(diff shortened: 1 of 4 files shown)


ldevulder:
AKS-CI and GKE-CI are failing on the sporadic TLS issue... and PR #964 should also be merged before to have a green test.

So, I manually tested on AKS with PR #964 also: it passed without any issue.

juadk (approved):

Issue #950, opened 2021-11-02
Use staging for letsencrypt cert in CI


ldevulder:
Because there are limitations/quotas with the production one, and we hit these limitations regularly.

ldevulder:
After many tests it is not so easy to use the staging part of letsencrypt because the CA is not validate by default (hacks are needed) and I also had the TLS issue with the staging.

I saw that if we don't have lot of tests each days then the tests are passing (this part) and also if we have running tests that use letsencrypt at the same time we have failures. So I will check the scheduling of the workflow to ensure that AKS and GKE tests are not run at the same time.

ldevulder:
Reopened because we still have this error:
```
"status": {
        "failureTime": "2021-11-10T14:39:46Z",
        "reason": "Failed to create Order: 429 urn:ietf:params:acme:error:rateLimited: Error creating new order :: too many certificates (5) already issued for this exact set of domains in the last 168 hours: epinio.ldevulder-epinio.suse.dev: see https://letsencrypt.org/docs/rate-limits/",
        "state": "errored"
    }
```

We should try to find a way to bypass or trust the staging CA to be able to use it for the whole test.

jimmykarily:
Let's postpone this until this story is done: https://github.com/epinio/epinio-helm-chart/issues/15

jimmykarily:
The CA issue is fixed with these:

- https://github.com/epinio/helm-charts/pull/43
- https://github.com/epinio/epinio/pull/1082

PR #969, opened 2021-11-08
Re-order CI workflow

ldevulder:
Just to ensure that Public Cloud workflow don't run at the same time.
This could help for sporadic letsencrypt issues on AKS and GKE.

This should improve (or fix?) #950.


juadk (approved):
LGTM

PR #970, opened 2021-11-08
feat: CORS

andreas-kupries:
- install option `--access-control-allow-origin` becomes server deployment env variable `ACCESS_CONTROL_ALLOW_ORIGIN`.
  - Above EV becomes server option `access-control-allow-origin`.
  - gin middleware places option value into response header `Access-Control-Allow-Origin`.

Default is the empty string. No header is set in that case.
Installer, or user has to choose the allowed domains (user can patch server deployment).

### DIFF
--- a/internal/cli/install.go
+++ b/internal/cli/install.go
@@ -110,6 +110,13 @@ var neededOptions = kubernetes.InstallationOptions{
 		Default:     true,
 		Value:       true,
 	},
+	{
+		Name:        "access-control-allow-origin",
+		Description: "Domain allowed to access the API. Comma-separated.",
+		Type:        kubernetes.StringType,
+		Default:     "",
+		Value:       "",
+	},
 	ingressServiceIPOption,
 	{
 		Name:        "s3-access-key-id",
--- a/assets/embedded-files/epinio/server.yaml
+++ b/assets/embedded-files/epinio/server.yaml
@@ -193,16 +193,18 @@ spec:
         - command: ["/epinio", "server"]
           args: ["--port", "80"]
           env:
-            - name: TRACE_LEVEL
-              value: "##trace_level##"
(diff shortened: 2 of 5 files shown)


internal/cli/install.go
```diff
@@ -110,6 +110,13 @@ var neededOptions = kubernetes.InstallationOptions{
 		Default:     true,
 		Value:       true,
 	},
+	{
+		Name:        "access-control-allow-origin",
+		Description: "Domain allowed to access the API. Comma-separated.",
+		Type:        kubernetes.StringType,
+		Default:     "*", // TODO: Better default than `*`, i.e. allow any ?
```
jimmykarily, line 117: I would say the default should the be most secure. That is the empty list (in which case we shouldn't set the header at all?).

andreas-kupries: Done. (f44ac7992f9bc0ea530703320dee7524cfa51bf7, collapsed history)

jimmykarily (approved):

andreas-kupries:
Manual testing:

- `epinio install` without acao option ... Verbose curl to the info endpoint - No header present :heavy_check_mark: 
- `epinio install --access-control-allow-origin '*,ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ'` ... Same curl. Header is present in the output: `< access-control-allow-origin: *,ZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZZ`. :heavy_check_mark:

andreas-kupries:
Reminder ticket to automate QA: #971

jimmykarily (approved):

juadk (approved):
Tested on GKE with and without the `access_control_allow_origin` option.
Curl outputs were good.
I need to think how to automate a more real case, first time with CORS.

Issue #321, opened 2021-04-23
API documentation


kkaempf:
Now that we have an API, we need to document it so others can consume it.

- [x] research on a good way to document APIs
- [x] start with API documentation
- [x] visible in the online documentation (done here: https://github.com/epinio/docs/pull/23)

depends #818 #806

kkaempf:
See https://github.com/epinio/epinio/pull/272

agracey:
#250 as well

jimmykarily:
done

PR #972, opened 2021-11-09
Add initial swagger spec generation for API docs

ldevulder:
Ref #321 

This adds swagger comments into a separate doc package, so they don't interfere with existing comments.  Models are scanned automatically, but routes are not and need to be added manually.

https://goswagger.io/use/spec.html

This is based on the closed PR #272.


internal/api/v1/docs/namespace.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+// swagger:route POST /namespaces/{namespace} namespace NamespaceCreate
+// Create a new named `namespace`.
+// responses:
+//   200: NamespaceCreateResponse
+
+// swagger:parameters NamespaceCreate
+type NamespaceCreateParam struct {
+	Namespace models.NamespaceCreateRequest
```
ldevulder, line 27: I'm not sure about this one: should it be `string` instead of `models.NamespaceCreateRequest`?

andreas-kupries: It should be `string`, and there should be a
```
// in: body
Body models.NamespaceCreateRequest
```
for the POST body.

internal/api/v1/docs/application.go
```diff
@@ -0,0 +1,242 @@
...
+
+// swagger:parameters AppLogs
+type AppLogsParam struct {
+	Namespace string
+	App       string
+}
+
+// swagger:response AppLogsResponse
+type AppLogsResponse struct {
+	// in: body
```
ldevulder, line 81: I haven't found what to add for the `LogsResponse`... Any hints?

internal/api/v1/docs/application.go
```diff
@@ -0,0 +1,242 @@
...
+
+// swagger:parameters StagingLogs
+type StagingLogsParam struct {
+	Namespace string
+	StageId   string
+}
+
+// swagger:response StagingLogsResponse
+type StagingLogsResponse struct {
+	// in: body
```
ldevulder, line 97: I haven't found what to add for the `LogsResponse`... Any hints?

internal/api/v1/docs/application.go
andreas-kupries: Heh. The log endpoints upgrade the connection to a websocket and then transfer log entries over it. They do not have a regular response per se I believe.

A very short google does not net me info on how to describe this for swagger. ... If you have more luck let us use that. And if not I would leave the response off, and make note of the websocket upgrade in the endpoint description.

internal/api/v1/docs/application.go
andreas-kupries: See above.

internal/api/v1/docs/application.go
ldevulder: Yes, it seems that swagger is not able to describe websocket connection.

So I can change the endpoint description to something like this: `Return logs of the named 'application' in the 'namespace' streamed over a websocket.`
WDYT?

ldevulder: I've pushed a comment change, let me know if this is not OK for you.

andreas-kupries: I am :+1: on that change.

jimmykarily (approved):

Makefile
```diff
@@ -147,6 +147,18 @@ embed_files: getstatik wrap_registry_chart
 help:
 	( echo _ _ ___ _____ ________ Overview ; epinio help ; for cmd in apps completion create-org delete help info install orgs push target uninstall ; do echo ; echo _ _ ___ _____ ________ Command $$cmd ; epinio $$cmd --help ; done ; echo ) | tee HELP
 
+
+########################################################################
+# Docs
+
+swagger:
+	swagger generate spec > docs/references/api/swagger.json
```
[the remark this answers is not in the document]
andreas-kupries: I use a binary from https://github.com/go-swagger/go-swagger/releases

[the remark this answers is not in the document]
andreas-kupries: Done. Commit c776ee0a6d909ed4bcab960db8b661c264f8d84a

jimmykarily (approved):

PR #978, opened 2021-11-10
CLI doc generation script, and hugo fixup

andreas-kupries:
No ticket. The code for generating the cli docs is moved into a script, as it is becoming to large for the Makefile.
Added a third post-processing step removing the hugo-related annotations from the front of the files.

Sibling to https://github.com/epinio/docs/pull/24


kkaempf (approved):
lgtm

PR #980, opened 2021-11-11
Persist app origin information to App CRD resource

### LINKED ISSUE #912: Persist information on application deployment method
Right now there is no way to tell how an existing application was deployed (staged from local code, container image or git).
This information is needed in the Web UI in order to decide wether to show a "restage" button or not. See here for more: https://github.com/epinio/ui/issues/10

This information should be part of the application API responses (list, get etc)

andreas-kupries:
Fix #912

CRD sibling PR https://github.com/epinio/application/pull/2

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,22 @@
 ########################################################################
 ## Development
 
-VERSION ?= $(shell git describe --tags)
+# Custom version suffix for use during development. The default, the
+# empty string, ensures that CI, releases, etc. are not affected by
+# this additional variable. During development on the other hand it
+# can be used to provide a string allowing developers to distinguish
+# different binaries built from the same base commit. A simple means
+# for that would be a timestamp. For example, via
+#
+# VSUFFIX="-$(date +%Y-%m-%dT%H-%M-%S)" make ...
+#
+# yielding, for example
+#
+# % ep version
+# Epinio Version: "v0.1.6-16-ge5ad0849-2021-11-18T10-00-27"
+
+VSUFFIX ?= 
+VERSION ?= "$(shell git describe --tags)$(VSUFFIX)"
 CGO_ENABLED ?= 0
 export LDFLAGS += -X github.com/epinio/epinio/internal/version.Version=$(VERSION)
 
--- a/Makefile
+++ b/Makefile
@@ -154,6 +154,11 @@ embed_files: getstatik wrap_registry_chart
 	statik -m -f -src=./assets/embedded-web-files/views -ns webViews -p statikWebViews -dest assets
 	statik -m -f -src=./assets/embedded-web-files/assets -ns webAssets -p statikWebAssets -dest assets
 
+update-app-crd:
+	./scripts/update-app-crd.sh
+# The script in the previous line expects the environment variable REV
+# to be set to the desired revision.
+
 ########################################################################
 # Docs
 
--- a/docs/references/api/swagger.json
+++ b/docs/references/api/swagger.json
@@ -1,8 +1,4 @@
-{
-  "info": {
-    "title": "Epinio",
-    "version": "1"
-  },
+{ "info": {"title": "Epinio", "version":"1"},
   "swagger": "2.0",
   "paths": {
     "/applications": {
@@ -943,6 +939,9 @@
         "meta": {
           "$ref": "#/definitions/AppRef"
         },
+        "origin": {
+          "$ref": "#/definitions/ApplicationOrigin"
+        },
         "status": {
           "$ref": "#/definitions/ApplicationStatus"
         },
@@ -1064,6 +1063,26 @@
       },
       "x-go-package": "github.com/epinio/epinio/pkg/api/core/v1/models"
     },
+    "ApplicationOrigin": {
+      "description": "ApplicationOrigin is the part of the manifest describing the origin of the application\n(sources). At most one of the fields may be specified / not empty.",
+      "type": "object",
+      "properties": {
+        "Container": {
+          "type": "string"
+        },
+        "Git": {
+          "$ref": "#/definitions/GitRef"
+        },
+        "Kind": {
+          "type": "integer",
+          "format": "int64"
+        },
+        "Path": {
+          "type": "string"
+        }
+      },
+      "x-go-package": "github.com/epinio/epinio/pkg/api/core/v1/models"
+    },
     "ApplicationStatus": {
       "type": "string",
       "x-go-package": "github.com/epinio/epinio/pkg/api/core/v1/models"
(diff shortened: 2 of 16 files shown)


andreas-kupries:
I have to work things out a in a bit more detail before I can start implementing things.

1. I found only two places writing to an app CRD resource (`routes` as guideline).
    1.  `internal/application/application.go`, `Create`. This creates the structure in memory and hands to the kube client.
         Via unstructured.
    2. `internal/api/v1/application/update.go`, `Update`. This applies a `Patch` via the kubeclient. @jimmykarily What other methods were available and discarded as not suitable ? (In most places we use kubeclient `Update`, for example).

2. We cannot save app origin on app create. `app create`, the command does not have the information, nor does the `ApplicationCreateRequest`. This is not via for `ap update` either, and the associated `ApplicationUpdateRequest`. We can only set/change origin via `app push`.

3. The `app push` command processes the origin locally, i.e. upload tarball, or from git, stage as needed, deploy. 
    1. `import-git` can be extended server side to save the gitref as origin
    2. `upload` can be extended client and server side to carry the origin path, and then save that as origin.
    3. Container mode is an issue. It has now upload/import endpoint, does no staging. And at the deploy state the server does not know if the image url is from container or staged ... While this could fixed by extending the deploy request to carry a flag to distinguish the cases it feels ... not right. ... I am more thinking to add a new endpoint as sibling of `import-git`, and `upload`, to signal container origin. This gives all three possible origins one endpoint for signaling their use.

All endpoints would call on a shared function to save the origin into the app CRD resource.

jimmykarily:
>        2. `internal/api/v1/application/update.go`, `Update`. This applies a `Patch` via the kubeclient. @jimmykarily What other methods were available and discarded as not suitable ? (In most places we use kubeclient `Update`, for example).

https://github.com/kubernetes/client-go/issues/236#issuecomment-320373496

jimmykarily:
The amount of `switch` and `if` statements we have around application Origin, shows that we need  refactoring. We have 3 different ways to push an app:
- from code
- from a git remote
- from an existing container image

We could tidy things up by creating a "class" for each type. We could have `ApplicationPusher` being an interface (or even `Application` itself) and then 3 implementations, one for each type. The push code would look something like this then (pseudo-code):

```
pusher := getApplicationPusher(appRef)
pusher.PrepareCode()
pusher.SetOrigin()
pusher.Stage()
pusher.Deploy()
```

Some methods on some implementations could be a no-op (e.g. `Stage` when pushing with existing image). If the 3 types of applications behave different in other places, other than push, then maybe Application itself should be the interface.

If you agree, we could do this refactoring here or in another PR. What do you think?

kkaempf:
:+1: for refactoring. Not sure if mixing this with the current PR is a simple task.

andreas-kupries:
Thinking about it I suspect that it will not be that simple ... I am coming down towards prefering a separate PR ... Right now the only thing I still want to do on this one is updating the code comments driving the swagger spec. I forgot about that part in the elation of finally having the tests working again.

Makefile
```diff
@@ -139,6 +140,10 @@ embed_files: getstatik wrap_registry_chart
 	statik -m -f -src=./assets/embedded-web-files/views -ns webViews -p statikWebViews -dest assets
 	statik -m -f -src=./assets/embedded-web-files/assets -ns webAssets -p statikWebAssets -dest assets
 
+update-app-crd:
+	./scripts/update-app-crd.sh
+# The form above expects the environment variable REV to be set to the desired revision
```
jimmykarily, line 145: what form? maybe `form` -> `script` or `form` -> `target` ?

docs/references/api/swagger.json
```diff
@@ -506,6 +502,41 @@
         }
       }
     },
+    "/namespaces/{Namespace}/applications/{App}/from-container": {
```
jimmykarily, line 505: I think this endpoint is not necessary. The other 2 types of application push, set the origin implicitly in code upload or when importing from git. We could set the origin to "app from container" if the deploy endpoint is called and origin hasn't been set yet.

That said, with or without the above, we have some strange situations. E.g.

- Client pushes an application with an existing image -> origin is now set to "container"
- User uploads code for the same application -> origin is now set to "path"
The state of the application now is inconsistent, the origin is set to path but the running application was never staged with that code.

similar situations can be reached with my suggestion above. Ideally, the endpoint that should set the application's origin, is the "deploy" endpoint. Because the origin should always match the currently running image. Now how do we keep track of where an image came from? One idea is to use the image tag. If it exists and if it matches the latest PipelineRun for that application then we know we staged some code. But we still don't know where that code came from (git or path?). Since the only input to the `deploy` endpoint is the image itself, it would be nice if that image had all the information we need. But even if it did, the `deploy` endpoint can't inspect the image because it doesn't even have credentials to pull it. Maybe a good trade-off is to trust the client and accept additional parameters on the `deploy` endpoint. As in: "Hey, deploy this image and trust me, it was built from source pulled from a git remote" or "Hey, deploy this image and trust me, it's coming from a remote git". This requires the client to keep track of how the image was built and make sure to provide that information to the `deploy` endpoint. A bad client could mess up things but that's always the case right?

jimmykarily (changes requested):

Makefile
```diff
@@ -1,7 +1,8 @@
 ########################################################################
 ## Development
 
-VERSION ?= $(shell git describe --tags)
+VSUFFIX ?= 
+VERSION ?= "$(shell git describe --tags)$(VSUFFIX)"
```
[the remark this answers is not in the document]
andreas-kupries: Addressed in rebase with comment in makefile, not just commit message

Makefile
andreas-kupries: Addressed in rebase, modified original commit

docs/references/api/swagger.json
andreas-kupries: Agreed. This is a larger rewrite of the PR ... It also means that the test changes can be dropped, mostly.

andreas-kupries: Done with commit 75a1bc85515b0e6c9667064a5b19cad8f6790f97 and predecessors (included a rebase to the new main without magic domain).

jimmykarily (approved):

jimmykarily:
When pushing with a path, the output shows the original directory: 


```
✔️  Details:
|        KEY        |                       VALUE                       |
|-------------------|---------------------------------------------------|
| Origin            | /home/opensuse/workspace/epinio/assets/sample-app |

```
this includes the username of the user that pushed the code. I'm not sure this is a problem or not. Let's keep it as it is for now. After all, it's likely that some CI pipeline will be doing the push in production setups so the user is not really important.

Issue #975, opened 2021-11-10
Spike: try to re-use the same image on every stage


jimmykarily:
I'm facing a bug (https://github.com/buildpacks/lifecycle/issues/213) and while reading up on docs (https://buildpacks.io/docs/concepts/components/lifecycle/analyze/) I see that:

```
the analyzer was responsible for analyzing the metadata from the cache and the previously built image (if available) to determine what layers can or cannot be reused. 
```

I'm wondering if the fact that every time we stage an application, the APP_IMAGE is different (has the stage id in the name) makes the whole process slower because the previous image is never there and no layers can be re-used (https://github.com/epinio/epinio/blob/main/assets/embedded-files/tekton/buildpacks-task.yaml#L138).

We did this to be able to control when the running app is restarted and which artifact it runs (rather than simply replacing the image under the hood). If this makes the staging process slower though, we should find some other solution.

Let's try to use the same image all the time (no stage id in the name) and see if that improves the performance of staging after the image has been built at least once.

jimmykarily:
NOTE: just in case someone lands here from the linked lifecycle issue, the bug in my case was that the username for the registry was empty (because I messed up the param name in the epinio cli).

jimmykarily:
If this proves to make any difference, we can simply use the running image as `previous-image` in the pipeline. The stage-id should be available on the running deployment.

jimmykarily:
With #991 fixed, we now need to find the previous image tag to be used as `previous-image`

andreas-kupries:
Research Notes:

  - `assets/embedded-files/tekton/buildpacks-task.yaml:138:        - "-previous-image=$(params.APP_IMAGE)"`
  - Above is in the create task.
  - No param for previous image. Yet. 
  - APP_IMAGE is specified by server, at
  - `internal/api/v1/application/stage.go:343:				{Name: "APP_IMAGE", Value: *str(app.ImageURL(app.RegistryURL))},`
  - `ImageUrl` constructs image name as `namespace-appname:stage-id`.
  - The stage id is the only varying part now. Namespace and app name are obviously fixed to the app.
  - We need the stage id of the previous run = The stage id of the deployed app
  - Although, we can have the situation that the app is scaled to zero, effectively stopped. And then push, with a restart implied. So, not necessarily deployed when looking for the old stage id.
  - When the app is deployed the stage id is available in the labels of the pod template in the deployment
[...]

PR #993, opened 2021-11-16
991 app images

### LINKED ISSUE #991: Subsequent image builds for the same app are stored in different image names
**Describe the bug**
When I moved my epinio install to using the new external registry feature, I noticed that each push is creating an entirely new image with the `:latest` tag.

This has a few issues:

- No way to optimize storage in the registry
- Never cache's layers even if the hashes match
- Breaks ability to GC
- Breaks DO's pricing model as now I'll get charged for every `epinio push` instead of per app. 

image

**To Reproduce**

1.  Install Epinio with external registry (I used digital ocean's container registry)
2. Push sample 12 factor app
3. Push again
4. Look at images that are produced

**Expected behavior**
I would expect applications with the same name to push to the same image name and only use different tags based on build id. 

**Additional context**
I'd imagine this bug is not new but only now visible with the ability to use an external registry.

andreas-kupries:
Fix #991 . See also #975 .

Modified the ImageURL handed to tekton staging for the app image to be saved into.
Dropped the StageID.
Hopefully makes our use of the container registry (internal, or external) less expensive.

### DIFF
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -46,7 +46,7 @@ type stageParam struct {
 // ImageID. The ImageURL is later used in app.yml and to send in the
 // stage response.
 func (app *stageParam) ImageURL(registryURL string) string {
-	return fmt.Sprintf("%s/%s:%s", registryURL, app.Name, app.Stage.ID)
+	return fmt.Sprintf("%s/%s-%s:%s", registryURL, app.Namespace, app.Name, app.Stage.ID)
 }
 
 // ensurePVC creates a PVC for the application if one doesn't already exist.
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -46,7 +46,7 @@ type stageParam struct {
 // ImageID. The ImageURL is later used in app.yml and to send in the
 // stage response.
 func (app *stageParam) ImageURL(registryURL string) string {
-	return fmt.Sprintf("%s/%s-%s", registryURL, app.Name, app.Stage.ID)
+	return fmt.Sprintf("%s/%s", registryURL, app.Name)
 }
 
 // ensurePVC creates a PVC for the application if one doesn't already exist.
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -46,7 +46,7 @@ type stageParam struct {
 // ImageID. The ImageURL is later used in app.yml and to send in the
 // stage response.
 func (app *stageParam) ImageURL(registryURL string) string {
-	return fmt.Sprintf("%s/%s", registryURL, app.Name)
+	return fmt.Sprintf("%s/%s:%s", registryURL, app.Name, app.Stage.ID)
 }
 
 // ensurePVC creates a PVC for the application if one doesn't already exist.
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -188,11 +188,13 @@ func (hc Controller) Stage(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err, fmt.Sprintf("failed to create pipeline run: %#v", o))
 	}
 
-	log.Info("staged app", "namespace", namespace, "app", params.AppRef, "uid", uid)
+	imageURL := params.ImageURL(params.RegistryURL)
+
+	log.Info("staged app", "namespace", namespace, "app", params.AppRef, "uid", uid, "image", imageURL)
 
 	response.OKReturn(c, models.StageResponse{
 		Stage:    models.NewStage(uid),
-		ImageURL: params.ImageURL(params.RegistryURL),
+		ImageURL: imageURL,
 	})
 	return nil
 }


andreas-kupries:
Setting up with dockerhub as registry looks to be fine ...

Logs from the first push (still in progress) `ep push --name fox --path assets/sample-app`

Server:
> 2021/11/16 08:53:54 fallback "level"=1 "msg"="staged app"  "app"={"name":"fox","namespace":"workspace"} __"image"="registry.hub.docker.com/a99k/fox"__ "org"="workspace" "uid"="f4dafe253c10586e"

Client:
> 🕞  [f4dafe253c10586e-stage-t4jsk-pod-rjdwj] step-create Saving registry.hub.docker.com/a99k/fox... 
> 🕞  [f4dafe253c10586e-stage-t4jsk-pod-rjdwj] step-create *** Images (sha256:da6e1e171a1741babebc8702541743e72681981ea696cbf2bd2d2c6643cd5aed):
> 🕞  [f4dafe253c10586e-stage-t4jsk-pod-rjdwj] step-create       registry.hub.docker.com/a99k/fox 

Image is present in registry ...

And now a second push `ep push --name fox --path assets/sample-app --bind dog` (service added to have some change)

Server:
> 2021/11/16 09:20:53 fallback "level"=1 "msg"="staged app"  "app"={"name":"fox","namespace":"workspace"} "image"="registry.hub.docker.com/a99k/fox" "org"="workspace" "uid"="ed7ddc5e0773b260"

Client:
> 🕞  [ed7ddc5e0773b260-stage-jznm6-pod-hj9qb] step-create Saving registry.hub.docker.com/a99k/fox... 
> 🕞  [ed7ddc5e0773b260-stage-jznm6-pod-hj9qb] step-create *** Images (sha256:df9d3e901069b5a1120c03f42f02c6bc8bfc167ccd089e20592fd6eef18f19d1):
> 🕞  [ed7ddc5e0773b260-stage-jznm6-pod-hj9qb] step-create       registry.hub.docker.com/a99k/fox 

The sha256 changes between the two pushes ... And I see the sha256 of the second push shown as digest in the hub, and push time has updated properly also ...

Ah ... Check the deployment spec ... Forgot that for the first push ...

>         image: registry.hub.docker.com/a99k/fox

Ok, no tag, no sha256 ... I can't say what image is now active.

> workspace         fox-554b5c6b8d-kvlhl                            2/2    Running    0         6m50s

Was not quick enough to check if the pod was restarted ... 
Saved deployment and pod data now for comparison after a 3rd push.

Ah ... looking over the pod (instead of the pod template in deployment) I see

>   image: registry.hub.docker.com/a99k/fox:latest
    imageID: registry.hub.docker.com/a99k/fox@sha256:df9d3e901069b5a1120c03f42f02c6bc8bfc167ccd089e20592fd6eef18f19d1

That is the sha256 of the new image from the second push. So, pod had to be restarted for that.

And now I also see in the client log of the second push the line

> 🕞  [ed7ddc5e0773b260-stage-jznm6-pod-hj9qb] step-create Reusing layers from image 'registry.hub.docker.com/a99k/fox@sha256:da6e1e171a1741babebc8702541743e72681981ea696cbf2bd2d2c6643cd5aed' 

And the `da6e1e` is the image from the first push, see way up top in this comment.

Third push confirms ... App pod is restarted ...

@jimmykarily The change I made with PR seems to be working. Regarding https://github.com/epinio/epinio/issues/975#issuecomment-964886709, i.e.

> If this proves to make any difference, we can simply use the running image as previous-image in the pipeline. The stage-id should be available on the running deployment.

I do not fully understand the comment. Could you explain a bit more ?
Note that I found the following in the `assets/embedded-files/tekton/buildpacks-task.yaml` file:

```
    - name: create
      image: $(params.BUILDER_IMAGE)
      imagePullPolicy: IfNotPresent
      command: ["/cnb/lifecycle/creator"]
      args:
        - (...)
        - "-previous-image=$(params.APP_IMAGE)"
        - (...)
        - "$(params.APP_IMAGE)"
```

kkaempf (dismissed):
lgtm

andreas-kupries:
Note @kkaempf Huddled with @jimmykarily and decided to try to keep the stage id in the image url, however as tag instead of as part of the name, i.e. `app-id` becomes `app:id` ... That way we would have a more explicit tag instead of the implicit `latest`.

Needs a second round of testing, of course.

And I see a conflict, so I will rebase also ...

kkaempf:
I'd guess we need to branch 0.1.6 for the 0.2.0 milestone release.

andreas-kupries:
Rebase ok.

First app push as expected. Properly tagged.

> 🕞  [a78fed7f40b3fe6d-stage-jfn6q-pod-rxzzf] step-create Saving registry.hub.docker.com/a99k/hen:a78fed7f40b3fe6d... 
🕞  [a78fed7f40b3fe6d-stage-jfn6q-pod-rxzzf] step-create *** Images (sha256:7cce1eb036320b3138bb3ccf821270b58e52384bf61354c8ce2443167415e4db): 
🕞  [a78fed7f40b3fe6d-stage-jfn6q-pod-rxzzf] step-create       registry.hub.docker.com/a99k/hen:a78fed7f40b3fe6d 

Second push. ... Tags show up properly, separately, in the hub for the image
( https://hub.docker.com/repository/docker/a99k/hen )

> 🕞  [13316cd766aaeff7-stage-xq4s4-pod-6r8v8] step-create Saving registry.hub.docker.com/a99k/hen:13316cd766aaeff7... 
🕞  [13316cd766aaeff7-stage-xq4s4-pod-6r8v8] step-create *** Images (sha256:dfe76b234814878519b30526ecd866bbbbbf7f891aaecae9c9cdd936178e8cdc): 
🕞  [13316cd766aaeff7-stage-xq4s4-pod-6r8v8] step-create       registry.hub.docker.com/a99k/hen:13316cd766aaeff7 

Comparing deployments and pods ...

The deployment's generation counter increments, the resourceVersion as well. Image is properly tagged in the spec.template.
Ditto for pod.

The `reuse` line from staging is gone. However as talked about with @jimmykarily in the huddle, that can be fixed as part of #975 . Here we focus on the image storage, and that works as well as the previous commit, arguably better, because now the app image revisions are properly tagged, and distinguishable in the registry dash board, while not considered as separate images.

@agracey I believe that this is what we should go forward with.

jimmykarily:
iirc, Epinio guarantees the uniqueness of the application name in a namespace. The image doesn't have the namespace in it: 

```
epinio-registry.10.86.4.38.omg.howdoi.website/apps/sample:53d85148e04a6712
```

Include the namespace name so that we avoid conflicts? I known it was like that before but it was almost impossible to happen due to the name having the stage ID in it.

jimmykarily:
> Include the namespace name so that we avoid conflicts? I known it was like that before but it was almost impossible to happen due to the name having the stage ID in it.

To be 100% correct, there is still no possibility for an app to run another app's image, because we include the tag in the image (not just `:latest`). But it will be confusing to push 2 different images under the same name (although with different tags)

andreas-kupries:
Yeah. ... Note, with routes we originally assumes uniques of app name in the system. With multi-route support we better get down on the unique per namespace.

jimmykarily (approved):

agracey:
LGTM

kkaempf:
This should go in before we release `v0.2.0`.
Setting labels accordingly.

PR #995, opened 2021-11-16
Rename "org" to "namespace" everywhere

jimmykarily:


andreas-kupries (approved):
LGTM after relatively quick scan ...

@richard-cox Merging this PR affects the API. Field names in requests/responses indicating `org`, have change to `namespace`.

PR #996, opened 2021-11-16
Don't break `epinio uninstall` after `epinio install-ingress`

### LINKED ISSUE #990: Uninstall breaks when using external S3 storage
**Describe the bug**
When I install with an external S3 storage then uninstall, I get this error: 

```
⚠️  Uninstall of minio-operator failed: Deleting minio/tenant.yaml failed:
error: unable to recognize "/tmp/epinio726000816": no matches for kind "Tenant" in version "minio.min.io/v2"
: exit status 1
```

**To Reproduce**
1. Install epinio with the `--s3-...` flags (I used digital ocean's spaces)
2. `epinio uninstall`

**Expected behavior**
The uninstaller should see the lack of minio tenants owned by it and skip as unneeded.

jimmykarily:
because Minio is not even installed. Although we set --ignore-not-found
in KubectlDeleteEmbeddedYaml, the removal of the Minio yamls fails when
Minio is not even installed. This commit moves the check for a Minio
namespace before the removal.

Fixes #990

### DIFF
--- a/deployments/minio.go
+++ b/deployments/minio.go
@@ -58,14 +58,6 @@ func (k Minio) PostDeleteCheck(ctx context.Context, c *kubernetes.Cluster, ui *t
 func (k Minio) Delete(ctx context.Context, c *kubernetes.Cluster, ui *termui.UI) error {
 	ui.Note().KeeplineUnder(1).Msg("Removing Minio...")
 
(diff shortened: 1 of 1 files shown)


agracey (approved):

agracey:
Easy enough of a change

PR #1002, opened 2021-11-17
Fix CI trigger on AKS, EKS, GKE and RKE

ldevulder:
CI will be triggered only if linter result is OK and when the PR is set to "Ready To Review". This will avoid long CI tests on each push and also limit the Public Cloud costs.

### DIFF
--- a/.github/workflows/aks.yml
+++ b/.github/workflows/aks.yml
@@ -2,6 +2,8 @@ name: AKS-CI
 
 on:
   pull_request:
+    branches: [ main ]
+    types: [ ready_for_review ]
     paths:
       - 'acceptance/install/scenario5_test.go'
       - '.github/workflows/aks.yml'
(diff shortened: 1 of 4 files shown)


juadk (approved):
Looks good to me, let's see later to factorize the linter job, I think it's possible using template.

.github/workflows/aks.yml
```diff
@@ -35,7 +37,34 @@ env:
   AKS_MACHINE_TYPE: 'Standard_D3_v2'
 
 jobs:
+  linter:
+    runs-on: self-hosted
```
juadk, line 41: I wonder why we are using self-hosted runners for that job. I expected to see ubuntu with github runners.
Maybe to avoid limitations on free github actions?

ldevulder: Maybe yes, we can check this outside of this PR anyway.

ldevulder:
AKS and GKE are failing as usual because of the TLS issue. So we can merge it as the other CI tests are green.

Issue #1005, opened 2021-11-22
Fix list all applications flake


jimmykarily:
Looking at the failed CI runs, there is a test that fails a lot (https://github.com/epinio/epinio/blob/main/acceptance/api/v1/applications_test.go#L717).

thardeck:
Steps to reproduce:

- run against at least 3 threads
- increase trace-level
- curl the api in a loop and wait until a 500 return

PR #1009, opened 2021-11-22
Simplify the creation of the Epinio private CA

jimmykarily:
we were creating the empty secret to be able to write the Kubed
annotation on it. This was needed for TraefikForwardAuth which we no
longer use. Now we let cert-manager create the secret.

This commit also fixed a log message refering to the wrong yaml

### DIFF
--- a/deployments/cert_manager.go
+++ b/deployments/cert_manager.go
@@ -14,7 +14,6 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/kyokomi/emoji"
 	"github.com/pkg/errors"
-	v1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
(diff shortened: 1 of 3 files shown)


deployments/cert_manager.go
```diff
@@ -297,7 +270,7 @@ func (cm CertManager) apply(ctx context.Context, c *kubernetes.Cluster, ui *term
 		"spec" : {
 			"isCA"       : true,
 			"commonName" : "epinio-ca",
-			"secretName" : "%s",
+			"secretName" : "epinio-ca-root",
```
andreas-kupries, line 273: I would (still) prefer to have this in the `rootCAName` constant, placed at the beginning, easy to find.

andreas-kupries (changes requested):
Only a slight change to make it easier on future us to find what is, essentially, system configuration.

deployments/cert_manager.go
jimmykarily: Done: https://github.com/epinio/epinio/pull/1009/commits/d71e84f0a2943e7ffb579e79a70a38090b549cd2

andreas-kupries (approved):

jimmykarily:
The regular flake #1005 . Merging.

PR #1015, opened 2021-11-23
Validate blob issued to staging (must match app)

### LINKED ISSUE #801: Validate blob belongs to the app when staging
Depends on #699

When the user uploads the application code, we [store the application metadata](https://github.com/epinio/epinio/pull/788/files#diff-4c92e03256b6a12e87b841840188789dbcf00fcb04087875af0779129ac3cf3bR79-R81) on the object itself. We do the same [when we pull from git](https://github.com/epinio/epinio/pull/788/files#diff-676e4bf1f445a3687fbd3165180a5d030da0b4438d7a5ae481f67ebb758e4054R83-R85)

We can use these metadata to validate that the blob the client sends on the stage request actually matches the application on the same request. This should prevent someone from staging an app with the sources of another app.

Given everyone can do everything on Epinio right now (we don't have user roles), this is mostly for validation against code mistakes on the client (e.g. the web UI).

After discussing it with @manno, we think it could even be a feature to be able to deploy applications with the blob of another app. E.g. Someone updates the sources of a wordpress application and everyone else is staging their apps using those (updated) sources. To make it a useful feature, it would require a lot more that simply allowing this.

TBD: To validate or not to validate?

andreas-kupries:
Fixes #801

feat: extended s3 manager with method to retrieve user meta data
feat: extended stage endpoint with validation of blob meta data

### DIFF
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -134,6 +134,48 @@ func (hc Controller) Stage(c *gin.Context) apierror.APIErrors {
 		return apierror.NewBadRequest("pipelinerun for image ID still running")
 	}
 
+	// Validate incoming blob id before attempting to stage
+
+	s3ConnectionDetails, err := s3manager.GetConnectionDetails(ctx, cluster, deployments.TektonStagingNamespace, deployments.S3ConnectionDetailsSecret)
+	if err != nil {
+		return apierror.InternalError(err, "failed to fetch the S3 connection details")
+	}
+
+	manager, err := s3manager.New(s3ConnectionDetails)
+	if err != nil {
+		return apierror.InternalError(err, "creating an S3 manager")
+	}
+
+	// Validate incoming blob id
+
+	blobMeta, err := manager.Meta(ctx, req.BlobUID)
+	if err != nil {
+		return apierror.InternalError(err, "querying blob id meta-data")
+	}
+
+	blobApp, ok := blobMeta["App"]
+	if !ok {
+		return apierror.NewInternalError("blob has no app name meta data")
+	}
+	if blobApp != req.App.Name {
+		return apierror.NewInternalError(
+			"blob app mismatch",
+			"expected: "+req.App.Name,
+			"found: "+blobApp)
+	}
+	blobNamespace, ok := blobMeta["Namespace"]
+	if !ok {
+		return apierror.NewInternalError("blob has no namespace meta data")
+	}
+	if blobNamespace != req.App.Namespace {
+		return apierror.NewInternalError(
+			"blob namespace mismatch",
+			"expected: "+req.App.Namespace,
+			"found: "+blobNamespace)
+	}
+
+	// Create uid identifying the staging pipeline to be
+
 	uid, err := randstr.Hex16()
 	if err != nil {
 		return apierror.InternalError(err, "failed to generate a uid")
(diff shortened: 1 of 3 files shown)


internal/s3manager/s3manager.go
```diff
@@ -158,6 +158,21 @@ region = %s
 	return secret, err
 }
 
+// Stat retrieves the meta data for the blob specified by it blobUID.
```
jimmykarily, line 161:
```suggestion
// Meta retrieves the meta data for the blob specified by it blobUID.
```

internal/s3manager/s3manager.go
```diff
@@ -158,6 +158,21 @@ region = %s
...
+func (m *Manager) Meta(ctx context.Context, blobUID string) (map[string]string, error) {
+	if err := m.EnsureBucket(ctx); err != nil {
```
jimmykarily, line 163: This line will make a request to S3 every time. If the bucket is not there yet, it will create it but this method will fail later because obviously the object doesn't exist (since the bucket didn't exist either). We could as well let it fail because of the missing bucket. After all, it's a bit unexpected for this "getter" method to create the bucket. One can argue that it's also the case for the `Upload` method which does the same but at least there, the creation of the bucket is happening so that the method doesn't fail.

In other words, we can remove this call to `EnsureBucket` here.

internal/api/v1/application/stage.go
```diff
@@ -134,6 +134,48 @@ func (hc Controller) Stage(c *gin.Context) apierror.APIErrors {
...
 	}
 
+	// Validate incoming blob id before attempting to stage
+
+	s3ConnectionDetails, err := s3manager.GetConnectionDetails(ctx, cluster, deployments.TektonStagingNamespace, deployments.S3ConnectionDetailsSecret)
+	if err != nil {
+		return apierror.InternalError(err, "failed to fetch the S3 connection details")
+	}
+
+	manager, err := s3manager.New(s3ConnectionDetails)
```
jimmykarily, line 144: All the additions below this point can be extracted to a function that does what the comment above says. E.g.

```
validateBlob(req.BlobUID, req.App, s3ConnectionDetails)
```

jimmykarily (changes requested):
Can we add a test for this? Something like:

```
Create app1 and upload its code
Create app2 and upload its code
Try to stage app2 with the blobUID of app1
Expect failure
```

internal/s3manager/s3manager.go
andreas-kupries: Ok. Will do. Will also look into writing a test.

internal/api/v1/application/stage.go
andreas-kupries: Will do.

andreas-kupries: commit 189a36981220ee32960d620612dadaf51204995d

internal/s3manager/s3manager.go
andreas-kupries: commit 189a36981220ee32960d620612dadaf51204995d

todo: test

andreas-kupries: test commit 96a76c10c5f8d2ca795b8463549f51fb92bb6d78

andreas-kupries:
The failed test is some cert issue in the installation

jimmykarily (approved):

Issue #1006, opened 2021-11-22
[Spike] Declarative Epinio installer


jimmykarily:
In order to better support:

- Upgrades
- Hotfixes
- Idempotency
- Inspection of components before installation

we can make the installation work with a manifest that described the installation and dependencies between components.

An example of a manifest that would install some of the Epinio components could look like this:

```
# User specified values (could come from a file)
values:
  system-domain: "{{ .Values.domain }}"
  loadbalancerIP: "{{ .Values.loadbalancerIP }}"

generate:
  - name: linkerdCerts
    type: cert
    tls:
      ca: linkerdCA
      crt: linkerdCrt
      key: linkerdKey

# The static part
components:
  - id: linkerd
    type: yaml
    source:
      path: assets/linkerd.yaml

  - id: traefik
    needs: linkerd
    linkerd-inject: true
    type: helm
    wait: true # tell helm to "--wait"
    waitComplete:
      - type: "pod"
        selector: "app.kubernetes.io/name=traefik"
      - type: "loadbalancer-has-service"
    source:
      url: https://...
      name: traefik
      chart: traefik/traefik
      version: 1.2.3
    values:
      - name: globalArguments
        value: ""
      - name: "deployment.podAnnotations.linkerd\\.io/inject"
        value: "enabled"
      - name: "ports.web.redirectTo"
        value": "websecure"
      - name: "ingressClass.enabled"
        value: "true"
      - name: "ingressClass.isDefaultClass"
        value: "true"
      - name: "service.spec.loadBalancerIP"
        value: "{{ .Values.loadbalancerIP }}"

  - id: cert-manager
    type: helm
    needs: traefik
    waitComplete:
      - type: "pod"
        selector: "app.kubernetes.io/name=webhook"
        namespace: "cert-manager"
      - type: "pod"
        selector: "app.kubernetes.io/name=cainjector"
        namespace: "cert-manager"
      - type: "pod"
        selector: "app.kubernetes.io/name=cert-manager"
        namespace: "cert-manager"

  - id: cluster-issuers
    type: yaml
    source:
      path: assets/cluster-issuers.yaml
    values: # Yes, flat yamls can be templates!
      - name: emailAddress
        value: "{{ .Values.emailAddress }}"
    needs: cert-manager

  - id: cluster-certificates
    type: yaml
    source:
      path: assets/cluster-certificates.yaml
    needs: cluster-issuers

  - id: tekton-pipelines
    needs: cert-manager
[...]

PR #1016, opened 2021-11-23
CI: use letsencrypt-staging

### LINKED ISSUE #950: Use staging for letsencrypt cert in CI

ldevulder:
This is to bypass the quotas on letsencrypt-production.

This should fix #950.

### DIFF
--- a/.github/workflows/aks.yml
+++ b/.github/workflows/aks.yml
@@ -101,23 +101,6 @@ jobs:
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
           aws-region: ${{ env.AWS_REGION }}
 
-      - name: Add letsencrypt-staging CA
-        # We need to use pwsh as a shell to add the certificates
-        shell: pwsh
-        run: |
-          $FILE_LIST = "letsencrypt-stg-int-r3.pem","letsencrypt-stg-int-e1.pem"
-          foreach ($CERT_FILE in $FILE_LIST)
-          {
-            Invoke-WebRequest -Uri https://letsencrypt.org/certs/staging/$CERT_FILE -OutFile $CERT_FILE
-            certutil –addstore -f CA $CERT_FILE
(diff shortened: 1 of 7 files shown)


andreas-kupries (dismissed):
lgtm

olblak (approved):

ldevulder:
After a debugging session with @jimmykarily we found that epinio blocks in the `ANALYZING` part during the app deployment because `getRegistryCAHash` returns an empty hash, and this is because there is no `ca.crt` secret: https://github.com/epinio/epinio/blob/f6cacb5c9fbdcf67d475665414d5e4d20e49ab31/deployments/tekton.go#L315
There was a [bug](https://github.com/jetstack/cert-manager/issues/2111) for this but it has been closed (why?).

This is linked top the issue we had in the installer, see https://github.com/epinio/epinio/issues/1006#issuecomment-981381251.

jimmykarily:
Since tekton only needs to trust the registry certificate (and not the CA in general), maybe we can fix it by mounting the registry certificate in the tekton pod. The certificate always exists, the CA may not (as explained in Loic's comment above).

We need to give that a try.

jimmykarily:
> There was a [bug](https://github.com/jetstack/cert-manager/issues/2111) for this but it has been closed (why?).

This was their answer: https://github.com/jetstack/cert-manager/issues/1571#issuecomment-486661887 (on a relevant issue).
People seem to disagree, but they don't seem to care :D

ldevulder:
Fixed by these PRs: https://github.com/epinio/epinio/pull/1082 and https://github.com/epinio/helm-charts/pull/43.

juadk (approved):
LGTM!

jimmykarily (approved):
AKS-CI is failing but we will fix it on another PR. Needs investigation, we can't reproduce it outside CI yet.

Issue #994, opened 2021-11-16
Test s3 storage feature in one of our CI scenarios


juadk:
If I'm not wrong, we don't have a scenario where we use S3 storage yet.
It can be good to test it in one of them.

juadk:
Successfully tested with Epinio 0.2.0 + GKE + Helm chart deployment + S3 (manually)

ldevulder:
PR merged, issue can be closed.

PR #1019, opened 2021-11-24
CI: add S3 storage tests in EKS

ldevulder:
Also in this commit:
- remove the 2 stages installation
- test the app deletion (on S3 storage)

This should fixe #994

**NOTE :** the removal of the 2 stages installation has been removed, as this adds an issue with Route53 DNS update. Anyway, this part will be changed with the new installer, so better to wait for it.

### DIFF
--- a/.github/workflows/eks.yml
+++ b/.github/workflows/eks.yml
@@ -35,6 +35,7 @@ on:
 env:
   SETUP_GO_VERSION: '^1.17.2'
   GINKGO_NODES: 1
+  FLAKE_ATTEMPTS: 1
   AWS_REGION: 'us-east-2'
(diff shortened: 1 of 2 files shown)


ldevulder:
I did a manual test (on a EKS cluster, tested like the CI with `make test-acceptance-install`) and I didn't had the error message I have on the CI check:
```
error setting into app environment: making the request: Post "https://epinio.id4490-***/api/v1/namespaces/workspace/applications/apps-578358000/environment": dial tcp: lookup epinio.id4490-*** on 1.1.1.1:53: no such host
```

The main difference is that I used my private AWS credentials for my manual test. I'm waiting for the CI credentials to be able to redo the test with them and check if there are some permissions issues.

juadk (approved):
LGTM

PR #1023, opened 2021-11-25
Provide previous app image to tekton for re-use.

### LINKED ISSUE #975: Spike: try to re-use the same image on every stage

andreas-kupries:
fix #975 

consider it draft, with some more testing needed to confirm that the proper information is passed around.
it does not seem to break pushing, and on second push the `image not message` seen in any first push is gone.

the second commit extends the code a bit to attach the previous stage-id to the new pipelinerun resource, as a label.
confirmed with this that the existing code truly puts in the previous stage id for the run to reuse.

### DIFF
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -250,18 +250,7 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
 // happened, yet there is no workload. Ee.g. by calling the "staging" endpoint but not calling the "deploy"
 // endpoint. Since our client doesn't support that scenario, this method doesn't support it either.
 func PreviousStageID(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
-	wl := NewWorkload(cluster, appRef)
-
-	deployment, err := wl.Deployment(ctx)
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return "", err
-		}
-		// App is inactive, no deployment, no workload, no id
-		return "", nil
-	}
-
-	return wl.GetStageID(ctx, deployment), nil
+	return NewWorkload(cluster, appRef).GetStageID(ctx)
 }
 
 // Unstage removes staging resources. It deletes either all PipelineRuns of the
@@ -414,20 +403,8 @@ func fetch(ctx context.Context, cluster *kubernetes.Cluster, app *models.App) er
 	// Check if app is active, and if yes, fill the associated parts.
 	// May have to straighten the workload structure a bit further.
 
-	wl := NewWorkload(cluster, app.Meta)
-
-	deployment, err := wl.Deployment(ctx)
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return err
-		}
-		// App is inactive, no deployment, no workload
-		return nil
-	}
-
-	app.Workload = wl.Get(ctx, deployment)
-
-	return nil
+	app.Workload, err = NewWorkload(cluster, app.Meta).Get(ctx)
+	return err
 }
 
 // calculateStatus sets the Status field of the App object.
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -245,7 +245,11 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
 		PersistentVolumeClaims(deployments.TektonStagingNamespace).Delete(ctx, appRef.MakePVCName(), metav1.DeleteOptions{})
 }
 
-func PreviousStageId(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
+// PreviousStageID returns the stage ID of the previous build, if one exists. It returns an empty string otherwise.
+// This method relies on the presence of a workload to get the previous id. There is the case that staging has
+// happened, yet there is no workload. Ee.g. by calling the "staging" endpoint but not calling the "deploy"
+// endpoint. Since our client doesn't support that scenario, this method doesn't support it either.
+func PreviousStageID(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
 	wl := NewWorkload(cluster, appRef)
 
 	deployment, err := wl.Deployment(ctx)
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -261,13 +261,7 @@ func PreviousStageID(ctx context.Context, cluster *kubernetes.Cluster, appRef mo
 		return "", nil
 	}
 
-	workload := wl.Get(ctx, deployment)
-	if workload == nil {
-		// No workload. No id
-		return "", nil
-	}
-
-	return workload.StageID, nil
+	return wl.GetStageID(ctx, deployment), nil
 }
 
 // Unstage removes staging resources. It deletes either all PipelineRuns of the
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -245,6 +245,27 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
 		PersistentVolumeClaims(deployments.TektonStagingNamespace).Delete(ctx, appRef.MakePVCName(), metav1.DeleteOptions{})
 }
 
+func PreviousStageId(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
+	wl := NewWorkload(cluster, appRef)
+
+	deployment, err := wl.Deployment(ctx)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return "", err
+		}
+		// App is inactive, no deployment, no workload, no id
+		return "", nil
+	}
+
+	workload := wl.Get(ctx, deployment)
+	if workload == nil {
+		// No workload. No id
+		return "", nil
+	}
+
+	return workload.StageID, nil
+}
+
 // Unstage removes staging resources. It deletes either all PipelineRuns of the
 // named application, or all but stageIDCurrent. It also deletes the staged
 // objects from the S3 storage.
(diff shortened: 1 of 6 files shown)


andreas-kupries:
# Previous Image - Timing Changes

## Expectations

  - With the changes it is expected that the time to push and deploy the app is notably lower for second to fourth push, versus the first push.

  - Without the changes it is expected that the Time to push and deploy the app is roughly the same for all pushes, without the supplication of the previous image.

## Results

The expectations were met for the changed setup, and not met for unchanged setup.

Basically, both setups show a much faster push time for the second and further pushes. I suspect that other caching exist in the system, like for the sources themselves causing this.

In a direct comparison I can however note that with the change the second and further pushes are consistently about 13-17 seconds faster still (wall clock) than the pushes without.

### Comparison

With change to the left, without to the right.

|#   |Wall Time|User Time|Sys Time|   |Wall Time|User Time|Sys Time|
|---:|---:     |---:     |---:    |---|---:     |---:     |---:    |
|1   |4m46.132s|0m0.072s |0m0.050s|   |3m0.134s |0m0.069s |0m0.030s|
|2   |1m01.922s|0m0.062s |0m0.019s|   |1m14.896s|0m0.045s |0m0.033s|
|3   |1m01.921s|0m0.063s |0m0.020s|   |1m17.716s|0m0.063s |0m0.014s|
|4   |1m01.934s|0m0.055s |0m0.027s|   |1m18.722s|0m0.060s |0m0.017s|
|5   |0m57.917s|0m0.069s |0m0.009s|   |1m16.729s|0m0.060s |0m0.020s|

## Sources

  - Testing of the change was done with commit 32710af4c2d106558796709a779ae4377e73058e.

  - Testing without the change was done with commit 8227440c5ffabb4fffd02facdf32caee18f5aac9.

## Methodology

The command

  - `time epinio push --name fox --path assets/sample-app`

was executed 5 times on the same cluster and epinio installation.

__ATTENTION__: The 5 second wait for user confirmation at the beginning of each `push` was __not__ shortened by manual intervention. This was done to ensure consistent timing. We can subtract that period from the final time at the end.

## Testing

### Changed - 32710af4c2d106558796709a779ae4377e73058e

|#   |Wall Time|User Time|Sys Time|Notes|
|---:|---:     |---:     |---:    |---	|
|1   |4m46.132s|0m0.072s |0m0.050s|[1]	|
|2   |1m01.922s|0m0.062s |0m0.019s|[2]	|
|3   |1m01.921s|0m0.063s |0m0.020s|	|
|4   |1m01.934s|0m0.055s |0m0.027s|	|
|5   |0m57.917s|0m0.069s |0m0.009s|	|

[1]
```
step-create Previous image with name "epinio-registry.172.19.0.4.omg.howdoi.website/apps/workspace-fox:b225d20068db4503" not found 
```

[2]
```
step-create Restoring metadata for "paketo-buildpacks/ca-certificates:helper" from app image 
step-create Restoring metadata for "paketo-buildpacks/httpd:httpd" from app image 
step-create Restoring metadata for "paketo-buildpacks/php-dist:php" from app image 
step-create Restoring metadata for "paketo-buildpacks/php-web:php-web" from app image 
```

### Unchanged - 8227440c5ffabb4fffd02facdf32caee18f5aac9

|#   |Wall Time|User Time|Sys Time|Notes|
|---:|---:     |---:     |---:    |---	|
|1   |3m0.134s |0m0.069s |0m0.030s|[1]	|
|2   |1m14.896s|0m0.045s |0m0.033s|[1]	|
|3   |1m17.716s|0m0.063s |0m0.014s|[1]	|
|4   |1m18.722s|0m0.060s |0m0.017s|[1]	|
|5   |1m16.729s|0m0.060s |0m0.020s|[1]	|

jimmykarily:
The reason all pushes are faster the second time is because of the `cache` volume: https://github.com/epinio/epinio/blob/main/internal/api/v1/application/stage.go#L357-L362 (search for `ensurePVC` too). What we are looking for here is additional improvement due to using layers from a previous build. It seems that something is happening indeed but I guess it all depends on the number of layer than can be re-used and such. I think we should set the previous image anyway and let the buildpack figure out how to make the best of it.

andreas-kupries:
I fully agree. Without setting the previous image the system does not even have a chance for reuse. And as setting it is certainly not worse by what I have seen, I see no reason not to.

internal/application/application.go
```diff
@@ -245,6 +245,27 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
 		PersistentVolumeClaims(deployments.TektonStagingNamespace).Delete(ctx, appRef.MakePVCName(), metav1.DeleteOptions{})
 }
 
+func PreviousStageId(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
```
jimmykarily, line 248:
```suggestion
// PreviousStageID return the stage ID of the previous build if one exists. It returns an empty string otherwise.
// This method relies on a workload to get the previous id. There is the case that a stage has happened but there
// is no workload (e.g. by calling the "staging" endpoint but not calling the "deploy" endpoint). Since our client
// doesn't support that, this method doesn't support that either.
func PreviousStageID(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (string, error) {
```

jimmykarily: Notice the capitalization of the method name. Needs the callers to changes as well. My vim plugins complain about Go conventions.

internal/application/application.go
```diff
@@ -245,6 +245,27 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
...
+	deployment, err := wl.Deployment(ctx)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return "", err
+		}
+		// App is inactive, no deployment, no workload, no id
+		return "", nil
+	}
+
+	workload := wl.Get(ctx, deployment)
```
jimmykarily, line 260: This `Get` is doing a little bit too much for what we need here (listing ingresses etc). Maybe it's better if implemented a method specifically for this use. E.g. `wl.GetStageID`. It should only get the needed field from the deployment (in one request against the kube api).

jimmykarily (changes requested):

internal/application/application.go
andreas-kupries: Oh, what ?

 ... Do I see this correctly that `Get` is given the deployment in question as argument, and then still asks the kube cluster for it by searching via a List() call ?

andreas-kupries: The only caller looks to be in the adjacent `application.go` and provides the deployment. This means the function can be massively simplified, it looks like.

andreas-kupries: Changes done ... Specialized get for stage id. Simplified `Get`. Rebased to current main. Commit 96cb80e8c4d01c448661588dec6ca0263a8af82a

internal/application/workload.go
```diff
@@ -353,48 +353,26 @@ func (a *Workload) Restarts(ctx context.Context) (int32, error) {
...
 
-	if err != nil {
-		status = pkgerrors.Wrap(err, "failed to get Deployment status").Error()
-	} else if len(deployments.Items) < 1 {
-		status = "0/0"
-	} else {
-		desiredReplicas = deployments.Items[0].Status.Replicas
-		readyReplicas = deployments.Items[0].Status.ReadyReplicas
+// Get returns the state of the app deployment encoded in the workload.
+func (a *Workload) Get(ctx context.Context, deployment *appsv1.Deployment) *models.AppDeployment {
```
jimmykarily, line 365: This cleanup was good but I think we can improve it further. The only consumer or `Get` (https://github.com/epinio/epinio/blob/main/internal/application/application.go#L400-L409), first finds the `Deployment` and then it passes it to `Get`. Can't `Get` simply find the deployment itself by calling `Deployment()` on the Workload object?

This way the signature of this method is simplified: `func (a *Workload) Get(ctx, context.Context) *models.AppDeployment`

Looking at the caller, we should just make sure we still return an `apierrors.IsNotFound(err)` error if a deployment doesn't exist.

jimmykarily: Now that we have `GetStageID` too, a caller could benefit from having the deployment in params by not having to fetch the deployment twice if both `GetStageID` and `Get` need to be called. Since we don't have a caller like that, we should probably not worry about this. If it's a problem in the future, we can maybe memoize the `Deployment()` result. But given the sideeffects and complication introduce by caching things, I would wait until that is needed.

andreas-kupries: Will look into this after the townhall. ... Might be easier to treat not found as `no error, return nil data`. That is what the caller expects and uses anyway when no deployment is present.

andreas-kupries: Commit 04b00c4a0f149b0e8310143fa38772565f52fed7

internal/application/application.go
```diff
@@ -250,18 +250,7 @@ func deleteStagePVC(ctx context.Context, cluster *kubernetes.Cluster, appRef mod
...
-	if err != nil {
-		if !apierrors.IsNotFound(err) {
-			return "", err
-		}
-		// App is inactive, no deployment, no workload, no id
-		return "", nil
-	}
-
-	return wl.GetStageID(ctx, deployment), nil
+	return NewWorkload(cluster, appRef).GetStageID(ctx)
```
jimmykarily, line 253: The only thing that makes this stage ID "the previous one" is the fact that the caller gets it before we stage again. The method itself simply returns the stage ID of the currently running app. It's up to the caller to decide when to call this. I think this method should be called just `StageID` . The caller can store it in a variable named "previousStageID" or something that makes sense.

andreas-kupries: True.

andreas-kupries: commit 9d3a5f190cfa088f86e69e91296b81e23833fe06

jimmykarily (approved):

PR #1024, opened 2021-11-26
Make epinio server unaware of the force-kube-internal-registry-tls flag

jimmykarily:
This will allow the declarative installer to use the registry localhost trick.
Now the epinio server relies only on the existence of a localhost url in the
connection details secret to decide wether to rewrite the image url or
not.

It's up to the installer (any of them) to decide if a localhost url
should be in the secret. Our current installer uses the force-kube**
flag, the declarative installer can simply allow an array of registry
urls to be passed through helm values.

### DIFF
--- a/internal/api/v1/application/deploy.go
+++ b/internal/api/v1/application/deploy.go
@@ -20,7 +20,6 @@ import (
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/gin-gonic/gin"
-	"github.com/spf13/viper"
 )
 
 const (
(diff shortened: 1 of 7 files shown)


internal/registry/registry.go
```diff
@@ -39,14 +39,16 @@ type DockerConfigJSON struct {
 }
 
 type ConnectionDetails struct {
-	*DockerConfigJSON
-	Namespace string
+	RegistryCredentials []RegistryCredentials
```
jimmykarily, line 42: Many of the changes in this PR are cause by this change here. I wanted to change the ConnectionDetails like this, because the json representation of the credentials is only important at the last minute, when we store the details as a Secret (or when we do the opposite). It feels more natural to work with an array of RegistryCredentials everywhere else. E.g. in `getInternalRegistryDetails`.

internal/api/v1/application/deploy.go
```diff
@@ -299,8 +313,8 @@ func replaceInternalRegistry(ctx context.Context, cluster *kubernetes.Cluster, i
 	if err != nil {
 		return imageURL, err
 	}
-	// If there is a local registry and kube is not supposed to access it through Ingress
-	if localURL != "" && !viper.GetBool("force-kube-internal-registry-tls") {
+
+	if localURL != "" {
```
jimmykarily, line 317: This is everything the epinio server has to know. The `force-kube-internal-registry-tls` is now only an instruction to the installer. The server doesn't care about that flag.

manno (approved):
So, `force-kube-internal-registry-tls` still defaults to false.
Which means, only if no external registry is used, we will create a node port service for the registry and use that in deployments. This connection, from the node to the registry won't be secured.

If `force-kube-internal-registry-tls` is true or an external registry is specified, will just use the registry and the node will fail to download the image, if it doesn't have proper TLS.  Therefore the setting forces Epinio to use a secured connection.

With the default settings, our internal registry will be accessed without transport encryption, which is fine on local clusters. Only public cloud and multi node clusters need transport encryption and are probably better of with an external registry. 
If we want to improve that, the registry's TLS setup should probably not happen in Epinio's code, but in the registries helm chart.

PR #1030, opened 2021-11-29
Enable CodeQL analysis

### LINKED ISSUE #1027: Automatically detect common vulnerabilities and coding errors
I would like to enable CodeQL to detect common vulnerabilities.

olblak:
Fix #1027 

Signed-off-by: Olivier Vernin <olivier.vernin@suse.com>

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,0 +1,43 @@
+name: "CodeQL"
+
+on:
+  workflow_dispatch:
+  push:
+    branches: [main]
+  pull_request:
+    # The branches below must be a subset of the branches above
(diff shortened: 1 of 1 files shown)


andreas-kupries (approved):

andreas-kupries:
Note, at least some of the reported issues are in testsuite code, and in code guarded by environment variable.

olblak:
> Note, at least some of the reported issues are in testsuite code, and in code guarded by environment variable.
> Which means that we will need a means of telling the system to ignore specific reports, with an explanation on why we believe > that it should do that, for future maintainers.

Let's give it a try, we can still revert If we realise that it introduces more noise than it really help

PR #1051, opened 2021-12-06
1028 helm chart module

### LINKED ISSUE #1048: Consume helm chart as git module in epinio
As per the comment here: https://github.com/epinio/epinio/issues/1028#issuecomment-985517302

Epinio should consume the epinio-helm-chart repository as a git module. This way we can work on the epinio/epinio repository even when changes on the helm charts are needed.

# Acceptance Criteria

- [x] [epinio-helm-chart repository](https://github.com/epinio/epinio-helm-chart) is a git module in epinio/epinio
- [x] All CI scripts use the git module directory to install Epinio
- [x] There is no other helm chart in the epinio/epinio repository
- [x] The installer is also a git submodule in epinio/epinio so that we can use the manifest and manifest assets from there

jimmykarily:
Fixes #1048


scripts/prepare-environment-k3d.sh
```diff
@@ -51,10 +53,23 @@ check_dependency kubectl helm
 create_docker_pull_secret
 
 echo "Preparing Epinio manifest"
-sed -i "s/10.86.4.38.omg.howdoi.website/$EPINIO_SYSTEM_DOMAIN/" assets/installer/manifest.yaml
+echo "Replacing the system domain"
```
jimmykarily, line 56: These `sed` commands may be gone in the future if the installer supports variables.

juadk (approved):
Mainly deletion of duplicate files, complicated to test all the things because the full stack is not very stable at the moment, we are still doing test beside this PR.

scripts/prepare-environment-k3d.sh
ldevulder: Yes, otherwise if this will be as is for some time it could be more "readable" to use something like `sed -i "s/%EPINIO_SYSTEM_DOMAIN%/$EPINIO_SYSTEM_DOMAIN/" assets/installer/manifest.yaml`

ldevulder (commented):
LGTM

PR #1052, opened 2021-12-07
Remove old ui and statik module

jimmykarily:
This UI was a PoC and we are moving to the new web UI in rancher. This is a cleanup.


juadk (approved):
It removes UI and CI is green.

PR #1090, opened 2021-12-21
Use blobUID and builderImage from previous staging when available

### LINKED ISSUE #1073: Stage endpoint without a blobuid
This is a follow up of this story: https://github.com/epinio/ui/issues/10

Currently the stage endpoint expect a [blobuid](https://github.com/epinio/epinio/blob/5988d5ce30bcfaefc71c67aa6479f8c5f2392098/pkg/api/core/v1/models/models.go#L153). That blod uid is the id of the blob that was previously uploaded. In order to allow restaging an application with the latest version of the code (e.g. from the web ui), we must make that parameter optional. If it's not there, the latest blob should be used, if one exists. Otherwise handle errors properly.

jimmykarily:
Fixes #1073 

TODO:

- [x] Write tests

Sibling PRs:

- https://github.com/epinio/helm-charts/pull/50
- https://github.com/epinio/application/pull/3

### DIFF
--- a/scripts/patch-epinio-deployment.sh
+++ b/scripts/patch-epinio-deployment.sh
@@ -73,7 +73,7 @@ spec:
   containers:
     - name: copier
       image: busybox:stable
-      command: ["/bin/sh", "-ec", "while :; do printf '.'; sleep 5 ; done"]
+      command: ["/bin/sh", "-ec", "trap : TERM INT; sleep infinity & wait"]
       volumeMounts:
         - mountPath: "/epinio"
           name: epinio-binary
--- a/acceptance/api/v1/application_deploy_test.go
+++ b/acceptance/api/v1/application_deploy_test.go
@@ -1,15 +1,12 @@
 package v1_test
 
 import (
-	"encoding/json"
 	"fmt"
-	"io/ioutil"
-	"net/http"
 	"strings"
 
 	"github.com/epinio/epinio/acceptance/helpers/catalog"
+	"github.com/epinio/epinio/acceptance/testenv"
 	"github.com/epinio/epinio/helpers"
-	v1 "github.com/epinio/epinio/internal/api/v1"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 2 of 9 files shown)


jimmykarily:
Did some preparation here: https://github.com/epinio/epinio/pull/1091

I will rebase when that is merged to add tests to the appropriate file.

jimmykarily:
This PR allows us to have an `epinio app restage` command. To keep this PR short, I suggest we split the implementation of that command to another PR.

scripts/patch-epinio-deployment.sh
```diff
@@ -73,7 +73,7 @@ spec:
   containers:
     - name: copier
       image: busybox:stable
-      command: ["/bin/sh", "-ec", "while :; do printf '.'; sleep 5 ; done"]
+      command: ["/bin/sh", "-ec", "trap : TERM INT; sleep infinity & wait"]
```
jimmykarily, line 76: allows faster removal of this pod (by listening for `TERM` signals)

andreas-kupries (dismissed):
Looks to be ok to me. The `*APIErrors` returns should possibly be commented on and explained.

jimmykarily:
> Looks to be ok to me. The `*APIErrors` returns should possibly be commented on and explained.

You mean like this one? https://github.com/epinio/epinio/pull/1090/files#diff-1d5683322c8ac645535eab16a4aa0ef4a516edecf543777e7de0b8ca5f4f7c7fR151

The error message itself documents each case enough, no?

jimmykarily:
> > Looks to be ok to me. The `*APIErrors` returns should possibly be commented on and explained.
> 
> You mean like this one? https://github.com/epinio/epinio/pull/1090/files#diff-1d5683322c8ac645535eab16a4aa0ef4a516edecf543777e7de0b8ca5f4f7c7fR151
> 
> The error message itself documents each case enough, no?

After talking on Slack, the issue if about the usage of pointers for errors. I did that to be able to return `nil` when no error occured. This is not needed tough because we can return `nil` for interfaces too. Fixed here: https://github.com/epinio/epinio/pull/1090/commits/19867de9bdbe351a3306c39bf176c6170082c20a

jimmykarily:
There is one test that failed. Maybe the 4s seconds are not always enough to show the application restarts. Doesn't seem relevant to this PR though.

PR #1100, opened 2022-01-10
Add app restart endpoint

### LINKED ISSUE #1074: Application restart endpoint
This is a follow up of this story: https://github.com/epinio/ui/issues/10

Sometimes, users want to restart applications. This can happen for various reasons (e.g. flush logs, expire caches, quickly fix problems like memory leaks before a code fix is in place, etc). 

Epinio currently doesn't have a way to simply restart an application. We need to implement an endpoint that will do something like the equivalent of this command on the app deployment: 

```kubectl rollout restart deployment app_deployment_here```

### PR DESCRIPTION
**Pull request by thardeck**
Fixes #1074

### DIFF
--- a/acceptance/api/v1/application_restart_test.go
+++ b/acceptance/api/v1/application_restart_test.go
@@ -1,0 +1,88 @@
+package v1_test
+
+import (
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+
+	"github.com/epinio/epinio/acceptance/helpers/catalog"
+	"github.com/epinio/epinio/helpers"
+	api "github.com/epinio/epinio/internal/api/v1"
+	v1 "github.com/epinio/epinio/internal/api/v1"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("AppRestart Endpoint", func() {
+	var (
+		namespace string
+	)
+	containerImageURL := "splatform/sample-app"
+
+	BeforeEach(func() {
+		namespace = catalog.NewNamespaceName()
+		env.SetupAndTargetNamespace(namespace)
+	})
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: acceptance/api/v1/application_restart_test.go [hunk 1]
```diff
@@ -0,0 +1,88 @@
...
+		env.SetupAndTargetNamespace(namespace)
+	})
+	AfterEach(func() {
+		env.DeleteNamespace(namespace)
+	})
+
+	It("restarts the app", func() {
+		app1 := catalog.NewAppName()
+		env.MakeContainerImageApp(app1, 1, containerImageURL)
+		defer env.DeleteApp(app1)
```
andreas-kupries, line 35: Should go into an `AfterEach` block. As not all tests (2 of 3) use that the (two) relevant tests have to go into a `Context` block.

pkg/api/core/v1/models/models.go
```diff
@@ -128,6 +128,12 @@ type ApplicationCreateRequest struct {
 	Configuration ApplicationUpdateRequest `json:"configuration" yaml:"configuration,omitempty"`
 }
 
+// ApplicationRestartRequest represents and contains the data needed to
```
andreas-kupries, line 131: This looks to be not used. Which is sensible, because the app name info is provided as part of the request url.
IMHO this should be removed.

andreas-kupries (changes requested):
Only small changes
- Cleanup in tests on how app deletion is done.
- Cleanup of a leftover unused structure

acceptance/api/v1/application_restart_test.go
andreas-kupries: Actually, the app creation could (should) then move into a `BeforeEach` block for that `Context` also.

andreas-kupries (approved):
LGTM now

PR #1119, opened 2022-01-14
Improve docker build speed

manno:
* don't compile all binaries
* don't copy helm charts into multi-stage
* (don't create extra layer, even though it's only in multi-stage)

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@ dist/
 docs/
 images/
 output/
+helm-charts/
 
 contrib/
 tmp/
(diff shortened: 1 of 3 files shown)


.dockerignore
```diff
@@ -3,6 +3,7 @@ dist/
 docs/
 images/
 output/
+helm-charts/
```
jimmykarily, line 6: maybe the installer too?

jimmykarily (dismissed):

.dockerignore
manno: I don't have an installer sub module in my tree.

jimmykarily: mea culpa. It has been removed.

jimmykarily (approved):

PR #1120, opened 2022-01-14
fix: unwanted retry of `namespace create` for existing namespace

### LINKED ISSUE #1081: Do not retry in Cli if the namespace already exists
Do not retry to create the namespace in Cli if it already exists. Follow up for the mentioned TODO:

https://github.com/epinio/epinio/blob/1cc932fb809132ad00212715aa98f882e3f1ac0d/internal/api/v1/namespace/create.go#L40-L43

andreas-kupries:
fix #1081 

- retry removed completely
- became superfluous again, with initial namespace creation moved to external installer

### DIFF
--- a/pkg/api/core/v1/client/namespaces.go
+++ b/pkg/api/core/v1/client/namespaces.go
@@ -2,14 +2,8 @@ package client
 
 import (
 	"encoding/json"
-	"fmt"
-	"time"
 
-	"github.com/avast/retry-go"
-
-	"github.com/epinio/epinio/helpers"
 	api "github.com/epinio/epinio/internal/api/v1"
-	"github.com/epinio/epinio/internal/duration"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 )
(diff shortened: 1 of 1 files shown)


pkg/api/core/v1/client/namespaces.go
```diff
@@ -22,33 +16,7 @@ func (c *Client) NamespaceCreate(req models.NamespaceCreateRequest) (models.Resp
 		return resp, nil
```
enrichman, line 16: It's not related to the PR but should be this returning `err` instead of `nil`?

andreas-kupries: Quite possible. I will make that change and see if that breaks CI.

andreas-kupries: Lets see what CI says about ea54e449a52827369fb0c5fa3158ea06622bd4d8

enrichman (approved):

PR #1125, opened 2022-01-18
Removed kexec package

enrichman:
Remove unnecessary kexec package


andreas-kupries:
Looking at the failures, which are in the BeforeSuites ... I suspect that a command's output is lost somewhere in the modified code, and this breaks the checks of said output for patterns.

enrichman:
> Looking at the failures, which are in the BeforeSuites ... I suspect that a command's output is lost somewhere in the modified code, and this breaks the checks of said output for patterns.

Yes, that seems the case. After those blind fixes let me try to replicate it locally.

jimmykarily (approved):
LGTM, merge!

PR #1126, opened 2022-01-18
fix: defend against bad path names

andreas-kupries:
Clean the paths from the archive to prevent zip slip attacks and the like.

Origin issue https://github.com/epinio/epinio/security/code-scanning/4?query=ref%3Arefs%2Fheads%2Fmain


helpers/untar.go
```diff
@@ -37,7 +38,7 @@ func Untar(tarFile, targetDir string) error {
 			continue
 		}
 
-		target := filepath.Join(targetDir, header.Name)
+		target := filepath.Join(targetDir, path.Clean(header.Name))
```
jimmykarily, line 41: Misses "../test" ? 

https://go.dev/play/p/yxluh7JFVxW

jimmykarily: I `targetDir` has slashes in it, we could run `path.Clean` after we do the `filepath.Join`. I feel there are better solutions though. If it's only the initial `..` this code isn't catching, maybe we can just handle that case separately.

andreas-kupries: Weird. The description let me believe that the leading `..` would be removed. Will investigate. Running it on the joined path is an idea.

jimmykarily: it removes `..` if the path has a leading `/`. e.g. `/../test`

andreas-kupries: Rereading the docs at https://pkg.go.dev/path#Clean I misread item 4. I read as any `../*` removed. It actually is only `/../*`. Note the leading `/`. So. Application to the joined path will fix that. Targetdir ... Can that be a relative path, and then have `..` or `.` ? ... Searching ....

Oh! The rest of epinio does __not use__ `untar`. The function has become *superfluous*.

New fix: Remove it.

andreas-kupries: Done, and rebased.

jimmykarily (approved):

PR #1129, opened 2022-01-19
fix: missing build of local helm-charts for public cloud CI

### LINKED ISSUE #1105: Build last helm chart locally for public cloud CI tests
When we install Epinio in public clouds (and RKE in our self host runners), we use the last release of the [Epinio Helm chart](https://github.com/epinio/epinio/blob/main/acceptance/helpers/epinio/epinio.go#L27).

But sometimes, a change is introduced in the epinio server and we also need a change in the helm chart, that's why we need to build the last helm chart locally (with chartmuseum) before triggering the installation. Like we do in the main CI with [chartmuseum](https://github.com/epinio/epinio/blob/main/scripts/setup-chart-museum.sh)

Issue : https://github.com/epinio/epinio/runs/4771873657?check_suite_focus=true

andreas-kupries:
Fix #1105


andreas-kupries:
Note from the slack:
> Yes that's the idea but this part of code has to be changed too
https://github.com/epinio/epinio/blob/main/acceptance/helpers/epinio/epinio.go#L25-L27

The `Install()` function currently receives only a set of strings, to become the args of the `helm install` command.

This actually includes the domain information, in a `domain=...` argument.
I am loath to parse the domain out of that.
As a first attempt I will see what breaks if I change the func to take the domain information as an explicit argument as well.

andreas-kupries:
- RKE failure - Domain looks ok, chartmuseum startup seems to fail, or is not properly recognized as started.
  - GKE failure - The domain looks broken.

Giving for today.

andreas-kupries:
Looking into the GKE flow I see that it is possible to have a secret value in the domain. I now believe that the domain is correct, and the weirdness in the log is just GH hiding the secret to prevent its leakage through the log.

With that said, GKE then has the same issue as RKE. I.e. the museum is either not coming up, or the code waiting for it is simply not managing to reach it.

andreas-kupries:
With the curl output visible GKE reports

>  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0curl: (6) Could not resolve host: chartmuseum.id17997-***

I suspect RKE is similar, except it ran into a cluster creation error, i.e. never reached the place of interest :(

andreas-kupries:
Lets see what I can get when running locally for comparison.

andreas-kupries:
That works, as expected. It does not really give a clue towards the issues with PC ... 

... Looking at the instructions for getting the CM url, they do not really look correct for PC ...
Having a look at `chartmuseum/templates/NOTES.txt` in the CM chart I notice that the CM entry service has a variety of options it can be configured with, i.e. `NodePort`, `LoadBalancer`, and `ClusterIP`.
The default is the last in that list, `ClusterIP`.
That does not seem to be good for PC.

Based on the text I suspect that I need `LoadBalancer`.

... On the other side, GKE does not even resolve the constructed cm url ... 
Maybe we simply need an entry in some kind of DNS instead ?

I will have to retry the RKE, given that it uses an IP in a magic domain to construct the url. Which worked when construction was done in the go testsuite code.

andreas-kupries:
RKE fails on getting micros os from the rancher repo ... That is something I cannot fix. Just wait until that repo is back ...

andreas-kupries:
Back to GKE I see management by Route53, and adding the loadbalancer there as part of the setup ... Ok ... Lets try to get the CM configured for loadbalancer and see if that helps.

andreas-kupries:
The service config looks mostly ok to me. The service exists, has a loadbalancer ip

>     "status": ***
        "loadBalancer": ***
            "ingress": [
                ***
                    "ip": "34.140.144.23"
                ***
            ]

The 
>"externalTrafficPolicy": "Local",
might bite us.

As before the `chartmuseum.id...` host does not resolve.
I am tending to see it more on the DNS side, that our Route53 setup needs an entry for the `chartmuseum` host.
If so the service type might not be relevant.

@ldevulder @juadk May I ask you to look over my latest code and comments above ?
I fear that it may need debugging outside of CI, where changes can be trialed faster.
And also somebody with access to Route53.

ldevulder:
> @ldevulder @juadk May I ask you to look over my latest code and comments above ?
> I fear that it may need debugging outside of CI, where changes can be trialed faster.
> And also somebody with access to Route53.

I will check this yes and try to test it on my lab.

juadk:
I just took a quick look at your code and the GKE issue.
I think you are right regarding the missing DNS entry for `chartmuseum`. However, I wonder if it could not be simpler to install `chartmuseum` in the github runner locally, this means a rework of the script.
What is the best between the 2 solutions...

andreas-kupries:
Instead of reworking the script maybe a new script first ? Less conditionals to confuse ... 
I cannot truly say which approach would be better. I do not have enough insight into either to make an estimate.

juadk:
Actually, we have the script somewhere already (https://github.com/epinio/helm-charts/blob/main/scripts/build-charts-ci.sh).
It is used in the helm chart repo to build chart locally in the github runner.
Otherwise, if we choose to update DNS, we will have to deal with tedious things like cleaning and mainly waiting for DNS update in route53.

andreas-kupries:
That sounds lovely ... Will have a look ...

andreas-kupries:
Having thought about having the CM in the runner instead of the cluster ... That would mean that all the work on the system domain was superfluous. Mind, that is not bad, from first glance it keeps things simpler. Just sad about the lost time.

It also means that I believe that I should restart the work from scratch.

Although ... I should keep the removal of the code adding a helm repo for the chart, in the epinio `Install()` helper. That looks to be still needed.

Ok. I seem to have a plan again.

juadk:
> Ok. I seem to have a plan again.

I totally agree with your new plan, locally will be simpler to manage.

andreas-kupries:
While GKE PCCI failed it looks to be "just" a mismatch in repository names.
The helm chart setup script creates a repo `chartmuseum` and the testsuite expects the repo `epinio-charts` instead.
That should be fixable.
To check viability first a hack which modifies the script in the action before I go for having a helm-charts PR adding customization features.

andreas-kupries:
Ok. While still a failure this may now be an issue with the charts we are testing, instead of the setup.

My reading of the reported

> Error: INSTALLATION FAILED: failed post-install: job failed: BackoffLimitExceeded

and log entries before it is that the testsuite is able to see the local chart museum, and apparently is able to pull the `epinio-installer` chart. And then fails to deploy the chart. This may be an issue in the current charts, or an issue caused by the patching of the `epinio-installer` chart done by the chart museum setup script. The `host.k3d.internal` magic.

Generally the new approach feels more viable tough, I am already much further along than with the first attempt.

andreas-kupries:
@juadk I believe it is the `host.k3d.internal` magic which is the issue here.

This is magic for a k3d cluster to talk to the localhost.
The clusters made by the CI for public clouds are not based on k3d.
In the trial I am currently doing the cluster is made by GKE.
And to such a cluster `host.k3d.internal` is most certainly not a proper host name for the gh action runner.
Neither would be localhost.

Regular CI, being based on k3d cluster is of course fine with this.

Do we actually allow access from the public cloud cluster to our action runner general ?
If yes, what name would that be ?

juadk:
It makes sense, I did not think about it but deployment in the GKE cluster can not reach the GH runner.... sounds obvious now...

> Do we actually allow access from the public cloud cluster to our action runner general ?

Unfortunately, I don't think so...

I'm thinking.... but this time, I wonder if we really can avoid installing chartmuseum in the cluster hosted in public cloud...

andreas-kupries:
Note: I have kept the latest of the previous attempt marked with a local branch. IOW I can rebase this PR and restore that.

andreas-kupries:
Restored old approach, with compressed history.
Recap of its state:

 - CM configured for loadbalancer - possibly not needed
 - CM not reachable from GHA runner - Likely DNS issue, __main problem__

juadk:
If we want to install `chartmuseum` in public cloud cluster, that means we have to deploy an ingress controller before.
Usually we use traefik deployed by Epinio but we need it before in that case, it also means we won't test the traefik installation by Epinio anymore.
Otherwise, we can use another type of service than loadbalancer and ingress for chartmuseum... sounds more and more complex...

andreas-kupries:
@ldevulder Given that you and @juadk are now deep into the guts of this, should I assign this PR over to you ?

ldevulder:
After more tests in Public Cloud I can confirm that it seems to be impossible to use the `epionio-installer` chart (aka. big helm chart). This is because `epinio-installer` also uses `helm` to install dependencies and in that case it sounds like these other helm installations are run inside some sandbox and in that case the port forwarding doesn't work...

So the solutions I see here:
- let is as is, which can lead to some failure when the helm charts are changed with important modifications
- use the [Method 2](https://docs.epinio.io/installation/install_epinio_manual.html) installation process: could be good as this one is not tested in CI
- use a public chartmuseum/pre-production helm repository

ldevulder:
@andreas-kupries you added some debugging code in your commits, do you think that some should be removed or not? My tests on RKE, GKE and EKS are OK, so I will re-read the code next week before putting the PR in the review column.

andreas-kupries:
@ldevulder IMHO the debug code should be removed.

juadk (approved):
LGTM

Issue #1101, opened 2022-01-10
AKS-CI is failing


ldevulder:
The AKS-CI test is failing during app deployment with this error:
```
Deploying application ...
  2022/01/09 03:47:00 EpinioApiClient "msg"="response is not StatusOK" "error"="Internal Server Error: invalid character 'a' in string escape code" "status"=500 
  error pushing app to server: can't deploy app: Internal Server Error: invalid character 'a' in string escape code
```

For example see this test: https://github.com/epinio/epinio/runs/4751235189?check_suite_focus=true

I was not able to reproduce the issue with manual test on Linux client as well as CI installation test running on a linux client. The CI test runs on a Windows client, maybe something is not exported correctly by Epinio which could explain this error?

manno:
The Epinio CLI complains about a 500 reply from the server. The json error parser fails, because the response is not JSON. It's likely the error message did not originate from the Epinio API server.

This might be a temporary issue, which could be fixed by retrying API requests: https://github.com/epinio/epinio/issues/807

ldevulder:
During my Rancher Desktop tests on Windows 10, I was able to reproduce the same kind of issue:
```
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create ===> EXPORTING
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'paketo-buildpacks/ca-certificates:helper'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'paketo-buildpacks/httpd:httpd'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'paketo-buildpacks/php-dist:php'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'paketo-buildpacks/php-web:php-web'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding 1/1 app layer(s)
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'launcher'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'config'
🕞  [efeda7c398a918db-stage-bbtkn-pod-p6v8g] step-create Adding layer 'process-types'
[...]

PR #1131, opened 2022-01-20
Log response body in case of error

enrichman:
This is not a complete fix for the #1101 but it will give us more insights onn what is happening.

### DIFF
--- a/pkg/api/core/v1/client/http.go
+++ b/pkg/api/core/v1/client/http.go
@@ -154,7 +154,12 @@ func (c *Client) do(endpoint, method, requestBody string) ([]byte, error) {
 	// TODO why is != 200 an error? there are valid codes in the 2xx, 3xx range
 	if response.StatusCode != http.StatusOK {
 		err := formatError(bodyBytes, response)
+
+		if respLog.V(15).Enabled() {
+			respLog = respLog.WithValues("body", string(bodyBytes))
+		}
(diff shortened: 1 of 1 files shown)


pkg/api/core/v1/client/http.go
```diff
@@ -154,7 +154,12 @@ func (c *Client) do(endpoint, method, requestBody string) ([]byte, error) {
 	// TODO why is != 200 an error? there are valid codes in the 2xx, 3xx range
 	if response.StatusCode != http.StatusOK {
 		err := formatError(bodyBytes, response)
+
+		if respLog.V(15).Enabled() {
+			respLog = respLog.WithValues("body", string(bodyBytes))
```
manno, line 159: can we provide more logging, maybe some http headers, that allow us to identify the server? In case it's a loadbalancer or DNS issue

enrichman: Yes, I've actually haven't added it because it was already there, but I've used the same logLevel

https://github.com/epinio/epinio/blob/bbe4a58416d43b263e83286dc0f203d37e590d31/pkg/api/core/v1/client/http.go#L247-L256

manno: So we didn't see it in the report, because the log level was not high enough. That's unfortunate.

I wonder how large the body can be for a request that's not of status code 200. Anyhow, 15 feels like a high log level.

manno (approved):

PR #1132, opened 2022-01-20
Address warning and deprecated messages on GKE CI

juadk:
This PR addresses the two following warnings:

```
acceptance-scenario2
google-github-actions/setup-gcloud is pinned at HEAD. We strongly advise against pinning to "@master" as it may be unstable. Please update your GitHub Action YAML from:

    uses: 'google-github-actions/setup-gcloud@master'

to:

    uses: 'google-github-actions/setup-gcloud@v0'

Alternatively, you can pin to any git tag or git SHA in the repository.
```

```
"service_account_key" has been deprecated. Please switch to using google-github-actions/auth which supports both Workload Identity Federation and Service Account Key JSON authentication. For more details, see https://github.com/google-github-actions/setup-gcloud#authorization
```


ldevulder (approved):
LGTM

ldevulder:
The 2 failing tests are because of sporadic issues... The code changed in this PR is validated, as it's in the GKE deployment stack and this one is done.

PR #1134, opened 2022-01-20
Centralize duplicated code

enrichman:
Many `ValidArgsFunction` were doing the same check, so this PR centralize the two most common func in the `commons` file.

### DIFF
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -77,6 +77,7 @@ var CmdAppList = &cobra.Command{
 		}
 
 		err = client.Apps(all)
+		// Note: errors.Wrap (nil, "...") == nil
 		return errors.Wrap(err, "error listing apps")
 	},
 }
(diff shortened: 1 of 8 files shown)


internal/cli/apps.go
```diff
@@ -78,19 +77,16 @@ var CmdAppList = &cobra.Command{
 		}
 
 		err = client.Apps(all)
-		if err != nil {
-			return errors.Wrap(err, "error listing apps")
-		}
-
-		return nil
+		return errors.Wrap(err, "error listing apps")
```
andreas-kupries, line 80: I had to remember that `errors.Wrap` returns `nil` if the `err` is `nil itself. Maybe add a small line about that before the command. Something like
> // __Note__: errors.Wrap (nil, "...") == nil

Not just here, all the changed places.

andreas-kupries (changes requested):
Comment and question.

PR #1135, opened 2022-01-21
Fix for windows paths

### LINKED ISSUE #1101: AKS-CI is failing

enrichman:
This should fix #1101 

I've also removed a couple of possible panics (if the `err` is nil then the `apierror.InternalError` will panic).

### DIFF
--- a/internal/application/origin.go
+++ b/internal/application/origin.go
@@ -2,11 +2,12 @@ package application
 
 import (
 	"context"
-	"fmt"
+	"encoding/json"
 
 	"github.com/epinio/epinio/helpers/kubernetes"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
@@ -50,8 +51,9 @@ func Origin(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppR
 			return result, err
 		}
 		if path == "" {
-			return result, apierror.InternalError(err, "Bad path origin, empty string")
+			return result, errors.New("bad path origin, empty string")
 		}
+
 		result.Kind = models.OriginPath
 		result.Path = path
 		return result, nil
@@ -63,8 +65,9 @@ func Origin(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppR
 			return result, err
 		}
 		if container == "" {
-			return result, apierror.InternalError(err, "Bad container origin, empty string")
+			return result, errors.New("bad container origin, empty string")
 		}
+
 		result.Kind = models.OriginContainer
 		result.Container = container
 		return result, nil
@@ -76,10 +79,8 @@ func Origin(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppR
 			return result, err
 		}
 		if repository == "" {
-			return result, apierror.InternalError(err, "Bad git origin, url is empty string")
+			return result, errors.New("bad git origin, url is empty string")
 		}
-		result.Kind = models.OriginGit
-		result.Git.URL = repository
 
 		// For git check for the optional revision as well.
 		revision, found, err := unstructured.NestedString(origin, "git", "revision")
@@ -88,11 +89,13 @@ func Origin(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppR
 				return result, err
 			}
 			if revision == "" {
-				return result, apierror.InternalError(err, "Bad git origin, revision is empty string")
+				return result, errors.New("bad git origin, revision is empty string")
 			}
 			result.Git.Revision = revision
 		}
 
+		result.Kind = models.OriginGit
+		result.Git.URL = repository
 		return result, nil
 	}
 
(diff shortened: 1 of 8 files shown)


internal/application/origin_test.go
```diff
@@ -0,0 +1,78 @@
+package application
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+)
+
+func Test_buildBodyPatch(t *testing.T) {
```
jimmykarily, line 10: I'm not sure our make target will run non-ginkgo tests: https://github.com/epinio/epinio/blob/main/Makefile#L71-L72

Also, although this is just a unit test, I would still stick to ginkgo syntax for consistency. E.g. like this one: https://github.com/epinio/epinio/blob/main/internal/s3manager/s3manager_test.go

I find the tests descriptions helpful when trying to understand what is being tested.

internal/application/origin.go
```diff
@@ -110,31 +113,58 @@ func SetOrigin(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
...
 		}
-		value = fmt.Sprintf(`"git": {%s}`, value)
 	}
 
-	// And enter into the app
-	patch := fmt.Sprintf(`[{"op": "replace", "path": "/spec/origin", "value": {%s}}]`, value)
+	operations := []PatchOperation{{
+		Op:    "replace",
+		Path:  "/spec/origin",
+		Value: patchOpValue,
```
andreas-kupries, line 149: Hm ... `patchOpValue` is the same type as `origin`, right ?
Could we use `origin` directly here ?
The preceding lines are effectively a more complicated `patchOpValue := origin` also ?! I.e could be dropped ?!

andreas-kupries: Oh, I see now. You created a local copy of the model type, and all the code above is doing the copying.
Does the local copy have a different tagging for json marhsalling ? ...
Yes, the `Kind` field is not present in the local copy.
Ok then, retracting my thoughts above.

andreas-kupries (approved):
You can resolve my comments without action ... Or maybe add some comments giving an overview of your solution ;)

internal/application/origin.go
enrichman: Yes, I was thinking to use the same but unfortunately the GitOrigin is slightly different (`repository` vs `url`), not sure why.

andreas-kupries: Please bring it up in the planning. I would like to have the code simple(r) here.
It may affect the server API, i.e. IIRC the model goes over the wire between client and server.
So a change would also affect @richard-cox

enrichman: As discussed I'm going to change the GitRef struct model to match the CRD

https://github.com/epinio/epinio/blob/dfc3aba919800f30a6a263949b9e7a813177359e/pkg/api/core/v1/models/app.go#L20-L23

https://github.com/epinio/helm-charts/blob/a3d66d6c0dda2d8c5752247f93a3e3fdf79064ec/chart/epinio/templates/app-crd.yaml#L59-L67

to

```golang
type GitRef struct {
	Revision   string `json:"revision,omitempty" yaml:"revision,omitempty"`
	Repository string `json:"repository"         yaml:"repository"`
}
```

I am going to update also the yaml, is that fine?

andreas-kupries: I am not sure right now ... The yaml may have to match the CRD spec. I would not touch it, in the interest of making a minimal change to get the desired effect.

Note that you only have to change the `json:"url"` to ``json:"repository"`.
There is no need to change the field name in Go.
I.e. the `URL` to `repository` change is superfluous as well.

enrichman: Ok, let's keep it at the minimum then. :)

andreas-kupries:
Note that we generally rebase the branch to the new head of main, instead of merging main into the branch.

enrichman:
Finally green :green_circle:

pkg/api/core/v1/models/models.go
```diff
@@ -90,10 +90,10 @@ type ApplicationStage struct {
 // ApplicationOrigin is the part of the manifest describing the origin of the application
 // (sources). At most one of the fields may be specified / not empty.
 type ApplicationOrigin struct {
-	Kind      int    `yaml:"-"` // Hidden from yaml. Type tag to simplify struct usage.
-	Container string `yaml:"container,omitempty"`
-	Git       GitRef `yaml:"git,omitempty"`
-	Path      string `yaml:"path,omitempty"`
+	Kind      int     `                           yaml:"-" ` // Hidden from json and yaml. Type tag to simplify struct usage.
```
andreas-kupries, line 93: Is this indeed hidden from JSON without a `json:"-"` clause ?

andreas-kupries (changes requested):
Have a question

pkg/api/core/v1/models/models.go
enrichman: Ah, I left the comment. I had to remove the `json:"-"` because it was actually used to marshal/unmarshal the Kind.

I don't really like to expose the property as-is

```
{
    "Kind": 1,
    "path": "C:\\Documents\\app"
}
```

but change it in other ways would need a bit of refactor. What do you think?

The best would be to have something like an enum mapping the Kind `int`, so exposing something more explicit:

```
{
    "type": "ORIGIN_PATH",
    "path": "C:\\Documents\\app"
}
```

andreas-kupries: That is weird wrt `json:"-"`. This is, AFAIK the proper way for excluding the field from (un)marshalling.

Looking at the tests you have I see that you have `Kind` in the JSON. That is what I am expecting for this tagging, without the `json:"-"` clause. Are you saying that when you add this clause the tests still pass, with `Kind` in the JSON ?

Go has no enums as we are used from C. See https://github.com/epinio/epinio/blob/5f3872d1e3c5eca11937f6020d459f8129502542/pkg/api/core/v1/models/models.go#L100-L105 for an example of a Go enum.

enrichman: No, I meant that originally I have added the clause but other tests were failing, because they were implicitly relying on the `Kind` field.
So I had to remove it and keep the marshaling as is (with the `Kind` exposed).

I know about the enum, unfortunately (is a feature that I actually miss). To add this kind of behaviour the only way would be to add a custom type/struct and implement the `Unmarshal` interface I suppose.
Probably overkill.

andreas-kupries: Oh. I see. ... Were the failing tests API tests ? Because IIRC it is exposed there. Makes the handling on each side nicer (albeit opens each side up to getting structure where kind and actual actual information do not match).

We could drop it from the API, close that attack vector ... Not in this issue and PR.

andreas-kupries: I recommend adding a note in the code about this issue with Kind not hidden.

Resolving as ok, will approve.

andreas-kupries (approved):

andreas-kupries (approved):

PR #1137, opened 2022-01-24
Cleanup acceptance cluster

### LINKED ISSUE #1123: Remove k3d at the end of each CI tests
To avoid performance and sporadic issues when we try to run other CI tests like E2E/UI tests on k3s as well as rke2 tests.

enrichman:
This should be enough to fix #1123


andreas-kupries (approved):
LGTM

Issue #1113, opened 2022-01-11
Faster builds


jimmykarily:
@agracey has some data that show that our staging is unreasonably slow to start. We need to investigate and find what is delaying the process. If it's Tekton, we can try to remove it (and maybe use kpack or pack directly?).

Goals:

- Find things are slow
- Find ways to make them faster
- Write stories or implement fixes

Idea: Implement an `.epinioignore` file to not upload files that are not needed (we can use it to ignore files if we ever implement `epinio watch app`. See here: https://github.com/epinio/epinio/discussions/852#discussioncomment-1848020)

agracey:
If you want to go down the path of removing Tekton, here's an example job that seems to work: https://gist.github.com/agracey/1acf78110787114a04e1e76f96111c4f 

This took 40 seconds from the time I created the job:
![image](https://user-images.githubusercontent.com/442274/149822117-dd09dfd7-5ad2-477f-9c3f-81ff67199c4b.png)


The one downside to this is that we lose tekton chains as an option to get cosign integration. That's pretty low priority right now but might be something to consider. IMO, This should really be integrated into the buildpack lifecycle itself so will ask in the upstream community if there's any research started already.

agracey:
We can ignore the cosign issue as it will be added to the buildpack lifecycle itself

jimmykarily:
> Am I right in my reading of that code and, if so, would it be possible to instead stream the tarball through to s3/minio and not need to cache it?

Yes, it's a 2 step process currently. By "stream through" you mean like a proxy? That would mean the client should speak the S3 protocol. I don't think javascript can do that (for the web ui). Otherwise we would have to implement a more complex solution where we don't store but simply translate from one protocol to the other on the fly.

In any case, the data would have to be transefre from the user's machine to the epinio server and from there to the s3 server. I think all we save is some handshakes and some headers, am I wrong?

It would make more sense if the data were copied from the user's machine directly to S3 but, leaving aside authentication and all, this again means the client should speak the S3 protocol.

agracey:
It's effectively proxying the data in the request while not exposing the s3 protocol to the client. The main benefit is timing and reduction in resource usage:

![image](https://user-images.githubusercontent.com/442274/150567809-e40031b5-d580-459b-ba51-2e1482bba020.png)


Here is the same question on Stack Overflow which has some example solutions. https://stackoverflow.com/questions/34177137/stream-file-upload-to-aws-s3-using-go

agracey:
[...]

PR #1138, opened 2022-01-24
feat: stream uploaded app sources directly into s3, without temp file

andreas-kupries:
ref #1113

### DIFF
--- a/internal/api/v1/application/upload.go
+++ b/internal/api/v1/application/upload.go
@@ -1,11 +1,6 @@
 package application
 
 import (
-	"io"
-	"io/ioutil"
-	"os"
-	"path"
-
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/helpers/tracelog"
 	"github.com/epinio/epinio/internal/api/v1/response"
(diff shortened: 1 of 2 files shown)


enrichman (approved):
LGTM

I thought a bit about the "Minio" uploading part, is it used just as storage for the sources/tarball?

At the moment we are forcing the user to use S3 (natively, or via Minio). I think it could make sense to abstract this part and extend it to other providers, having just an _Uploader_ interface that could be binded to other external sources.

i.e.: if I'm using a GKE or AKS cluster maybe I want to use the blob storage provided by Google or Azure. Just food for thought, maybe I should tag @agracey :slightly_smiling_face:

andreas-kupries:
Thank you. Do the GKE/AKS blob storages do not provide an S3 compatible API ? If they do we can use them as they are.

Not merging yet, there looks to be a true failed test

> when staging the same app with no blob defined
   /home/sles/actions-runner/_work/epinio/epinio/acceptance/api/v1/application_stage_test.go:100
    stages with the previous blob [It]

which looks to require a fix.

andreas-kupries:
It was flake. CI :heavy_check_mark: now ...

enrichman:
> Thank you. Do the GKE/AKS blob storages do not provide an S3 compatible API ? If they do we can use them as they are.

Ah, I didn't investigate that. IOt seems that Google does, while Azure not. But I think that if that's the case the priority will go down.

PR #1139, opened 2022-01-24
Moving code from fork to branch

agracey:
This should let me build on CI, right?


andreas-kupries:
Not quite. As a fork it needs the `safe to test` label to run CI. IIRC. Added that label now. If that does not work the change has to be made as branch of the epinio rep instead of as a fork.

andreas-kupries:
Also please try to branch from cd7d9a9328a97569b5e421282984582fd00315a7 (older main).
The current main is currently more testing of CI for a switch of the release scripts to `goreleaser`. I.e. not in a working state.

andreas-kupries:
The label does not have seem to be done anything.

andreas-kupries:
Oh ... Something I did not see before. The destination is epinio:tekton ... Not main.
So, if I were to merge this then it should be a local epinio branch ...

Ok. I will approve and merge, and then I can look towards rebasing to an older main given how head looks to be flaky.

PR #1140, opened 2022-01-24
Remove tekton

andreas-kupries:
Experimental replacement of tekton with a plain Job calling on the buildpack.
Code written by @agracey .


agracey:
Pushed again as tekton-removal

PR #1156, opened 2022-01-28
Make github repositories configurable and add notes

svollath:
Tested with terraform already. Scripts can be configured to work with any repository, while e.g.
`export GITHUB_REPOSITORY_URL=https://github.com/epinio` will make it become an "Organization" runner, or
`export GITHUB_REPOSITORY_URL=https://github.com/epinio/epinio` a runner within the epinio repo
and in general, scripts can be reused for any repository.

### DIFF
--- a/scripts/remove-github-action-runner.sh
+++ b/scripts/remove-github-action-runner.sh
@@ -2,7 +2,7 @@
 
 # This script can be used to remove a Github Action Runner on an openSUSE or SLE
 # distro. It will unconfigure the service and unregister the runner.
-# Copy the scritpt to runner:/home/<user> and run it as <user>.
+# Copy the script to runner:/home/<user> and run it as <user>.
 # It requires GITHUB_REPOSITORY_URL (https) and GITHUB_RUNNER_TOKEN to be set
 # e.g. export GITHUB_REPOSITORY_URL=https://github.com/epinio/epinio
 # and  export GITHUB_RUNNER_TOKEN=<current token from github settings/actions/runners/new>
--- a/scripts/remove-github-action-runner.sh
+++ b/scripts/remove-github-action-runner.sh
@@ -11,20 +11,15 @@
 
 set -e
 
-if [ -z "$GITHUB_REPOSITORY_URL" ]; then
-  echo "The variable GITHUB_REPOSITORY_URL is empty. Exiting"
+if [ -z "$GITHUB_REPOSITORY_URL" ] || [ -z "$GITHUB_RUNNER_TOKEN" ]; then
+  echo "Script requires GITHUB_REPOSITORY_URL and GITHUB_RUNNER_TOKEN to be set. Exiting"
   exit 1
 fi
 
-if [ -z "$GITHUB_RUNNER_TOKEN" ]; then
-  echo "The variable GITHUB_RUNNER_TOKEN is empty. Exiting"
-  exit 1
-fi
-
-REPOSITORY_NAME=$(echo $GITHUB_REPOSITORY_URL | cut -d '/' -f 4- | sed -e 's|/$||g' -e 's|/|-|g')
-ACTIONS_RUNNER_SERVICE=actions.runner.$REPOSITORY_NAME.`hostname`.service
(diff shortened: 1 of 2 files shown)


scripts/remove-github-action-runner.sh
```diff
@@ -2,15 +2,29 @@
 
 # This script can be used to remove a Github Action Runner on an openSUSE or SLE
 # distro. It will unconfigure the service and unregister the runner.
+# Copy the scritpt to runner:/home/<user> and run it as <user>.
```
jimmykarily, line 5:
```suggestion
# Copy the script to runner:/home/<user> and run it as <user>.
```

juadk (dismissed):

scripts/remove-github-action-runner.sh
```diff
@@ -2,15 +2,29 @@
...
   echo "The variable GITHUB_RUNNER_TOKEN is empty. Exiting"
   exit 1
 fi
 
+REPOSITORY_NAME=$(echo $GITHUB_REPOSITORY_URL | cut -d '/' -f 4- | sed -e 's|/$||g' -e 's|/|-|g')
+ACTIONS_RUNNER_SERVICE=actions.runner.$REPOSITORY_NAME.`hostname`.service
+
 cd actions-runner
-sudo systemctl stop actions.runner.epinio-epinio.`hostname`.service
+sudo systemctl stop $ACTIONS_RUNNER_SERVICE
```
manno, line 28: I don't see how globs or whitespaces can end up in $ACTIONS_RUNNER_SERVICE, but quoting might be safer.

scripts/remove-github-action-runner.sh
```diff
@@ -2,15 +2,29 @@
...
+  echo "The variable GITHUB_REPOSITORY_URL is empty. Exiting"
+  exit 1
+fi
+
 if [ -z "$GITHUB_RUNNER_TOKEN" ]; then
   echo "The variable GITHUB_RUNNER_TOKEN is empty. Exiting"
   exit 1
 fi
 
+REPOSITORY_NAME=$(echo $GITHUB_REPOSITORY_URL | cut -d '/' -f 4- | sed -e 's|/$||g' -e 's|/|-|g')
```
manno, line 24:
```suggestion
REPOSITORY_NAME=$(echo $GITHUB_REPOSITORY_URL | cut -d '/' -f 4- | sed -e 's|//*$||' -e 's|/|-|g')
```
If you want to remove multiple dashes at the end of the string.

scripts/remove-github-action-runner.sh
```diff
@@ -2,15 +2,29 @@
...
+ACTIONS_RUNNER_SERVICE=actions.runner.$REPOSITORY_NAME.`hostname`.service
```
manno, line 25: String should be clean, but could be  `"$REPOSITORY_NAME"`

scripts/remove-github-action-runner.sh
```diff
@@ -2,15 +2,29 @@
...
+
 cd actions-runner
-sudo systemctl stop actions.runner.epinio-epinio.`hostname`.service
+sudo systemctl stop $ACTIONS_RUNNER_SERVICE
 sudo ./svc.sh uninstall
 ./config.sh remove --token $GITHUB_RUNNER_TOKEN
```
manno, line 30: Not your change, but I wonder if GITHUB_RUNNER_TOKEN can ever contain a string like `/*/*/*/*/../../../../*/*/*/*/../../../../*/*/*/*`. Quoting it might be safer ;)

svollath: Good jint. Will incorporate quoting where needed in the scripts.

scripts/remove-github-action-runner.sh
svollath: A bit too expensive to fully handle that here - also epinio//epinio isn't covered - in the end it's about URL checking and how action.runner config will compose service names, which we would need to match in any case then ...

manno (approved):

scripts/remove-github-action-runner.sh
manno: In that case you can drop the g flag, I think :)

PR #1164, opened 2022-02-01
Fix application update api

thardeck:
Fixes https://github.com/epinio/ui/issues/53

The main issue is already addressed only the handling of non existing service bindings will be added.


pkg/api/core/v1/models/models.go
```diff
@@ -133,9 +133,10 @@ type ApplicationCreateRequest struct {
 // run, and the services bound to it.
 // Note: Instances is a pointer to give us a nil value separate from
 // actual integers, as means of communicating `default`/`no change`.
+// The same goes for Services - nil is the default and means no change.
 type ApplicationUpdateRequest struct {
 	Instances   *int32         `json:"instances"   yaml:"instances,omitempty"`
-	Services    []string       `json:"services"    yaml:"services,omitempty"`
+	Services    *[]string      `json:"services"    yaml:"services,omitempty"`
```
enrichman, line 139: Why a pointer to the slice? I don't think we need it, we could just add the `omitempty` to the JSON tag I suppose.

enrichman (changes requested):

pkg/api/core/v1/models/models.go
andreas-kupries: I disagree.

We have to distinguish the cases of:

  - Do not change the set of services with this update - That is `nil`, and
  - Remove all services - That is the empty array/slice.

enrichman: Ok, I agree on that, but it's something that you can check also without the pointer (as you said without the `omitempty`).
The only difference is that the JSON will contain the key with the `null` value: https://go.dev/play/p/aRCZ144dKk6

```
{"services":["foobar"]}

{"services":[]}
false // Services == nil

{"services":null}
true // Services == nil
```

If we have to send the JSON without the key (I don't know why) we can keep it as-is.

andreas-kupries: Do I understand correctly that a variable of slice type can be compared to nil ?
I.e
```
var foo []string
if foo == nil ...
```

? If so, then yes, the pointer would be superfluous.
Googling ... Oh https://medium.com/@habibridho/golang-nil-vs-empty-slice-87fd51c0a4d says exactly that.

So, first, apologies to Tim for saying that we need a pointer here.
Not needing it should simplify things.

thardeck: Thanks @enrichman.

enrichman (approved):

andreas-kupries (dismissed):

andreas-kupries (approved):

Issue #1163, opened 2022-02-01
Port forward


enrichman:
The `port-forward` functionality is needed in order to enable the possibility of remote debugging (#1099).

PR #1172, opened 2022-02-01
Port Forward command and endpoint

enrichman:
https://github.com/epinio/epinio/issues/1163

- added port-forward command
- updated some k8s apis
- added portforward endpoint
- fix for env command not showing usage


internal/api/v1/application/portforward.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	// app exists but has no workload to connect to
+	if app.Workload == nil {
+		return apierror.NewAPIError("Cannot connect to application without workload",
+			"", http.StatusBadRequest)
+	}
+
+	// TODO: Find podName from application and params (e.g. instance number etc).
+	// The application may have more than one pods.
```
jimmykarily, line 46: This should be doable now, you can already implement this. Check this for an example: https://github.com/epinio/epinio/pull/1184

enrichman: Ah yes, I forgot to ask you about this, but when I've read the TODO I thought about it! Let's try

pkg/api/core/v1/client/apps.go
```diff
@@ -413,3 +415,79 @@ func (c *Client) AppExec(namespace string, appName string, tty kubectlterm.TTY)
...
+	dialer := gospdy.NewDialer(upgradeRoundTripper, &http.Client{Transport: wrapper}, "GET", portForwardURL)
+	fw, err := portforward.NewOnAddresses(dialer, opts.Address, opts.Ports, opts.StopChannel, opts.ReadyChannel, opts.Out, opts.ErrOut)
+	if err != nil {
+		return err
+	}
+
+	return fw.ForwardPorts()
+}
+
+func (c *Client) addAuthTokenToUrl(url *url.URL) error {
```
manno, line 482: Note: 
Hm, this doesn't just add the auth token to the URL, it also makes a whole new GET request to fetch the auth token. So the func name does not suggest that.

I probably would not do this extraction for the sake of de-duplication.  I preferred the direct return of values (instead of modifying a pointer arg) and having the request logic/error handling in the top level function.

On the other hand it seems wrong to repeat the `values.Add("authtoken", token)` logic. That's an implementation detail of the auth endpoint 🤷

internal/api/v1/application/controller.go
```diff
@@ -1,6 +1,28 @@
...
+	"github.com/epinio/epinio/helpers/kubernetes"
+	"github.com/epinio/epinio/internal/namespaces"
+	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
+)
+
 // Controller represents all functionality of the API related to applications
 type Controller struct {
 }
+
+func (c Controller) validateNamespace(ctx context.Context, cluster *kubernetes.Cluster, namespace string) apierror.APIErrors {
```
manno, line 16: Wow, this was repeated a lot. I expected more validations in this func, but since it returns an error, the name is a good fit.

internal/cli/apps.go
```diff
@@ -1,32 +1,21 @@
...
-var ()
-
 // CmdApp implements the  command: epinio app
 var CmdApp = &cobra.Command{
 	Use:           "app",
 	Aliases:       []string{"apps"},
 	Short:         "Epinio application features",
 	Long:          `Manage epinio application`,
-	SilenceErrors: true,
-	SilenceUsage:  true,
```
manno, line 22: I don't understand  much about the SilenceUsage flag, when this is absent, the command will just print the usage and we can get rid of lines 24-29?

internal/api/v1/application/controller.go
enrichman: Yes, actually I was trying to add some other validations in the controller or utility but it was "too much" for this PR.

I would like also to create only one controller in the Routes, but also that is something completely unrelated atm.

manno:
There are some wip and fix commits in this PR. I don't think we have any guidelines for that, but I prefer what's described under ["Commit Guidelines"](https://git-scm.com/book/en/v2/Distributed-Git-Contributing-to-a-Project)

> Capitalized, short (50 chars or less) summary

Describes the change concisely, maybe using the imperative. If something isn't revertible on it's own, without breaking tests, I put it together with another commit. Anyhow, could just squash in the end :)

internal/cli/apps.go
enrichman: That's correct, in this case is fine to leave everything out, so the `epinio app` will just print the usage, as it was done in the `RunE`.

internal/cli/apps.go
```diff
@@ -51,6 +40,10 @@ func init() {
 	CmdApp.AddCommand(CmdAppList)
 	CmdApp.AddCommand(CmdAppLogs)
 	CmdApp.AddCommand(CmdAppExec)
+
+	CmdApp.AddCommand(CmdAppPortForward)
+	CmdAppPortForward.Flags().StringSliceVar(&portForwardAddress, "address", []string{"localhost"}, "Addresses to listen on (comma separated). Only accepts IP addresses or localhost as a value. When localhost is supplied, kubectl will try to bind on both 127.0.0.1 and ::1 and will fail if neither of these addresses are available to bind.")
```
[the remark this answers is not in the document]
enrichman: That's fine, altough I put it here to keep it close where the command was added. :smile:

pkg/api/core/v1/client/apps.go
enrichman: The drawback is that if we are going to change the auth logic (i.e. using a JWT, adding an header, or even just changing the query param name) then we would need to change it everywhere.

The main reason of not returning the value is that in this case you are kind of telling the caller that you have probably created a new struct and you should rely on that one. And also you can keep it shorter checking the error in one line.

For ref the oauth library for example is modifying the request in order to add the token:
https://github.com/golang/oauth2/blob/d3ed0bb246c8d3c75b63937d9a5eecff9c74d7fe/token.go#L74-L81

enrichman: I was thinking that we could "split" the logic, adding a function to the Client like `Authenticate` that will do the request and saving the token (or wich auth method we will decide to use) and then a `SetAuthenticationToken` to the request, so at least the fetch/set logic is separate and the flow will be something like

```
func (c *Client) addAuthTokenToUrl(url *url.URL) error {
    token, err := c.AuthToken()
    if err != nil {
	return err
    }

    c.authToken = token
    return nil
}

func (c *Client) SetAuth(url *url.URL) {
    values := url.Query()
    values.Add("authtoken", token)
    url.RawQuery = values.Encode()
}
```

```
if err := c.Authenticate(); err != nil {
	return err
}

c.SetAuth(portForwardURL)
```

manno: > The drawback is that if we are going to change the auth logic (i.e. using a JWT, adding an header, or even just changing the query param name) then we would need to change it everywhere.

Well, yes, but "everywhere" is only two places right now. The auth token can only be used for websocket connections. 
The client code does a lot of this low level stuff, e.g. like the instance query param is handled or setting headers. 

But I'm fine with the addAuthTokenToURL ([URL](https://github.com/golang/go/wiki/CodeReviewComments#initialisms)) extraction, it's a bit surprising and the name is a bit long, but the content is short enough and it's not widely used.

As a method on the client I wouldn't want to signal that this can be used for auth in general. Maybe we need a websocket client if we use more of them.

manno (approved):

Issue #1175, opened 2022-02-03
Extract the Job staging scripts into ConfigMaps


andreas-kupries:
See https://github.com/epinio/epinio/pull/1141#pullrequestreview-871449152

@jimmykarily  wrote:
> I believe it's better to have these scripts defined in a ConfigMap and mounted on the Job. It will make them more readable. Also by not embedding them in the binary, it will make it possible to change them without building a new binary (e.g. to debug or change the behavior).

andreas-kupries:
Helmchart PR: https://github.com/epinio/helm-charts/pull/93
Epinio PR: To Come.

PR #1181, opened 2022-02-03
Allow re-staging when previous staging failed - job-based stager issue

### LINKED ISSUE #1179: Unable to re-stage when previous staging failed
The current logic for the application state does not work correctly for the new job-based stager and prevents re-staging when a previous staging run failed. It claims that this failed staging run is still active.

Edit: Folding https://github.com/epinio/epinio/pull/1141#pullrequestreview-871446865 into here also.

> Look into using `Job` states for staged detection, and possible failure, instead of `Pod` state.

Edit: How useful would `k8s.io/kubectl/pkg/util/podutils` be ? Is there something similar for jobs too ?

andreas-kupries:
Fix #1179

### DIFF
--- a/acceptance/api/v1/helpers_test.go
+++ b/acceptance/api/v1/helpers_test.go
@@ -17,6 +17,7 @@ import (
 	"github.com/epinio/epinio/acceptance/testenv"
 	v1 "github.com/epinio/epinio/internal/api/v1"
 	"github.com/epinio/epinio/internal/helmchart"
+	"github.com/epinio/epinio/internal/names"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/pkg/errors"
 
@@ -95,8 +96,8 @@ func stageApplication(appName, namespace string, stageRequest models.StageReques
 	err = json.Unmarshal(b, stage)
 	Expect(err).NotTo(HaveOccurred())
 
-	waitForStaging(stage.Stage.ID)
-
+	jobName := names.GenerateResourceName("stage", namespace, appName, stage.Stage.ID)
+	waitForStaging(jobName)
 	return stage
 }
 
@@ -122,11 +123,11 @@ func deployApplication(appName, namespace string, request models.DeployRequest) 
 	return *deploy
 }
 
-func waitForStaging(stageID string) {
+func waitForStaging(jobName string) {
 	Eventually(func() string {
 		out, err := proc.Kubectl("get", "job",
 			"--namespace", helmchart.StagingNamespace,
-			stageID,
+			jobName,
 			"-o", "jsonpath={.status.conditions[0].status}")
 		Expect(err).NotTo(HaveOccurred(), out)
 		return out
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -369,11 +369,13 @@ func newJobRun(app stageParam) (*batchv1.Job, *corev1.Secret) {
 
 	unpackScript := fmt.Sprintf("echo Unpacking _ _ __ ___ _____ $(whoami) ; mkdir /workspace/source/app ; tar -xvf /workspace/source/%s -C /workspace/source/app/; rm /workspace/source/%s; chown -R 1000:1000 /workspace ; cp -vL /platform/appenv/* /platform/env/ ; ls -la /platform/env ; echo _ _ __ ___ _____ Unpacked", app.BlobUID, app.BlobUID)
 
-	buildpackScript := fmt.Sprintf("echo Creating _ _ __ ___ _____ $(whoami) ; ls -la /platform/env ; /cnb/lifecycle/creator -app=/workspace/source/app -cache-dir=/workspace/cache -uid=1000 -gid=1000 -layers=/layers -platform=/platform -report=/layers/report.toml -process-type=web -skip-restore=false -previous-image=%s %s && ( curl -X POST http://localhost:4191/shutdown || true )",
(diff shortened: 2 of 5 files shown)


acceptance/api/v1/helpers_test.go
```diff
@@ -126,7 +126,7 @@ func waitForStaging(stageID string) {
 	Eventually(func() string {
 		out, err := proc.Kubectl("get", "job",
 			"--namespace", helmchart.StagingNamespace,
-			stageID,
+			"-l", fmt.Sprintf("epinio.suse.org/stage-id=%s", stageID),
```
manno, line 129: note; I see this wasn't a constant before. Would be nice to have labels somewhere, so we know which labels we use and only need to make changes in one place. However we don't want to introduce to many package imports..

internal/api/v1/application/stage.go
```diff
@@ -369,11 +366,13 @@ func newJobRun(app stageParam) (*batchv1.Job, *corev1.Secret) {
 
 	unpackScript := fmt.Sprintf("echo Unpacking _ _ __ ___ _____ $(whoami) ; mkdir /workspace/source/app ; tar -xvf /workspace/source/%s -C /workspace/source/app/; rm /workspace/source/%s; chown -R 1000:1000 /workspace ; cp -vL /platform/appenv/* /platform/env/ ; ls -la /platform/env ; echo _ _ __ ___ _____ Unpacked", app.BlobUID, app.BlobUID)
 
-	buildpackScript := fmt.Sprintf("echo Creating _ _ __ ___ _____ $(whoami) ; ls -la /platform/env ; /cnb/lifecycle/creator -app=/workspace/source/app -cache-dir=/workspace/cache -uid=1000 -gid=1000 -layers=/layers -platform=/platform -report=/layers/report.toml -process-type=web -skip-restore=false -previous-image=%s %s && ( curl -X POST http://localhost:4191/shutdown || true )",
+	buildpackScript := fmt.Sprintf(`trap "curl -X POST http://localhost:4191/shutdown || true" EXIT ; echo Creating _ _ __ ___ _____ $(whoami) ; ls -la /platform/env ; /cnb/lifecycle/creator -app=/workspace/source/app -cache-dir=/workspace/cache -uid=1000 -gid=1000 -layers=/layers -platform=/platform -report=/layers/report.toml -process-type=web -skip-restore=false -previous-image=%s %s`,
```
manno, line 369: hmmm, does this have to be one line?
The back quote is often used for multi-line strings, but maybe multiple lines confuse the shell.

acceptance/api/v1/helpers_test.go
andreas-kupries: I switched this back to a direct name again, constructed in the caller.
Making consts from the labels I see as separate cleanup issue.

internal/api/v1/application/stage.go
andreas-kupries: It started as one line in the PoC and I did not want to touch that ... See #1175 about this.

andreas-kupries: I used the back quote so that I could use `"..."` for the `trap` without having to quote them.

internal/api/v1/application/stage.go
```diff
@@ -268,33 +269,27 @@ func (hc Controller) Staged(c *gin.Context) apierror.APIErrors {
...
-	}
-	if len(podList.Items) == 0 {
-		return apierror.InternalError(fmt.Errorf("no pods in %s with selector %s", namespace, selector))
+	if len(jobList.Items) == 0 {
+		return apierror.InternalError(fmt.Errorf("no jobs in %s with selector %s", namespace, selector))
 	}
 
-	for _, pod := range podList.Items {
-		failed, err := cluster.IsPodFailed(ctx, pod.Name, helmchart.StagingNamespace)
+	for _, job := range jobList.Items {
```
jimmykarily, line 285: I think this is wrong. We don't care about older job being failed, we only care about the latest one.

internal/application/application.go
```diff
@@ -85,19 +85,23 @@ func Exists(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef)
...
+	jobList, err := cluster.ListJobs(ctx, helmchart.StagingNamespace, selector)
 	if err != nil {
 		return false, err
 	}
 
-	// assume that completed jobs are from the past and have a CompletionTime
-	for _, pr := range l.Items {
-		if pr.Status.CompletionTime == nil {
+	for _, job := range jobList.Items {
+		done, err := cluster.IsJobDone(ctx, job.Name, helmchart.StagingNamespace)
```
jimmykarily, line 100: This is `N+1` requests (one to fetch all jobs and 1 request per Job to check if it's done). I believe the Jobs in `jobList.Items` already hold the information needed in order to tell if they are done or not. This can be solved if `IsJobDone` takes a Job as an argument instead of a string. This means, the caller will have to fetch the Job.

jimmykarily: We delete old Jobs so it should be a maximum of 3 requests but I think it still makes sense to fix.

internal/api/v1/application/stage.go
andreas-kupries: The ListJob's selector in front should only find the current.

internal/application/application.go
andreas-kupries: Checking. Note that we should have `N == 1`.

internal/api/v1/application/stage.go
andreas-kupries: See https://github.com/epinio/epinio/blob/961e0db2b5fff331898528cd3b67bccd726d4405/internal/api/v1/application/stage.go#L274-L277

internal/application/application.go
andreas-kupries: Easier to inline, also, can undo the complexification in the helper code.

andreas-kupries: Done. Commit 8b803e28fde413bad1103bc10b813f448b7484a9

internal/api/v1/application/stage.go
jimmykarily: true. That's ok then.

jimmykarily (approved):

PR #1184, opened 2022-02-03
1151 shell to specific instance 2

### LINKED ISSUE #1151: Shell to specific app instance
A follow up of : https://github.com/epinio/epinio/issues/1098 (Depends on it)

## Acceptance Criteria

- The user should be able to list instances of an app (as pod names? TBD)
- The user should be able to pass the desired instance to `epinio app exec` and get a shell to that one. This parameter can be optional. If not set, we give a shell to a random instance.
- [Optional] Tab completion for the above parameter. This way the user doesn't have to run one command to discover instances and one to get a shell.

We probably have to implement a new endpoint to list application instances or make it part of `AppShow`. The UI will need some information about instances too so let's make sure we are in sync. cc @richard-cox

jimmykarily:
Fixes #1151

### DIFF
--- a/acceptance/api/v1/application_exec_test.go
+++ b/acceptance/api/v1/application_exec_test.go
@@ -24,88 +24,144 @@ var _ = Describe("AppExec Endpoint", func() {
 
 	containerImageURL := "splatform/sample-app"
 
-	BeforeEach(func() {
-		namespace = catalog.NewNamespaceName()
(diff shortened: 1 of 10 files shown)


andreas-kupries (changes requested):
Missing swagger spec update

enrichman (dismissed):

andreas-kupries (approved):

PR #1185, opened 2022-02-04
Bump ginkgo v2

jimmykarily:


jimmykarily:
Unit tests are failing because github actions is using the workflow definition from `main`, so it tries to run the tests with the old ginkgo version. Locally tests run fine. I suggest we merge to `main` and see if everything is fine.

andreas-kupries (approved):

PR #1189, opened 2022-02-04
Try to deduplicate the staging logs display in tests

### LINKED ISSUE #1177: Simplify handling of staging logs in the testsuite
See https://github.com/epinio/epinio/pull/1141#discussion_r797779699

In reference to `OnStageFailureShowStagingLogs`
@manno wrote:
> It's unfortunate that we have to copy it everywhere. But I guess a more generic FailHandler (https://github.com/epinio/epinio/blob/tekton-removal/acceptance/suite_test.go#L69-L82) wouldn't have access to the context.
>
> Maybe having a more specialized env.EpinioPush(...) that includes this functionality would work.

manno:
Fix #1177

Trying to make sure this is done for every epinio push, however there
was one special case.

### DIFF
--- a/acceptance/apps_env_test.go
+++ b/acceptance/apps_env_test.go
@@ -68,9 +68,7 @@ var _ = Describe("apps env", func() {
 
 			It("is not present in the pushed workload", func() {
 				appDir := "../assets/sample-app"
-				out, err := env.Epinio(appDir, "apps", "push",
-					"--name", appName)
-				env.OnStageFailureShowStagingLogs(err, out, appName)
+				out, err := env.EpinioPush(appDir, appName, "--name", appName)
 				Expect(err).ToNot(HaveOccurred(), out)
 
 				Expect(deployedEnv(namespace, appName)).ToNot(MatchRegexp("MYVAR"))
@@ -104,9 +102,7 @@ var _ = Describe("apps env", func() {
 
 			It("is injected into the pushed workload", func() {
 				appDir := "../assets/sample-app"
-				out, err := env.Epinio(appDir, "apps", "push",
-					"--name", appName)
-				env.OnStageFailureShowStagingLogs(err, out, appName)
+				out, err := env.EpinioPush(appDir, appName, "--name", appName)
 				Expect(err).ToNot(HaveOccurred(), out)
 
 				Expect(deployedEnv(namespace, appName)).To(MatchRegexp("MYVAR"))
@@ -117,9 +113,7 @@ var _ = Describe("apps env", func() {
 	Describe("deployed app", func() {
 		BeforeEach(func() {
 			appDir := "../assets/sample-app"
-			out, err := env.Epinio(appDir, "apps", "push",
-				"--name", appName)
-			env.OnStageFailureShowStagingLogs(err, out, appName)
+			out, err := env.EpinioPush(appDir, appName, "--name", appName)
 			Expect(err).ToNot(HaveOccurred(), out)
 		})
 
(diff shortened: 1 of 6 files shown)


andreas-kupries (approved):
Looks as envisioned

acceptance/apps_env_test.go
```diff
@@ -68,9 +68,7 @@ var _ = Describe("apps env", func() {
 
 			It("is not present in the pushed workload", func() {
 				appDir := "../assets/sample-app"
-				out, err := env.Epinio(appDir, "apps", "push",
-					"--name", appName)
-				env.OnStageFailureShowStagingLogs(err, out, appName)
+				out, err := env.Epinio(appDir, appName)
```
andreas-kupries, line 71: Should this be `EpinioPushDir` ?

acceptance/helpers/machine/apps.go
```diff
@@ -18,11 +18,9 @@ func (m *Machine) MakeApp(appName string, instances int, deployFromCurrentDir bo
 }
 
 func (m *Machine) MakeContainerImageApp(appName string, instances int, containerImageURL string) string {
-	pushOutput, err := m.Epinio("", "apps", "push",
-		"--name", appName,
+	pushOutput, err := m.Epinio("", "apps", "push", appName,
```
andreas-kupries, line 21: Should this be EpinioPushDir ?

acceptance/apps_env_test.go
manno: According to @jimmykarily there is no staging log for container image pushes, so I removed that.

Honestly, the way it works right now (if an error happens, look for staging in the error message), it might just be easier to always check the error in the old `env.Epinio` and not add any new helpers.

acceptance/helpers/machine/apps.go
manno: This is also used to push container imagers.

acceptance/apps_test.go
```diff
@@ -498,8 +488,7 @@ configuration:
 
 				By("pushing the app specified in the manifest")
 
-				out, err := env.Epinio("", "apps", "push", manifestPath)
-				env.OnStageFailureShowStagingLogs(err, out, appName)
+				out, err := env.EpinioPush("", appName, manifestPath)
```
andreas-kupries, line 491: I guess this case is why we have to repeat the `--name appName` in all the other cases, instead of having `Push` adding it automatically to the command line, right ?

andreas-kupries (changes requested):

acceptance/apps_test.go
```diff
@@ -266,10 +260,10 @@ configuration:
 		It("uses the custom builder to stage", func() {
 			By("Pushing a golang app")
 			appDir := "../assets/golang-sample-app"
-			pushLog, err := env.Epinio(appDir, "apps", "push",
+			pushLog, err := env.EpinioPush(appDir,
 				"--name", appName,
+				appName,
```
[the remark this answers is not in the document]
manno: Ah thanks for catching that.

acceptance/apps_test.go
manno: Yes, the app name arg is now duplicated in most requests and it's not really apparent why, unless one looks into the EpinioPush func. However it does shorten the code nicely, at least around pushing, the other epinio commands are still verbose.

I also tried this https://github.com/epinio/epinio/compare/dedup-staginglogs-handler-alt4. Doesn't read as nice, but maybe less surprising?

andreas-kupries (approved):

PR #1199, opened 2022-02-08
Fix 1175 - Move staging scripts to config map

### LINKED ISSUE #1175: Extract the Job staging scripts into ConfigMaps

andreas-kupries:
Fix #1175 See also https://github.com/epinio/helm-charts/pull/93

feat: import helm chart having the stage scripts as config map.
feat: support function to retrieve config maps
feat: constants describing the config map and keys for stage scripts
feat: replace builtin scripts with scripts loaded from config map

### DIFF
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -351,77 +351,31 @@ func newJobRun(app stageParam) (*batchv1.Job, *corev1.Secret) {
 		protocol = "https"
 	}
 
-	// TODO: Extract scripts to config map -- https://github.com/epinio/epinio/issues/1175
 	// TODO: Simplify env setup -- https://github.com/epinio/epinio/issues/1176
 
 	// runtime: AWSCLIImage
-	awsScript := fmt.Sprintf(`echo By _ _ __ ___ _____ $(whoami)
-aws --endpoint-url %s://%s s3 cp s3://%s/%[4]s /workspace/source/%[4]s
-echo _ _ __ ___ _____ Done`,
-		protocol,
-		app.S3ConnectionDetails.Endpoint,
-		app.S3ConnectionDetails.Bucket,
-		app.BlobUID)
+	awsScript := fmt.Sprintf(`PROTOCOL="%s"
+ENDPOINT="%s"
+BUCKET="%s"
+BLOBID="%s"
+source "/stage-support/%s"
+`,
+		protocol, app.S3ConnectionDetails.Endpoint, app.S3ConnectionDetails.Bucket,
+		app.BlobUID, helmchart.EpinioStageDownload)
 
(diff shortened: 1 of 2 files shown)


internal/api/v1/application/stage.go
```diff
@@ -211,6 +219,9 @@ func (hc Controller) Stage(c *gin.Context) apierror.APIErrors {
 		Username:            username,
 		RegistryCAHash:      registryCertificateHash,
 		RegistryCASecret:    registryCertificateSecret,
+		DownloadScript:      stageMap.Data[helmchart.EpinioStageDownload],
+		UnpackScript:        stageMap.Data[helmchart.EpinioStageUnpack],
+		BuildScript:         stageMap.Data[helmchart.EpinioStageBuild],
```
jimmykarily, line 224: Can't we just mount the ConfigMap into the Job? https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#populate-a-volume-with-data-stored-in-a-configmap

andreas-kupries: Note that each script takes parameters. ... On the other hand, while writing this I came up with a way to pass them in via stub scripts which then simply source the mounted files. I.e. the hardwired script becomes

```
BLOB="%s"
source /stage/download
```

And the `%s` is filled via `fmt.Sprintf`.

andreas-kupries: Done. Commit a996c22f33b7f3f07a8af6791c6468951a2acc64
(Also squashed history, and rebased to latest main)

jimmykarily: Can't we pass the needed parameters as argument to the script here? https://github.com/epinio/epinio/pull/1199/files#diff-1d5683322c8ac645535eab16a4aa0ef4a516edecf543777e7de0b8ca5f4f7c7fR572-R576

We don't need a second script to wrap it then. Just `/stage/download %s` . Alternatively, these needed parameters could be set as env variables on the Job container itself. So by simply invoking the script, it would work.

andreas-kupries: Right. My thinking was still to complex. I believe using EV is best, less issue with quoting, and done only once.
Well, at least at this point I only have to update epinio, not the charts.

andreas-kupries: Done. d9a870b37312c87ce2644b97eaffe4834b48f486

jimmykarily (approved):

PR #1200, opened 2022-02-08
Add non-letsencrypt tests on AKS and GKE

ldevulder:
This is to ensure that a default Epinio installation still works on these Public Cloud.

### DIFF
--- a/.github/workflows/aks-basic.yml
+++ b/.github/workflows/aks-basic.yml
@@ -1,0 +1,162 @@
+name: AKS-BASIC-CI
+
+on:
+  pull_request:
+    branches: [ main ]
+    types: [ ready_for_review ]
+    paths:
+      - 'acceptance/install/scenario6_test.go'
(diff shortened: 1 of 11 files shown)


ldevulder:
`AKS-CI` is failing "as usual", with the same error that we need to investigate. But the "basic" tests added to AKS and GKE are green, so this should help us to try to find the issue.

juadk (approved):
LGTM overall
1. Should we update our status badges with these 2 new tests for AKS and GKE instead of having the failed status given by the two others?
2. Can you also update this please? https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)

ldevulder:
> * Should we update our status badges with these 2 new tests for AKS and GKE instead of having the failed status given by the two others?

No better to add the new ones for now, I want to see what is working or nor for now. But I added the new tests.

> * Can you also update this please? https://github.com/epinio/epinio/wiki/Release-Test-Plan(s)

Done!

Issue #1130, opened 2022-01-19
No message after a 500 server error


enrichman:
**Describe the bug**

After have received a 500 error from the API I cannot see any info in the logs.

```
Method: POST | Path: /api/v1/namespaces/workspace/applications/sample/store | User: admin | IP: 10.42.0.1 | Timestamp: 2022-01-19T16:51:27Z | Latency: 1.298522129s | Status: 500 | Message:  
```

In order to see something I had to increase the TRACE_LEVEL.

**To Reproduce**
In my case there was an error with the S3 authentication

**Expected behavior**
The message should report some info about the failure.

PR #1206, opened 2022-02-10
Add error to the Gin context

enrichman:
Fix for https://github.com/epinio/epinio/issues/1130

This PR adds the error to the Gin context, so that the LogFormatter can log them.

### DIFF
--- a/internal/api/v1/response/response.go
+++ b/internal/api/v1/response/response.go
@@ -44,6 +44,15 @@ func Error(c *gin.Context, responseErrors errors.APIErrors) {
 		"origin", c.Request.URL.String(),
 		"error", responseErrors)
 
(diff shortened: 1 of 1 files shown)


enrichman:
Yeah sure, I've put the log in the Logging discussion but obviously I've should added also here:

```
Method: POST | Path: /api/v1/namespaces/workspace/applications/samplejava/stage | User: admin | IP: 10.42.0.1 | Timestamp: 2022-02-10T08:56:54Z | Latency: 1.984465ms | Status: 400 | Message: Error #01: thi was on porpouse  
```

This is handled by the Gin logger, so I guess if multiple errors are thrown then we will have Error #01, Error #02 and so on.

andreas-kupries (approved):

PR #1207, opened 2022-02-10
Mount CA & TLS in staging pod, not only the CA

ldevulder:
See https://github.com/jetstack/cert-manager/issues/2111.

Note: this was already fixed in tekton, but tekton has been removed.


andreas-kupries (dismissed):

ldevulder:
I successfully tested Scenario5 and Scenario6 installation tests on AKS (so with `epinio-ca` and `letsencrypt-staging` CA).

PR #1211, opened 2022-02-14
Reduce epinio docker image size

### LINKED ISSUE #1145: Switch to smaller docker image
The Epinio server uses the [epinio-base image](https://github.com/epinio/epinio/blob/main/images/baseimage-Dockerfile), however we don't need any of the installed tools.

We could switch to distroless or "from scratch" and reduce the image size by about 200mb (260mb vs 60mb). 

This is a good cleanup, because we don't need to build and maintain the base image anymore.

* epinio-server listens on port 80, image needs root
* patch script for CI needs to be adapted
* talks to k8s, does SSL work?
* same image in CI as in [release](https://github.com/epinio/epinio/blob/main/.goreleaser.yml#L71-L108)

thardeck:
We use `FROM scratch` instead of OpenSUSE as base for epinio-server. The image size is reduced from 236MB in v0.4.0 to 47MB.

scratch does not provide a readable write dir so we provide one via
volume mount.

This does have the advantage that we properly split the tmp dir from the
node.
The tmp dir is needed to pull an app from git.

The remaining penalty of scratch is that we can not shell into it
easily, but there are workarounds for it which we can document.

Fixes #1145

### DIFF
--- a/scripts/prepare-environment-k3d.sh
+++ b/scripts/prepare-environment-k3d.sh
@@ -50,10 +50,8 @@ helm upgrade --install \
 	epinio-installer epinio-chartmuseum/epinio-installer \
 	--wait
 
-echo "Importing locally built epinio server image"
-k3d image import -c epinio-acceptance ghcr.io/epinio/epinio-server:latest
-
 # Patch Epinio
+./scripts/build-images.sh
 ./scripts/patch-epinio-deployment.sh
 
 "${EPINIO_BINARY}" config update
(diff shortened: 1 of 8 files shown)


scripts/prepare-environment-k3d.sh
```diff
@@ -50,6 +50,9 @@ helm upgrade --install \
 	epinio-installer epinio-chartmuseum/epinio-installer \
 	--wait
 
+echo "Importing locally built epinio server image"
+k3d image import -c epinio-acceptance splatform/epinio-server:latest
```
manno, line 54: Please let's not use splatform anymore

manno (changes requested):

thardeck:
This pr requires https://github.com/epinio/helm-charts/pull/103 to be merged.

andreas-kupries (approved):

PR #1212, opened 2022-02-14
feat: move stage id storage and retrieval to app cr, out of workload

### LINKED ISSUE #1162: Unable to get staging logs of an app which never staged successfully
**Describe the bug**

It is not possible to get the staging logs of an app which never staged successfully.
Retrieval of any logs, regular or staging currently requires a workload and an app which never staged does not have such.
For regular the workload is the origin of the logs, so that limit is ok.
For staging logs that limit does not make sense, as the staging pods are the origin and exist.

The cause of the limit are the two facts that
 1. We need the stage id, and
 2. That information is currently only stored in the workload.

Storing the id of the last stage attempt, or of all stage attempts in the App CRD should get around this limitation.

andreas-kupries:
Fix #1162 

See also
  - https://github.com/epinio/helm-charts/pull/98 for the helm chart changes, and
  -  https://github.com/epinio/application/pull/4 for the App CRD work,

### DIFF
--- a/acceptance/helpers/machine/machine.go
+++ b/acceptance/helpers/machine/machine.go
@@ -33,37 +33,7 @@ func New(dir string, user string, password string, root string, epinioBinaryPath
 }
 
 func (m *Machine) ShowStagingLogs(app string) {
-	namespace := m.getNamespace()
-	_, _ = proc.Run("", true, "bash", "-c", fmt.Sprintf(`
-app="%s"
-namespace="%s"
-sns=epinio-staging
-
-echo ""
-echo ___ XXXX APP "($app)" SPACE "($namespace)" ___
-echo ___ LOGS _ _ __ ___ _____ ________ _____________
-for pod in $(kubectl get pods -n "${sns}" -l "app.kubernetes.io/component=staging,app.kubernetes.io/part-of=${namespace},app.kubernetes.io/name=${app}" -o 'jsonpath={.items[*].metadata.name}')
-do
-    for container in $(kubectl get pods -n "${sns}" "${pod}" -o jsonpath='{.spec.initContainers[*].name}')
-    do
-	case ${container} in
-	    *linkerd*) continue ;;
-	esac
-	echo ___ INIT POD "($pod)" C "($container)" _ _ __ ___ _____ ________ _____________
-	kubectl logs -n "${sns}" -c "${container}" "${pod}"
-    done
-    for container in $(kubectl get pods -n "${sns}" "${pod}" -o jsonpath='{.spec.containers[*].name}')
-    do
-	case ${container} in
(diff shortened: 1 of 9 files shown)


jimmykarily:
There is only one issue:

- Deploy an app that stages successfully
- Deploy the app again with a change that breaks staging
- Run `epinio app show`

```
✔️  Details:
|        KEY        |                          VALUE                           |
|-------------------|----------------------------------------------------------|
| Origin            | /home/dimitris/workspace/epinio/epinio/assets/sample-app |
| Status            | 1/1                                                      |
| Username          | admin                                                    |
| StageId           | 8744d27d86bacbd2                                         |
| Age               | 1m10s                                                    |
| Active Routes     |                                                          |
|                   | sample2.172.18.0.2.omg.howdoi.website                    |
| Desired Instances |                                                        1 |
| Bound Services    |                                                          |
| Environment       |                                                          |
```

Stage id `8744d27d86bacbd2` above, it the id of the failed staging. But the application is still running the old version. By looking at this output, there is no way to tell that this stage ID doesn't match the running app or that it failed. I'm wondering if we should keep that information too. E.g. store the currently running stage ID too, when the 2 differ, it means the last staging process failed to stage or deploy. We will update the currently running stage ID when we deploy successfully. 

Another option may be to just store the status of the last stage ID, whether it was a success or not. We won't be able to tell which stage ID matches the currently running workload but I don't think that information can be used anyway. We don't support setting the stage ID to any log fetching endpoint, so not sure that we need it.

Maybe there are other solutions to this problem that I missed?

jimmykarily:
All good now:


```
|        KEY        |                          VALUE                           |
|-------------------|----------------------------------------------------------|
| Origin            | /home/dimitris/workspace/epinio/epinio/assets/sample-app |
| Status            | 1/1                                                      |
| Username          | admin                                                    |
| Running StageId   | 7e7f06536d62e6b6                                         |
| Last StageId      | 22e1af5101ef945c                                         |
| Age               | 3m17s                                                    |
| Active Routes     |                                                          |
|                   | sample2.172.18.0.2.omg.howdoi.website                    |
| Desired Instances |                                                        1 |
| Bound Services    |                                                          |
| Environment       |                                                          |

✔️  Instances: 
|           NAME           | READY |  MEMORY  | MILLICPUS | RESTARTS |  AGE  |
|--------------------------|-------|----------|-----------|----------|-------|
| sample2-5999444446-9z7cp | true  | 44.4 MiB |         0 |        0 | 3m17s |

```

jimmykarily (approved):

andreas-kupries:
Thanks. Will start merging from the subordinate PR's on up. May wait on the CI to be green again here.

andreas-kupries:
Did not wait on CI.

PR #1215, opened 2022-02-15
Return staging logs in order when fetched by `app logs` command

### LINKED ISSUE #1196: Staging logs appear in random order
`epinio app logs --staging myapp` returns the logs from the various containers in random order. The reason is [the way we select the containers](https://github.com/epinio/epinio/blob/4db57ba9306acc69a455331faa6af311a0f20e66/internal/application/application.go#L345-L358) doesn't specify a specific order.

Staging happens using 2 init containers which [according to the docs](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) run in order: `the kubelet runs the Pod's init containers in the order they appear in the Pod's spec` and one container which run after the init container. So the order is predefined and known.

We can add some logic to the logs endpoint, to return the staging container logs in the order in which the container run.

Logs that come out of order can be useless or even confusing to the user.

andreas-kupries:
Fix #1196 

feat: Extend tailer backend with ordered mode for fetching logs
note: stream mode ignores ordering flag
feat: activate ordered mode when retrieving staging logs
      (affects only log command, which fetches)
      (streamed staging logs on push ignore this)

  - Containers are retrieved in order (init before main)
  - In ordered mode the logs are fetched and returned serially, not concurrently

### DIFF
--- a/helpers/kubernetes/tailer/setup.go
+++ b/helpers/kubernetes/tailer/setup.go
@@ -32,6 +32,7 @@ type Config struct {
 	LabelSelector         labels.Selector
 	TailLines             *int64
 	Template              *template.Template // Template to apply to log entries for formatting
+	Ordered               bool               // Featch/stream logs in container order, synchronously
 }
 
 // ContainerLogLine is an object that represents a line from the logs of a container.
@@ -54,6 +55,7 @@ func FetchLogs(ctx context.Context, logChan chan ContainerLogLine, wg *sync.Wait
 		return errors.New("no namespace set for tailing logs")
 	}
 
+	logger.Info("list pods")
 	podList, err := cluster.Kubectl.CoreV1().Pods(namespace).List(
 		ctx, metav1.ListOptions{LabelSelector: config.LabelSelector.String()})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


helpers/kubernetes/tailer/setup.go
```diff
@@ -86,22 +88,45 @@ func FetchLogs(ctx context.Context, logChan chan ContainerLogLine, wg *sync.Wait
 		return true
 	}
 
+	logger.Info("filter pods, containers")
+
 	for _, pod := range podList.Items {
-		for _, c := range pod.Spec.Containers {
+		for _, c := range pod.Spec.InitContainers {
```
jimmykarily, line 94: I wonder if it's guaranteed that `InitContainers` come always in the right order. By trying out it seems so and it's also logical to be the case, since they are supposed to be executed in order. I guess Kubernetes is using this very same slice to execute them in order.

jimmykarily (approved):

helpers/kubernetes/tailer/setup.go
andreas-kupries: Yep.

PR #1218, opened 2022-02-16
Add timeout for zypper to succeed and option cleanup

svollath:
Already tested within terraform. Zypper command can only run exclusive. With the timeout - instead of the script exiting - zypper will wait up to 300s for any other zypper process to finish.


ldevulder (approved):
LGTM.

NOTE: add this PR in Epinio project and in `To Review` state then, easier for people to see that the PR needs to be reviewed ;-)

PR #1220, opened 2022-02-18
Revert "Rebuild image instead of copying binary during patching"

thardeck:
This fixes the issue of failing CI on public cloud providers.

Even though we are reverting the patching, the git defined Docker image
is still tested on our regular github actions ci runners, so the
testing coverage is better then before.

This reverts commit 9e372d96ae363ba7dbc73537d7ac13837016151c.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -133,7 +133,7 @@ fmt:
 check:
 	golangci-lint run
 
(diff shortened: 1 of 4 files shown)


ldevulder (approved):
Tested successfully on my RKE lab.

PR #1225, opened 2022-02-21
Deploy, update, and remove applications via helm

### LINKED ISSUE #1224: Epinio uses helm to create the workload's deployment
This is a follow up. Ref #361 

As a first, we want to replace the deployment manifests with a generic helm chart.
See #851 the previous attempt for a generic helm chart that can be used.
We don't embed secrets and services. It is fine for the chart to work only on the Epinio cluster.

Later stories will deal with customization of the chart and being able to take (generated) charts to another cluster.

Work is done in
  * #1225
  * https://github.com/epinio/helm-charts/pull/105
  * https://github.com/epinio/application/pull/5

andreas-kupries:
Fix #1224 

Deploy, update, and remove applications via helm.

See also
* https://github.com/epinio/application/pull/5  - New app CRD field `imageURL`.
* https://github.com/epinio/helm-charts/pull/105 - app chart, and changes to epinio chart.


internal/api/v1/application/deploy.go
```diff
@@ -64,14 +51,30 @@ func (hc Controller) Deploy(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err, "failed to get access to a kube client")
 	}
 
+	return DeployCurrentAppState(c, log, ctx, cluster, req.App, username, req.ImageURL, req.Stage.ID, &req.Origin)
+}
+
+func DeployCurrentAppState(c *gin.Context, log logr.Logger, ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef, username, imageURL, stageID string, origin *models.ApplicationOrigin) apierror.APIErrors {
```
enrichman, line 57: We probably don't need the gin.Context, and the logger can/should be fetched from the `context.Context`:
```golang
log := requestctx.Logger(ctx)
```

andreas-kupries: Agree on `log` ... gin is currently used for the OKReturn at the end ... However that likely has to be moved out as it looks to be specific to the deploy endpoint and should not be used for the updating endpoints.

andreas-kupries: It was quick work yesterday between meetings, not too much thought went into the signature then.

andreas-kupries: Commit d6059482226188462b3cd17b7fba44e5004c035a

andreas-kupries:
Currently blocked on failure to deploy any kind of application.
The go helm client used (https://github.com/mittwald/go-helm-client (MGHC)) reports:
> release installed successfully: ear.haddock/epinio-application-0.1.6

and then nothing happens. IOW none of the expected resources are created.

When this happened earlier this week it was thought to be a permission issue.
As test and fix the SA `epinio:epinio-server` was given a new CRB linking it to CR `cluster-admin` (CA), i.e. permissions: __yes__.

This actually worked. Resources were created, and MGHC was reporting more things, i.e. that it would create this and that, that something had not changed (when updating), etc.

Except for apps using `--container-image`. They behaved as before, with nothing happening.
And now the regular apps are also back to nothing happening.

Confirmed that the CRB for CA is present in the cluster. I.e. permissions should be no issue any longer.

Manually deploying using a regular helm client from outside the api server pod works.
__Also__ working is using a regular helm client in a debug pod with the same SA (.i.e. permissions) and other setup.
Used images from https://hub.docker.com/r/alpine/helm and https://hub.docker.com/r/dtzar/helm-kubectl .
This second thing is of course again from within the cluster (because pod). 

Currently stumped again ... In part frustrated that there is no good error reporting when the underlying helm/helm libraries (HHL) have trouble (I cannot say if that is an issue of the HHL, or of MGHC).

Ideas for testing so far

  - change the debug helm pod to an SA without cluster-admin permissions, just the base api server.
    see if that triggers the issue from such a pod also.
  - get help from MGHC wrt more detailed error reporting

---

Went with changing the debug pod. This ran indeed into a permissions issue (listing replicasets).
After adding this to the epinio-server CR deployment from the debug pod worked.
Deploying from the epinio-server pod is still not happening.
Still added the change to the system: See 9da376c24197d4c8416d12c25750a79df13bd119

andreas-kupries:
Trial: Run epinio server outside the cluster, see if that unbreaks the helm

andreas-kupries:
Push started working again despite no material changes (added debug in the helm code, via `vendor/` directory. Switching back to regular packages did __not break__ push again).

Some testsuite failures from local run for -api and -cli. See below.
Fix added for the release name length limit.

```
Summarizing 3 Failures:
  [FAIL] AppDeploy Endpoint with non-staging using custom container image deploying an app with custom routes [It] the app Ingress matches the specified route
  /home/work/SUSE/dev/epinio/acceptance/api/v1/application_deploy_test.go:186
  [FAIL] AppUpdate Endpoint routes have changed [It] synchronizes the ingresses of the application with the new routes list
  /home/work/SUSE/dev/epinio/acceptance/api/v1/application_update_test.go:112
  [FAIL] AppLogs Endpoint [It] should follow logs
  /home/work/SUSE/dev/epinio/acceptance/api/v1/application_logs_test.go:119

Ran 94 of 94 Specs in 1091.501 seconds
FAIL! -- 91 Passed | 3 Failed | 0 Pending | 0 Skipped

Summarizing 5 Failures:
  [FAIL] Apps push and delete [It] should not fail for a max-length application name
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:532
  [FAIL] Apps push and delete [It] deploys an app from the current dir
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:446
  [FAIL] Apps logs [It] follows logs
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:870
  [FAIL] Apps pushing with custom route flag [It] creates an ingress matching the custom route
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:231
  [FAIL] Apps push and delete [It] deploys a golang app
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:420
```

Running through CI now to see if there are material differences wrt test failures.

I suspect that the Ingress issues are true. Because the chart simply creates all ingresses as configured. And on update I do not believe that it will drop the ingresses for the removed routes. This may require specific options for upgrade.

andreas-kupries:
Ingress notes:
  - I forgot to properly separate the ingresses in the range/iteration. Result looks to be a merge where only the last is essentially created.
  - I forgot to create the certs for the ingresses
  - I forgot to handle the ingress class name (if specified)

This requires
  - tls issuer, and
  - ingress (class name)

as additional API values.

andreas-kupries:
Somewhat better ... Still issue with `too long resource names` ... See if we can do the truncation with md5 in helm. If not we have to pass in the names of all derived resources ... If we do not wish to give up having (part of) the app name in the resource name. If we do that we can simply compute and pass in an md5 hash and use that to derived short names all over.

> ternary x (print (trunc x (sub 62 (len (adler32sum x)))) "-" (adler32sum x)) (lt (len x) 63)

My current best attempt. md5 not available. sha256sum available, not suitable, the armored hash alone is 64 characters.

The failure to follow logs may be not having enough time for the ingress to be present and active ... helm is a slower than direct creation of our resources.

andreas-kupries:
Actually, sha256sum can be made suitable, just trunc it.

acceptance/api/v1/application_portforward_test.go
```diff
@@ -92,16 +92,19 @@ var _ = Describe("AppPortForward Endpoint", func() {
 			})
 		})
 
-		When("you you specify a specific instance", func() {
+		When("you specify a specific instance", func() {
 			var conn httpstream.Connection
 			var connErr error
 
+			// Use separate application instead of the main of the suite
+			appName := "portforward"
```
jimmykarily, line 100:
```
This detail is really important. Ginkgo requires, by default, that specs be fully independent. This allows Ginkgo to shuffle the order of specs and run specs in parallel. We'll cover this in more detail later on but for now embrace this takeaway: "Declare in container nodes, initialize in setup nodes".
```
from here: https://onsi.github.io/ginkgo/#extracting-common-setup-beforeeach

internal/api/v1/deploy/deploy.go
```diff
@@ -0,0 +1,192 @@
+package deploy
```
manno, line 1: hm, this package is straight forward, but a [package comment](https://github.com/golang/go/wiki/CodeReviewComments#package-comments) could be added.

acceptance/api/v1/application_portforward_test.go
andreas-kupries: Extended comment with `Bug fix:` prefix, and also added the link.

helpers/authtoken/token_test.go
```diff
@@ -43,6 +43,6 @@ var _ = Describe("Token", func() {
 
 		_, err := authtoken.Validate(token)
 		Expect(err).To(HaveOccurred())
-		Expect(err.Error()).To(ContainSubstring("token is expired by"))
+		Expect(err.Error()).To(ContainSubstring("token is expired"))
```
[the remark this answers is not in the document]
andreas-kupries: I did not investigate the cause, just noticed the breakage.
An updated JWT library is IMHO a very good candidate for the cause.
No action.

internal/api/v1/deploy/deploy.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	// determine number of desired instances
+	instances, err := application.Scaling(ctx, cluster, app)
+	if err != nil {
+		return nil, apierror.InternalError(err, "failed to access application's desired instances")
+	}
+
+	// determine runtime environment, if any
+	// TODO: Used in helm.Deploy with environment.List().Assignments() - Simplify
```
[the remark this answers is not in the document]
andreas-kupries: No, I do not believe that this TODO is needed anymore. Not as an issue either.

My thinking has changed.
I was not satisfied with having a chain of transforms for the environment
(map(name->value) -> slice (pair) -> slice (yaml-string)).
It is growing on me however, as clearer to read/understand.
Getting the data ordered and converted to yaml are kept separate, instead of mingled.

Action: Removing the comment.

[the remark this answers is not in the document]
andreas-kupries: Done, commit e713fc21cfce6b2f7905a1577a64640418d41899

internal/api/v1/deploy/deploy.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/epinio/epinio/internal/application"
+	"github.com/epinio/epinio/internal/cli/server/requestctx"
+	"github.com/epinio/epinio/internal/helm"
+	"github.com/epinio/epinio/internal/helmchart"
+	"github.com/epinio/epinio/internal/registry"
+	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+)
+
+func DeployCurrentAppState(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef, username, imageURL, stageID string, origin *models.ApplicationOrigin) ([]string, apierror.APIErrors) {
```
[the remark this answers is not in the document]
andreas-kupries: Done, commit e713fc21cfce6b2f7905a1577a64640418d41899

internal/api/v1/deploy/deploy.go
andreas-kupries: Done, commit e713fc21cfce6b2f7905a1577a64640418d41899

acceptance/api/v1/application_portforward_test.go
andreas-kupries: Done, commit e713fc21cfce6b2f7905a1577a64640418d41899

jimmykarily: sorry, what I meant was that it's bad practice to initialize variables in "container" blocks (when, describe, etc). This is explained in the link I sent. This initialization should be moved inside the before block.

andreas-kupries: Ok.

andreas-kupries: Done, 0d2335dad901577d605fdb1aa441c88a93533a02

jimmykarily (changes requested):

internal/helm/helm.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/epinio/epinio/internal/names"
+	"github.com/epinio/epinio/internal/routes"
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	"github.com/go-logr/logr"
+	hc "github.com/mittwald/go-helm-client"
+	"github.com/spf13/viper"
+)
+
+const (
+	StandardChart = "https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.9/epinio-application-0.1.9.tgz"
```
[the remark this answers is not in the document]
andreas-kupries: I believe that the helm client caches this. At least I have set the necessary config elements to make it possible:

> internal/helm/helm.go:153:			RepositoryCache:  "/tmp/.helmcache", // Hopefully reduces chart downloads.
internal/helm/helm.go:154:			RepositoryConfig: "/tmp/.helmrepo",  // s.a.

internal/helm/helm.go
```diff
@@ -0,0 +1,169 @@
...
+	Chart         string                // Chart to use for deployment
+	ImageURL      string                // Application Image
+	Username      string                // User causing the (re)deployment
+	Instances     int32                 // Number Of Desired Replicas
+	StageID       string                // Stage ID for ImageURL
+	Owner         metav1.OwnerReference // App CRD Owner Information
+	Environment   models.EnvVariableMap // App Environment
+	Services      []string              // Bound Services (list of names)
+	Routes        []string              // Desired application routes
+	Start         *int64                // Nano-epoch of deployment. Optional.
```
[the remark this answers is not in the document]
andreas-kupries: Agreed.

internal/api/v1/application/update.go
```diff
@@ -178,10 +154,13 @@ func (hc Controller) Update(c *gin.Context) apierror.APIErrors { // nolint:gocyc
 		if err != nil {
 			return apierror.InternalError(err)
 		}
+	}
 
-		_, err = application.SyncIngresses(ctx, cluster, appRef, username)
```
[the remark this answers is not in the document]
andreas-kupries: I had not done that already ? ... Huh, indeed. Will be removed.

internal/api/v1/deploy/deploy.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+		return nil, apierror.InternalError(err, "failed to get the application resource")
+	}
+
+	stageID, err := application.StageID(applicationCR)
+	if err != nil {
+		return nil, apierror.InternalError(err, "failed to retrieve application's stage id")
+	}
+	if expectedStageID != "" && expectedStageID != stageID {
+		return nil, apierror.InternalError(err, "stage id mismatch")
```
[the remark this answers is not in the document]
andreas-kupries: ... Checking, ah, `StatusBadRequest` ... Agreed.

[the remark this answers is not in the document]
andreas-kupries: Done. a85cf7ace1392a0512bd071519a7cfd2d925d7c3

internal/api/v1/application/update.go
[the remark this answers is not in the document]
andreas-kupries: Done. 992fa13efa18ae5e6001ced2d1db8cdacb3f609a

internal/helm/helm.go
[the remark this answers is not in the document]
andreas-kupries: Done 7a217d55cda8c0cbfdd8b40d8c9a7fdb3f3ab32f

internal/api/v1/deploy/deploy.go
```diff
@@ -0,0 +1,189 @@
...
+		return nil, apierror.InternalError(err, "failed to access application's runtime environment")
+	}
+
+	// determine bound services, if any
+	services, err := application.BoundServiceNames(ctx, cluster, app)
+	if err != nil {
+		return nil, apierror.InternalError(err, "failed to access application's bound services")
+	}
+
+	// NOTE: This retrieves the `applicationCR` again, internally. See if we can simplify.
```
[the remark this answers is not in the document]
andreas-kupries: Done. 2007bbb3b2640f82bdd220561af3b312b56067b6

andreas-kupries:
Ok. The latest changes ( c0eb370 ) are passing the local testsuite just fine. Modulo the S3 cleanup.
The failures in CI all look to happen in startup, i.e. before the helm deployment stuff is in control.

I can only retry CI and hope that whatever issue there is settles and goes away.

internal/helm/helm.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+type ChartParameters struct {
+	models.AppRef                       // Application: name & namespace
+	Cluster       *kubernetes.Cluster   // Cluster to talk to.
+	Chart         string                // Chart to use for deployment
+	ImageURL      string                // Application Image
+	Username      string                // User causing the (re)deployment
+	Instances     int32                 // Number Of Desired Replicas
+	StageID       string                // Stage ID for ImageURL
```
jimmykarily, line 36: The comment seems wrong. The StageID is only [used as a label on the Deployment](https://github.com/epinio/helm-charts/pull/105/files#diff-778ac2f21904c0d549f62adec97acb398f3544de76019c2f7a485de5848bdb21R27). It doesn't relate to the ImageURL as far as I can tell.

internal/helm/helm.go
```diff
@@ -0,0 +1,168 @@
...
+  ingress: %[11]s
+  replicaCount: %[1]d
+  routes: %[8]s
+  services: %[6]s
+  stageID: "%[3]s"
+  tlsIssuer: "%[12]s"
+  username: "%[5]s"
+  %[9]s
+`, parameters.Instances,
+		parameters.Owner.UID,
```
jimmykarily, line 109: Given [we now remove with helm](https://github.com/epinio/epinio/pull/1225/files#diff-44a311773090a2abd70d10d9f026c5d98e1a8f88f37381e50f8b4a30fe784854R240), I wonder if we still need to set owner references on the various resources. After all, if someone deletes the AppCRD without calling `helm uninstall` on the app release, the resulting state will be invalid from the helm's perspective. We can clean this up in another PR.

jimmykarily (dismissed):

internal/helm/helm.go
andreas-kupries: This is the id of the staging run which created the image referenced by the image url.
So maybe `id of the staging run which created the image referenced by the image url` is a suitable comment ?

internal/helm/helm.go
andreas-kupries: True that. Would definitely simplify the chart. There was already an issue with ownership, the certs for the ingress' is now owned by the app, not the ingress as before.

andreas-kupries:
Rebased to catch up to main, in both epinio and helm-charts. Squashed history.

internal/helm/helm.go
jimmykarily: or `Stage ID that produced ImageURL` for short

jimmykarily (approved):

internal/helm/helm.go
andreas-kupries: Done. 3e786567250304aecd59ee8a0219993c09c4b4d6

jimmykarily (approved):

PR #1226, opened 2022-02-21
Execute release drafter on main branch change

### LINKED ISSUE #1223: Add release-drafter to main branch, so we can see a draft release

olblak:
Fix #1223 

This pull request moves the release-drafter from the release GitHub action to a specific one that runs each time a pull request land on the main branch.
The goal is to see at any moment what the release changelog would like to look for the next release.


.github/release-drafter.yml
```diff
@@ -26,8 +26,10 @@ exclude-labels:
   - question
   - QA
 
-change-template: '- $TITLE (#$NUMBER)'
+change-template: '- $TITLE @$AUTHOR (#$NUMBER)'
```
jimmykarily, line 29: We had the author in the past and decided to remove it as not useful information

jimmykarily:
Does this change the release process? I mean, we still need to trigger the release pipeline in order to build binaries, build and push images, open a PR on the helm-charts repo etc. What happends to the draft PRs then?

olblak:
> Does this change the release process? I mean, we still need to trigger the release pipeline in order to build binaries, build and push images, open a PR on the helm-charts repo etc. What happens to the draft PRs then?

Well, they are subtle changes depending on the way we trigger the release between manual tag creation from the command line or via the GitHub interface.

Via the GitHub interface, Release drafter creates a draft release and detects the next version depending on pull request labels. When we publish the release, GitHub automatically creates the tag associated with the release which triggers the release work flow as usual.
The risk is if we push a tag different than the one from the draft release.

So I would suggest slightly modifying the release process by using the GitHub interface instead of manually pushing the tag.
I can demo how I do the release on a different project 

By changing the workflow, we would have access  in advance to the coming release changelog and also reducing the risk of using the wrong version

.github/release-drafter.yml
olblak: I guess it's a taste,  I can remove it

olblak: I usually like to see who's behind a pull request but  I guess it's a personal taste

olblak: I removed the author

olblak:
I am also adding a new Github workflow in this PR that tests releasing Epinio but without anything.
The purpose is to identify early breaking change in the release process

olblak:
@jimmykarily  I am looking for another review :)

.github/release-drafter.yml
```diff
@@ -28,6 +28,8 @@ exclude-labels:
 
 change-template: '- $TITLE (#$NUMBER)'
 change-title-escapes: '\<*_&' # You can add # and @ to disable mentions, and add ` to disable code blocks.
+tag-template: 'v$RESOLVED_VERSION'
+name-template: 'v$RESOLVED_VERSION 🌈'
```
jimmykarily, line 32: what's with the rainbow?

jimmykarily (approved):

PR #1227, opened 2022-02-21
Simplify CI and fix outside_pr CI

ldevulder:
- Use `fetch-depth` to get the whole repo
- Properly clean all at the end of the tests on self-hosted runners
- Make sure that k3d cluster is *really* deleted at the end of the tests
- Make `outside-pr` workflow working again

### DIFF
--- a/.github/workflows/aks.yml
+++ b/.github/workflows/aks.yml
@@ -47,9 +47,6 @@ jobs:
         with:
           submodules: recursive
 
-      - name: Get All Git Tags
-        run: git fetch --force --prune --unshallow --tags
-
       - name: Setup Go
         uses: actions/setup-go@v2
         with:
@@ -71,6 +68,10 @@ jobs:
       - name: Lint Epinio
         run: make lint
 
+      - name: Clean all
+        if: always()
+        uses: colpal/actions-clean@v1
+
   acceptance-scenario6:
     needs:
       - linter
@@ -81,9 +82,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
-
-      - name: Get All Git Tags
-        run: git fetch --force --prune --unshallow --tags
+          fetch-depth: 0
 
       - name: Setup Go
         uses: actions/setup-go@v2
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,7 @@ on:
     - cron:  '0 0 * * *'
 
 env:
-  SETUP_GO_VERSION: '1.17'
+  SETUP_GO_VERSION: '^1.17.2'
   GINKGO_NODES: 8
 
 jobs:
@@ -28,9 +28,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
-
-      - name: Get All Git Tags
-        run: git fetch --force --prune --unshallow --tags
+          fetch-depth: 0
 
       - name: Setup Go
         uses: actions/setup-go@v2
@@ -56,6 +54,10 @@ jobs:
       - name: Unit Tests
         run: make test
 
+      - name: Clean all
+        if: always()
+        uses: colpal/actions-clean@v1
+
   acceptance-cli:
     needs:
       - linter
@@ -66,9 +68,7 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
-
-      - name: Get All Git Tags
-        run: git fetch --force --prune --unshallow --tags
+          fetch-depth: 0
 
       - name: Setup Go
         uses: actions/setup-go@v2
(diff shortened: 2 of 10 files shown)


.github/workflows/aks.yml
```diff
@@ -71,6 +68,10 @@ jobs:
       - name: Lint Epinio
         run: make lint
 
+      - name: Clean all
+        if: always()
```
jimmykarily, line 72: I see [in examples](https://github.com/colpal/actions-clean#usage) curlies being used `${{}}`, are they needed here?

.github/workflows/main.yml
```diff
@@ -103,12 +103,18 @@ jobs:
           REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
           EPINIO_TIMEOUT_MULTIPLIER: 3
         run: |
-          make acceptance-cluster-delete
```
jimmykarily, line 106: We put the cleanup before a job runs to be able to inspect the cluster in case of failures. Personally I never had to do that so I'm fine with this change. I don't know if anyone else found it useful though.

jimmykarily: Hmm and then we added a second cleanup below :D. I guess nodoby ever debugged a CI cluster then. All good.

.github/workflows/outside-pr.yml
```diff
@@ -18,35 +22,57 @@ jobs:
...
         uses: actions/cache@v2
         with:
           path: ${{ github.workspace }}/tools
           key: ${{ runner.os }}-tools
+
+      - name: Login to Docker Hub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
```
jimmykarily, line 45: Being logged it to dockerhub was supposed to be optional. If you are not logged in you are subject to rate limiting but we thought it would be ok for the few external PRs. At some point we broke CI and it didn't succeed without dockerhub credentials.

In any case, external PRs don't have access to our secrets. I don't know how external contributors should define these then (define the secrets on their fork?). With us having moved away from dockerhub for many of our images, maybe we are fine to skip login all together. In the main CI pipeline, I think we even use a mirror registry like we do locally.

You can leave it as it is and handle dockerhub login on another PR if you prefer.

Makefile
```diff
@@ -96,7 +96,7 @@ acceptance-cluster-setup-kind:
 	@./scripts/acceptance-cluster-setup-kind.sh
 
 test-acceptance: showfocus
-	ginkgo -nodes ${GINKGO_NODES} --slow-spec-threshold ${GINKGO_SLOW_TRESHOLD}s -randomize-all --flake-attempts=${FLAKE_ATTEMPTS} --fail-on-pending acceptance/. acceptance/api/v1/. acceptance/apps/.
+	ginkgo -nodes ${GINKGO_NODES} -slow-spec-threshold ${GINKGO_SLOW_TRESHOLD}s -randomize-all --flake-attempts=${FLAKE_ATTEMPTS} --fail-on-pending acceptance/. acceptance/api/v1/. acceptance/apps/.
```
jimmykarily, line 99: I see `--` in `ginkgo run --help`, accidental change?

.github/workflows/aks.yml
ldevulder: We have without them in the UI tests and according to the [docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idif) in that case (no variable) they can be omitted. But I can add them if your prefer, for consistency with the other steps.

.github/workflows/main.yml
ldevulder: We added twice because there was some suspicions one time about k3d cluster still running and causing issues. Not sure that this is a good idea to keep a k3d cluster as the runners are shared for other tests now (rke, UI for example).

.github/workflows/outside-pr.yml
ldevulder: > I don't know how external contributors should define these then (define the secrets on their fork?).
Yes, they should be defined in their fork, it will work in that case.

> You can leave it as it is and handle dockerhub login on another PR if you prefer.
Maybe better yes, as it may have more impacts.

Makefile
ldevulder: Not accidental, it simply because the others `test-acceptance-*` that seem to run well on main CI have `-slow-spec-threshold` and not `--slow-spec-threshold` :-). So it seems that the others are faulty in that case ;-) I will fix this.

jimmykarily (approved):

.github/workflows/outside-pr.yml
ldevulder: >>You can leave it as it is and handle dockerhub login on another PR if you prefer.
>
>Maybe better yes, as it may have more impacts.

I created an issue for this: https://github.com/epinio/epinio/issues/1229

PR #1228, opened 2022-02-22
Single chart - minio TLS

manno:
This refers to #1207

The chart PR is https://github.com/epinio/helm-charts/pull/101

### DIFF
--- a/internal/s3manager/s3manager.go
+++ b/internal/s3manager/s3manager.go
@@ -107,8 +107,6 @@ func New(connectionDetails ConnectionDetails) (*Manager, error) {
 // GetConnectionDetails retrieves s3 details from an ini file stored in a
 // secret.
 // Note: The CLI tool in the staging job will use the ini file directly.
-// TODO: Enrico: we removed the NewXXX constructor since it was not used, but this Get was used instead.
-// Probably it would be clearer to refactor this part.
 func GetConnectionDetails(ctx context.Context, cluster *kubernetes.Cluster, secretNamespace, secretName string) (ConnectionDetails, error) {
 	details := ConnectionDetails{}
 	secret, err := cluster.GetSecret(ctx, secretNamespace, secretName)
--- a/internal/s3manager/s3manager.go
+++ b/internal/s3manager/s3manager.go
@@ -27,18 +27,13 @@ type Manager struct {
 }
 
 type ConnectionDetails struct {
-	Endpoint string
-	// UseSSL toggles the use of SSL for the s3 connection. If we deploy
-	// our own thing, linkerd will take care of this and we set UseSSL to
-	// false.
-	// External S3 implementations should use SSL.
-	UseSSL              bool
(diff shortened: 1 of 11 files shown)


internal/s3manager/s3manager.go
```diff
@@ -101,7 +111,11 @@ func New(connectionDetails ConnectionDetails) (*Manager, error) {
 // GetConnectionDetails retrieves s3 details from an ini file stored in a
 // secret.
 // Note: The CLI tool in the staging job will use the ini file directly.
+// TODO: Enrico: we removed the NewXXX constructor since it was not used, but this Get was used instead.
```
manno, line 114: TODO

internal/s3manager/s3manager.go
```diff
@@ -126,6 +140,15 @@ func GetConnectionDetails(ctx context.Context, cluster *kubernetes.Cluster, secr
 	}
 	details.Bucket = string(secret.Data["bucket"])
 
+	// load ca cert from minio-tls
+	// if secret is missing just log and skip
+	secret, err = cluster.GetSecret(ctx, "epinio", "minio-tls")
+	if err != nil {
+		logger.Error(err, "error while getting minio-tls secret. Cannot add CA")
+	} else {
+		details.CA = secret.Data["ca.crt"]
```
manno, line 149: we might need 'tls.crt' too?

internal/s3manager/s3manager.go
```diff
@@ -126,6 +135,15 @@ func GetConnectionDetails(ctx context.Context, cluster *kubernetes.Cluster, secr
 	}
 	details.Bucket = string(secret.Data["bucket"])
 
+	// load ca cert from minio-tls
+	// if secret is missing just log and skip
+	secret, err = cluster.GetSecret(ctx, "epinio", "minio-tls")
```
jimmykarily, line 140: for now it's fine but this should be an argument we pass to the epinio server. Maybe the user deployed some S3 compatible solution and they have another secret we should trust instead.

jimmykarily: Also epinio server shouldn't have hardcoded minio references. It shouldn't care what S3 it's talking to.

internal/s3manager/s3manager.go
jimmykarily: Added that too.

internal/s3manager/s3manager.go
jimmykarily: added a setting in values.yaml in the helm chart and an argument in `epinio server` command here: https://github.com/epinio/epinio/pull/1228/commits/ab0c4b71dffcf53a1fe0fcb8fe98bcf21dc88cde

jimmykarily:
Only one test failing for irrelevant reason. I created an issue to fix: https://github.com/epinio/epinio/issues/1235 . We can merge.

jimmykarily (approved):

PR #1237, opened 2022-02-23
Fix public cloud provider installation tests to use the epinio chart

jimmykarily:
Needed: https://github.com/epinio/helm-charts/pull/107


acceptance/helpers/epinio/epinio.go
```diff
@@ -32,22 +30,14 @@ func (e *Epinio) Install(args ...string) (string, error) {
...
-	// Install Epinio - Use the epinio-install chart repo directly
 	opts := []string{
 		"upgrade",
 		"--install",
-		"--set", "epinioChart=http://" + internal_ip + ":8080/charts/epinio-0.1.0.tgz",
-		"epinio-installer",
-		"helm-charts/chart/epinio-installer",
+		"-n",
+		"epinio",
+		"--create-namespace",
```
ldevulder, line 38: If you don't use `default` namespace, in that case you also have to add `-n epinio` in `Uninstall` function.

ldevulder (dismissed):
LGTM.
I saw that you removed `skipCertManager=true`, but is CertManager still installed by default if not present? Or we have to add this in all tests?

acceptance/helpers/epinio/epinio.go
jimmykarily: true, fixing

jimmykarily:
> LGTM. I saw that you removed `skipCertManager=true`, but is CertManager still installed by default if not present? Or we have to add this in all tests?

cert-manager became an external dependency, it has to be installed before Epinio. I had a look on one test and it was installing cert-manager, I though all of them did (after all they were skipping the installation)

ldevulder:
> cert-manager became an external dependency, it has to be installed before Epinio. I had a look on one test and it was installing cert-manager, I though all of them did (after all they were skipping the installation)

No some of them use the default installation, but I can do this if you want, and add it to this PR.

enrichman (approved):

.github/workflows/aks-letsencrypt.yml
```diff
@@ -148,7 +148,6 @@ jobs:
           # available version is the one provided by GitHub. So calling a bash script from a
           # Makefile lead to this error: 'Windows Subsystem for Linux has no installed distributions.'
           # The workaround is to execute the bash script directly
-          ./scripts/setup-chart-museum.sh
```
ldevulder, line 151: The comments before can be removed here, there were for chartmuseum ;-)

.github/workflows/aks.yml
```diff
@@ -148,7 +148,6 @@ jobs:
           # available version is the one provided by GitHub. So calling a bash script from a
           # Makefile lead to this error: 'Windows Subsystem for Linux has no installed distributions.'
           # The workaround is to execute the bash script directly
-          ./scripts/setup-chart-museum.sh
```
ldevulder, line 151: Same for the comments.

ldevulder (dismissed):
@enrichman LGTM, just a small modification ;-)

.github/workflows/aks-letsencrypt.yml
enrichman: Ah, I haven't read them throughly :sweat_smile: Thanks

ldevulder:
I saw some small issues that can be fixed later (not related to CI): https://github.com/epinio/epinio/issues/1240

ldevulder (approved):
EKS latest issue will be investigated outside of this PR, it was OK with the last run.

Issue #1236, opened 2022-02-23
Cloud providers CIs are failing for missing steps in Makefile


enrichman:
We have removed from the Makefile the `setup_chart_musuem` step, and added the `install-cert`manager`, but we need to need to update the public CIs workflow.

https://github.com/epinio/epinio/runs/5300760854?check_suite_focus=true

PR #1238, opened 2022-02-23
Fix public cloud CIs

enrichman:
Removed chart museum references and scripts, and added the installation of cert-manager #1236


enrichman:
Added the changes to https://github.com/epinio/epinio/pull/1237

PR #1239, opened 2022-02-24
Ignore services that were deleted in the middle of an "index" request

### LINKED ISSUE #1235: Don't fail when a service is removed in the middle of a request

jimmykarily:
Fixes #1235 

E.g. This happens in tests when a ginkgo node is listing all services
from all namespaces while some other ginkgo node is deleting a service
in that list. Our "index" code, first fetchese the list of services and
then finds details for each one of them. If a service is gone between
listing and fetching the details, the old version would make the whole
request fails. Now the service is simply ignored.

### DIFF
--- a/internal/api/v1/service/index.go
+++ b/internal/api/v1/service/index.go
@@ -11,6 +11,7 @@ import (
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/gin-gonic/gin"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
 // Index handles the API end point /namespaces/:namespace/services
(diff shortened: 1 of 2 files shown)


internal/api/v1/service/index.go
```diff
@@ -56,10 +57,13 @@ func makeResponse(ctx context.Context, appsOf map[string][]string, services serv
 	response := models.ServiceResponseList{}
 
 	for _, service := range services {
-
 		serviceDetails, err := service.Details(ctx)
 		if err != nil {
-			return models.ServiceResponseList{}, err
+			if apierrors.IsNotFound(err) {
```
enrichman, line 62: You can keep it together and avoid the `continue`

```golang
if err != nil && !apierrors.IsNotFound(err){
    return models.ServiceResponseList{}, err
}
```

jimmykarily: good one, fixing

jimmykarily: Hm, I need the continue because I don't want the rest of the code in this block to run

enrichman: Ah, I haven't see the rest of the block. Then it's fine :+1:

enrichman (approved):

PR #1243, opened 2022-02-24
Add `--wait` option to Helm install/upgrade in CI

ldevulder:
This could help to "resolve" sporadic issues we have sometimes in Public Cloud CI. Some of them are because the helm command finish too quickly.

### DIFF
--- a/acceptance/helpers/epinio/epinio.go
+++ b/acceptance/helpers/epinio/epinio.go
@@ -38,6 +38,7 @@ func (e *Epinio) Install(args ...string) (string, error) {
 		"--create-namespace",
 		"epinio",
 		"helm-charts/chart/epinio",
+		"--wait",
 	}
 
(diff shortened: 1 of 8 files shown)


olblak:
Would it make sense to use a different timeout value than the default 5min

olblak:
The reason why I am asking is that sometimes it takes more than 5minutes to create provision loadbalancers.
It could make sense to increase the timeout value to 10min

ldevulder:
> The reason why I am asking is that sometimes it takes more than 5minutes to create provision loadbalancers.

Not sure that waiting for more than 5 minutes is a good idea, and anyway I added code to test that we get an IP for the LB. As in RKE for example we can't have LB configured after Traefik installation, as the LB (metallb) is installed after.

olblak (approved):

ldevulder:
EKS issue is something else, it will be invested separately.

PR #1245, opened 2022-02-25
Badges should target main branch instead of schedule

juadk:
I think that's better with badges targeting main branch instead of schedule event. 
Otherwise you will have to wait for the next schedule to get up-to-date badge status.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,16 +3,16 @@
 Opinionated platform that runs on Kubernetes to take you from Code to URL in one step.
 
 [![godoc](https://pkg.go.dev/badge/epinio/epinio)](https://pkg.go.dev/github.com/epinio/epinio/internal/api/v1)
-[![CI](https://github.com/epinio/epinio/workflows/CI/badge.svg?event=schedule)](https://github.com/epinio/epinio/actions/workflows/main.yml?query=event%3Aschedule)
-[![AKS-CI](https://github.com/epinio/epinio/actions/workflows/aks.yml/badge.svg?event=schedule)](https://github.com/epinio/epinio/actions/workflows/aks.yml?query=event%3Aschedule)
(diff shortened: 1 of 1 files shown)


ldevulder (approved):

ldevulder (approved):

manno:
@juadk after changing this to main, the badges will no longer show flakes, problems with updated versions at the providers , etc.  
They will only be updated if main gets a push, currently that's often, but that might change during vacation time or if work shifts to other repos. 
After this change you can no longer use the badges to see if tests are healthy in the morning.

In theory nothing should be merged into main without being green first, so if test setup is a problem, can't you re-trigger the scheduled tests to control the badges?

juadk:
> @juadk after changing this to main, the badges will no longer show flakes, problems with updated versions at the providers , etc. They will only be updated if main gets a push, currently that's often, but that might change during vacation time or if work shifts to other repos. After this change you can no longer use the badges to see if tests are healthy in the morning.

Scheduled CI is also targeting main branch, so my understanding is that they are also included in the event filter.
https://github.com/epinio/epinio/actions?query=branch%3Amain
And this morning, if I click on a badge, I can see results of tests which were scheduled in the night.

That's not the same as:
```
on:
  push:
    branches: [ main ]
```

It's just an event filter to track tests which are targeting main branch, whatever what trigger the test (manual, scheduled, push etc...)


> In theory nothing should be merged into main without being green first, so if test setup is a problem, can't you re-trigger the scheduled tests to control the badges?
I tried to do that but If I re-trigger the scheduled test, it doesn't take the last code. It restarts it with the code status when it was triggered for the first time. So from my understanding,  you can't push a fix and restart last scheduled test, you have to wait next trigger.

I hope it's more clear what I wanted to do.

juadk:
I think it's ok like that because AKS has failed last night and we can see it:
image

juadk:
After re-triggering the failed test, all is green.

PR #1283, opened 2022-03-07
Added restart comand to cli

enrichman:
Added the `epinio app restart [appName]` command to the `epinio` cli.

This PR adds also a couple of unit tests to the cli, using a mock client.

I've added also a couple of middleware that wrap the NotFound and NoMethod routes, so that we are going to receive JSON responses also in case of 404 or 405.

### DIFF
--- a/pkg/api/core/v1/client/apps_test.go
+++ b/pkg/api/core/v1/client/apps_test.go
@@ -1,0 +1,61 @@
+package client_test
+
+import (
+	"fmt"
+	"net/http"
+	"net/http/httptest"
+
+	"github.com/epinio/epinio/helpers/tracelog"
+	"github.com/epinio/epinio/pkg/api/core/v1/client"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Client Apps unit tests", func() {
+	var epinioClient *client.Client
+	var statusCode int
+	var responseBody string
+
+	BeforeEach(func() {
+		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(statusCode)
+			fmt.Fprint(w, responseBody)
+		}))
+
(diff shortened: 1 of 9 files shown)


andreas-kupries (approved):

andreas-kupries:
Is

> • [0.003 seconds]
Client Apps unit tests AppRestart something bad happened it returns an error
/home/work/SUSE/dev/epinio/pkg/api/core/v1/client/apps_test.go:55
>
>  Begin Captured StdOut/StdErr Output >>
    2022/03/08 10:17:18 "msg"="response is not StatusOK" "error"="Internal Server Error: Error title" "status"=500
  << End Captured StdOut/StdErr Output

expected ? This is from the new unit tests, I believe.

enrichman:
@andreas-kupries nope, but it doesn't happen locally, let's see if it will happen again after this run!

pkg/api/core/v1/client/apps_test.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+var _ = Describe("Client Apps unit tests", func() {
+	var epinioClient *client.Client
+	var statusCode int
+	var responseBody string
+
+	BeforeEach(func() {
+		srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			w.WriteHeader(statusCode)
```
jimmykarily, line 21:
```
When Ginkgo runs a spec it runs through all the BeforeEach closures that appear in that spec's hierarchy from the outer-most to the inner-most.
```
from here: https://onsi.github.io/ginkgo/#organizing-specs-with-container-nodes

If you want to DRY the code this way, I think you need `JustBeforeEach` which allows you to create the `srv` object after you set the `statusCode` and still keep it at the top level block.

enrichman: It works in both cases but it's cleaner with the `JustBeforeEach`, probably because the `statusCode` value is not fetched during the creation of the server but when the handlerFunc is executed.

I'll check the failing tests and add the fix!

jimmykarily (approved):

PR #1301, opened 2022-03-16
Fix for env variable.

enrichman:
This PR fix a bug occuring during the parsing of the env variables

```bash
-> % epinio app create --env "BP_GO_BUILD_LDFLAGS=-X main.Version=myVersion" -- samplego
unable to get app configuration: Bad --env assignment `BP_GO_BUILD_LDFLAGS=-X main.Version=myVersion`, expected `name=value` as value
```

There was a check that an env option should be splitted in exactly two, divided by a `=` sign.
This was failing trying to set a variable like this `VAR=foo=bar`, it was not possible to set the value of `foo=bar` to `VAR`.

```golang
environment := models.EnvVariableMap{}
for _, assignment := range evAssignments {
	pieces := strings.Split(assignment, "=")
	if len(pieces) != 2 {
		return manifest, errors.New("Bad --env assignment `" + assignment + "`, expected `name=value` as value")
	}
	environment[pieces[0]] = pieces[1]
}
```

Moved the `strings.Split` to a `strings.SplitN(s, sep, 2)` so we can split at most at two values.

```golang
environment := models.EnvVariableMap{}
for _, assignment := range evAssignments {
	pieces := strings.SplitN(assignment, "=", 2)
	if len(pieces) < 2 {
		return manifest, errors.New("Bad --env assignment `" + assignment + "`, expected `name=value` as value")
	}
	environment[pieces[0]] = pieces[1]
}
```

I've also renamed the `UpdateISE` to `UpdateICE`, and refactored the `UpdateICE` and the `UpdateBSN` splitting the three logic in different functions.

### note
I don't feel that we should pass the cobra.Command, the `manifest` package should probably just perform logic on the manifest and should not be aware of where the values are coming from, but I didn't want to change too many things.

I'm available to move that logic though. :)

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -74,7 +74,9 @@ var _ = Describe("Apps", func() {
 					"--bind", configurationName,
 					"--instances", "2",
 					"--env", "CREDO=up",
-					"--env", "DOGMA=no")
+					"--env", "DOGMA=no",
+					"--env", "COMPLEX=-X foo=bar",
+				)
 				Expect(err).ToNot(HaveOccurred(), out)
 				Expect(out).To(MatchRegexp("Ok"))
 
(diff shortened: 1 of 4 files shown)


andreas-kupries (dismissed):
LTGM. The only nit I have is that there is no cli test demonstrating a complex `--env` where the changed split makes the difference. Not a blocker tough.

andreas-kupries (approved):

PR #1308, opened 2022-03-21
Goreleaser homebrew

enrichman:
This PR adds the automation needed to bump the Homebrew formulas for `epinio`.

According to the [Goreleaser documentation](https://goreleaser.com/customization/homebrew/?h=homebrew) it cannot create a Formula in the official homebrew repo:

> Note that GoReleaser does not generate a valid homebrew-core formula. The generated formulas are meant to be published as [homebrew taps](https://docs.brew.sh/Taps.html), and in their current form will not be accepted in any of the official homebrew repositories.

so the release will update our new Tap repo: https://github.com/epinio/homebrew-tap

In order to install `epinio` with the latest version the users should install it using the full path

```
brew install epinio/tap/epinio
```

To update also the Formula in the https://github.com/Homebrew/homebrew-core I've added this [bump-homebrew-formula](https://github.com/marketplace/actions/bump-homebrew-formula) Github action.
It will use a `COMMITTER_TOKEN` secret to create a fork and open a PR against it.

After the merge the users will be able to fetch the latest `epinio` also from the core repo with

```
brew install epinio
```

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -55,3 +55,15 @@ jobs:
           repository: epinio/helm-charts
           event-type: epinio-release
           client-payload: '{"ref": "${{ steps.get_tag.outputs.TAG }}"}'
+
+      - name: Bump Homebrew formula
+        uses: mislav/bump-homebrew-formula-action@v1
+        if: "!contains(github.ref, '-')" # skip prereleases
+        with:
+          download-url: https://github.com/epinio/epinio/archive/refs/tags/${{ steps.get_tag.outputs.TAG }}.tar.gz
+          commit-message: |
+            {{formulaName}} {{version}}
+
+            Created by https://github.com/mislav/bump-homebrew-formula-action
+        env:
+          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -89,3 +89,29 @@ dockers:
     # This field does not support wildcards, you can add an entire folder here
     # and use wildcards when you `COPY`/`ADD` in your Dockerfile.
     extra_files: []
+
+brews:
+  - name: epinio
+    description: "CLI for Epinio, the Application Development Engine for Kubernetes"
+    homepage: "https://epinio.io/"
+    license: "Apache-2.0"
+
+    tap:
+      owner: epinio
+      name: homebrew-tap
+
+    folder: Formula
+    url_template: "https://github.com/epinio/epinio/releases/download/{{ .Tag }}/{{ .ArtifactName }}"
+
+    # If set to auto, the release will not be uploaded to the homebrew tap
+    # in case there is an indicator for prerelease in the tag e.g. v1.0.0-rc1
+    # Default is false.
+    skip_upload: "auto"
+
+    test: |
+      output = shell_output("#{bin}/epinio version 2>&1")
+      assert_match "Epinio Version: #{version}", output
+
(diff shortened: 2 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -55,3 +55,15 @@ jobs:
           repository: epinio/helm-charts
           event-type: epinio-release
           client-payload: '{"ref": "${{ steps.get_tag.outputs.TAG }}"}'
+
+      - name: Bump Homebrew formula
+        uses: mislav/bump-homebrew-formula-action@v1
+        if: "!contains(github.ref, '-')" # skip prereleases
+        with:
+          download-url: https://github.com/epinio/epinio/archive/refs/tags/${{ steps.get_tag.outputs.TAG }}.tar.gz
```
olblak, line 63: The download url seems weird to me, because for example the x86 for darwin is :
https://github.com/epinio/epinio/releases/download/v0.6.1/epinio-darwin-x86_64

And I don't see tar.gz file published for version v0.6.0

olblak:
I am wondering if we should bother with the formula thing

.github/workflows/release.yml
enrichman: This is because it's the link of the sources ([https://github.com/epinio/epinio/archive/refs/tags/v0.6.1.tar.gz](https://github.com/epinio/epinio/archive/refs/tags/v0.6.1.tar.gz)).  
[Homebrew doesn't like binary formulae](https://docs.brew.sh/Acceptable-Formulae#we-dont-like-binary-formulae), they prefer to build the binaries from sources during installation, as you can see in the Formula: https://github.com/Homebrew/homebrew-core/blob/master/Formula/epinio.rb#L17-L21

I had a look at `goreleaser` as an example and they do the same, so they have the build approach for the core repo, while they use the binaries in their custom tap:

- https://github.com/Homebrew/homebrew-core/blob/master/Formula/goreleaser.rb
- https://github.com/goreleaser/homebrew-tap/blob/master/Formula/goreleaser.rb


Link to v0.6.0: https://github.com/epinio/epinio/archive/refs/tags/v0.6.0.tar.gz

enrichman:
@olblak with the formula thing you mean bothering updating the version in the core repo?

It seems nice to me, but we can remove it from the main project if you feel so. I will try to use `updatecli` with my fork or another repo to do that, eventually. :stuck_out_tongue:

enrichman:
@olblak I've seen that you used the `homebrew-updatecli` name for the tap, in order to do a simpler `brew tap updatecli/updatecli`.

Should I move to `homebrew-epinio` and so to `brew tap epinio/epinio`?

.github/workflows/release.yml
olblak: Thanks a lot for the clear explanation

olblak:
> It seems nice to me, but we can remove it from the main project if you feel so. I will try to use `updatecli` with my fork or another repo to do that, eventually. :)

It seems nice to me as well but I am concerned by the time it takes to merge pullrequest on  https://github.com/Homebrew/homebrew-core/pulls

olblak:
> @olblak I've seen that you used the `homebrew-updatecli` name for the tap, in order to do a simpler `brew tap updatecli/updatecli`.
> 
> Should I move to `homebrew-epinio` and so to `brew tap epinio/epinio`?

As far as I know running `brew tap` need to be done only once which seems to be a good middle ground, especially considering that we release at least once a week

olblak:
But since you already did most of the work, we can try your approach and disable it in several weeks if we realize that it introduces too many burden

manno (approved):

Issue #1281, opened 2022-03-04
[Spike] Service Marketplace (powered by Helm)


jimmykarily:
The devops team needs to be able to create Service Custom resources on the epinio cluster which will describe how a helm chart should be installed. The information needed is:

- The url for the helm chart (with a version)
- The default values
- A whitelist of values (or a range?) the user is allowed to override
- A description of the service (free text for the listing)
- A way to define what the output of the provisioning process is. It should probably be a secret but some charts create a secret using the name of the release. This is not always a static string, maybe we want a "variable" (e.g. "release_name"). TBD.

More info here: https://github.com/epinio/epinio/issues/1012

jimmykarily:
With @enrichman we gave flux crds a try:

- https://fluxcd.io/docs/components/source/helmrepositories/
- https://fluxcd.io/docs/components/source/helmcharts/
- https://fluxcd.io/docs/components/helm/helmreleases/

with a couple of controllers installed:
- https://github.com/fluxcd/helm-controller
- https://github.com/fluxcd/source-controller

which we installed with `flux install` because the flat yamls (see the github release of each controller) were missing rbac resources so needed more work.

It turns out we could use list `helmcharts` as the available things that can be provisioned and create `helmreleases` in order to provision the services.

We also gave the [rancher/helm-controller](https://github.com/k3s-io/helm-controller) a try. It comes installed by default on k3d. Although it's very simple and does the job of provisioning a helm chart, there is no resource that can be listed that acts as the "available helm charts". In other words, before you create a helmchart (rancher one, no flux), that helm chart is nowhere on the cluster. The moment you create that resource, the helm chart is installed.

I think the flux resources make a better candidate. We now want to see if they have some field there is way to implement:

- service description
- whitelist of values the user is allowed to change
- standard secret output

Regarding the whitelist of values, flux supports [values overrides](https://fluxcd.io/docs/components/helm/helmreleases/#values-overrides) with `spec.values` always winning. Any values defined by the operator in `spec.values` will not be available to the user to override, but this makes it a blacklist instead of a whitelist.

jimmykarily:
If we don't find a way to implement the above features using flux crds, we can create our own Service CRD that has a these attributes and when the user requests a service to be provisioned, we can use the rancher helm-controller to install the helm chart. No flux resources in this solution.

jimmykarily:
The following image describes the solution using a Service CRD and rancher helm-controller:

[...]

PR #1310, opened 2022-03-21
Simplify Configuration names and lookup

jimmykarily:
Part of #1281 
Needs this: https://github.com/epinio/helm-charts/pull/145/files

### DIFF
--- a/internal/configurations/configurations.go
+++ b/internal/configurations/configurations.go
@@ -1,4 +1,11 @@
 // Package configurations encapsulates all the functionality around Epinio configurations
+// A Configuration is essentially a Secret with some Epinio specific labels.
+// This allows us to use any Secret as a Configuration as long as someone labels
+// it as such. In the future, we will use this to expose secrets created by Service
+// helm charts as Configurations (https://github.com/epinio/epinio/issues/1281).
+// Since we don't control the name of the produced secret in that case, we will
+// need some method to tie a Configuration to a Service. This can be solved with
+// some labels on the Configuration or the Service instance resource.
 package configurations
 
 import (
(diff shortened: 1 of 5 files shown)


internal/configurations/configurations.go
```diff
@@ -1,4 +1,11 @@
 // Package configurations encapsulates all the functionality around Epinio configurations
+// A Configuration is essentially a Secret with some Epinio specific labels.
+// This allows us to use any Secret as a Configuration as long as someone labels
+// it as such. In the future, we will use this to expose secrets created by Service
+// helm charts as Configurations (https://github.com/epinio/epinio/issues/1281).
+// Since we don't control the name of the produced secret in that case, we keep
+// the name of the Configuration in a label instead of just using the Secret's name,
+// to be able to have predictable Configuration names when a Service is provisioned.
```
jimmykarily, line 8:
```suggestion
// A Configuration is essentially a Secret with some Epinio specific labels.
// This allows us to use any Secret as a Configuration as long as someone labels
// it as such. In the future, we will use this to expose secrets created by Service
// helm charts as Configurations (https://github.com/epinio/epinio/issues/1281).
// Since we don't control the name of the produced secret in that case, we will
// need some method to tie a Configuration to a Service. This can be solved with
// some labels on the Configuration or the Service instance resource.
```

internal/helm/helm.go
```diff
@@ -18,7 +18,8 @@ import (
 )
 
 const (
-	StandardChart = "https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.13/epinio-application-0.1.13.tgz"
+	StandardChart = "https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.15/epinio-application-0.1.15.tgz"
+
```
jimmykarily, line 22:
```suggestion
```

internal/configurations/configurations.go
```diff
@@ -124,13 +118,11 @@ func CreateConfiguration(ctx context.Context, cluster *kubernetes.Cluster, name,
 		sdata[k] = []byte(v)
 	}
 
-	err = cluster.CreateLabeledSecret(ctx, namespace, secretName, sdata,
+	err = cluster.CreateLabeledSecret(ctx, namespace, name, sdata,
 		map[string]string{
-			// "epinio.suse.org/configuration-type": "custom",
-			"epinio.suse.org/configuration": name,
-			"epinio.suse.org/namespace":     namespace,
-			"app.kubernetes.io/name":        "epinio",
```
andreas-kupries, line 132: I believe that this label should be kept, as a means of declaring (to the world) that this configuration was created by the epinio server.

andreas-kupries (changes requested):
Just a small thing.

internal/configurations/configurations.go
jimmykarily: ok fixing.

PR #1311, opened 2022-03-23
Get Staging Support Image References From Kube/Chart

### LINKED ISSUE #1305: epinio chart and cli ignore the `image.{bash,awscli}.*` settings.
**Describe the bug**

This is a sub issue for #1248.

Rancher requires chart support to support airgapped installation and the like. A chart has to specify images in a specific format rancher will then use to redirect all image requests from the chart to its own registry.

While the epinio chart contains most basic support for the `awscli` and `bash` images the server uses during staging this support is incmplete. The `values.yaml` only contains the necessary repository and tag settings. Missing:

  1. The chart does not use these settings, i.e. they are not stored in cluster resources, and
  2. The epinio binary actually has the images baked into it, i.e. is not reading the information from cluster resources.

**Expected behavior**

The epinio server has to read the image information from cluster resources, and the chart has to pass the settings into whatever cluster resources the reads from.

**Proposed change**

  1. Pass the image references into the same config map the server reads the staging scripts from.
  2. Extend the epinio server to read the image references at the same time it reads the scripts.

andreas-kupries:
Fix #1305 

Read staging image references (paketo, awscli, bash) from kube resources.
See chart PR https://github.com/epinio/helm-charts/pull/148 for the change supplying the information to the resources.

### DIFF
--- a/helpers/kubernetes/cluster.go
+++ b/helpers/kubernetes/cluster.go
@@ -604,6 +604,11 @@ func (c *Cluster) Exec(namespace, podName, containerName string, command, stdin 
 	return strings.TrimSpace(stdout.String()), strings.TrimSpace(stderr.String()), err
 }
 
+// GetConfigMap gets a configmap's values
+func (c *Cluster) GetConfigMap(ctx context.Context, namespace, name string) (*v1.ConfigMap, error) {
+	return c.Kubectl.CoreV1().ConfigMaps(namespace).Get(ctx, name, metav1.GetOptions{})
+}
+
 // GetSecret gets a secret's values
 func (c *Cluster) GetSecret(ctx context.Context, namespace, name string) (*v1.Secret, error) {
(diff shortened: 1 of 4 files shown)


andreas-kupries:
Ok, the issues in CI I am seeing here look to be a consequence of the #1310 merge.

I.e. push app, create config, bind config to app. The redeploy of the app fails in the container creating stage

>   Warning  FailedMount  13s (x8 over 77s)  kubelet, k3d-epinio-acceptance-server-0  MountVolume.SetUp failed for volume "rabbit" : secret "configuration.namespace-workspace.conf-rabbit" not found

I still suspect it is somewhere in my code ... Ah ! This needs app chart 0.1.15, and my code is on 0.1.13 ...
And looking back in the history it seems that this change to 0.1.15 is missing from #1310.
Nothing for it, will fix for myself now.

jimmykarily (approved):

PR #1320, opened 2022-03-30
Bump golangci-lint release to 1.45

thardeck:


juadk:
The following linters are disabled because of go1.18:
```
gosimple
staticcheck
structcheck
unparam
```
As we especially enabled these linters [here](https://github.com/epinio/epinio/blob/main/.golangci.yml#L44-L65), I can not say if we can trust the CI without them.

Issue #1253, opened 2022-02-28
Export helm chart via epinio cli


agracey:
**Is your feature request related to a problem? Please describe.**
Once a developer is at a state that they want to deploy to production, if the target cluster isn't running Epinio, they might need to produce all the pieces required to be picked up by other systems (e.g. Fleet). 

**Describe the solution you'd like**
Add a command `epinio app helm <app-name> <dir>` that pulls the currently deployed chart and values and writes to a directory.

**Describe alternatives you've considered**
Currently, this would have to be done manually or through some convoluted paths in Rancher.

jimmykarily:
The second solution could also export the image (to a tarball inside the `dir`?). For the first iteration we could assume the image is stored on some external registry that will be accessible from any cluster where the app will be deployed. In the future we can export the image or push it to other registries as part of the command.

agracey:
Sounds good. Do you think the two solutions that I described are different enough to make it work doing both?

jimmykarily:
> Sounds good. Do you think the two solutions that I described are different enough to make it work doing both?

If I understand the question correctly, I think one command could cover all cases with some additional flags in the future.

andreas-kupries:
Done.

PR #1324, opened 2022-03-31
Implementation of `app export` command

andreas-kupries:
Ref #1253 

- [x] Client side 
- [x] Server side
- [x] Tests
- [x] Swagger
- [x] values.yaml
- [x] chart tarball

### DIFF
--- a/internal/api/v1/application/part.go
+++ b/internal/api/v1/application/part.go
@@ -10,6 +10,7 @@ import (
 	"github.com/epinio/epinio/internal/cli/server/requestctx"
 	"github.com/epinio/epinio/internal/helm"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/gin-gonic/gin"
 )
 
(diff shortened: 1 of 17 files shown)


internal/cli/usercmd/app.go
```diff
@@ -201,6 +204,59 @@ func (c *EpinioClient) AppShow(appName string) error {
...
+	if err != nil {
+		return err
+	}
+
+	err = c.API.AppGetPart(c.Settings.Namespace, appName, "chart", filepath.Join(directory, "app-chart.tar.gz"))
+	if err != nil {
+		return err
+	}
+
+	//	err = c.API.AppGetPart(c.Settings.Namespace, appName, "image", filepath.Join(directory,"image"))
```
olblak, line 252: Any reason for this block to be commented or is it just an error?

pkg/api/core/v1/client/apps.go
```diff
@@ -138,6 +138,60 @@ func (c *Client) AppShow(namespace string, appName string) (models.App, error) {
...
+
+	defer response.Body.Close()
+	reqLog.V(1).Info("request finished")
+
+	// Create the file
+	out, err := os.Create(destinationPath)
+	if err != nil {
+		return err
+	}
+	defer out.Close()
```
olblak, line 185: Don't you want to catch the potential error using something like 

```
defer func(){
 err = out.Close()
  if err != nil {
    fmt.Println(err)
  } 
}()
```

olblak: It's sometimes usefull to catch error from defer execution

internal/cli/usercmd/app.go
andreas-kupries: `image` is the one part I currently have no full idea how to handle server side. So I blocked it client side for now, while also not wanting to forget the possibility.

pkg/api/core/v1/client/apps.go
andreas-kupries: Hm.

andreas-kupries: Not taking it.

internal/api/v1/application/part.go
```diff
@@ -0,0 +1,93 @@
...
+	if app == nil {
+		return apierror.AppIsNotKnown(appName)
+	}
+
+	if app.Workload == nil {
+		// While the app exists it has no workload, and therefore no chart to export
+		return apierror.NewBadRequest("No chart available for application without workload")
+	}
+
+	if partName == "chart" {
```
jimmykarily, line 52: Split in 3 function? E.g.

```
    switch partName {
    case "chart":
        fetchChart(ctx)
    case "image":
        fetchImage(ctx)
    case "values":
        fetchValues(ctx)
    }
```

internal/api/v1/application/part.go
```diff
@@ -0,0 +1,93 @@
...
+		requestctx.Logger(c.Request.Context()).Info("OK",
+			"origin", c.Request.URL.String(),
+			"returning", fmt.Sprintf("%d bytes %s as is", contentLength, contentType),
+		)
+		c.DataFromReader(http.StatusOK, contentLength, contentType, reader, nil)
+		return nil
+	}
+
+	if partName == "image" {
+		return apierror.NewInternalError("image part not yet supported")
```
jimmykarily, line 76: Maybe an `HTTP 400` is more suitable here? `500`s are unexpected things usually.

internal/api/v1/application/part.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	if partName == "values" {
+		log := requestctx.Logger(ctx)
+
+		yaml, err := helm.Values(cluster, log, app.Meta)
+
```
jimmykarily, line 83:
```suggestion
```

internal/cli/usercmd/app.go
```diff
@@ -201,6 +204,59 @@ func (c *EpinioClient) AppShow(appName string) error {
...
+	// 1. Create directory
+	// 2. Retrieve values.yaml for application
+	// 3. Retrieve chart tarball for application
+	// (4.) Retrieve application image (staged sources)
+	// Save all to the directory (in each step)
+	//
+	// While the `values.yaml` is likely quite small it is
+	// essentially a file like all the other parts (chart tarball,
+	// image). For consistency just handle all of them as files,
+	// in the same manner (web socket ?)
```
jimmykarily, line 235:
```suggestion
```

The code is rather explanatory itself (like it should). I don't think we need to maintain comments as well.

internal/api/v1/application/part.go
andreas-kupries: Done with commit 0f85444f1587cd6e32799c0ac2b12c093be2bd86

internal/cli/usercmd/app.go
andreas-kupries: Done with commit 0f85444f1587cd6e32799c0ac2b12c093be2bd86

internal/api/v1/application/part.go
andreas-kupries: Done with commit 0f85444f1587cd6e32799c0ac2b12c093be2bd86

internal/api/v1/application/part.go
andreas-kupries: Done with commit 0f85444f1587cd6e32799c0ac2b12c093be2bd86

jimmykarily (approved):

gitricko:
@andreas-kupries : I tested the cli and saw the export. But i did not see oci images being exported. How do i use this command ?

andreas-kupries:
The OCI image of the app is currently not exported. At the moment only the chart tarball and the configuration (values.yaml) are. IIRC @jimmykarily mentioned it in the demo that the image is not exported.

jimmykarily:
@gitricko for now this feature is only useful if:
- The registry is external (thus the container images are accessible from anywhere)
- The image is exported to another registry manually + the helm chart is edited to point to that other registry

We decided it was ok for this the first iteration but we can improve it more by exporting the image too in the future (see my comment here: https://github.com/epinio/epinio/issues/1253#issuecomment-1066572188). Trying to think how users would use this feature, I feel it is a natural decision to keep the container registry accessible by both the epinio cluster and the other one (non-epinio) because it is simpler and doesn't have any serious drawbacks. If you think this assumption is wrong, please let us know and we will try to prioritize the exporting of the image.

jimmykarily:
@gitricko feel free to add your comments here: https://github.com/epinio/epinio/issues/1352

PR #1338, opened 2022-04-05
Bump subcharts and dependencies to fix security vulnerabilities

jimmykarily:
reported by artifacthub scanning

This PR was created to test this one: https://github.com/epinio/helm-charts/pull/164


thardeck (approved):

thardeck:
The helm chart was not rebased on 0.7.3 so the cluster could not came up because the Nginx version was moved to values.yaml. I had to bump it to 0.8.0 though because two service related tests were failing.

andreas-kupries (approved):

PR #1339, opened 2022-04-06
1332 service instances

### LINKED ISSUE #1332: Service CRUD
Follow up of: https://github.com/epinio/epinio/issues/1281

We now have a catalog of services. We need endpoints and command to create, delete, list, show service instances.

## Acceptance Criteria

- [x] There is an `epinio service create` command that creates services from the catalog
- [x] There is an `epinio service show` command that displays detailed information about a service instance (e.g. the service it belongs to, whether it's ready/deployed or not, etc)
- [x] There is an `epinio service bind` command that binds the service's secrets to an application and annotates them to makes them epinio "Configurations"
- [x] No relics are left behind. The custom resource `helmchart` should be deleted when a service instance is deleted. The same should happen when a namespace is deleted with `epinio namespace delete`. - #1350 

- [x] There is an `epinio service delete` command that deletes provisioned services - #1354 
- [x] There is an `epinio service list` command that shows a list of all service instances (in the currently targeted namespace). The catalog service they belong to is also visible. - #1349 

## Notes

- No `update` endpoint yet. For now we won't implement "user defined values" for helm charts. If the operator wants to allow different set of values, they will have to create a different catalog service.

jimmykarily:
Fixes #1332


jimmykarily:
NOTE: Let's not forget to generate docs (swagger)

internal/api/v1/router.go
```diff
@@ -130,6 +130,14 @@ var Routes = routes.NamedRoutes{
 	// Services
 	"ServiceCatalog":     get("/services", errorHandler(service.Controller{}.Catalog)),
 	"ServiceCatalogShow": get("/services/:servicename", errorHandler(service.Controller{}.CatalogShow)),
+	"ServiceCreate":      post("/namespaces/:namespace/services", errorHandler(service.Controller{}.Create)),
+	"ServiceShow":        get("/namespaces/:namespace/services/:service", errorHandler(service.Controller{}.Show)),
+
+	// Bind a service to/from applications
+	"ServiceBindingCreate": post(
+		"/namespaces/:namespace/services/:servicename/configurationbindings",
```
jimmykarily, line 138: I think we shouldn't refer to "configurations" in the url. Maybe this instead?

```
/namespaces/:namespace/services/:servicename/bind
```

internal/api/v1/service/create.go
```diff
@@ -0,0 +1,86 @@
...
+TODO2:
+A service may output more than one secrets, so `epinio service bind` could bind all of them at once.
+
+TODO3:
+Show the type of the configuration and the service it belongs to (if any) in the `epinio configuration list/show`
+
+TODO4:
+remove the hardcoded service/release name during the `epinio service create`
+
+*/
```
jimmykarily, line 36:
```suggestion
```

internal/configurations/configurations.go
```diff
@@ -212,6 +215,41 @@ func (s *Configuration) GetSecret(ctx context.Context) (*v1.Secret, error) {
 	return secret, nil
 }
 
+// LabelServiceSecrets will look for the Opaque secrets released with a service, looking for the
+// app.kubernetes.io/instance label, then it will add the Configuration labels to "create" the configurations
+func LabelServiceSecrets(ctx context.Context, kubeClient *kubernetes.Cluster, namespace, name string) ([]v1.Secret, error) {
+	secretSelector := labels.Set(map[string]string{
+		"app.kubernetes.io/managed-by": "Helm",
+		"app.kubernetes.io/instance":   name,
```
jimmykarily, line 223: I think this needs to be `models.ServiceHelmChartName(name, namespace)` because we had to prefix the name of the helmchart with the target namespace to avoid collisions when 2 services with the same name exist on 2 different namespaces (the helmcharts are all created in the `epinio` namespaces because of how the helm-controller works). CI being green means we don't test this, it should be broken.

internal/helm/helm.go
```diff
@@ -161,6 +162,20 @@ epinio:
...
+	if err != nil {
+		return "", err
+	}
+
+	var r *helmrelease.Release
+	if r, err = client.GetRelease(releaseName); err != nil {
+		return "", err
+	}
+
+	return string(r.Info.Status), nil
```
jimmykarily, line 176: Can `Info` be nil?

internal/services/catalog.go
```diff
@@ -11,9 +11,13 @@ import (
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 )
 
-func NewServiceFromJSONMap(m map[string]interface{}) (*models.Service, error) {
+const (
+	ServiceLabelKey = "application.epinio.io/catalog-service-name"
```
jimmykarily, line 15:
```suggestion
        # Only Helmcharts with this label are considered Epinio "Services". Used to filter out Helmcharts created
        # by other means (manually, k3s etc).
	ServiceLabelKey = "application.epinio.io/catalog-service-name"
```

internal/api/v1/router.go
jimmykarily: done

internal/helm/helm.go
jimmykarily: added a check

jimmykarily:
Needs commit squashing but it's ready to be reviewed.

jimmykarily:
Moved docs here: https://github.com/epinio/docs/issues/91

enrichman (approved):

PR #1340, opened 2022-04-08
Added support for multiple application charts

### LINKED ISSUE #1270: Multiple app helm charts
An operator might want to allow developers to pick from a few different flavors of application to build. 

We should allow:
- Operators to add/remove charts
- Operators to set default chart
- Developers to list charts
- Developers to specify target chart on push

Charts should also have descriptions or labels that we can show in the list.

You could use multiple epinio installs but the overhead is too large to be realistic 

### TODO

- [x] Extend AppCRD to hold chart (URL!) used to deploy app. -- https://github.com/epinio/application/pull/7
- [x] Define CRD for holding chart references (name, url, description, k8s labels for recognition)
- [x] Create CR for the above CRD, to reference the standard chart. -- https://github.com/epinio/helm-charts/pull/167
- [x] Going for local default initially ~~Define and create ConfigMap to hold the global information about default chart, set to standard. (Check if we already have a config map for global epinio state) -- :construction: Only if we wish to support a global default.~~
- [x] Provide app chart CR name from client. ~~Change app deployment to read default from global map~~.
- [x] Extend API app push endpoint to save to app CR. -- https://github.com/epinio/epinio/pull/1340
- [x] Read chart repo/url and use for deployment.
- [x] Rewrite app export to take app chart information saved in app CR.
- [x] Implement API endpoint and cli command to list app chart CRs.
- [x] Tests for above
- [x] Implement API endpoint and cli command to create app chart CRs.
- [x] Tests for above
- [x] Implement API endpoint and cli command to delete app chart CRs.
- [x] Tests for above
- [x] Implement cli command to change default app chart (local config).
- [x] Tests for above
- [x] Extend `app push` to set app chart by name
- [x] Tests for above
- [x] Extend `app create` to set app chart by name
- [x] Extend API app create endpoint to save app chart name
- [x] Tests for above
- [x] Extend `app update` to change named app chart (undeployed app only)
- [x] Tests for above
- [x] Extend `app manifest` to export app chart name
- [x] Tests for above
- [x] Extend `app show` to display app chart name.
- [x] Modify server app fetch to provide app chart
- [x] Tests for above
- [x] Extend swagger defs in `internal/api/v1/docs/`, regenerate swagger json.
- [x] Fix `app export` for app chart specified by name+helmrepo.

Note: Saving the chart url instead of a name means that (re)deployment still has a chance of using the chart even if the name for it is removed (the ConfigMap). Fail only if the chart itself is not accessible anymore.

andreas-kupries:
Fix #1270

See also 

  - https://github.com/epinio/application/pull/7
  - https://github.com/epinio/helm-charts/pull/167

### DIFF
--- a/acceptance/api/v1/appchart_create_test.go
+++ b/acceptance/api/v1/appchart_create_test.go
@@ -79,24 +79,17 @@ var _ = Describe("ChartCreate Endpoint", func() {
 			Expect(response.StatusCode).To(Equal(http.StatusCreated), string(bodyBytes))
 
 			out, err := proc.Kubectl("get", "appcharts", "-n", "epinio",
-				"standard.direct", "-o", "jsonpath={.spec.description}")
+				"standard.direct", "-o", "json")
 			Expect(err).ToNot(HaveOccurred(), out)
-			Expect(out).To(Equal("Direct standard"))
 
-			out, err = proc.Kubectl("get", "appcharts", "-n", "epinio",
-				"standard.direct", "-o", "jsonpath={.spec.shortDescription}")
-			Expect(err).ToNot(HaveOccurred(), out)
-			Expect(out).To(Equal("Direct url to tarball standard"))
+			var result map[string]interface{}
+			json.Unmarshal([]byte(out), &result)
+			spec := result["spec"].(map[string]interface{})
 
-			out, err = proc.Kubectl("get", "appcharts", "-n", "epinio",
-				"standard.direct", "-o", "jsonpath={.spec.helmChart}")
-			Expect(err).ToNot(HaveOccurred(), out)
-			Expect(out).To(Equal(standardBall))
-
-			out, err = proc.Kubectl("get", "appcharts", "-n", "epinio",
-				"standard.direct", "-o", "jsonpath={.spec.helmRepo}")
-			Expect(err).ToNot(HaveOccurred(), out)
-			Expect(out).To(Equal(""))
+			Expect(spec["description"].(string)).To(Equal("Direct standard"))
+			Expect(spec["shortDescription"].(string)).To(Equal("Direct url to tarball standard"))
+			Expect(spec["helmChart"].(string)).To(Equal(standardBall))
+			Expect(spec["helmRepo"]).To(BeNil())
 		})
 	})
 })
--- a/acceptance/api/v1/appchart_create_test.go
+++ b/acceptance/api/v1/appchart_create_test.go
@@ -18,6 +18,39 @@ var _ = Describe("ChartCreate Endpoint", func() {
 
 	standardBall := "https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.15/epinio-application-0.1.15.tgz"
 
+	When("creating a duplicate chart", func() {
+		BeforeEach(func() {
+			out, err := env.Epinio("", "apps", "chart", "create", "duplicate.api", "fox")
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+
+		AfterEach(func() {
+			out, err := env.Epinio("", "apps", "chart", "delete", "duplicate.api")
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+
+		FIt("fails to create the chart", func() {
+			request := models.ChartCreateRequest{
+				Name:      "duplicate.api",
+				HelmChart: "placeholder",
+			}
+			b, err := json.Marshal(request)
+			Expect(err).ToNot(HaveOccurred())
+
+			url := serverURL + v1.Root + "/" + v1.Routes.Path("ChartCreate")
+			response, err := env.Curl("POST", url, strings.NewReader(string(b)))
+
+			Expect(err).ToNot(HaveOccurred())
+			Expect(response).ToNot(BeNil())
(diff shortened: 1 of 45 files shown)


andreas-kupries:
Notes

  - https://github.com/PrasadG193/helm-clientgo-example/blob/master/main.go is a small example using the low-level helm packages to perform operations ...
  - https://github.com/helm/helm/blob/main/pkg/action/pull.go is the implementation of the low-level `helm pull` action which retrieves the tarball of a chart.

acceptance/api/v1/appchart_create_test.go
```diff
@@ -0,0 +1,102 @@
...
+			Expect(out).To(Equal("Direct url to tarball standard"))
+
+			out, err = proc.Kubectl("get", "appcharts", "-n", "epinio",
+				"standard.direct", "-o", "jsonpath={.spec.helmChart}")
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(Equal(standardBall))
+
+			out, err = proc.Kubectl("get", "appcharts", "-n", "epinio",
+				"standard.direct", "-o", "jsonpath={.spec.helmRepo}")
+			Expect(err).ToNot(HaveOccurred(), out)
```
jimmykarily, line 98: We can get the json once an parse it to avoid 4 requests.

acceptance/appcharts_test.go
```diff
@@ -0,0 +1,254 @@
...
+			Expect(err).ToNot(HaveOccurred(), out)
+
+			By("unsetting default")
+			out, err = env.Epinio("", "apps", "chart", "default", "")
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("Unset Default Application Chart"))
+
+			out, err = env.Epinio("", "apps", "chart", "default")
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("Name: not set, system default applies"))
```
jimmykarily, line 251: what is the "system default"? Is that "standard" or something written in the epinio config?

acceptance/apps_test.go
```diff
@@ -205,6 +209,57 @@ configuration:
...
+						Expect(out).To(MatchRegexp("Ok"))
+					})
+
+					AfterEach(func() {
+						env.DeleteApp(appName1)
+					})
+
+					It("respects the desired app chart", func() {
+						out, err := env.Epinio("", "app", "update", appName1,
+							"--app-chart", "standard")
```
jimmykarily, line 250: The app was created with the "standard" chart. This update didn't do anything afaict. Maybe the app should be created with the `chartName` one, check that it was respected and then try the `update` and check that is was changed.

internal/api/v1/appchart/fullindex.go
```diff
@@ -0,0 +1,29 @@
...
+	"github.com/epinio/epinio/internal/api/v1/response"
+	"github.com/epinio/epinio/internal/appchart"
+	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
+
+	"github.com/gin-gonic/gin"
+)
+
+// FullIndex handles the API endpoint GET /appcharts
+// It lists all the known appcharts in all namespaces
+func (hc Controller) FullIndex(c *gin.Context) apierror.APIErrors {
```
jimmykarily, line 14: There is only one "index" and it's the full one. I will rename this to just "Index" (the file too)

internal/cli/chart.go
```diff
@@ -0,0 +1,170 @@
...
+		desc, err := cmd.Flags().GetString("desc")
+		if err != nil {
+			return errors.Wrap(err, "error reading option --desc")
+		}
+		repo, err := cmd.Flags().GetString("helm-repo")
+		if err != nil {
+			return errors.Wrap(err, "error reading option --helm-repo")
+		}
+
+		// TODO: process options, pass options
```
jimmykarily, line 112: done?

acceptance/appcharts_test.go
jimmykarily: https://github.com/epinio/epinio/blob/94b00d0d52d1fc3f52659d12a03ef33dbbbd93b4/internal/api/v1/application/create.go#L91

internal/helm/helm.go
```diff
@@ -141,9 +146,37 @@ epinio:
...
 
+	client, err := GetHelmClient(parameters.Cluster.RestConfig, logger, parameters.Namespace)
+	if err != nil {
+		return err
+	}
+
+	helmChart := appChart.HelmChart
+	helmVersion := ""
+
+	// See also part.go, fetchAppChart
```
jimmykarily, line 157: There is duplication with this code: https://github.com/epinio/epinio/blob/1270-multi-app-chart-support/internal/api/v1/application/part.go#L158

That method needs to be split in functions with smaller scope. Maybe then we can reuse code here.

internal/cli/chart.go
jimmykarily, line 112:
```suggestion
```

internal/helm/helm.go
jimmykarily: Let's not keep 3 PRs open for this. We can try that on a separate PR.

jimmykarily (approved):

PR #1342, opened 2022-04-08
Add ginkgo info and make it verbose by default

juadk:


jimmykarily (approved):

juadk:
Tested here: https://github.com/epinio/epinio/runs/5886941717?check_suite_focus=true

jimmykarily (approved):

PR #1349, opened 2022-04-12
Service list endpoint

enrichman:
(ref #1332)

This PR adds the `service list` command and endpoint, and a couple of small refinements.

### DIFF
--- a/acceptance/api/v1/service_list_test.go
+++ b/acceptance/api/v1/service_list_test.go
@@ -3,29 +3,27 @@ package v1_test
 import (
 	"encoding/json"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"strings"
 
 	"github.com/epinio/epinio/acceptance/helpers/catalog"
 	"github.com/epinio/epinio/acceptance/helpers/proc"
 	v1 "github.com/epinio/epinio/internal/api/v1"
-	"github.com/epinio/epinio/internal/services"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	helmapiv1 "github.com/k3s-io/helm-controller/pkg/apis/helm.cattle.io/v1"
 )
 
 var _ = Describe("ServiceList Endpoint", func() {
-	var namespace string
+	var namespace1, namespace2 string
 	var catalogService models.CatalogService
 
 	BeforeEach(func() {
-		namespace = catalog.NewNamespaceName()
-		env.SetupAndTargetNamespace(namespace)
+		namespace1 = catalog.NewNamespaceName()
+		env.SetupAndTargetNamespace(namespace1)
+
+		namespace2 = catalog.NewNamespaceName()
+		env.SetupAndTargetNamespace(namespace2)
 
 		catalogService = models.CatalogService{
 			Name:      catalog.NewCatalogServiceName(),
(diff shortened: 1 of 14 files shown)


acceptance/api/v1/service_list_test.go
```diff
@@ -0,0 +1,174 @@
...
+	When("helmchart list is empty", func() {
+		It("returns a 404", func() {
+			endpoint := fmt.Sprintf("%s%s/namespaces/%s/services", serverURL, v1.Root, namespace)
+			response, err := env.Curl("GET", endpoint, strings.NewReader(""))
+			Expect(err).ToNot(HaveOccurred())
+
+			b, err := ioutil.ReadAll(response.Body)
+			Expect(err).ToNot(HaveOccurred())
+
+			fmt.Printf("NS: %s - %+v\n", namespace, string(b))
```
jimmykarily, line 56:
```suggestion

```

acceptance/api/v1/service_list_test.go
```diff
@@ -0,0 +1,174 @@
...
+		createCatalogService(catalogService)
+	})
+
+	AfterEach(func() {
+		deleteCatalogService(catalogService.Name)
+		env.DeleteNamespace(namespace)
+	})
+
+	When("helmchart list is empty", func() {
+		It("returns a 404", func() {
```
jimmykarily, line 48: I think it should be HTTP 200 with empty list (see here too: https://stackoverflow.com/questions/13366730/proper-rest-response-for-empty-table)

acceptance/api/v1/service_list_test.go
```diff
@@ -0,0 +1,174 @@
...
+		var serviceName1, serviceName2 string
+
+		BeforeEach(func() {
+			serviceName1 = catalog.NewServiceName()
+			serviceName2 = catalog.NewServiceName()
+		})
+
+		When("helmchart is not labeled", func() {
+			BeforeEach(func() {
+				helmChart := helmapiv1.HelmChart{
```
jimmykarily, line 72: We did this manually for `show` tests because we wanted to control the labels. In this case you can use this, if we merge and you rebase: https://github.com/epinio/epinio/pull/1350/files#diff-d25fdeedd459208ed788b4626c53fff0c528c84a2a1d7ce233682743c74e673dR7

acceptance/api/v1/service_list_test.go
```diff
@@ -0,0 +1,174 @@
...
+			It("returns a 404", func() {
+				endpoint := fmt.Sprintf("%s%s/namespaces/%s/services", serverURL, v1.Root, namespace)
+				response, err := env.Curl("GET", endpoint, strings.NewReader(""))
+				Expect(err).ToNot(HaveOccurred())
+
+				Expect(response.StatusCode).To(Equal(http.StatusNotFound))
+			})
+		})
+
+		When("helmchart is labeled", func() {
```
jimmykarily, line 104: You don't seem to test the other case. E.g. one labeled and one not. It should only list the labeled in that case.

acceptance/api/v1/service_list_test.go
```diff
@@ -0,0 +1,174 @@
...
+			// Cleanup for all sub-cases
+			AfterEach(func() {
+				out, err := proc.Kubectl("delete", "helmchart", "-n", "epinio", models.ServiceHelmChartName(serviceName1, namespace))
+				Expect(err).ToNot(HaveOccurred(), out)
+
+				out, err = proc.Kubectl("delete", "helmchart", "-n", "epinio", models.ServiceHelmChartName(serviceName2, namespace))
+				Expect(err).ToNot(HaveOccurred(), out)
+			})
+
+			When("helmchart is ready", func() {
```
jimmykarily, line 142: what happens in a service is not ready?

internal/services/instances.go
```diff
@@ -32,16 +32,10 @@ func (s *ServiceClient) Get(ctx context.Context, namespace, name string) (*model
...
 	}
 
-	catalogServiceName := ""
-	for k, v := range srv.GetLabels() {
-		if k == ServiceLabelKey {
-			catalogServiceName = v
-			break
-		}
-	}
+	catalogServiceName, found := srv.GetLabels()[ServiceLabelKey]
```
jimmykarily, line 35: Good one :+1:

internal/services/instances.go
```diff
@@ -117,3 +111,93 @@ func (s *ServiceClient) Create(ctx context.Context, namespace, name string, cata
 		ctx, unstructureHelmChart, metav1.CreateOptions{})
 	return errors.Wrap(err, "error creating helm chart")
 }
+
+func (s *ServiceClient) List(ctx context.Context, namespace string) ([]*models.Service, error) {
+	serviceList := []*models.Service{}
+
+	unstructuredServiceList, err := s.helmChartsKubeClient.Namespace(helmchart.Namespace()).List(ctx, metav1.ListOptions{})
```
jimmykarily, line 118: You can use a label selector here to filter only the helmcharts that have the label key. Simply setting the label key, checks for existence of that label (with no `=`). The `kubectl` equivalent:

```
kubectl get helmcharts -A -l "application.epinio.io/catalog-service-name"
```

internal/services/instances.go
```diff
@@ -117,3 +111,93 @@ func (s *ServiceClient) Create(ctx context.Context, namespace, name string, cata
...
+	helmChartList, err := convertUnstructuredListIntoHelmCharts(unstructuredServiceList)
+	if err != nil {
+		return nil, errors.Wrap(err, "error converting unstructured list to helm charts")
+	}
+
+	filteredServiceList := filterNamespacedHelmCharts(helmChartList, namespace)
+
+	for _, srv := range filteredServiceList {
+		var catalogServicePrefix string
+		catalogServiceName := srv.GetLabels()[ServiceLabelKey]
```
jimmykarily, line 135: this label key is not guaranteed to exist unless you filter like in my comment above

internal/services/instances.go
```diff
@@ -117,3 +111,93 @@ func (s *ServiceClient) Create(ctx context.Context, namespace, name string, cata
...
+	filteredServiceList := []helmapiv1.HelmChart{}
+
+	for _, helmChart := range helmCharts {
+		// check if Service was deployed in the targeting namespace
+		isInTargetNamespace := helmChart.Spec.TargetNamespace == namespace
+
+		// check if it's an Epinio Service
+		_, isEpinioService := helmChart.GetLabels()[ServiceLabelKey]
+
+		if isInTargetNamespace && isEpinioService {
```
jimmykarily, line 197: I had to solve similar issues in my PR here: https://github.com/epinio/epinio/pull/1350/files#diff-c885830d4eb2eb56c3a09cc1c222c860dd316495e927f36e6ac56eb6918293cfR171

I would be way easier, at least in my case, if we had the targetNamespace as a label. I could just delete with `DeleteCollection` and I would avoid the go routine mess. I think is would also simplify the listing here, you would only need to use a LabelSelector when listing. Should I make that change in my PR so you can use it here too? I suggesting we merge mine first because I had another suggestion for a re-use in tests above. What do you think?

internal/services/instances.go
enrichman: The service are manually filtered in the above function, so only the labeled one are there. But I agree that a filter in the get it's better. :smile:

internal/services/instances.go
enrichman: Agree, I was going to propose you the same. :+1:

acceptance/api/v1/service_list_test.go
enrichman: Yes, absolutely. It was my understanding that we were not doing like this, but it was not about collections, only for specific resources. :+1:

internal/services/instances.go
```diff
@@ -133,3 +128,81 @@ func (s *ServiceClient) DeleteAll(ctx context.Context, targetNamespace string) e
...
+		}
+		return nil, errors.Wrap(err, "fetching the service instance")
+	}
+
+	helmChartList, err := convertUnstructuredListIntoHelmCharts(unstructuredServiceList)
+	if err != nil {
+		return nil, errors.Wrap(err, "error converting unstructured list to helm charts")
+	}
+
+	for _, srv := range helmChartList {
```
jimmykarily, line 157: This loop is making 2N+1 queries to the kube API (1 to list all helm charts and then 1 queries per helmchart to find the catalog service and the status). I'm pretty sure this will perform [worse than the namespace listing](https://github.com/epinio/epinio/issues/1353). 

On small improvement would be to fetch all catalog services once, create a lookup map and check against that for existence of the catalog service. This will still be N+2 queries though. If we skip the status of the services in the list view we go down to 2 queries (one for service + one for catalog services).

It would be nice to see the status of the helm release in the listing page but not against performance. I think it should be the helm-controller's job to update the helmchart status to match the one of the helm release (maybe it does?).

enrichman: I've pushed the small improvement that will fetch all the services and then will look for the `Missing` services with a lookup map. I've left the release status, I'm not sure we are going to have many services running in the same workspace, we can drop the status or see if it's going to be a problem in the future.

jimmykarily: Let's keep it as it is for now

jimmykarily (approved):

PR #1350, opened 2022-04-13
Delete services instances belonging to namespace when it's deleted

jimmykarily:
Also refactor the parallel removal of applications to use the panjf2000/ants
library.


andreas-kupries (approved):
LGTM. The concurrent app removal looks definitely simpler too, and is easier to understand. With the details all hidden in the ant pool.

PR #1351, opened 2022-04-13
Enable codecov for unit tests

### LINKED ISSUE #1347: Code coverage
It will be nice to have a code coverage badge, to get an idea of our level of unit tests coverage.

juadk:
Fix #1347


juadk:
First upload appears [here](https://github.com/epinio/epinio/runs/6010109725?check_suite_focus=true).
@enrichman if you want to take a look 😄 
Also this [one](https://codecov.io/gh/epinio/epinio/tree/4d2e15cca4a34ffd8995b52aef446f15c584f7d4) is useful.

juadk:
We should increase the coverage percentage before adding a Codecov badge in the README.

agracey:
It would be nice if they counted code paths that are validated in the integration testing...

juadk:
> It would be nice if they counted code paths that are validated in the integration testing...

That's why I wonder if it makes sense to use Codecov... I will check deeper tomorrow.

enrichman:
@agracey @juadk I did some tests and it seems possibile, while not easy. We should refactor a couple of things, move some stuff around, at least for the tests.

[This article](https://medium.com/@manabie/test-coverage-of-go-services-during-integration-tests-6ff1bdbe33e0) was enlightening, and these are my discoveries about Epinio.

The first problem is that we are starting from a `scratch` Docker image, that is probably too limited. In order to do a couple of tricks we should probably move to an `alpine` (this is what I did in my tests). As I said this can be done only for the integrations tests if we want to keep the `scratch`.

Then I've created a `main_test.go` file, that will have only the command to run the server, so I've moved the stuff inside the `internal/cli/server.go$CmdServer` into a `RunServer()` func, called by the test:

```go
func TestMain(t *testing.T) {
	log.Println(cli.RunServer())
}
```

With `go test -c` is possible to compile a binary that will run this func:

```
GOARCH="amd64" GOOS="linux" CGO_ENABLED=0 go test -c -cover -covermode=count -coverpkg=./...
```

and then copy/patched the generated binary:

```sh
cp epinio.test dist/epinio-linux-amd64 && make patch-epinio-deployment
```

This ALMOST works, the testing server is running into the pod and if called with the `-test.coverprofile=cover.out` it will generate the coverprofile:

```sh
kubectl exec -it epinio-server-65cfd9f8f-4lkb7 -n epinio -- sh
/ $ cd epinio-binary/
/epinio-binary $ ./epinio -test.coverprofile=cover.out
2022/04/15 15:20:23 error creating listener: listen tcp :8030: bind: address already in use
PASS
coverage: 3.2% of statements in ./...
/epinio-binary $ ls
cover.out  epinio
```

Now the issues are that we need a killswitch (like the one in the article) or a different strategy to preserve the coverprofile (maybe a PV would work?). But I'm not sure that killing the server will a `kill 1` will preserve the coverprofile.

To run the server with the correct flag we would need to edit the Deployment

```yaml
...
    spec:
      containers:
      - args:
        - -test.coverprofile=cover.out
        - --port
        - "8030"
```

and I've also used an hardcoded port, because since we are fetching it with `port := viper.GetInt("port")` I guess that it is not working properly while starting the server without the cli.

Still far from getting something to work properly, but it's interesting.

jimmykarily:
@enrichman the epinio server can run outside the cluster with some minimal configuration:

```
SESSION_KEY=test epinio server
```

does this help?

enrichman:
> @enrichman the epinio server can run outside the cluster with some minimal configuration:
> 
> ```
> SESSION_KEY=test epinio server
> ```
> 
> does this help?

Not really, the integration tests are heavily dependent on it running in the cluster, and it still need to be compiled and run with `go test`.

enrichman:
I think that for the original scope of the issue we can add the coverage only for the unit tests, and maybe add another story to add them for the end2end, if we want to.

enrichman (approved):

enrichman (approved):

enrichman (approved):

PR #1354, opened 2022-04-14
1332 delete service instances

jimmykarily:
Part of: #1332


internal/api/v1/service/delete.go
```diff
@@ -0,0 +1,43 @@
...
+	err = kubeServiceClient.Delete(ctx, namespace, serviceName)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
+			return apierror.NewNotFoundError("service not found")
+		}
+
+		return apierror.InternalError(err)
+	}
+
+	response.OKReturn(c, nil)
```
enrichman, line 40: If we are not returning data we can use just `response.OK()`

acceptance/services_test.go
```diff
@@ -0,0 +1,42 @@
...
+			env.SetupAndTargetNamespace(namespace)
+
+			service = catalog.NewServiceName()
+
+			out, err := env.Epinio("", "service", "create", "mysql-dev", service)
+			Expect(err).ToNot(HaveOccurred(), out)
+
+			out, err = env.Epinio("", "service", "show", service)
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(MatchRegexp("Name.*|.*service-instance-268892044"))
```
enrichman, line 25: Does it work with the "fixed" name?

acceptance/api/v1/service_delete_test.go
```diff
@@ -0,0 +1,154 @@
...
+		When("helmchart exists", func() {
+			var serviceName string
+
+			BeforeEach(func() {
+				serviceName = catalog.NewServiceName()
+			})
+
+			When("helmchart is not labeled", func() {
+				BeforeEach(func() {
+					helmChart := helmapiv1.HelmChart{
```
enrichman, line 75: Should we use the MakeInstance func that you already added with the other PR?

enrichman (commented):
Some comments

acceptance/api/v1/service_delete_test.go
jimmykarily: I need to control the labels in this case (see the `when` description)

jimmykarily: I can use the helper in the "labeled" case below though. Changing it.

internal/api/v1/service/delete.go
jimmykarily: true, changing

acceptance/services_test.go
jimmykarily: Stupid me. It only worked because the `|` needs to be escaped (otherwise it's a regex "or"). Fixing

enrichman (approved):

PR #1387, opened 2022-04-25
Add swagger docs for `epinio service`

thardeck:
There seems to be some inconsistency in the parameter naming. Sometimes we use the term `servicename` and some times just `service`. Should we align this? https://github.com/epinio/epinio/blob/main/internal/api/v1/router.go#L133

There might be also a confusion between the naming `Epinio service` for a `catalog service` and a `service instance`.
I have tried to add the term `instance` everywhere  an instance is addressed and add the term `Catalog` to `Epinio Service` to make the differentiation more clear, because even a service instance is a service managed by Epinio.
If you agree with this naming scheme we should also update the cli description of `service`.

Fixes [#91](https://github.com/epinio/docs/issues/91)

### DIFF
--- a/internal/cli/services.go
+++ b/internal/cli/services.go
@@ -113,7 +113,7 @@ var CmdServiceDelete = &cobra.Command{
 }
 var CmdServiceBindCreate = &cobra.Command{
 	Use:   "bind SERVICENAME APPNAME",
-	Short: "Bind a service SERVICENAME to an Epinio app APPNAME",
+	Short: "Bind a service instance SERVICENAME to an Epinio app APPNAME",
 	Args:  cobra.ExactArgs(2),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
--- a/internal/cli/services.go
+++ b/internal/cli/services.go
@@ -70,7 +70,7 @@ var CmdServiceCreate = &cobra.Command{
 		serviceName := args[1]
 
 		err = client.ServiceCreate(catalogServiceName, serviceName)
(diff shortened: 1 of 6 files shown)


jimmykarily:
While working on this we did some back and forth with naming but in the end we agreed to use "Service" for an instance and "CatalogService" for the catalog service. If that's not consistent then we need to fix but I would like us to use the shorter term "service" for one of the two and the instances seem like a better fit.

thardeck:
> While working on this we did some back and forth with naming but in the end we agreed to use "Service" for an instance and "CatalogService" for the catalog service. If that's not consistent then we need to fix but I would like us to use the shorter term "service" for one of the two and the instances seem like a better fit.
All is fine with me as long as it is consistence, so I will adapt it accordingly. Thanks.

internal/cli/services.go
```diff
@@ -70,13 +70,13 @@ var CmdServiceCreate = &cobra.Command{
 		serviceName := args[1]
 
 		err = client.ServiceCreate(catalogServiceName, serviceName)
-		return errors.Wrap(err, "error creating Epinio Service")
+		return errors.Wrap(err, "error creating Epinio catalog service")
```
andreas-kupries, line 73: Catalog service is the input, service is what is created here, if I read the preceding chunk correctly.
```suggestion
		return errors.Wrap(err, "error creating service")
```

andreas-kupries (changes requested):
Mostly ok, except for one error message apparently talking about a catalog service (class) instead of the service (instance).

andreas-kupries (approved):

PR #1403, opened 2022-04-28
Authentication cleanup

enrichman:
## Note

This PR needs the new users from this helm-charts PR
- https://github.com/epinio/helm-charts/pull/184

It would work also without it, but I guess it's better to merge that before.

---

This is a cleanup/refactor made before the authorization spike. It should not change the behaviour of the API.

The main changes are in the `internal/auth/auth.go`.

Some unused structs and func were removed:
- PasswordAuth
- HashBcrypt()
- RandomPasswordAuth()

Then a new `auth.User` struct was added,  and all the methods were adapted accordingly. This struct will match an Epinio User (or account maybe?), and it will be the one that will be returned from the different functions and package (I don't think we should move too much around with secrets and the kubernetes internal, generally speaking). This struct will be also the one that is going to be keep around in the context and session.

```go
// User is a struct containing all the information of an Epinio User
type User struct {
	Username   string
	Password   string
	CreatedAt  time.Time
	Role       string
	Namespaces []string
}
```

So a `NewUserFromSecret` func was added, and the other `GetUserAccounts`, `GetFirstUserAccount` and `GetUserSecretsByAge` were updated accordingly.

It's probably easier to have a look at the result: https://github.com/epinio/epinio/blob/auth-cleanup/internal/auth/auth.go

Other important changes are in the `internal/cli/server/server.go`, so that now we are going to persist the `auth.User` in the session and context. I've also removed the BasicAuth header decoding, that can be done with just

```go
username, password, ok = ctx.Request.BasicAuth()
```

### DIFF
--- a/internal/cli/server/requestctx/context.go
+++ b/internal/cli/server/requestctx/context.go
@@ -5,6 +5,7 @@ import (
 	"context"
 
 	"github.com/epinio/epinio/helpers/tracelog"
+	"github.com/epinio/epinio/internal/auth"
 	"github.com/go-logr/logr"
 )
 
@@ -18,13 +19,17 @@ type UserKey struct{}
 type LoggerKey struct{}
 
 // WithUser adds the user name to the context
-func WithUser(ctx context.Context, val string) context.Context {
+func WithUser(ctx context.Context, val auth.User) context.Context {
 	return context.WithValue(ctx, UserKey{}, val)
 }
 
 // User returns the user name from the context
-func User(ctx context.Context) string {
-	return extractString(ctx, UserKey{})
+func User(ctx context.Context) auth.User {
+	user, ok := ctx.Value(UserKey{}).(auth.User)
+	if !ok {
+		return auth.User{}
+	}
+	return user
 }
 
 // WithID adds the request ID to the context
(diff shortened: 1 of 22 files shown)


andreas-kupries (approved):
Look ok.

andreas-kupries (approved):
looks ok.

wondering however if our tests actually check the various code paths of auth, i.e. try to trigger the various auth failures.

enrichman:
Thanks @andreas-kupries! Actually yes, there are few tests that are checking the basic and session authentication (they failed me for a couple of things, so it was interesting to see that the auth is working fine!).

In the next PR I will add a couple of tests to check also the authorization.

internal/cli/server/requestctx/context.go
```diff
@@ -18,13 +19,17 @@ type UserKey struct{}
 type LoggerKey struct{}
 
 // WithUser adds the user name to the context
-func WithUser(ctx context.Context, val string) context.Context {
+func WithUser(ctx context.Context, val auth.User) context.Context {
 	return context.WithValue(ctx, UserKey{}, val)
 }
 
 // User returns the user name from the context
```
jimmykarily, line 26: Comment is now outdated (sorry for the post-merge review)

Issue #1389, opened 2022-04-26
[Spike] Epinio AuthZ


jimmykarily:
This is an investigation on how [AuthZ in Epinio](https://github.com/epinio/epinio/issues/1375) can be implemented. We don't need to implement all persmissions. We just need to come up with a PoC that would work.

enrichman:
I'll keep this open until we document a bit the authorization.

PR #1418, opened 2022-05-03
Authorization Spike (custom)

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1389

---

This PR adds a middleware for the authorization.

This MW will permit everything to a user with an `admin` role, while it will check if the targeted namespace is owned by the user in case of a `user` role.

### DIFF
--- a/acceptance/helpers/machine/users.go
+++ b/acceptance/helpers/machine/users.go
@@ -1,7 +1,7 @@
 package machine
 
 import (
-	"fmt"
+	"encoding/json"
 	"os"
 	"strings"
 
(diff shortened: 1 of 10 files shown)


internal/api/v1/router.go
```diff
@@ -69,6 +69,9 @@ func put(path string, h gin.HandlerFunc) routes.Route {
 	return routes.NewRoute("PUT", path, h)
 }
 
+// AdminRoutes is the list of restricted routs, only accessible by admins
```
jimmykarily, line 72:
```suggestion
// AdminRoutes is the list of restricted routes, only accessible by admins
```

acceptance/helpers/machine/users.go
```diff
@@ -11,20 +12,25 @@ import (
...
+func (m *Machine) CreateEpinioUser(role string, namespaces []string) (string, string) {
 	user, password := catalog.NewUserCredentials()
+	ns := strings.Join(namespaces, "\n")
+
 	secretData := fmt.Sprintf(`apiVersion: v1
 stringData:
   username: "%s"
   password: "%s"
+  namespaces: |
+    %s
```
andreas-kupries, line 24: The `Join` at line 17 uses only `"\n"`. That means here that the second and further namespaces will not be properly indented. Join should use `"\n...."` (replace each `.` with space, (*)).

This is likely not realized because the usages I see later do not specify more than one namespace.
Thus not hitting the issue.

(*) I used `.`'s to make the count clear.

acceptance/users_test.go
```diff
@@ -107,11 +108,72 @@ var _ = Describe("Users", func() {
...
+			request, err := http.NewRequest("GET", uri, nil)
+			Expect(err).ToNot(HaveOccurred())
+
+			request.SetBasicAuth(user, password)
+			resp, err := env.Client().Do(request)
+			Expect(err).ToNot(HaveOccurred())
+
+			Expect(resp.StatusCode).To(Equal(http.StatusUnauthorized))
+		})
+	})
```
andreas-kupries, line 154: Given the issue above with `CreateEpinioUser` I recommend adding a test where the user is linked to two or more namespaces.

internal/api/v1/auth.go
```diff
@@ -0,0 +1,68 @@
...
+
+	var authorized bool
+	switch user.Role {
+	case "admin":
+		authorized = authorizeAdmin(logger)
+	case "user":
+		authorized = authorizeUser(logger, user, path, namespace)
+	}
+
+	logger.Info(fmt.Sprintf("user [%s] authorized [%t]", user.Username, authorized))
```
andreas-kupries, line 33: Recommend to show role information, and maybe also namespace.

internal/api/v1/auth.go
```diff
@@ -0,0 +1,68 @@
...
+
+func authorizeAdmin(logger logr.Logger) bool {
+	logger.V(1).WithName("authorizeAdmin").Info("user admin is authorized")
+	return true
+}
+
+func authorizeUser(logger logr.Logger, user auth.User, path, namespace string) bool {
+	logger = logger.V(1).WithName("authorizeUser")
+
+	// check if the requested path is not restricted
```
andreas-kupries, line 50:
```suggestion
	// check if the requested path is restricted
```

andreas-kupries (changes requested):

acceptance/helpers/machine/users.go
enrichman: Ah, yes, good one! Thanks!

acceptance/users_test.go
enrichman: Agree, let's do this.

internal/api/v1/auth.go
enrichman: I've logged the role in line 23, I will add also the namespaces, I don't think they will be many. :slightly_smiling_face:

andreas-kupries: I referred here only to __the__ namespace, i.e. the targeted one. Not the list of allowed namespaces

enrichman: Ah, I see. It can be seen in the path then, but it's nice to have it also here for clarification!

acceptance/helpers/machine/users.go
enrichman: :smile: 

```
error converting YAML to JSON: yaml: line 8: could not find expected ':'
```

andreas-kupries: How does the formatted YAML look like ?
Hopefully that makes the misformatting obvious.

andreas-kupries: Ah, this was about confirming the issue. I thought the error was happening for my proposed fix.
Agreed that serialization of a properly constructed struct is safer.

andreas-kupries (approved):

PR #1419, opened 2022-05-03
Prevent `app log` panic / message channel double-close

### LINKED ISSUE #1417: Epinio "apps log" panics because channel is closed twice
This came up during #1416 and blocks it. 
Tried to fix it with a mutex, but then apps log [unit](https://github.com/epinio/epinio/blob/main/acceptance/apps_test.go#L997) and [acceptance tests](https://github.com/epinio/epinio/blob/main/acceptance/apps_test.go#L997) would not terminate.

```
🕞  [test-799dfb8759-5p787] test [29-Apr-2022 14:20:58] NOTICE: ready to handle connections
2022/04/29 16:25:01 EpinioApiClient: "level"=4 "msg"="error reading message" "error"="websocket: close 1000 (normal)"
2022/04/29 16:25:01 EpinioClient/Apps: "level"=4 "msg"="return" "Namespace"="workspace" "Application"="test"
PASS
2022/04/29 16:25:01 EpinioApiClient: "level"=4 "msg"="error sending close message to websocket" "error"="websocket: close sent"
panic: close of closed channel
```

Channels are also garbage collected. However client code seems to rely on either `close()`. We have to make sure to only close the channel once: https://github.com/epinio/epinio/blob/main/pkg/api/core/v1/client/apps.go#L378-L410

andreas-kupries:
Fix #1417 : make app logging closer goroutine on-demand and disable it for `app logs` command.

Recap: For `app logs` context cancellation is late.
Closer races against process termination
When closer wins the race a panic occurs, double close of the message channel.

However, closer is not required, because websocket is closed by server!
Closing by user would be sigterm (^C).
That kills the process and goroutines with no time for this kind of finalization.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,9 @@ build-linux-amd64: build-amd64
 build-amd64:
 	GOARCH="amd64" GOOS="linux" CGO_ENABLED=$(CGO_ENABLED) go build $(BUILD_ARGS) -ldflags '$(LDFLAGS)' -o dist/epinio-linux-amd64
 
+build-cover:
+	GOARCH="amd64" GOOS="linux" CGO_ENABLED=0 go test -c -covermode=count -coverpkg ./... -o dist/epinio-linux-amd64-coverage
+
 build-windows-amd64: build-windows
 build-windows:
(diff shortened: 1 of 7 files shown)


andreas-kupries:
Closing in favor of #1420. That PR significantly simplifies and cleans the client log receiver. My attempt here complexifies it.

PR #1421, opened 2022-05-04
PHP buildpack became more clever. Simplify the sample app

jimmykarily:
Signed-off-by: Dimitris Karakasilis <DKarakasilis@suse.com>


enrichman (approved):

juadk:
Next thing is to fix https://github.com/epinio/example-wordpress (removing `buildpack.yml` etc)

PR #1422, opened 2022-05-04
1370 epinio names

### LINKED ISSUE #1370: Epinio names
Relevant issue: https://github.com/epinio/helm-charts/pull/171

In Epinio we are naming things using [a helper function](https://github.com/epinio/epinio/blob/main/internal/names/names.go#L39), sometimes [calling it directly](https://github.com/epinio/epinio/blob/main/pkg/api/core/v1/models/app.go#L132) some others [using another helper](https://github.com/epinio/epinio/blob/cd6a7c2a290a9781362fdbd7598d0d5c7345ad4c/internal/names/names.go#L22).

There are 2 problems with this approach:

- Adding characters after the `GenerateResourceName` has been called, can result in names longer that allowed
- The generated name is not validated for usage as a Label or Secret name etc. [There are rules](https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names) regarding how resources can be named in Kubernetes. That's how we ended up with https://github.com/epinio/helm-charts/pull/171. 

In most cases we are trying to put too much information in a name, even when it's not needed. E.g. an ingress named like `i-sample-sample.172.18.0.2.omg.howdoi.website` makes no sense. `i-` for Ingress is useless (we know it's an Ingress) and the host is already a field on the Ingress resource itself. Most of the time, the application name is enough. In some cases we may need a random part (as we do with staging Job names) but that can be handled before we call the `GenerateResourceName`.

We can port the Sanitize method from quarks: https://github.com/cloudfoundry-incubator/quarks-utils/blob/master/pkg/names/names.go#L43 to make sure we have valid names (problem #2). We should also change all calls to `GenerateResourceName` to not add text after the call.

We need to fix this in other places too in the helm charts: https://github.com/epinio/helm-charts/blob/0ca9c2953166a38c651db62031abb0386cb2dcfd/chart/application/templates/ingress.yaml#L45
We can fix this either by implementing `GenerateResourceName` in helm or some other mechanism that produces "safe" names (just a checksum of the ingress "id" ?)

jimmykarily:
Fixes #1370 

Sibling PR: https://github.com/epinio/helm-charts/pull/187

### DIFF
--- a/internal/application/workload.go
+++ b/internal/application/workload.go
@@ -261,7 +261,7 @@ func (a *Workload) generatePodInfo(pods []corev1.Pod) map[string]*models.PodInfo
 	for i, pod := range pods {
 		restarts := int32(0)
 		for _, cs := range pod.Status.ContainerStatuses {
-			if cs.Name == a.app.Name {
+			if cs.Name == a.deployment.Name {
 				restarts += cs.RestartCount
 			}
 		}
--- a/internal/application/workload.go
+++ b/internal/application/workload.go
@@ -106,8 +106,17 @@ func (b AppConfigurationBindList) ToNames() []string {
 func (a *Workload) Deployment(ctx context.Context) (*appsv1.Deployment, error) {
 	var err error
 	if a.deployment == nil {
-		a.deployment, err = a.cluster.Kubectl.AppsV1().
-			Deployments(a.app.Namespace).Get(ctx, a.app.Name, metav1.GetOptions{})
+		depList, err := a.cluster.Kubectl.AppsV1().
+			Deployments(a.app.Namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("app.kubernetes.io/component=application,app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s", a.app.Name, a.app.Namespace),
+		})
+		if err != nil {
+			return nil, err
+		}
+		if len(depList.Items) < 1 {
+			return nil, apierrors.NewNotFound(appsv1.Resource("deployment"), a.app.Name)
+		}
+		a.deployment = &depList.Items[0]
 	}
 
 	return a.deployment, err
@@ -202,6 +211,7 @@ func (a *Workload) Get(ctx context.Context) (*models.AppDeployment, error) {
 	}
 
 	return &models.AppDeployment{
+		Name:            deployment.Name,
 		Active:          true,
 		CreatedAt:       createdAt.Format(time.RFC3339), // ISO 8601
 		Replicas:        replicas,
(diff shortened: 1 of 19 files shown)


internal/names/names.go
```diff
@@ -5,46 +5,48 @@ package names
...
+	name = strings.TrimRight(name, "-")
+	return name
 }
 
 // GenerateResourceName joins the input strings with dots (".")  and
 // returns the result, suitably truncated to the maximum length of
 // kube resource names.
 func GenerateResourceName(names ...string) string {
-	return TruncateMD5(strings.Join(names, "."), 63)
+	name := DNSLabelSafe(strings.Join(names, "-"))
```
jimmykarily, line 33: As said here: https://github.com/epinio/helm-charts/pull/187/files#diff-7f0e1fad4fc65b5c6294727baa852f191201feafe9d27c9c0ff5dd28da4ba9beR28
replacing characters could result in collisions. I start thinking that we should simplify that down to just a  md5sum (truncated if needed, although it's always 32 characters long so should be small enough).

andreas-kupries: Maybe always append an MD5, even if the incoming string is short enough ?
In that case the collision in the prefix does not matter because the md5 should be computed from the unmodified string, thus be different, and resolve the collision.

jimmykarily: As we agreed while pairing, we will always use the sha1sum of the original name and prefix it with as many characters are still free from the original name (with invalid characters removed).

internal/application/workload.go
```diff
@@ -106,8 +106,17 @@ func (b AppConfigurationBindList) ToNames() []string {
...
-		a.deployment, err = a.cluster.Kubectl.AppsV1().
-			Deployments(a.app.Namespace).Get(ctx, a.app.Name, metav1.GetOptions{})
+		depList, err := a.cluster.Kubectl.AppsV1().
+			Deployments(a.app.Namespace).List(ctx, metav1.ListOptions{
+			LabelSelector: fmt.Sprintf("app.kubernetes.io/component=application,app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s", a.app.Name, a.app.Namespace),
+		})
+		if err != nil {
+			return nil, err
+		}
+		if len(depList.Items) < 1 {
```
andreas-kupries, line 116: Should we check for `> 1` also ? I.e. ambiguous result, multiple matches.

andreas-kupries (approved):
Approved with a question open.

internal/application/workload.go
jimmykarily: There are only 2 consumers of this method. One of the 2 (line 184 below in this file) ignores all errors except "not found". I will add the check for the other consumer though.

jimmykarily: done below

andreas-kupries (approved):

Issue #1359, opened 2022-04-15
Deleting a service should unbind it


jimmykarily:
Depends on: https://github.com/epinio/epinio/issues/1390 (to reuse code)

- Create an app
- Create a service instance
- Bind the 2
- Delete the service

The service is deleted without a warning or anything. The application still has the secret mounted although the secret is now deleted (?!).

Deleting the bound service should behave the same as the deletion of bound Configurations. The user should either force it or unbind it first. Forcing it, should first unbind it and all apps that had it bound, should be restarted.

jimmykarily:
Relevant story: https://github.com/epinio/epinio/issues/1390

PR #1443, opened 2022-05-12
Fail service unbind if bound to apps and `--unbind` not specified. New option.

andreas-kupries:
Fix: #1359 

:construction: Will not pass CI. Tests not adapted yet.

### DIFF
--- a/internal/api/v1/service/delete.go
+++ b/internal/api/v1/service/delete.go
@@ -9,6 +9,8 @@ import (
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 )
 
+// Delete handles the API end point /namespaces/:namespace/services/:service (DELETE)
+// It deletes the named service
 func (ctr Controller) Delete(c *gin.Context) apierror.APIErrors {
 	ctx := c.Request.Context()
 	namespace := c.Param("namespace")
--- a/internal/api/v1/service/delete.go
+++ b/internal/api/v1/service/delete.go
@@ -1,7 +1,6 @@
 package service
 
 import (
-	"errors"
 	"fmt"
 	"strings"
 
@@ -11,17 +10,12 @@ import (
 	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/cli/server/requestctx"
 	"github.com/epinio/epinio/internal/configurations"
-	"github.com/epinio/epinio/internal/helm"
-	"github.com/epinio/epinio/internal/names"
 	"github.com/epinio/epinio/internal/services"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/gin-gonic/gin"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
-
-	helmrelease "helm.sh/helm/v3/pkg/release"
-	helmdriver "helm.sh/helm/v3/pkg/storage/driver"
 )
 
 // Delete handles the API end point /namespaces/:namespace/services/:service (DELETE)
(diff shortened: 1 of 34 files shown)


internal/api/v1/service/delete.go
```diff
@@ -23,6 +48,76 @@ func (ctr Controller) Delete(c *gin.Context) apierror.APIErrors {
...
+		if errors.Is(err, helmdriver.ErrReleaseNotFound) {
+			return apierror.NewNotFoundError(fmt.Sprintf("%s - %s", err.Error(), releaseName))
+		}
+		return apierror.InternalError(err)
+	}
+
+	logger.Info(fmt.Sprintf("service found %+v\n", serviceName))
+	if srv.Info.Status != helmrelease.StatusDeployed {
+		return apierror.InternalError(err)
+	}
```
andreas-kupries, line 71: @enrichman I added this block of code to have the `delete service` endpoint check that the instance to be deleted actually exists.

As is done by `bind` and `unbind`. After adapting the delete tests to provide the new delete request structure I consistently get `404 release not found` coming from this code.

Was it intentional to not have this validation in the delete endpoint ?
If not then I will have to dig again into the test setup to see if/why no release is present ...
And/or test cli first, as that should match on all sides.

(side note: should be factored into a helper function)

andreas-kupries: Yes, it is definitely this piece of code tripping the test now.

andreas-kupries:
Big rebase, and insertion of fixes.

The big issue was the truncation of the name for the service helm chart CR.

The way we do that it actually expands the name to the given limit. And then comes the chart controller and derives the name of the release from that, which in turn derives pod names etc. from that. And neither seems to perform proper truncation.
Thus we have to make the name of the helm hart CR short enough to leave space for all these to append their stuff without running the kube limits. Under the length of the shasum. Made that possible now, forced a leading letter. Believing that the chance of collision is still low enough.

With this done everything creating a service as part of a test should now work instead of timing out because the service was not deployed without any error messages.

Still leaves issues with possibly broken tests.

andreas-kupries:
The -cli failure
> In [AfterEach] at: /home/sles/actions-runner/_work/epinio/epinio/acceptance/services_test.go:98
is for 
> /home/sles/actions-runner/_work/epinio/epinio/acceptance/services_test.go:103
The same test passes locally.

andreas-kupries:
-api failure

  - app stage is likely unrelated
  - service show I have to look into - something in the setup
  - service delete - @enrichman I hope you can look at that one. It should fail locally also. And I am unclear why. There is something in that test I am not understanding.

andreas-kupries:
Ah. Service Show is because I added `wait for helm release` to `CreateHelmChart`. And in that test it has to fail.
Ok. Make waiting conditional, and disable for that one case.

andreas-kupries:
Yay. A single failing test.

>  STEP: creating a namespace: namespace-849391446 05/17/22 09:49:04.849
    STEP: targeting a namespace: namespace-849391446 05/17/22 09:49:23.86
    STEP: MSI service-428751036 -> service-instance-787772250 05/17/22 09:49:25.788
    STEP: MSI/ok 05/17/22 09:49:30.695
    STEP: locate helm chart x90de42de07a57973259f04cc2b9e 05/17/22 09:49:30.695
    STEP: assemble url 05/17/22 09:49:30.817
    STEP: assemble request for https://epinio.192.168.96.4.omg.howdoi.website/api/v1/namespaces/namespace-849391446/services/service-instance-787772250 05/17/22 09:49:30.817
    STEP: curl request 05/17/22 09:49:30.817
    STEP: read response 05/17/22 09:49:33.078
    STEP: decode response {"errors":[{"status":404,"title":"release: not found - x90de42de07a57973259f04cc2b9e","details":""}]} 05/17/22 09:49:33.078
    STEP: check status 05/17/22 09:49:33.078
    STEP: deleting a namespace: namespace-849391446 05/17/22 09:49:33.627
  << End Captured GinkgoWriter Output

>  {"errors":[{"status":404,"title":"release: not found - x90de42de07a57973259f04cc2b9e","details":""}]}
  Expected
      <int>: 404
  to equal
      <int>: 200
  In [It] at: /home/sles/actions-runner/_work/epinio/epinio/acceptance/api/v1/service_delete_test.go:157

It is the one where I do not understand why it fails.
A service (instance) is created.
When deleting that service the validation of the request finds the service, yet not the helm release for it.
Thus is reports the `404 release not found`.
Yet the setup should have created the release, even waited for it to be deployed, IIUC.

@enrichman ^ Please have a look.

enrichman (approved):

Issue #1455, opened 2022-05-16
Auth unit test non-deterministic failure


andreas-kupries:
**Describe the bug**

Intermittent failure of auth unit test for auth package. See with commit 27a98279141f48dfcd2f5f12c7107b2845bde2ba

> Auth users RemoveNamespaceFromUsers when users have the namespace [It] will be removed
> /home/work/SUSE/dev/epinio/internal/auth/auth_test.go:168  Expected
>       <string>: user3
>   to equal
>       <string>: user2
>   In [It] at: /home/work/SUSE/dev/epinio/internal/auth/auth_test.go:191

Rerun the tests and it likely passes. 

It looks as if the expected call order of `user2 user3` does not always happen.
It looks as if `user3 user2` is also possible.

I suspect that a go map is involved. Their iteration order is not deterministic.

**To Reproduce**

Go to the referenced commit. Compile epinio. Run the unit tests until the error is seen. multiple times might be required given the non-determinism.

**Expected behavior**

Always pass, do not fail.

andreas-kupries:
A first attempt at fixing this revealed that the bad case comes with a call count of 3, instead of 2. There is something deeper going on than just the unexpected order.

enrichman:
This should be fixed now that #1460 is merged.

Issue #1459, opened 2022-05-17
Services endpoint rename to CatalogServices


enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1449
---

The `services` endpoints needs to be renamed to `catalogservices`, so we can use the `services` to be used for the Service instances.

## Acceptance Criteria

- There is a `/api/v1/catalogservices` endpoint that returns all the catalog services 
- swagger is updated

Issue #1451, opened 2022-05-16
Implement service instance list-all endpoint


jimmykarily:
Depends on: https://github.com/epinio/epinio/issues/1449

## Acceptance Criteria

- There is a `/api/v1/services` endpoint that returns all service instances in all namespaces **the user has access to** (no namespace defined)
- swagger is updated

PR #1460, opened 2022-05-17
ListAll Services endpoint and CatalogService rename

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1451
- https://github.com/epinio/epinio/issues/1459
---

- [x] implementation
- [x] tests
- [x] swagger
- [x] documentation https://github.com/epinio/docs/pull/116

This PR renames the `/services` endpoint to `/catalogservices` for the CatalogServices, and adds the `/services` endpoints to list all the service instances.
When the user use the `--all` flag the instances are filtered for the namespaces where the user has the authorization.

I've also added an helper that can be used with the MatchRegexp in the tests, to check in a cleaner and more resilient way the table row from the output of the CLI:

```go
Expect(out).To(MatchRegexp(regex.TableRow(namespace1, service1, "mysql-dev", "not-ready")))
```

```go
func TableRow(args ...string) string {
	reg := `\|`
	for _, arg := range args {
		reg = fmt.Sprintf(`%s[\s]+%s[\s]+\|`, reg, arg)
	}
	return reg
}
```

Fixes also https://github.com/epinio/epinio/issues/1455

### DIFF
--- a/acceptance/helpers/regex/regex.go
+++ b/acceptance/helpers/regex/regex.go
@@ -1,0 +1,13 @@
+package regex
+
+import "fmt"
+
+// TableLine returns a regular expression that will match a line of a CLI table
+// The arguments passed should contain the expected text of the cell (or regex, i.e. you can pass '.*' to match anything)
+func TableLine(args ...string) string {
+	reg := `\|`
+	for _, arg := range args {
+		reg = fmt.Sprintf(`%s[\s]+%s[\s]+\|`, reg, arg)
+	}
+	return reg
+}
--- a/acceptance/helpers/regex/regex.go
+++ b/acceptance/helpers/regex/regex.go
@@ -2,9 +2,9 @@ package regex
 
 import "fmt"
 
-// TableLine returns a regular expression that will match a line of a CLI table
+// TableRow returns a regular expression that will match a line of a CLI table
 // The arguments passed should contain the expected text of the cell (or regex, i.e. you can pass '.*' to match anything)
-func TableLine(args ...string) string {
+func TableRow(args ...string) string {
 	reg := `\|`
 	for _, arg := range args {
 		reg = fmt.Sprintf(`%s[\s]+%s[\s]+\|`, reg, arg)
--- a/acceptance/helpers/regex/regex.go
+++ b/acceptance/helpers/regex/regex.go
@@ -1,6 +1,13 @@
 package regex
 
-import "fmt"
+import (
+	"fmt"
+)
+
+const (
+	// DateRegex will check for a date in the '2022-05-19 13:49:20 +0000' UTC format
+	DateRegex = "[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} [+][0-9]{4} [A-Z]{3,4}"
+)
 
 // TableRow returns a regular expression that will match a line of a CLI table
 // The arguments passed should contain the expected text of the cell (or regex, i.e. you can pass '.*' to match anything)
--- a/acceptance/helpers/regex/regex.go
+++ b/acceptance/helpers/regex/regex.go
@@ -2,6 +2,7 @@ package regex
 
 import (
 	"fmt"
+	"strings"
 )
 
 const (
@@ -12,9 +13,15 @@ const (
 // TableRow returns a regular expression that will match a line of a CLI table
 // The arguments passed should contain the expected text of the cell (or regex, i.e. you can pass '.*' to match anything)
 func TableRow(args ...string) string {
-	reg := `\|`
-	for _, arg := range args {
-		reg = fmt.Sprintf(`%s[\s]+%s[\s]+\|`, reg, arg)
+	if len(args) == 0 {
+		return ""
 	}
-	return reg
+
+	var b strings.Builder
+	for _, arg := range args {
+		fmt.Fprintf(&b, `[|][\s]+%s[\s]+`, arg)
+	}
+	b.WriteString(`[|]`)
+
+	return b.String()
 }
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -216,6 +216,11 @@ var _ = FDescribe("Services", func() {
 				return out
 			}, "2m", "5s").Should(MatchRegexp(regex.TableLine(namespace1, service1, "mysql-dev", "deployed")))
 
+			Eventually(func() string {
+				out, _ := env.Epinio("", "service", "list", "--all")
+				return out
+			}, "2m", "5s").Should(MatchRegexp(regex.TableLine(namespace2, service2, "mysql-dev", "deployed")))
+
 			By(fmt.Sprintf("%s/%s up", namespace1, service1))
 		})
 
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -123,7 +123,7 @@ var _ = FDescribe("Services", func() {
 			Expect(out).To(MatchRegexp("Listing Services"))
 			Expect(out).To(MatchRegexp("Namespace: " + namespace))
 
-			Expect(out).To(MatchRegexp(regex.TableRow(service, "mysql-dev", "not-ready")))
+			Expect(out).To(MatchRegexp(regex.TableRow(service, regex.DateRegex, "mysql-dev", "not-ready")))
 
 			By("wait for deployment")
 			Eventually(func() string {
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -123,7 +123,7 @@ var _ = FDescribe("Services", func() {
 			Expect(out).To(MatchRegexp("Listing Services"))
 			Expect(out).To(MatchRegexp("Namespace: " + namespace))
 
-			Expect(out).To(MatchRegexp(regex.TableLine(service, "mysql-dev", "not-ready")))
+			Expect(out).To(MatchRegexp(regex.TableRow(service, "mysql-dev", "not-ready")))
 
 			By("wait for deployment")
 			Eventually(func() string {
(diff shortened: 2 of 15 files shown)


acceptance/helpers/regex/regex.go
```diff
@@ -0,0 +1,20 @@
...
+	// DateRegex will check for a date in the '2022-05-19 13:49:20 +0000' UTC format
+	DateRegex = "[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} [+][0-9]{4} [A-Z]{3,4}"
+)
+
+// TableRow returns a regular expression that will match a line of a CLI table
+// The arguments passed should contain the expected text of the cell (or regex, i.e. you can pass '.*' to match anything)
+func TableRow(args ...string) string {
+	reg := `\|`
+	for _, arg := range args {
+		reg = fmt.Sprintf(`%s[\s]+%s[\s]+\|`, reg, arg)
```
andreas-kupries, line 17: If an `arg` is the empty string, i.e. `""`, the resulting regexp will contain `[\s]+[\s]+\|`. I.e. it will expect two spaces. Not sure if that is guaranteed by the table output itself. Might be sensible to check for that case and then use `reg = fmt.Sprintf(`%s[\s]+\|`, reg)` instead. Although that could be written as `reg = reg + `[\s]+\|``.

andreas-kupries (changes requested):

acceptance/services_test.go
```diff
@@ -78,6 +81,187 @@ var _ = Describe("Services", func() {
...
+
+			// show all the services (we are admin, good to go)
+			By("show it")
+			out, err = env.Epinio("", "service", "list", "--all")
+			Expect(err).ToNot(HaveOccurred(), out)
+
+			Expect(out).To(MatchRegexp("Listing all Services"))
+
+			Expect(out).To(MatchRegexp(regex.TableRow(namespace1, service1, "mysql-dev", "not-ready")))
+			Expect(out).To(MatchRegexp(regex.TableRow(namespace2, service2, "mysql-dev", "not-ready")))
```
andreas-kupries, line 211: I expect these to fail due to the new creation timestamps shown by the list.
If they do not fail I suspect that the new code is missing the timestamps.

acceptance/services_test.go
```diff
@@ -78,6 +81,187 @@ var _ = Describe("Services", func() {
...
+				out, _ := env.Epinio("", "service", "list", "--all")
+				return out
+			}, "2m", "5s").Should(MatchRegexp(regex.TableRow(namespace1, service1, "mysql-dev", "deployed")))
+
+			Eventually(func() string {
+				out, _ := env.Epinio("", "service", "list", "--all")
+				return out
+			}, "2m", "5s").Should(MatchRegexp(regex.TableRow(namespace2, service2, "mysql-dev", "deployed")))
+
+			By(fmt.Sprintf("%s/%s up", namespace1, service1))
```
andreas-kupries, line 224: Should mention namespace2, service2 also ?

acceptance/services_test.go
```diff
@@ -78,6 +81,187 @@ var _ = Describe("Services", func() {
...
+
+			// show only owned namespaces (we are user2, only namespace2)
+			By("show it")
+			out, err = env.Epinio("", "service", "list", "--all", "--settings-file", tmpSettingsPath)
+			Expect(err).ToNot(HaveOccurred(), out)
+
+			Expect(out).To(MatchRegexp("Listing all Services"))
+
+			Expect(out).NotTo(MatchRegexp(regex.TableRow(namespace1, service1, "mysql-dev", "not-ready")))
+			Expect(out).To(MatchRegexp(regex.TableRow(namespace2, service2, "mysql-dev", "not-ready")))
```
andreas-kupries, line 253: Same here wrt timestamps.

acceptance/services_test.go
```diff
@@ -78,6 +81,187 @@ var _ = Describe("Services", func() {
...
+
+			By("wait for deployment")
+			Eventually(func() string {
+				out, _ := env.Epinio("", "service", "list", "--all", "--settings-file", tmpSettingsPath)
+				return out
+			}, "2m", "5s").Should(MatchRegexp(regex.TableRow(namespace2, service2, "mysql-dev", "deployed")))
+
+			By(fmt.Sprintf("%s/%s up", namespace1, service1))
```
andreas-kupries, line 261: Ditto namespace2, service2

internal/api/v1/router.go
```diff
@@ -131,13 +131,16 @@ var Routes = routes.NamedRoutes{
...
-	"ServiceCatalogShow": get("/services/:catalogservice", errorHandler(service.Controller{}.CatalogShow)),
-	"ServiceCreate":      post("/namespaces/:namespace/services", errorHandler(service.Controller{}.Create)),
-	"ServiceList":        get("/namespaces/:namespace/services", errorHandler(service.Controller{}.List)),
-	"ServiceShow":        get("/namespaces/:namespace/services/:service", errorHandler(service.Controller{}.Show)),
-	"ServiceDelete":      delete("/namespaces/:namespace/services/:service", errorHandler(service.Controller{}.Delete)),
+	"AllServices":   get("/services", errorHandler(service.Controller{}.FullIndex)),
+	"ServiceCreate": post("/namespaces/:namespace/services", errorHandler(service.Controller{}.Create)),
+	"ServiceList":   get("/namespaces/:namespace/services", errorHandler(service.Controller{}.List)),
+	"ServiceShow":   get("/namespaces/:namespace/services/:service", errorHandler(service.Controller{}.Show)),
+	"ServiceDelete": delete("/namespaces/:namespace/services/:service", errorHandler(service.Controller{}.Delete)),
```
andreas-kupries, line 143: General side note, need a ticket that the service show/delete commands do not have auto-completion, and for this `Match` endpoints are needed here, similar to what we have for other resources.

internal/auth/auth_test.go
```diff
@@ -19,11 +19,12 @@ import (
...
 	var authService *auth.AuthService
-	var fake = &authfakes.FakeSecretInterface{}
-
-	rand.Seed(time.Now().UnixNano())
+	var fake *authfakes.FakeSecretInterface
 
 	BeforeEach(func() {
+		rand.Seed(time.Now().UnixNano())
+
+		fake = &authfakes.FakeSecretInterface{}
```
andreas-kupries, line 27: This fixes #1455 ? If yes, should be mentioned as such in the PR description to ensure auto-close of ticket.

internal/cli/usercmd/service.go
```diff
@@ -251,3 +253,30 @@ func (c *EpinioClient) ServiceList() error {
...
+	if err != nil {
+		return errors.Wrap(err, "service list failed")
+	}
+
+	if len(resp.Services) == 0 {
+		c.ui.Normal().Msg("No services found")
+		return nil
+	}
+
+	msg := c.ui.Success().WithTable("Namespace", "Name", "Catalog Service", "Status")
```
andreas-kupries, line 275: As suspected, this is missing the creation timestamps added to the regular list command.

andreas-kupries (changes requested):
Some parts not updated towards the new creation timestamps.

acceptance/services_test.go
andreas-kupries: Confirmed that the new `--all` code in the cli is missing printing of the timestamps.

acceptance/services_test.go
andreas-kupries: Confirmed that the new `--all` code in the cli is missing printing of the timestamps.

acceptance/helpers/regex/regex.go
enrichman: I've tried with a small test and it seems that spaces are always added:

```go
func TestTable(t *testing.T) {
	msg := termui.NewUI().Normal()
	msg.WithTable("", "", "", "", "").WithTableRow("", "").KeepLine().Msg("")
}
```
```
=== RUN   TestTable
|  |  |  |  |  |
|--|--|--|--|--|
|  |  |
--- PASS: TestTable (0.00s)
```

but I've seen that an empty row will not be printed at all

```go
func TestTable(t *testing.T) {
	msg := termui.NewUI().Normal()
	msg.WithTable("", "", "", "", "").WithTableRow().KeepLine().Msg("")
}
```
```
=== RUN   TestTable
|  |  |  |  |  |
|--|--|--|--|--|
--- PASS: TestTable (0.00s)
PASS
```

So at the moment I rewrote it like this (with a guard clause and the `strings.Builder`):

```go
func TableRow(args ...string) string {
	if len(args) == 0 {
		return ""
	}

	var b strings.Builder
	for _, arg := range args {
		fmt.Fprintf(&b, `[|][\s]+%s[\s]+`, arg)
	}
	b.WriteString(`[|]`)

	return b.String()
}
```

What do you think?

acceptance/services_test.go
enrichman: That's right! Fixed

acceptance/services_test.go
enrichman: Good catch!

internal/auth/auth_test.go
enrichman: Edited PR description! :+1:

internal/api/v1/router.go
enrichman: Done! https://github.com/epinio/epinio/issues/1476 :slightly_smiling_face:

internal/cli/usercmd/service.go
```diff
@@ -244,10 +244,12 @@ func (c *EpinioClient) ServiceList() error {
 
 	msg := c.ui.Success().WithTable("Name", "Created", "Catalog Service", "Status")
 	for _, service := range resp.Services {
-		msg = msg.WithTableRow(service.Meta.Name,
-			fmt.Sprintf("%v", service.Meta.CreatedAt),
+		msg = msg.WithTableRow(
+			service.Meta.Name,
+			service.Meta.CreatedAt.String(),
```
andreas-kupries, line 249: Oh! I will have to make a PR for all the other places showing stamps. I used `fmt.Sprintf("%v", ...)`. This is much simpler.

acceptance/helpers/regex/regex.go
enrichman: Ah, now I get what you're saying. Well, it should be true, but it seems to work anyway :thinking: 

https://regex101.com/r/pVSouv/1

enrichman: If you group them you can see that they are going to "share" the spaces.

https://regex101.com/r/ik9bRP/1

andreas-kupries: Testing with ginkgo shows that it works for the go regex engine as well. No bug. Ok.

andreas-kupries (approved):

acceptance/helpers/regex/regex.go
enrichman:
```go
Describe("", func() {
	It("Regex", func() {
		Expect("| yee |  |  |").To(MatchRegexp(regex.TableRow("yee", "", "")))
		Expect("| foo |").To(MatchRegexp(regex.TableRow("foo")))
		Expect("|  |").To(MatchRegexp(regex.TableRow("")))
	})
})
```

PR #1477, opened 2022-05-20
Ignore trailing slash in Origin header

jimmykarily:
like we do in allowed origins

I encountered this problem while debugging the logs in the UI. The origin header had a trailing `/` and the code wasn't matching the allowed header.


mmartin24 (approved):
Description of change make sense.
Validations pass.

andreas-kupries (approved):

PR #1478, opened 2022-05-20
Track and show bound services

### LINKED ISSUE #1450: "service" list/show endpoints should return bound applications too
To avoid too many requests against the kube api, consider caching the information on the helmchart CRD (annotation or label?)
We need to keep this updated, e.g. every time an application is bound/unbound/deleted/etc

Investigate if labels or annotations are betters, one of them may trigger controllers and such.

andreas-kupries:
Fix #1450 

Replicate machinery tracking bound configurations to track bound services.

:construction: TODO

- [ ] tracking secret and functions for manipulation and access
- [ ] old tests
- [ ] new tests?
- [ ] swagger (extended service information structure)
- [ ] extended `service list`
- [ ] extended `service show`
- [ ] extended `app show`
- [ ] extended `app list` ? (This might be too much, given that bound configs are shown)


richard-cox:
@andreas-kupries I've had a scroll through and it looks like the remaining items from the referenced issues are resolved. Can't comment on the actual code changes though

internal/application/services.go
```diff
@@ -0,0 +1,228 @@
...
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/util/retry"
+)
+
+// ServicesBoundApps is an extension of ServicesBoundAppsNames after it, to retrieve a map from
+// services to the full data of the applications bound to them. It uses ServicesBoundAppsNames
+// internally to quickly determine the applications to fetch.
+
```
jimmykarily, line 19:
```suggestion
```

internal/application/services.go
```diff
@@ -0,0 +1,228 @@
...
+	fetched := map[string]*models.App{}
+
+	for key, appNames := range bindings {
+		serviceName, namespace := DecodeServiceKey(key)
+		for _, appName := range appNames {
+			app, ok := fetched[appName]
+			if !ok {
+				meta := models.NewAppRef(appName, namespace)
+				app := meta.App()
+				err := fetch(ctx, cluster, app)
```
jimmykarily, line 39: At some point we should find a way to avoid these N+1 queries. Maybe by implementing a version of `fetch` that fetches multiple apps at once, with as few requests as possible. These endpoints are not very uncommon so we may have multiple users running these commands are the same time, not to mention the Web UI polling which will generate quite some traffic even if the user takes no action. Not to be solved here, but let's start thinking about it.

internal/application/services.go
```diff
@@ -0,0 +1,228 @@
...
+			result[serviceName] = append(result[serviceName], *app)
+		}
+	}
+
+	return result, nil
+}
+
+// ServicesBoundAppsNamesFor is a specialization of ServicesBoundAppsNames after it, to retrieve the
+// names of the applications bound to a single service, specified by name.
+
```
jimmykarily, line 57:
```suggestion
```

internal/application/services.go
```diff
@@ -0,0 +1,228 @@
...
+	return result, nil
+}
+
+// ServicesBoundAppsNames returns a map from the names of services in the specified namespace, to
+// the names of the applications they are bound to. The keys of the map are always a combination of
+// namespace name and service name, to distinguish same-named services in different namespaces (See
+// `ServiceKey` below). The application names never contain namespace information, as they are
+// always in the same namespace as the service referencing them.
+
```
jimmykarily, line 95:
```suggestion
```

internal/application/services.go
andreas-kupries: List, via label selectors, on each of the application-associated secrets, maybe. Similar to what we do for the tracking of services/configurations vs app.

andreas-kupries: I will make a ticket.

jimmykarily:
This PR looks ok but there is a lack of new tests. There is new functionality which needs to be tested. Maybe we can modify some existing tests to check if the additional information is there (apps using services, bounds apps etc).

internal/application/services.go
```diff
@@ -0,0 +1,228 @@
...
+// DecodeServiceKey splits the given key back into name and namespace.  The name is the first
+// result, the namespace the second.
+func DecodeServiceKey(key string) (string, string) {
+	parts := strings.Split(key, "\0000")
+	return parts[0], parts[1]
+}
+
+// BoundServiceNameSet returns the service names for the services bound to the application by a
+// user, as a map/set
+func BoundServiceNameSet(ctx context.Context, cluster *kubernetes.Cluster, appRef models.AppRef) (NameSet, error) {
```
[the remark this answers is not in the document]
andreas-kupries: Huh. This function is actually not used. Because the process of binding is sufficiently different from Configurations, where the original code is used. Will remove this thing.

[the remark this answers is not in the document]
andreas-kupries: Done, commit a03d0ca6272ba0604f9c242ce72162bb6efc7a52

Issue #1452, opened 2022-05-16
List-all endpoints (app/configuration) should filter by accessible namespaces


jimmykarily:
Web UI uses the "AllApps" and "AllConfigurations" endpoints without filtering namespaces the user doesn't have access too. We need to fix that

## Acceptance Criteria

- The user cannot see app or configurations that belong to namespace that are not assigned to them

PR #1481, opened 2022-05-24
Filter non accessible resources

enrichman:
Ref:
- #1452 
---

This PR moves the filterServices func into the `auth` package, and makes use of Generics in order to be used by other resources. The `FilterResources` needs just a struct that implements the `NamespacedResource` interface, to get the `Namespace`:

```go
type NamespacedResource interface {
	Namespace() string
}

func FilterResources[T NamespacedResource](user User, resources []T) []T {
	if user.Role == "admin" {
		return resources
	}

	namespacesMap := make(map[string]struct{})
	for _, ns := range user.Namespaces {
		namespacesMap[ns] = struct{}{}
	}

	filteredResources := []T{}
	for _, resource := range resources {
		if _, allowed := namespacesMap[resource.Namespace()]; allowed {
			filteredResources = append(filteredResources, resource)
		}
	}

	return filteredResources
}
```

### DIFF
--- a/acceptance/api/v1/application_index_all_test.go
+++ b/acceptance/api/v1/application_index_all_test.go
@@ -84,7 +84,7 @@ var _ = Describe("AllApps Endpoints", func() {
 			[]string{app2, namespace2}))
 	})
 
-	It("will not lists the applications belonging to other namespaces", func() {
+	It("doesn't list applications belonging to non-accessible namespaces", func() {
 		endpoint := fmt.Sprintf("%s%s/applications", serverURL, v1.Root)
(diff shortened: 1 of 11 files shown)


jimmykarily (commented):
Shouldn't there be a test for the service instance list filtering too?

enrichman:
@jimmykarily fixed comments. :slightly_smiling_face: 
Actually the tests for the services were already implemented in the services PR https://github.com/epinio/epinio/pull/1460 
They are in the `list` file though, not in their own.

jimmykarily (approved):

PR #1489, opened 2022-05-25
Allow streaming of staging logs after staging is done

jimmykarily:
Fixes https://github.com/epinio/ui/issues/93

- Re-use existing method for Logs
- Don't filter out containers that are not in "running" state when
  streaming logs (e.g. finished staging containers)
- Don't disable "follow" when staging logs are requested

### DIFF
--- a/internal/cli/usercmd/app.go
+++ b/internal/cli/usercmd/app.go
@@ -626,9 +626,5 @@ func (c *EpinioClient) AppRestage(appName string) error {
 	log.V(1).Info("wait for job", "StageID", stageID)
 	// blocking function that wait until the staging is done
 	_, err = c.API.StagingComplete(app.Meta.Namespace, stageID)
-	if err != nil {
-		return errors.Wrap(err, "waiting for staging failed")
-	}
-
-	return nil
+	return errors.Wrap(err, "waiting for staging failed")
 }
--- a/internal/cli/usercmd/app.go
+++ b/internal/cli/usercmd/app.go
@@ -618,9 +618,17 @@ func (c *EpinioClient) AppRestage(appName string) error {
 	}
 
 	log.V(3).Info("stage response", "response", stageResponse)
-
 	stageID := stageResponse.Stage.ID
-	log.V(1).Info("start tailing logs", "StageID", stageID)
 
-	return c.stageLogs(log.V(1), app.Meta, stageID)
+	log.V(1).Info("start tailing logs", "StageID", stageID)
+	c.stageLogs(app.Meta, stageID)
+
+	log.V(1).Info("wait for job", "StageID", stageID)
+	// blocking function that wait until the staging is done
+	_, err = c.API.StagingComplete(app.Meta.Namespace, stageID)
+	if err != nil {
+		return errors.Wrap(err, "waiting for staging failed")
+	}
+
+	return nil
 }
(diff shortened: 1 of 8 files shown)


jimmykarily:
`ginkgo -race internal/cli/usercmd` complains about data race. The reason is this:

```
// SetVerbosity sets the global level against which all info logs will be
// compared.  If this is greater than or equal to the "V" of the logger, the
// message will be logged.  A higher value here means more logs will be written.
// The previous verbosity value is returned.  This is not concurrent-safe -
// callers must be sure to call it from only one goroutine.
func SetVerbosity(v int) int {
```

internal/cli/usercmd/app.go
```diff
@@ -618,9 +618,17 @@ func (c *EpinioClient) AppRestage(appName string) error {
...
+
 	log.V(1).Info("start tailing logs", "StageID", stageID)
+	c.stageLogs(app.Meta, stageID)
+
+	log.V(1).Info("wait for job", "StageID", stageID)
+	// blocking function that wait until the staging is done
+	_, err = c.API.StagingComplete(app.Meta.Namespace, stageID)
+	if err != nil {
+		return errors.Wrap(err, "waiting for staging failed")
+	}
```
enrichman, line 631: You can just return the error with Wrap (if `nil` it will return `nil`).

internal/cli/usercmd/push.go
```diff
@@ -202,10 +200,17 @@ func (c *EpinioClient) Push(ctx context.Context, params PushParams) error { // n
 		log.V(3).Info("stage response", "response", stageResponse)
 
 		details.Info("start tailing logs", "StageID", stageResponse.Stage.ID)
-		err = c.stageLogs(details, appRef, stageResponse.Stage.ID)
+		c.stageLogs(appRef, stageResponse.Stage.ID)
 		if err != nil {
 			return err
```
enrichman, line 205: Maybe we should Wrap also this `err`.

internal/cli/usercmd/push.go
```diff
@@ -202,10 +200,17 @@ func (c *EpinioClient) Push(ctx context.Context, params PushParams) error { // n
...
 		}
+
+		details.Info("wait for job", "StageID", stageID)
+		// blocking function that wait until the staging is done
+		_, err := c.API.StagingComplete(appRef.Namespace, stageID)
+		if err != nil {
+			return errors.Wrap(err, "waiting for staging failed")
+		}
```
enrichman, line 213: Same as before

enrichman (commented):
Some comments

internal/cli/usercmd/app.go
jimmykarily: done

internal/cli/usercmd/push.go
jimmykarily: this `err` shouldn't exist. The `stageLogs` method no longer returns an error. It's running in a separate go routine and simply prints the error if it occurs. I'm removing this `if`.

internal/cli/usercmd/push.go
jimmykarily: here I don't want to return. There is more code to be run below.

enrichman: lol sorry, I missed that!

enrichman (approved):

PR #1490, opened 2022-05-25
Fail fast targeting unauthorized namespaces

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1375
---

Last point for the authorization layer.

This PR adds the check for the existence of a namespace, to fail-fast with a 404 in case of a missing one, and a 401 in case of an unauthorized.
As for the comments in the issue we had the authorization middleware before the business logic that was checking the validity of the namespace, and this check was repeated among all the controllers and methods that needed to do this validation.

Adding a `NamespaceMiddleware` before the auth layer simplified and moved all the logic just before entering the controller.

- [x] implementation
- [x] tests
- [x] ~~cli documentation~~
- [x] ~~documentation~~
- [x] ~~swagger documentation~~

### DIFF
--- a/acceptance/api/v1/configurations_mutation_test.go
+++ b/acceptance/api/v1/configurations_mutation_test.go
@@ -240,8 +240,8 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 
 		It("returns a 'not found' when the namespace does not exist", func() {
 			endpoint := fmt.Sprintf(
-				"%s%s/namespaces/%s/configurations/%s",
-				serverURL, api.Root, namespace, configuration,
+				"%s%s/namespaces/idontexist/configurations/%s",
+				serverURL, api.Root, configuration,
 			)
 			response, err := env.Curl("DELETE", endpoint, strings.NewReader(`{ "unbind": false }`))
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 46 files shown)


jimmykarily:
In my opinion, the right order should be:

```
authentication -> namespace existence -> authorization
```

with this being also valid:

```
authentication -> authorization -> namespace existence
```

the question between the 2 is whether we want users to see 403 or 404 when they don't have permissions on a namespace. `404` would be more appropriate (as in "none of you business") if we didn't indirectly expose the existence of namespaces when the user tries to create an existing one (namespace names are globally unique). For that reason, I think a 403 is fine thus the second suggestion is ok too.

Regarding `401` (the `authentication` step), I think it should come first because we are not referring to the specific endpoint but rather to the API as a whole. The API is "private" and none of the endpoint is accessible to unauthenticated users. Thus a general `401` is fine when user doesn't have an account.

enrichman:
I agree with the first order, it was what I was actually trying to do but I've put the NamespaceMiddleware before the authentication, so let me swap that.

I will also add the `403` instead of the `401` for the authorization part, to fix the #1491 .

jimmykarily (dismissed):

jimmykarily (approved):

Issue #1361, opened 2022-04-18
Password is stored in plaintext


agracey:
**Describe the bug**
By default, the password is stored as plaintext in both the local config and the secret in k8s. We should switch to some other way of storing the authN. 

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Install Epinio
2. `epinio settings show`

**Expected behavior**
The password should not be stored as plaintext.

jimmykarily:
We can do base64. The secret is already base64 encoded.

jimmykarily:
The command `epinio settings show` exists to allow the user to see settings, including the password (which the operator probably set up for them). Should we hide the password by default and only return it when some specific flag is set?

The other option is to let operator and developers handle the transferring of passwords outside Epinio and change Epinio to only check against password hashes (e.g. MD5). This means, the password is not recoverable. If the user user forgets the password, someone will have to change the user's secret and put a new password there.

agracey:
Yeah, in general passwords should not be recoverable. I'm not sure about having it be in the settings.yaml on the user's system but I can ask the security team for recommendations. 

And yes, until we deal with upstream authentication, a user losing their password would require someone with the right k8s permissions to reset it.

enrichman:
I don't see any other options, every cli that needs some kind of authentication needs to store the passwords somewhere (i.e. the `aws` in the `.aws/credentials` or `kubectl` with the `.kube/config`).

If we want just to prevent [shoulder surfing](https://en.wikipedia.org/wiki/Shoulder_surfing_(computer_security)) we can hide the password from the `settings show` command, and maybe just show it explicitly with a flag.

agracey:
The other option is to use something like X.509 or another id mechanism. 

Since we are hoping to use upstream auth eventually, I don't want to go too far down the rabbit hole, but it would be good to at least salt/hash the secrets and remove from the `settings show` command.

agracey:
_Adding comment to correct issue now..._

The more important part of this is what is stored in kubernetes.

The password field there should be stored salted and hashed in a non-reversable way. Then when an auth request comes, you compare using the verify function instead of equality.

For the human side of this, the admin would hash a password, store the output while giving the input to the developer. Or the developer would give the hashed password to the admin to store.

This weirdness is why we need to eventually work with an upstream auth provider (such as Dex) and not handle passwords ourselves. That and people are going to ask how to hook into their existing LDAP server.

jimmykarily:
[...]

PR #1499, opened 2022-05-30
Hide password from show command

enrichman:
Ref:
- #1361 
---

This PR will hide the password from the `epinio settings show` command.

It can be shown specifying the `--show-password` flag.

### DIFF
--- a/acceptance/settings_test.go
+++ b/acceptance/settings_test.go
@@ -74,6 +74,23 @@ var _ = Describe("Settings", func() {
 					WithRow("Current Namespace", ".*"),
 					WithRow("Certificates", "Present"),
 					WithRow("API User Name", env.EpinioUser),
+					WithRow("API Password", "[*]+"),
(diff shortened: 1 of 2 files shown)


andreas-kupries (dismissed):
Looks ok. Also looks to be in conflict with the new table matchers

enrichman:
@andreas-kupries that's correct, that's why I asked for the review of that one! Now that is merged I'll rebase it and resolve the conflicts. :+1:

jimmykarily:
Can we maybe hash the password in the `settings.yaml` file too? Just in case someone is editing the file while someone else is looking. Could happen on another PR.

andreas-kupries (changes requested):
Looks good. The change request is for checking and updating the epinio [docs](https://github.com/epinio/docs) around adding and removing users.

enrichman:
I'm going to update doc about users in a separate PR (lots of thing moving around auth and users these days), and I will address the password encoding in another one as well.

PR #1501, opened 2022-05-30
Use Bcrypt for authentication

enrichman:
Ref:
- #1361 
---

This PR will use Bcrypt to check for password validity, instead of plaintext.

It also removes the `epinio settings update-ca` (with some unused methods) and it adds the `epinio settings update-ca` command.

**Helm charts:**
- https://github.com/epinio/helm-charts/pull/208

**Docs:**
- https://github.com/epinio/docs/pull/128

### DIFF
--- a/scripts/prepare-environment-k3d.sh
+++ b/scripts/prepare-environment-k3d.sh
@@ -33,6 +33,27 @@ function create_docker_pull_secret {
 	fi
 }
 
+function retry {
+  retry=0
+  maxRetries=$1
+  retryInterval=$2
+  local result
+  until [ ${retry} -ge ${maxRetries} ]
+  do
+    echo -n "."
+    result=$(eval "$3") && break
+    retry=$[${retry}+1]
+    sleep ${retryInterval}
+  done
+
+  if [ ${retry} -ge ${maxRetries} ]; then
+    echo "Failed to run "$3" after ${maxRetries} attempts!"
+    exit 1
+  fi
+
+  echo " ✔️"
+}
+
 # Ensure we have a value for --system-domain
 prepare_system_domain
 # Check Dependencies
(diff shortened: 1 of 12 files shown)


jimmykarily:
`epinio settings update` is doing 2 things:

- fetch the credentials from the cluster and update the settings.yaml file
- update the server's certificate in the settings.yaml file so that it's always trusted.

By storing the password's hash on the kubernetes side, we can no longer retrieve that password. I suggest we do this instead:


- we get rid for the `epinio settings update` command
- users update the credentials on the settings file by calling `epinio login`
- when the epinio cli encounters an invalid certificate it offers the user to accept it either for this request or permanently. The permanent choice updates the settings file. Users already have a flag `--skip-ssl-verification` to avoid prompts in CI and such.

scripts/prepare-environment-k3d.sh
```diff
@@ -89,4 +87,6 @@ if [ ${retry} -ge ${maxRetries} ]; then
   exit 1
 fi
 
+"${EPINIO_BINARY}" info
```
jimmykarily, line 90: Strange but I got this:

```
✔️  Login successful

❌  error retrieving Epinio environment information: cannot parse JSON response: 'Bad Gateway': invalid character 'B' looking for beginning of value
make: *** [Makefile:189: prepare_environment_k3d] Error 255
```

How is it possible to successfully login if the server is not ready yet. Tryiing `epinio info` manually after the error, works.

jimmykarily: Maybe we can define a function like this:

```
function retry {
  retry=0
  maxRetries=$1
  retryInterval=$2
  local result
  until [ ${retry} -ge ${maxRetries} ]
  do
    result=$(eval "$3") && break
    retry=$[${retry}+1]
    sleep ${retryInterval}
  done

  if [ ${retry} -ge ${maxRetries} ]; then
    echo "Failed to run "$3" after ${maxRetries} attempts!"
    exit 1
  fi

  echo $result
}
```

which can then be called like:

```
retry 5 1 "${epinio} info"
```

(copied it from another script I had locally)

jimmykarily (approved):

scripts/prepare-environment-k3d.sh
enrichman: Done! :+1:

jimmykarily (approved):

PR #1521, opened 2022-06-02
Don't expect a not-ready service

jimmykarily:
because it may deploy fast enough to be ready on the first check

### DIFF
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -147,7 +147,7 @@ var _ = Describe("Services", func() {
 			Expect(out).To(
 				HaveATable(
 					WithHeaders("NAME", "CREATED", "CATALOG SERVICE", "STATUS", "APPLICATIONS"),
-					WithRow(service, WithDate(), catalogService.Meta.Name, "not-ready|deployed", ""),
+					WithRow(service, WithDate(), catalogService.Meta.Name, "(not-ready|deployed)", ""),
 				),
 			)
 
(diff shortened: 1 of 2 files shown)


acceptance/services_test.go
```diff
@@ -147,7 +147,7 @@ var _ = Describe("Services", func() {
 			Expect(out).To(
 				HaveATable(
 					WithHeaders("NAME", "CREATED", "CATALOG SERVICE", "STATUS", "APPLICATIONS"),
-					WithRow(service, WithDate(), catalogService.Meta.Name, "not-ready", ""),
+					WithRow(service, WithDate(), catalogService.Meta.Name, "not-ready|deployed", ""),
```
andreas-kupries, line 150: Please check the resulting regex. We may need parens around the alternation to get something sensible, i.e. `(not-ready|deployed)`, to separate it from the surrounding elements (Which check for a sequence of spaces, using `*`).

andreas-kupries (changes requested):
Actually requesting a check that the generated/internal regex is good.

acceptance/services_test.go
jimmykarily: Good point, fixed here: https://github.com/epinio/epinio/pull/1521/commits/9d104ede0469453cbd61a5b155e5efe01791995e

andreas-kupries: Thanks.

andreas-kupries (approved):

Issue #1380, opened 2022-04-21
Test Upgrades


agracey:
**Is your feature request related to a problem? Please describe.**

We should have an automated way to test a PR against the current stable release to see if there are any issues upgrading.

juadk:
https://confluence.suse.com/display/Epinio/epinio+upgrade

juadk:
Each night, a draft binary is created and pushed to sandbox registry. The draft binary contains the latest epinio (main branch):
https://github.com/epinio/epinio/actions/workflows/release-qa.yml (https://github.com/epinio/epinio/blob/main/.github/workflows/release-qa.yml). We use a specific [goreleaser](https://github.com/epinio/epinio/blob/main/.goreleaser-qa.yml) file to create it.
Now we got the binary, I can try to update from the latest release to it.

juadk:
I tried to test upgrade from` 0.9.0` to `1.0.0-rc1` but the UI is broken in 1.0.0-rc1 ( reported [here](https://github.com/epinio/ui/issues/138))

juadk:
### Upgrade process
Test triggered for every PR in the helm-chart repo
1. Build latest [epinio-server-qa](https://github.com/epinio/epinio/actions/workflows/release-qa.yml) and [epinio-ui-qa ](https://github.com/epinio/epinio-end-to-end-tests/actions/workflows/build_latest_epinioUI.yml)
2. Install latest release of epinio
3. Deploy an app
4. Upgrade epinio by

   - Downloading latest helm-chart
   - Updating the file to use the qa binary
   - Running the helm command to upgrade
5. Check if the app is still alive

PR #1523, opened 2022-06-02
Test - Add workflow to build draft binaries (QA)

juadk:
Step1 of #1380 
This is a test to try building draft binaries in order to test epinio upgrade.
I cannot test it without merging the PR.

### DIFF
--- a/.github/workflows/release-qa.yml
+++ b/.github/workflows/release-qa.yml
@@ -1,0 +1,51 @@
+name: Build draft binaries - QA
+
+on:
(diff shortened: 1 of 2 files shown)


mmartin24 (approved):
Looks good to me.
After [1 error](https://github.com/epinio/epinio/runs/6764333560?check_suite_focus=true#step:10:3346) on acceptance test, retried and passed:

Initial error:
image

Tests passing rightafter:
image

PR #1528, opened 2022-06-06
Namespace routing refactor

enrichman:
:construction:


enrichman:
This can be closed. Too old and other refactorings are taking over.

PR #1534, opened 2022-06-07
Filtered namespaces

### LINKED ISSUE #1437: Namespace#list endpoint should return only assigned namespaces for non-admin users
Currently, a non-admin user can list all namespaces with the command:

```
epinio namespace list
```
(and by using the relevant endpoint). The reason is that this route is not namespaced. There is no reason why the user should see namespaces that they cannot access. Maybe it's better if that endpoint only returns namespace which exist and are assigned to the user.

We need to consider what happens when the user tries to create a namespace but that namespace already exists and is not assigned to the user. That again reveals existing namespaces in a indirect way. Also it's confusing to the user because they had no way of knowing before-hand.

enrichman:
#### Ref
- #1437 
---

This PR adds the filtering for the not owned namespaces when listing all the namespaces.

Fix #1437

### DIFF
--- a/internal/namespaces/namespaces.go
+++ b/internal/namespaces/namespaces.go
@@ -20,6 +20,10 @@ type Namespace struct {
 	CreatedAt metav1.Time
 }
 
+func (n Namespace) Namespace() string {
+	return n.Name
+}
+
 func List(ctx context.Context, kubeClient *kubernetes.Cluster) ([]Namespace, error) {
 	listOptions := metav1.ListOptions{
 		LabelSelector: kubernetes.EpinioNamespaceLabelKey + "=" + kubernetes.EpinioNamespaceLabelValue,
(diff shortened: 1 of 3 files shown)


internal/namespaces/namespaces.go
```diff
@@ -20,6 +20,10 @@ type Namespace struct {
 	CreatedAt metav1.Time
 }
 
+func (n Namespace) Namespace() string {
```
andreas-kupries, line 23: I am unable to see where this is used.

andreas-kupries (changes requested):

internal/namespaces/namespaces.go
enrichman: It is used to implement the `NamespacedResource` interface, so that we can use the `auth.FilterResources` func. :slightly_smiling_face: 

https://github.com/epinio/epinio/blob/20f1495fc9d366657765b5083be0c34f96d8ee6b/internal/auth/auth.go#L152-L175

andreas-kupries: :+1: Retracting change request.

andreas-kupries (approved):

PR #1539, opened 2022-06-09
Bump helm-charts to change how default users are created

jimmykarily:
Fixes https://github.com/epinio/helm-charts/issues/214


andreas-kupries (commented):
The references helmchart change does not remove the default users. It only changes them from a fixed single default user to an array holding two default users. The operator installing epinio still has to override them, and may forget to do so.

jimmykarily:
> The references helmchart change does not remove the default users. It only changes them from a fixed single default user to an array holding two default users. The operator installing epinio still has to override them, and may forget to do so.

I can't decide what is best:

- Don't create any users by default. Possibly also add a new boolean flag to create the default users (and admin and a regular user).
- Create the users but allow the operator to override them (what the PR is doing currently)

The main problem in the original issue is that the operator can't skip the creation of the default `epinio` user even if they wanted to. The operator can only control what the default admin user credentials would be.

This PR allows the operator to:

- skip user creation altogether (and handle that whatever way they want)
- create default users with default credentials (e.g. for local development environments)
- create default users with "strong" credentials (for productions environments where no other sophisticated user management is needed)

In that regards the main problem is fixed. The question is about default behavior. Should we create users by default or not? I was hoping that simply documenting the need to override the default users in production environments may be enough. If it's not, then I think we should have a `--set api.createDefaultUsers=true` flag.

andreas-kupries (approved):

PR #1543, opened 2022-06-10
Removed unnecessary interface

enrichman:
Removed unnecessary interface used as a wrapper for `counterfeiter`.

It's possible to specify also external interfaces, so we can just use the kubernetes one.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -23,13 +23,10 @@ var (
 	ErrUserNotFound = errors.New("user not found")
 )
 
-//counterfeiter:generate . SecretInterface
-type SecretInterface interface {
-	typedcorev1.SecretInterface
-}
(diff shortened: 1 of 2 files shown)


andreas-kupries (changes requested):
Where does the chart change come from ? 
It does not seem to have anything to do with the description of the PR.

enrichman:
@andreas-kupries that's true, I took the opportunity to bump it to the latest `main` version, but I haven't add it to the description.

Should I add it, or just leave it in a separate PR?

andreas-kupries:
... I would honestly prefer a separate PR to not mix things, i.e. cleanup, and chart bump.
Despite the price in CI time.

enrichman:
Agree! I was unsure, but as you said it's better to have a clear view on what a PR does. :slightly_smiling_face: 
Reverted!

andreas-kupries (approved):

PR #1556, opened 2022-06-15
Simplify how we install tools in the CI

### LINKED ISSUE #1295: Revamp how we install tools for the project (scripts/tools)
As we discussed in the retro meeting, we would like something easier and more "dummy" for installing tools needed by the project.
Current implementation was done at the beginning of Epinio: https://github.com/epinio/epinio/tree/main/scripts/tools

jimmykarily:
Fixes #1295


juadk:
> @juadk I've removed all other "target" OSes, other than Linux. Are we running any pipelines on Mac or Windows that would rely on those tools being installed?

I don't think so, if we look at [EKS](https://github.com/epinio/epinio/blob/main/.github/workflows/eks.yml) for instance, we install the tools [here](https://github.com/epinio/epinio/blob/main/.github/workflows/eks.yml#L46) but it happens in the self-hosted runner (Linux). Then we use [macos](https://github.com/epinio/epinio/blob/main/.github/workflows/eks.yml#L82) but we only install go and ginkgo there, we don't install the tools so I think it shouldn't break the test.

I can trigger a public cloud test with your branch though.

juadk (approved):
LGTM, public cloud tests ok

PR #1558, opened 2022-06-15
Add crds to the release artifacts

### LINKED ISSUE #1553: Add CRDs to the release artifacts in epinio/epinio
Because every epinio binary version only works with specific versions on the CRDs. This makes the helm chart an "optional" installation method. One could just take the artifacts from an epinio/epinio release and deploy epinio whatever way they prefer (including the CRDs).

As agreed on a call, this is needed in order to stick to `epinio/epinio` as the official/main repository for epinio. The epinio's version is the epinio/epinio repository release version.

jimmykarily:
Fixes #1553

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -194,3 +194,6 @@ brews:
       output = shell_output("#{bin}/epinio settings update-ca 2>&1")
       assert_match "failed to get kube config", output
       assert_match "no configuration has been provided", output
+release:
+  extra_files:
+    - glob: ./helm-charts/chart/epinio/crds/*


.goreleaser.yml
```diff
@@ -194,3 +194,6 @@ brews:
       output = shell_output("#{bin}/epinio settings update-ca 2>&1")
       assert_match "failed to get kube config", output
       assert_match "no configuration has been provided", output
+release:
+  extra_files:
+    - glob: ./helm-charts/chart/epinio/crds/*
```
jimmykarily, line 199: I'm not sure if the git submodule is available here

jimmykarily: I guess this should do the trick: https://github.com/epinio/epinio/blob/151121bfb2aed5fb89a8a851d1c38f187a10e7ee/.github/workflows/release.yml#L18

enrichman: I'll approve (LGTM) but you can try it locally if you need to, maybe just checking out a different submodule version.

The `goreleaser --snapshot --rm-dist` command should do the trick.

enrichman (approved):

Issue #1551, opened 2022-06-14
EKS CI is failing with the new epinio login command


juadk:
All other Public Clouds work fine, only EKS fails and the main difference is that EKS is running in a Macos github runner.
https://github.com/epinio/epinio/runs/6881101827?check_suite_focus=true

```
epinio login -u admin -p password --trust-ca https://epinio.id32098-epi.epinio-ci.suse.dev
....
error while checking CA: error while checking CA: x509: "epinio.id7188-***" certificate is not trusted
```

I can reproduce it locally with my own Mac, I also tested from my linux laptop and it works well.

I installed an epinio instance in my linux server and I tried to do the epinio login command on my mac and I got the same error.

juadk:
Tested successfully on [EKS](https://github.com/epinio/epinio/runs/6915676971?check_suite_focus=true).

PR #1561, opened 2022-06-16
Fix for SSL Verify on MacOS

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1551
---

This PR fixes the bug that happens during the `Verify` and its error check with the MacOS architecture.
We actually don't need the type check, so we can probably skip it like this.
We could eventually do a check only for the `darwin/ios` arch, but it seems overkill to me.

A brief explanation.

The SSL verification in the `windows` and `darwin/ios` platforms is done with the system verifier:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/x509/verify.go;l=748-761
```go
// Use platform verifiers, where available, if Roots is from SystemCertPool.
if runtime.GOOS == "windows" || runtime.GOOS == "darwin" || runtime.GOOS == "ios" {
	if opts.Roots == nil {
		return c.systemVerify(&opts)
	}
	if opts.Roots != nil && opts.Roots.systemPool {
		platformChains, err := c.systemVerify(&opts)
		// If the platform verifier succeeded, or there are no additional
		// roots, return the platform verifier result. Otherwise, continue
		// with the Go verifier.
		if err == nil || opts.Roots.len() == 0 {
			return platformChains, err
		}
	}
}
```

In the Windows implementation the `UnknownAuthorityError` error is checked and returned:

https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/x509/root_windows.go;l=91-107

```go
// checkChainTrustStatus checks the trust status of the certificate chain, translating
// any errors it finds into Go errors in the process.
func checkChainTrustStatus(c *Certificate, chainCtx *syscall.CertChainContext) error {
	if chainCtx.TrustStatus.ErrorStatus != syscall.CERT_TRUST_NO_ERROR {
		status := chainCtx.TrustStatus.ErrorStatus
		switch status {
		case syscall.CERT_TRUST_IS_NOT_TIME_VALID:
			return CertificateInvalidError{c, Expired, ""}
		case syscall.CERT_TRUST_IS_NOT_VALID_FOR_USAGE:
			return CertificateInvalidError{c, IncompatibleUsage, ""}
		// TODO(filippo): surface more error statuses.
		default:
			return UnknownAuthorityError{c, nil, nil}
		}
	}
	return nil
}
```

but in the Darwin arch it's probably not possible and a generic `errorString` is returned:
https://cs.opensource.google/go/go/+/refs/tags/go1.18.3:src/crypto/x509/root_darwin.go

### DIFF
--- a/internal/cli/usercmd/login.go
+++ b/internal/cli/usercmd/login.go
@@ -119,9 +119,8 @@ func checkAndAskCA(ui *termui.UI, address string, trustCA bool) (string, error) 
 
 	cert, err := checkCA(address)
 	if err != nil {
-		_, isUnknownAuthority := err.(x509.UnknownAuthorityError)
 		// something bad happened while checking the certificates
-		if !isUnknownAuthority {
+		if cert == nil {
 			return "", errors.Wrap(err, "error while checking CA")
 		}
 


juadk (approved):
LGTM
I triggered [EKS test](https://github.com/epinio/epinio/runs/6915676971?check_suite_focus=true) from your branch and :champagne: 
```
STEP: Connecting to Epinio 06/16/22 09:52:31.135
STEP: Checking Epinio info command 06/16/22 09:52:32.963
STEP: Creating the application 06/16/22 09:52:33.607
```

gitricko:
Does this work in macosx ? I tested 0.90 and linux works... but it did not work in osx.

gitricko:
I think 1.0.0-rc1 is not in brew tap yet

jimmykarily:
> I think 1.0.0-rc1 is not in brew tap yet

https://github.com/epinio/epinio/blob/02d3e28408eec7be805c1f9c10d434e037e43d6e/.goreleaser.yml#L185

PR #1566, opened 2022-06-16
Recreate the helm release when pushing

### LINKED ISSUE #1393: Can't restage a failed app
I tried to stage [an app](https://github.com/gofiber/recipes/tree/master/heroku). I created this Procfile:

```
web: heroku
```

and the app fails to start with this error:

```
bash: heroku: command not found
```

I tried to fix the application by changing the Procfile but when I push again I get:

```
❌  error pushing app to server: can't deploy app: Internal Server Error: another operation (install/upgrade/rollback) is in progress
```

This is Epinio failing to call `helm install` . The only way to fix the problem is delete the app and push again. It should be possible to re-stage a failed app without deleting it.

jimmykarily:
because if helm status is `pending-install` (e.g. because the Pod is
failing with an error), then the user can no longer push the app unless
they delete it first.

There are also the "Force" and the "Replace" attributes that could possibly be
used.

Fixes #1393

### DIFF
--- a/internal/cli/usercmd/push.go
+++ b/internal/cli/usercmd/push.go
@@ -226,6 +226,9 @@ func (c *EpinioClient) Push(ctx context.Context, params PushParams) error { // n
 		deployRequest.Stage = models.StageRef{ID: stageID}
 	}
 
+	s := c.ui.Progress("Waiting for deployment")
+	defer s.Stop()
+
 	deployResponse, err := c.API.AppDeploy(deployRequest)
 	if err != nil {
 		return err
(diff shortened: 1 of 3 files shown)


andreas-kupries:
Docs: https://pkg.go.dev/github.com/mittwald/go-helm-client#ChartSpec.Recreate

andreas-kupries:
The force and replace flags are for special situations IIRC and generally use is not recommended.
At the moment unable to refind the docs I am basing this on.
Maybe I left links in the relevant issue and PR where we introduced helm for apps.

enrichman (approved):

PR #1583, opened 2022-06-21
fix: path display across platforms.

### LINKED ISSUE #1578: epinio CLI on Windows doesn't display path names correctly
Example:

```
C:\Users\Jan>epinio settings update-ca

🚢  Updating CA in the stored credentials from the current cluster
Settings: /.config/epinio/settings.yaml

✔️  Ok
```

The settings path is `C:\.config\epinio\settings.yaml`. It should use backslashes and include the drive letter.

andreas-kupries:
fix #1578 

Use a filepath function to ensure that the path is absolute.
And hopefully transform into the platform-specific form as well.

Should check where else we display paths for the user to see.
(`push` comes to mind, location of the app sources)

### DIFF
--- a/helpers/string.go
+++ b/helpers/string.go
@@ -1,5 +1,18 @@
 package helpers
 
+import (
+	"fmt"
+	"path/filepath"
+)
+
+func AbsPath(path string) string {
+	absLocation, err := filepath.Abs(path)
+	if err != nil {
+		return fmt.Sprintf("<<error: %s>>", err.Error())
+	}
+	return absLocation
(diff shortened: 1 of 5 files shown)


jimmykarily:
Output looks good:

```
PS C:\Users\Dimitris\epinio> epinio settings update-ca

🚢  Updating CA in the stored credentials from the current cluster
Settings: C:\.config\epinio\settings.yaml

✔️  Ok
```

jimmykarily (approved):

PR #1586, opened 2022-06-21
Use the `adrg/xdg` library to find OS specific path for config

### LINKED ISSUE #1579: Settings file on Windows should not be created at the drive root
The settings are stored at `/.config/epinio/settings.yaml`, which is not an absolute path (it is missing the drive letter), and is outside the users profile directory.

It should be stored under `%APPDATA%` (the location of the roaming profile, that travels with the user to different machines in the same domain), or under `%LOCALAPPDATA%` if the settings are only valid on the local machine (or contain large files).

jimmykarily:
Because on Windows we used 'C:\.config\epinio\settings.yaml' which
is not the place where apps write their configuration on Windows.

Probably did the wrong thing on Mac as well.

Sibling PR: https://github.com/epinio/helm-charts/pull/240

Fixes #1579

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,6 +3,7 @@ module github.com/epinio/epinio
 go 1.18
 
 require (
+	github.com/adrg/xdg v0.4.0
 	github.com/alron/ginlogr v0.0.4
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/briandowns/spinner v1.18.1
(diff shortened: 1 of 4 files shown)


jimmykarily:
On Windows 10:

```
PS C:\Users\User> epinio settings show

🚢  Show Settings
Settings: C:\Users\User\AppData\Local\epinio\settings.yaml

...
```

mmartin24:
On Mac:

```
➜  Downloads ./epinio settings show

🚢  Show Settings
Settings: /Users/mamartin/Library/Application Support/epinio/settings.yaml

✔️  Ok
|        KEY        |    VALUE     |
|-------------------|--------------|
| Colorized Output  | true         |
| Current Namespace | workspace    |
| Default App Chart |              |
| API User Name     |              |
| API Password      |              |
| API Url           |              |
| WSS Url           |              |
| Certificates      | None         |
|                   | defined      |
```
image

jimmykarily:
The prepare scripts fail now because the server pod can't start:


```
panic: could not create any of the following paths: /.config/epinio, /etc/xdg/epinio                                                                                                                             
```

enrichman (approved):

Issue #1360, opened 2022-04-16
Improve OpenAPI/swagger specification


enrichman:
I was trying writing an Epinio API client just to try out the new Go 1.18 generics, and I was having a look at some tools to mock/fake the API, to avoid writing each test case manually and have some random data to test. I had a look at Prism and it seems pretty cool, but looking at the responses and even the routes we are probably missing some specification in the Swagger file. For example there is no `/API/v1` in front of the routes, and we are missing some example. Calling the namespaces route is not returning anything but I should have a deeper look on it, just test it quickly. :)

enrichman:
With the https://github.com/epinio/epinio/pull/1594 I've added some improvements, but as written in the description we have some issues to fis:
- the catalog service response has some strange output with helm templates?
- some catalog service apis are under the service section
- missing errors declaration/specification (or more broadly every non-200 responses)
- some output spec are wrong (i.e. the created time seems to be an object but it's a UTC string

PR #1594, opened 2022-06-23
[Swagger] Fixed host and missing path, and new UI

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1360
---

This PR fixes the missing path `/api/v1` and adds some metadata to our swagger.json file (auth, version and so on).

It also adds an handler to the epinio server on the `/api/swagger.json` path that will serve the specification.
This will be used by the new UI (https://github.com/swagger-api/swagger-ui) during the `make swagger-serve` and adds the possibility to "try out" the endpoints.

image

image

## Authentication

It's now possible to authenticate with the Basic Auth (still missing the token auth):

image

In the following screenshot an example of the output

image

From this I've seen that we have some "mismatch" or strange things to fix.
The output of the values is the helm template, we are missing the errors declaration, and some catalog services APIs are under the "Services" section.

### DIFF
--- a/docs/howtos/swagger.md
+++ b/docs/howtos/swagger.md
@@ -1,0 +1,35 @@
+# Swagger
+
+A Swagger endpoint is available on `/api/swagger.json`. This will expose the OpenAPI spec for the Epinio API.
+
+## Generate the `swagger.json` file
+
+To generate the `swagger.json` file you need `github.com/go-swagger/go-swagger`.
+
+You can just run the `make swagger` command and it will install the tool if needed, and generate the `swagger.json` file.
+
+
+## Serve the documentation
+
+### Development mode
+
+To see the documentation of a local deployment you can run the `make swagger-serve` command.
+It will query your running Epinio for its swagger file and you will be able to try out the APis.
+
+### External Epinio instance
+
+If you want run the documentation on a running Epinio deployment you just need to specify the URL:
+
+```
+docker run -p 8080:8080 -e SWAGGER_JSON_URL=https://epinio.<EPINIO_SYSTEM_DOMAIN>/api/swagger.json swaggerapi/swagger-ui
+```
+but remember that since you are running the documentatio locally then you would need to enable CORS on your deployment.
+You can do it specifying the `server.accessControlAllowOrigin: "*"` value in the Helm values.
+
+### Locally
+
+If you just want to see the documentation without a running instance of Epinio you can do it with:
+
+```
+docker run -p 8080:8080 -e SWAGGER_JSON=/docs/swagger.json -v $(pwd)/docs/references/api:/docs swaggerapi/swagger-ui
+```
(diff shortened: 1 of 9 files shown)


docs/howtos/swagger.md
```diff
@@ -0,0 +1,35 @@
...
+It will query your running Epinio for its swagger file and you will be able to try out the APis.
+
+### External Epinio instance
+
+If you want run the documentation on a running Epinio deployment you just need to specify the URL:
+
+```
+docker run -p 8080:8080 -e SWAGGER_JSON_URL=https://epinio.<EPINIO_SYSTEM_DOMAIN>/api/swagger.json swaggerapi/swagger-ui
+```
+but remember that since you are running the documentatio locally then you would need to enable CORS on your deployment.
```
andreas-kupries, line 26:
```suggestion
but remember that since you are running the documentation locally then you would need to enable CORS on your deployment.
```

docs/howtos/swagger.md
```diff
@@ -0,0 +1,35 @@
...
+You can do it specifying the `server.accessControlAllowOrigin: "*"` value in the Helm values.
+
+### Locally
+
+If you just want to see the documentation without a running instance of Epinio you can do it with:
+
+```
+docker run -p 8080:8080 -e SWAGGER_JSON=/docs/swagger.json -v $(pwd)/docs/references/api:/docs swaggerapi/swagger-ui
+```
```
andreas-kupries, line 35: How is that different from `make swagger-serve` ?

andreas-kupries: Ah. I see now. `make swagger-serve` was changed to query the running epinio installation.
Then the question better is, what is the difference to `swagger serve docs/references/api/swagger.json`.

docs/references/api/swagger.json
```diff
@@ -1,5 +1,15 @@
...
+    "http",
+    "https"
+  ],
   "swagger": "2.0",
+  "info": {
+    "description": "This is the API specification of the Epinio API",
+    "title": "Epinio API",
+    "version": "1.0.0"
+  },
+  "basePath": "/api/v1",
```
andreas-kupries, line 12: What swagger version does generate all this ?
I have 0.28.0 and it does not generate `info`. 
Which is why the Makefile contained `sed -i 's/^{/{ "info": {"title": "Epinio", "version":"1"},/' docs/references/api/swagger.json` as post processing after `swagger generate`.

andreas-kupries (changes requested):

docs/howtos/swagger.md
enrichman: It's a different client. The `swagger` command is the one that we use to generate the json file from the comments and annotation, but the UI that he offers is not "great", and it doesn't have the "try out" functionality. :slightly_smiling_face:

docs/references/api/swagger.json
andreas-kupries: Oh! Ok.

PR #1599, opened 2022-06-24
Error refinement [NotFound and Conflict]

enrichman:
I was having a look at the `errors` package of the APIs and it's probably a bit "over complicated", or maybe just verbose.

I wanted to simplify it a bit, and I've started just doing a small couple of things for the NotFound and Conflict errors.

I've changed the `NewNotFoundError` signature to accept an `error` instead of just a message. In my opinion it makes sense to create an API error from another error, that has more context about what happened and where. So I just needed to modify the callers adding a Wrap or using a simple errors.New.

I've also added a `NewConflictError`, and used it (along with the `NewNotFoundError`) to construct the other "domain" errors.

### DIFF
--- a/internal/api/v1/service/validate.go
+++ b/internal/api/v1/service/validate.go
@@ -2,7 +2,6 @@ package service
 
 import (
 	"context"
-	"errors"
 	"fmt"
 
 	"github.com/epinio/epinio/helpers/kubernetes"
@@ -10,9 +9,9 @@ import (
 	"github.com/epinio/epinio/internal/names"
 	"github.com/epinio/epinio/internal/services"
 	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
 
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
-
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
 
 	helmrelease "helm.sh/helm/v3/pkg/release"
@@ -38,14 +37,14 @@ func ValidateService(
 	theService, err := kubeServiceClient.Get(ctx, namespace, service)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			return apierror.NewNotFoundError("service not found")
+			return apierror.NewNotFoundError(errors.Wrap(err, "service not found"))
 		}
 
 		return apierror.InternalError(err)
 	}
 	// See internal/services/instances.go - not found => no error, nil structure
 	if theService == nil {
-		return apierror.NewNotFoundError("service not found")
+		return apierror.NewNotFoundError(errors.New("service not found"))
 	}
 
 	logger.Info("getting helm client")
@@ -61,7 +60,7 @@ func ValidateService(
 	srv, err := client.GetRelease(releaseName)
 	if err != nil {
 		if errors.Is(err, helmdriver.ErrReleaseNotFound) {
-			return apierror.NewNotFoundError(fmt.Sprintf("%s - %s", err.Error(), releaseName))
+			return apierror.NewNotFoundError(errors.Wrapf(err, "release %s not found", releaseName))
 		}
 		return apierror.InternalError(err)
 	}
--- a/internal/api/v1/service/validate.go
+++ b/internal/api/v1/service/validate.go
@@ -37,14 +37,14 @@ func ValidateService(
 	theService, err := kubeServiceClient.Get(ctx, namespace, service)
 	if err != nil {
 		if k8serrors.IsNotFound(err) {
-			return apierror.NewNotFoundError(errors.Wrap(err, "service not found"))
+			return apierror.NewNotFoundError(err.Error()).WithDetailsf("service '%s' not found", service)
 		}
 
 		return apierror.InternalError(err)
 	}
 	// See internal/services/instances.go - not found => no error, nil structure
 	if theService == nil {
-		return apierror.NewNotFoundError(errors.New("service not found"))
+		return apierror.NewNotFoundError("service not found")
 	}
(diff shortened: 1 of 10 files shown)


andreas-kupries:
> So I just needed to modify the callers adding a Wrap or using a simple errors.New.

Which makes all the callers more verbose.

internal/api/v1/service/validate.go
```diff
@@ -61,7 +60,7 @@ func ValidateService(
 	srv, err := client.GetRelease(releaseName)
 	if err != nil {
 		if errors.Is(err, helmdriver.ErrReleaseNotFound) {
-			return apierror.NewNotFoundError(fmt.Sprintf("%s - %s", err.Error(), releaseName))
+			return apierror.NewNotFoundError(errors.Wrapf(err, "release %s not found", releaseName))
```
andreas-kupries, line 63: I can see this being more sensible than what we had, i.e. providing the entire base error instead of just its message.
OTOH the 

```
-		return apierror.NewNotFoundError("service not found")
+		return apierror.NewNotFoundError(errors.New("service not found"))
```

in line 47 above just becomes more verbose than necessary.

How about having two functions, with slightly differing signatures, i.e. the old function, keeping its signature

> NewXError (msg string, ...)

and a new

> NewXFromError (err error, ...)

with the new signature for where we wish to pass in an underlying error ?

And coming back to this place, as written even that is actually to verbose, as it does not use the `...string` argument of the old and new function, while it could have:
```suggestion
			return apierror.NewNotFoundError(err, "release", releaseName, "not found")
```

Although that runs afoul of the `", "` separator used in the underlying `strings.Join`. A `" "` separator would be nicer IMHO.

Looking at other places using `fmt.Errorf`, `fmt.Sprintf`, and the like a simple reorg of the error message could get rid of many of these I believe. I.e. instead of `fmt.Errorf("Foo '%s' does not exist", X)` use `"Foo does not exist:", X`, i.e. the dynamic part is just another detail argument of the `NewX...` function.

andreas-kupries (changes requested):
See the one place where I commented.

enrichman:
> > So I just needed to modify the callers adding a Wrap or using a simple errors.New.
> 
> Which makes all the callers more verbose.

That's true, but an error usually has more information than a simple string (like the stacktrace), and it's not the common case to return an error without an error (usually an error occurred, and you need to create an HTTP error from it).

https://github.com/pkg/errors/blob/master/errors.go#L16-L18

>// The errors.Wrap function returns a new error that adds context to the
>// original error by recording a stack trace at the point Wrap is called,
>// together with the supplied message.

So I think that this is a valid fix:

```go
// BEFORE
if err != nil {
	if k8serrors.IsNotFound(err) {
		return apierror.NewNotFoundError("service not found")
=	}
}

// AFTER
if err != nil {
	if k8serrors.IsNotFound(err) {
		return apierror.NewNotFoundError(errors.Wrap(err, "service not found"))
	}
}
```

while sometimes it's not very useful when you need to return an error without errors (but as I said it's not the most common case)

```go
// BEFORE
if theService == nil {
	return apierror.NewNotFoundError("service not found")
}

// AFTER
if theService == nil {
	return apierror.NewNotFoundError(errors.New("service not found"))
}
```

So if you think that adding the `errors.New` is sometimes too verbose I'm fine with keeping both the implementation. :+1: 


Also about the stacktrace I've seen that the `pkg/errors` is archived and no longer maintained, and they suggest to move to the standard `fmt` lib actually: https://github.com/pkg/errors/issues/245

But I think it's not really an issue, it's just something that I've noticed and that we could add in the backlog to remove a dependency.

internal/api/v1/service/validate.go
enrichman: Ah, I've commented before have read this comment. :smile: 
I'm fine with having the two signatures! :+1:

andreas-kupries: Note that `NewX` functions seem to use only the `err.Error()` of the passed error. The `errors.Wrapf` in the callers effectively devolve into an extended error message. and that is something we already can have without the additional Wrap, through the `details` argument.

enrichman:
@andreas-kupries  As you pointed out we are not really using the error, we are losing the information with `err.Error()`.
And also the `details` are a bit strange, with the `, ` separator, while sometimes we want to format the error or the details in a different way.

Also in many places we are not sure on what to put into the details, and we are passing a blank string.

I was thinking a bit about it, and I wanted to simplify everything a bit more.

I'm going to push a new implementation, with only the string message, and without the need to specify the details. We can add the details later with a `WithDetails` and `WithDetailsf` func.

This seems to simplify everything, and it looks a bit less verbose.

Please note that I'm changing only the `NotFound` and `Conflict` errors, but if that's a feasible approach I will obviosuly port these changes also in the other errors (to avoid too many changes all at once).

andreas-kupries:
:+1: Will wait until the new changes are and then revisit. I am guessing that the `removed details ...` is not everything coming.

andreas-kupries (approved):
Good enough to me now.

The nits I saw can be done in a separate PR.

Example:
> NewNotFoundError(fmt.Sprintf("Application Chart '%s' does not exist", app))

could be written as
> NewNotFoundError("Application Chart", app)

with a slight change to the signature of the new helper. And the `fmt.Sprintf` moves inside.

Similar for `NewConflictError`.

Maybe as part of #1600 ?

enrichman:
Thanks, yes, I'll add those changes in the other PR, so I can merge this.

I was thinking to add a `NewXXXErrorf`, as the std and other libraries have, that adds the formatting capability.

PR #1600, opened 2022-06-24
Error refinement [BadRequest]

enrichman:
Following the https://github.com/epinio/epinio/pull/1599

This PR changes the signature of the `NewBadRequest` to add more context to the BadRequests errors, and use it in our domain errors helpers.

### DIFF
--- a/pkg/api/core/v1/errors/errors.go
+++ b/pkg/api/core/v1/errors/errors.go
@@ -163,11 +163,6 @@ func AppChartIsNotKnown(appChart string) APIError {
 	return NewNotFoundError("application chart", appChart)
 }
 
-// UserNotFound constructs an API error for when the user name is not found in the header
-func UserNotFound() APIError {
-	return NewBadRequestError("user not found in the request header")
-}
-
 /////////////////////////
 //
 // Conflict (409) errors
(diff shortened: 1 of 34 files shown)


andreas-kupries (changes requested):
See #1599 (https://github.com/epinio/epinio/pull/1599#discussion_r913550373) for my general issues with these changes.

enrichman:
I've changed the PR addressing the comments of the #1599.

I've added the `NewXXXErrorf` func to use internally the `fmt.Sprintf`, and I've changed the func name from `BadRequest` to `NewBadRequestError` to match the other constructors.

pkg/api/core/v1/errors/errors.go
```diff
@@ -132,62 +142,60 @@ func NewConflictError(msg string) APIError {
 
 // NamespaceIsNotKnown constructs an API error for when the desired namespace does not exist
 func NamespaceIsNotKnown(namespace string) APIError {
-	return NewNotFoundError(fmt.Sprintf("Targeted namespace '%s' does not exist", namespace))
+	return NewNotFoundErrorf("targeted namespace '%s' does not exist", namespace)
```
andreas-kupries, line 145: When I proposed

> NewNotFoundError ("Application", app)

i.e.

> func NewNotFoundError (kind, instance string) ... {
    ... fmt.Sprintf("%s '%s' does not exist", kind, instance)
}

or

>  ... fmt.Sprintf("%s does not exist: '%s'", kind, instance)

I did not really want a `...f` function for generic formatting. I wanted a restricted function taking just the minimal data needed for the error and hiding the form of the message. Ensuring the caller not having to care about the exact form.
As is the caller still has to write the entire message, and formatting.
Removing that need was the intent behind the proposed change.

The same for the `NewConflict`. This is also a very restricted form for the messages.
`BadRequest` on the other hand can not be restricted in that way, so for that having the generic formatting is good.

andreas-kupries (changes requested):

andreas-kupries (dismissed):

enrichman:
@andreas-kupries thanks for the approval, sorry but I've seen that some errors were not used and so I dismissed your review!

andreas-kupries (approved):

andreas-kupries (approved):

Issue #1603, opened 2022-06-28
epinio push --skip-ssl-verification crash if epinio is SSL-enabled


gitricko:
**Describe the bug**

If your epinio deployment is configure with TLS from let's encrypt and when you do `epinio push --skip-ssl-verification --name APPNAME`, you will get below error. Removing `--skip-ssl-verification` will make epinio push working again.

Question why ignore certificate is causing this issue.

```
....
⚠️  Hit Enter to continue or Ctrl+C to abort (deployment will continue automatically in 5 seconds)


Create the application resource ...

Application exists, updating ...

Collecting the application sources ...

Uploading application code ...

Staging application with code...

🕞  Running staging

🚢  Streaming application logs
Namespace: workspace
Application: expense-mgt-dev
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x5bcae3f]

goroutine 40 [running]:
github.com/epinio/epinio/pkg/api/core/v1/client.(*Client).AppLogs(0xc000114a80, {0xc000bf0620, 0x9}, {0x7ff7bfeff62b, 0xf}, {0xc000a1ca50, 0x10}, 0x1, 0xc0001356f8)
        github.com/epinio/epinio/pkg/api/core/v1/client/apps.go:395 +0x61f
github.com/epinio/epinio/internal/cli/usercmd.(*EpinioClient).AppLogs(0xc000bd65c0, {0x7ff7bfeff62b, 0xf}, {0xc000a1ca50, 0x10}, 0x0?)
        github.com/epinio/epinio/internal/cli/usercmd/app.go:390 +0x69f
github.com/epinio/epinio/internal/cli/usercmd.(*EpinioClient).stageLogs.func1()
        github.com/epinio/epinio/internal/cli/usercmd/push.go:256 +0x4a
created by github.com/epinio/epinio/internal/cli/usercmd.(*EpinioClient).stageLogs
        github.com/epinio/epinio/internal/cli/usercmd/push.go:255 +0xef

```

Hint: Make sure to search through the existing issues, as they often contain important information on recent changes.

**To Reproduce**
Steps to reproduce the behavior, e.g.:

see above

**Expected behavior**
Even if LetsEncrypt certificate is configure correctly in epinio deployment, ignore ssl should cause epinio push to fail.

**Logs**
If applicable, add logs to help explain your problem.

[...]

PR #1620, opened 2022-07-05
Guard against a missing (nil) response from a websocket dialer problem

andreas-kupries:
Ref #1603. The PR might fix the obvious issue. It would not do anything about the possible underlying LE cert issue.

### DIFF
--- a/acceptance/install/scenario2_test.go
+++ b/acceptance/install/scenario2_test.go
@@ -155,6 +155,34 @@ var _ = Describe("<Scenario2> GKE, Letsencrypt-staging, Zero instance", func() {
 			Expect(out).To(Equal("'letsencrypt-staging'"))
 		})
 
+		By("Pushing an app with zero instances, and not verifying certs", func() {
+			appName2 := appName + "-ssv"
+			out, err := epinioHelper.Run("push",
+				"--skip-ssl-verification",
+				"--name", appName2,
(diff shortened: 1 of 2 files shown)


enrichman (approved):

andreas-kupries:
The #1621 test result shows me that `--skip-ssl-verification` works for a properly set up LE issuer and certs.
The #1603 crash likely originates in a setup error. I suspect that no fix beyond the guard added here will be required.

andreas-kupries:
@gitricko reports that the deployment works for him now, modulo the stage logging not working.
His excerpted push log is

```
[...]
🚢  Streaming application logs
Namespace: workspace
Application: sslbug3

❌  failed to tail logs: Failed to connect to websockets endpoint. Response was = <nil>
The error is: malformed HTTP response "\x00\x00\x18\x04\x00\x00\x00\x00\x00\x00\x05\x00\x10\x00\x00\x00\x03\x00\x00\x00\xfa\x00\x06\x00\x10\x01@\x00\x04\x00\x10\x00\x00"

Deploying application ...
.........
✔️  App is online.
[...]
```

Note the error when trying to get staging logs. That is the error which was hidden by the crash.
I suspect something is wrong in his setup of LE certs.
Given that our scenario 2 tests (GKE-LETSENCRYPT-CI) pass. Including the new case I added with skip-verify.

I will now merge this PR.

PR #1621, opened 2022-07-05
Testing / Issue reproduction attempt -- Extended scenario 2 test with --skip-ssl-verification sub case

andreas-kupries:
This PR is an adjunct to #1620 for #1603.
It contains only the extended test case.
No attempts at fixing are made.
This is an attempt to reproduce the #1603 in our CI.
Created as draft PR, and the change to ready triggers the relevant workflow.


andreas-kupries:
The test shows that `--skip-ssl-verification` works for a properly set up LE issuer and certs. The #1603 crash likely originates in a setup error. I suspect that no fix beyond the guard will be required.

PR #1624, opened 2022-07-06
Ui143 validate archive type

jimmykarily:
Partly fixes: https://github.com/epinio/ui/issues/143 (missing the front-end validation)


internal/api/v1/application/upload.go
```diff
@@ -9,8 +13,20 @@ import (
...
 )
 
+// Should match the supported types:
+// https://github.com/epinio/helm-charts/blob/3954c214de3d7b957cfc2054ba4fa4bfa140f5a3/chart/epinio/templates/stage-scripts.yaml#L27-L62
+var validArchiveTypes = []string{
+	"application/zip",
+	"application/x-tar",
+	"application/gzip",
+	"application/x-bzip2",
+	"application/x-xz",
```
jimmykarily, line 27: https://github.com/h2non/filetype#supported-types

jimmykarily: Added that as a comment in the code

internal/api/v1/application/upload.go
```diff
@@ -31,6 +48,16 @@ func (hc Controller) Upload(c *gin.Context) apierror.APIErrors {
 	}
 	defer file.Close()
 
+	// TODO: Does this break streaming of the file? We need to get the whole file
+	// before we can check its type
```
jimmykarily, line 52: Since we are only reading 512 bytes from the file and the returning the cursor back to 0, I think this is fine. We are not reading the whole file nor do we store anything on the filesystem.

jimmykarily, line 52:
```suggestion
```

mmartin24:
I checked it on branch and verified that now there is a warning on the upload step if the file is not supported:
image

@jimmykarily I understand here is where the front-end validation was placed and not earlier, right?

jimmykarily:
@mmartin24 this PR implements only the backend validation. The front-end one (which should come eariler) will happen on the UI repository.

mmartin24:
thank you @jimmykarily, then I guess this ticket (https://github.com/epinio/ui/issues/143) should be closed once that earlier ui implementation is ready, unless you think is better to split it.

andreas-kupries (approved):

PR #1625, opened 2022-07-06
Loading extensible Domain/Secret mapping

### LINKED ISSUE #1605: New TLS cert on every new app pushed
**Describe the bug**
In every push of a different app, a new certificate request is sent to Let'sEncrypt and a new challenge is created. There are two problems with this:
1. It is slow. Each certificate request takes about 1-2 minutes to fully complete, and until then the application is insecure
2. It is not scalable. Let'sEncrypt limits the amount of requests to the same DNS, so in high load situations, this limit can be surpassed and all future requests for that DNS will be blocked. The exact limits are documented here: https://letsencrypt.org/docs/rate-limits/

**To Reproduce**
1. Push any application to Epinio
2. Watch as a new challenge is issued (kubectl get challenge -n epinio) or
3. See that a new certificate was created (kubectl get certificates -n epinio)

**Expected behavior**
Maybe there should be a way of deploying Epinio using wildcard DNS, so all new applications use the same certificate from the top domain.

**Logs**
This is a log from when the LetsEncrypt limit was surpassed

```The certificate request has failed to complete and will be retried: Failed to wait for order resource "epinio-n72n45-2651626544" to become ready: order is in "errored" state: Failed to create Order: 429 urn:ietf:params:acme:error:rateLimited: Error creating new order :: too many certificates (5) already issued for this exact set of domains in the last 168 hours: epinio.example.com: see https://letsencrypt.org/docs/duplicate-certificate-limit/```

It can be acquired describing the challenge

**Cluster:**
 - Provider: K3D
 - Kubernetes Version: v1.23.6

 - OS: Ubuntu 18.04
 - Epinio Version: 1.0.0
 - LoadBalancer: Happens when using both Nginx and Traefik

### PR DESCRIPTION
**Pull request by andreas-kupries**
May fix #1605

Done so far

- [x] loading domain map from secrets. secrets identified by label
- [x] matching domain against map. unit tests

To do

- [ ] perf tests for map loading using larger number of secrets
- [x] loading and using map on app deployment, passing result into app helm chart
- [x] extending standard app chart to use passed-in secrets
- [ ] trialing

### DIFF
--- a/helpers/cahash/cahash.go
+++ b/helpers/cahash/cahash.go
@@ -17,7 +17,7 @@ import (
 // -----------------------------------------------------------------------------------
 
 func GenerateHash(certRaw []byte) (string, error) {
-	cert, err := decodeOneCert(certRaw)
+	cert, err := DecodeOneCert(certRaw)
 	if err != nil {
 		return "", fmt.Errorf("failed to decode certificate\n%w", err)
 	}
@@ -36,7 +36,7 @@ func GenerateHash(certRaw []byte) (string, error) {
 
 // Iterates over pem blocks until a non-CA certificate if found or no other PEM
 // blocks exist.
-func decodeOneCert(raw []byte) (*x509.Certificate, error) {
+func DecodeOneCert(raw []byte) (*x509.Certificate, error) {
 	byteData := raw
 	for len(byteData) > 0 {
 		block, rest := pem.Decode(byteData)
--- a/internal/api/v1/deploy/deploy.go
+++ b/internal/api/v1/deploy/deploy.go
@@ -10,6 +10,7 @@ import (
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/cli/server/requestctx"
+	"github.com/epinio/epinio/internal/domain"
 	"github.com/epinio/epinio/internal/helm"
 	"github.com/epinio/epinio/internal/helmchart"
 	"github.com/epinio/epinio/internal/registry"
@@ -41,6 +42,7 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 	imageURL := appObj.ImageURL
 	routes := appObj.Configuration.Routes
 	chartName := appObj.Configuration.AppChart
+	domains := domain.MatchMapLoad(ctx)
 
 	deployParams := helm.ChartParameters{
 		Context:        ctx,
@@ -54,6 +56,7 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 		Username:       username,
 		StageID:        stageID,
 		Routes:         routes,
+		Domains:        domains,
 		Start:          start,
 	}
 
--- a/internal/api/v1/deploy/deploy.go
+++ b/internal/api/v1/deploy/deploy.go
@@ -42,7 +42,7 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 	imageURL := appObj.ImageURL
 	routes := appObj.Configuration.Routes
 	chartName := appObj.Configuration.AppChart
-	domains := domain.MatchMapLoad(ctx)
+	domains := domain.MatchMapLoad(ctx, app.Namespace)
 
 	log.Info("domain map begin")
 	for k, v := range domains {
--- a/internal/api/v1/deploy/deploy.go
+++ b/internal/api/v1/deploy/deploy.go
@@ -44,11 +44,12 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 	chartName := appObj.Configuration.AppChart
 	domains := domain.MatchMapLoad(ctx, app.Namespace)
 
-	log.Info("domain map begin")
+	maplog := log.V(1)
+	maplog.Info("domain map begin")
 	for k, v := range domains {
-		log.Info("domain map", k, v)
+		maplog.Info("domain map", k, v)
 	}
-	log.Info("domain map end")
+	maplog.Info("domain map end")
 
 	deployParams := helm.ChartParameters{
 		Context:        ctx,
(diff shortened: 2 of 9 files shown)

### CODE REVIEW

**Comment by andreas-kupries** [comment]
The feature seems to work in local testing ...
  - :warning: And in writing this up I realize, that it does not, and why. Will come back to that after the test sequence.

There are no CI test cases for this yet.
Performance tests with large number of secrets to load (10, 100, 1000) has not been done either yet.

Local testing was done like so

1. Deployed cluster, deployed epinio.
2. Pushed an application as normal. This generates a Cert and Secret for the route.
3. Used kube commands to get that Secret from the cluster, as yaml
4. Edit this secret
     1. Rename it 
     2. Change the namespace to `epinio` ! The server searches only in this namespace
     3. Removed cert manager annotations
     4. Added label `epinio.io/routing`
     5. Removed other generated stuff blocking an upload of the changed secret:
          - `managedFields`,
          - `creationTimestamp`,
          - `resourceVersion`,
          - `uid`.
5. Applied the changed secret.
6. Re-pushed the application.
7. The server log should now show a non-empty domain map and successful match - grep `domain ma` (The `ma` matches `map` and `match`).
8. The helm manifest for the app's release should show the ingress using the new secret, by name.
9. The generated secret should be gone, as well as the Cert resource.

:warning: Regarding this not working - The secret with the domain information is searched for in the `epinio` namespace.
The app ingress is in some other namespace (here: `workspace`).
The ingress refers to a secret outside of its own namespace! :boom: 
The secret will not be used.

**Comment by andreas-kupries** [comment]
Possible solutions, of the cuff ...

1. Search secrets in the destination namespace for the app, instead of epinio .
2. Copy secrets to the app namespace, somehow ? (Read secret into memory, add secret into the deployed chart in some way ?)
3. Anything else ... ?

I actually begin to like 1 above. It basically will domain info per namespace/user, and the number should be small. Even if there is a larger number of secrets scattered among namespaces. That should keep the perf impact low.
The app's namespace is also available at the time of deployment.

... Implemented 1, works.

**Comment by andreas-kupries** [comment]
:notes: Using 1 and searching only in the destination namespace is actually no big limitation at all.
One of epinio's components is `kubed`, a controller capable of copying secrets between namespaces based on annotations on the secret.

For example, placing a routing secret `FOO` into the `epinio` namespace and annotating it with `kubed.appscode.com/sync="app.kubernetes.io/component=epinio-namespace"` will cause `kubed` to copy `FOO` to all namespace matching the label selector `app.kubernetes.io/component=epinio-namespace`. These are all the namespaces under epinio's control.

For more limited distribution the operator can add custom labels to the desired namespaces and then add matching annotations on the relevant secret(s).

This requires documentation, of course.

File: helpers/cahash/cahash.go [hunk 1]
```diff
@@ -36,7 +36,7 @@ func GenerateHash(certRaw []byte) (string, error) {
 
 // Iterates over pem blocks until a non-CA certificate if found or no other PEM
 // blocks exist.
-func decodeOneCert(raw []byte) (*x509.Certificate, error) {
+func DecodeOneCert(raw []byte) (*x509.Certificate, error) {
```
enrichman, line 39: I would also use a guard close to keep the indentation cleaner, so a check for the byte length before the decoding:

```
if len(byteData) == 0 {
    return nil, errors.New("failed find PEM data")
}
// decode and parse
```

Then having a look at the [original func](https://github.com/paketo-buildpacks/ca-certificates/blob/main/cacerts/certs.go) it was expecting only one cert, while we are returning the first one.
I'm not sure about it but I guess it's fine (?).

internal/helm/helm.go
```diff
@@ -98,12 +100,32 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
...
+			var routeInfo string
+			if err != nil || domainSecret == "" {
+				// No secret found, no secret passed
+				routeInfo = fmt.Sprintf(`{"id":"%s","domain":"%s","path":"%s"}`,
+					rdot, r.Domain, r.Path)
+			} else {
+				// Pass the found secret
+				routeInfo = fmt.Sprintf(`{"id":"%s","domain":"%s","path":"%s","secret":"%s"}`,
+					rdot, r.Domain, r.Path, domainSecret)
+			}
```
enrichman, line 126: I'm not a big fan of handwritten JSON, but I see that we have an entire YAML a few lines after, so... let it be.
I will probably open a couple of issue/enhancement to do to move these to structs and/or use some libraries to do the marshaling.

enrichman (approved):

enrichman (changes requested):
Just a couple of small things

internal/helm/helm.go
andreas-kupries: Yeah, this was keeping with the existing code (which IIRC I am also responsible for, i.e. wrote it :( ). Cleanup issues to come are good.

internal/api/v1/deploy/deploy.go
```diff
@@ -40,6 +41,13 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 	imageURL := appObj.ImageURL
 	routes := appObj.Configuration.Routes
 	chartName := appObj.Configuration.AppChart
+	domains := domain.MatchMapLoad(ctx, app.Namespace)
+
+	log.Info("domain map begin")
```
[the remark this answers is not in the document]
andreas-kupries: Sure.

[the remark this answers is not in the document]
andreas-kupries: Commit d2b5ff0ca014cb577ce011fe7dc8fc488876c4c7

helpers/cahash/cahash.go
andreas-kupries: I can insert the guard, sure. I am not sure if that is better however, because then the loop will do the same check and fall through anyway.

Point is, the loop __cannot be eliminated__.

Your statement `we are returning the first one.` is not true as such.

Note the `if !cert.IsCA { ...` check in line 57. The code actually delivers the `first cert, which is not a CA cert`.

This desired cert may be deep into the PEM data, thus the loop.

enrichman: Yes, sorry, what I meant is that if the PEM contains multiple certs it will return the first "not-CA" one in the chain, but I'm not an experte of certs/ssl stuff, so I might be wrong here. :smile:

enrichman (approved):

PR #1650, opened 2022-07-19
Spike: Dex re-integration

### LINKED ISSUE #1571: [Spike] Allow Epinio users to login with any OIDC provider
Relevant: https://github.com/epinio/epinio/issues/686

We ran a spike around a year ago (see link above). Let's check if there are other alternatives and how the previous findings would work with the current Epinio. There [is a branch](https://github.com/epinio/epinio/tree/486-multiple-users-dex) which should have a working PoC of the old Spike.

## What we need

- Users should be able to login to Epinio using existing accounts on some OIDC provider (GitHub, Google, etc)
- Operators should be able to enable/disable OIDC providers
- If there is any authz information provided by the upstream provider, see if we can make any use of it (e.g. to replace our namespace based authz).
- In any case, operators should be able to promote users to admins or revoke admin rights.

andreas-kupries:
Fix #1571 - import work on helm chart for dex integration

Chart PR https://github.com/epinio/helm-charts/pull/252


andreas-kupries:
Confirmed `dex` as properly running and accessible by retrieving `https://dex.<domain>/.well-known/openid-configuration`.
(discovery endpoint)

andreas-kupries:
Trying the dex example-app fails when using https access, with a `509: certificate signed by unknown authority`.
This happens even when adding (?)/using the epinio ca cert we get from `epinio settings update-ca` and placing it into its own file.

Talking unsecured http ... Works to a point ... While the app is then able to retrieve the discovery data and later fails on an issuer mismatch, http vs https.

Going back to an unsecured ingress ... That works ... The example-app can be used to trial auth workflow.

andreas-kupries:
Flow refs
  - https://darutk.medium.com/diagrams-and-movies-of-all-the-oauth-2-0-flows-194f3c3ade85
  - https://auth0.com/docs/get-started/authentication-and-authorization-flow/authorization-code-flow
  - https://auth0.com/docs/get-started/authentication-and-authorization-flow/which-oauth-2-0-flow-should-i-use
  - https://auth0.com/docs/get-started/authentication-and-authorization-flow/which-oauth-2-0-flow-should-i-use#is-the-client-a-native-mobile-app-

Reading the old spike in light of the above it seems that

  - The epinio server serves as the client app.
  - Dex serves as the provider.
  - The /login endpoint receives the login form, redirects to dex, starting the flow
      - Unclear where the form comes from initially
  - The /dex/callback endpoint is where flow results from dex end up.
  - Was the cli serving as the user posting the /login request ? No relevant cli code in the spike.
  - There is traefik involved with its forward auth plugin.
      - Maybe an automatic thing where traefik uses a missing id/access token as signal to initiate the flow ? Also to auto refresh the main token from the refresh token when needed ?
  - Still puzzling how this works for the cli, as the main flow sort of assumes a browser.

Or the heat is getting to me.

  - It seems that cli generally has to
      - Launch an internal web server to receive the callback
      - Launch/open a browser window for the parts of the flow it should not see, initially pointing at the provider auth page (here `dex`) ... This browser window then goes through the flow until approval is granted, which goes to the callback ... At that point the internal web server gets the auth code, get the tokens from the provider via it, and then close the server ...
      - From then on the tokens are used taking to the API, ...
      - The token has to be stored locally
      - Seems that only the `login` command would require this.
      - All other cli commands would assume to have a valid token.

  - The epinio UI, being a regular web application has it easier for the flow, the browser is already present.
  - I see now way to do such a flow from the server. At least not correctly ... There is no way to perform the necessary user interaction.

jimmykarily:
NOTE: Don't let the epinio server fail on startup when dex is not reachable. Require dex lazily (only when needed).

enrichman:
@jimmykarily 

> - [x] Don't let the epinio server fail on startup when dex is not reachable. Require dex lazily (only when needed).

:heavy_check_mark: now there is a lazy initialization of the provider, done during the first call to the authMiddleware. I'm not sure if it need to be "refreshed" for the key rotations or such, but a restart of the Epinio service will suffice.

> - [x]  Fix prepare script to do the login correctly

:heavy_check_mark: the `login.sh` script was moved to the `scripts` folder and modified a bit to get 3 args. It can be launched also with a `-v` flag to get a more detailed log.
Also the Go login cmd is available with the same arguments, and can be triggered with a `go run`

```
./scripts/login.sh admin@epinio.io password https://auth.172.21.0.4.omg.howdoi.website
```
```
go run cmd/login/main.go admin@epinio.io password https://auth.172.21.0.4.omg.howdoi.website
```

> - [x]  epinio login should also allow the user to copy-paste the code (in case epinio is running on a remote machine)

:heavy_check_mark: now the copy-pasting is available again logging in with the `--prompt` flag, so the
```
epinio login --prompt https://epinio.172.21.0.4.omg.howdoi.website
```
will not spin up the server but it will wait for a user interaction.

> - [ ]  Create additional (fixture) users for tests that need more than one (by patching the configmap before tests start). Some admins and some regular ones.

This is the only missing part. Now some tests are passing but we need to create some fixtures and use them in our test. We also need to map the static dex users (with email) with some epinio users (at least the admin, the other users will be created automatically after the first login).

We need also to update the info message of the Helm charts, because the username and password flags are not available anymore.

acceptance/login_test.go
```diff
@@ -177,3 +227,38 @@ var _ = Describe("Login", func() {
...
 	})
 })
+
+// getAuthorizationCode will read the output from reader looking for the authentication url where to login,
+// Then it will try to login, get the authentication code and write it to the stdin
+func getAuthorizationCode(reader io.Reader, stdinPipe io.Writer) {
+	defer GinkgoRecover()
+
+	// wait a second for the command to execute
+	time.Sleep(time.Second)
```
jimmykarily, line 237: maybe read from reader until the output we expect appears? I'm not sure if we can immediately start reading and how the final code will look but let's try to avoid magic sleeps

internal/dex/code_verifier.go
```diff
@@ -0,0 +1,39 @@
...
+// ChallengeS256 returns an encoded SHA256 code_challenge of the code_verifier
+func (c *CodeVerifier) ChallengeS256() string {
+	h := sha256.New()
+	h.Write([]byte(c.Value))
+	hash := h.Sum(nil)
+
+	encoded := base64.StdEncoding.EncodeToString(hash)
+	encoded = strings.Replace(encoded, "+", "-", -1)
+	encoded = strings.Replace(encoded, "/", "_", -1)
+	encoded = strings.Replace(encoded, "=", "", -1)
```
jimmykarily, line 36: What's with the replacements?

internal/dex/dex.go
```diff
@@ -0,0 +1,142 @@
...
+	token, err := verifier.Verify(ctx, rawIDToken)
+	if err != nil {
+		return nil, errors.Wrap(err, "verifying rawIDToken")
+	}
+	return token, nil
+}
+
+func newVerifier(ctx context.Context, clientID string) (*oidc.IDTokenVerifier, error) {
+	// Server should always trust the dex url certificate
+	// TODO: Mount the actual certificate and ExtendLocalTrust?
```
jimmykarily, line 91: done?

internal/dex/dex.go
```diff
@@ -0,0 +1,142 @@
...
+	provider, err := oidc.NewProvider(ctx, fmt.Sprintf("https://auth.%s", domain))
+	if err != nil {
+		return nil, errors.Wrap(err, "setting up the provider")
+	}
+
+	// Create an ID token parser, but only trust ID tokens issued to "example-app"
+	return provider.Verifier(&oidc.Config{ClientID: clientID}), nil
+}
+
+// TODO (ec) Keeping this for the TODOs
```
jimmykarily, line 107: Remove this function?

pkg/api/core/v1/client/apps.go
```diff
@@ -634,7 +640,8 @@ func (c *Client) AppPortForward(namespace string, appName, instance string, opts
 		PingPeriod: time.Second * 5,
 	})
 
-	wrapper := transport.NewBasicAuthRoundTripper(c.Settings.User, c.Settings.Password, upgradeRoundTripper)
+	// TODO: does it work?
```
jimmykarily, line 643: It seems so. I tried:

```
epinio app exec myapp
$ nc -l 9000
```

and from another terminal:

```
epinio app port-forward 9000
```

Opening localhost:9000 works and it prints things in the app exec terminal

We can remove the TODO

acceptance/helpers/auth/auth.go
```diff
@@ -0,0 +1,119 @@
...
+		return "", err
+	}
+
+	// get auth code from HTML
+	html, err := io.ReadAll(res.Body)
+	if err != nil {
+		return "", err
+	}
+
+	// this is a very ugly/hacky way, but it's just for testing
```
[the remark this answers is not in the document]
enrichman: Yes, this was going to look for the value of the input, without parsing the HTML and adding an external library.
It's quite brittle, so I'm going to change it and use a regexp.
I'm not a huge fan of regular expression but in this case it seems the best option!

acceptance/helpers/auth/dex.go
```diff
@@ -0,0 +1,76 @@
...
+	Expect(err).ToNot(HaveOccurred(), out)
+}
+
+func restartDex() {
+	// sometimes this fails for empty patch
+	_, _ = proc.Kubectl("rollout", "restart", "-n", "epinio", "deployment/dex")
+	out, err := proc.Kubectl("rollout", "status", "-n", "epinio", "deployment/dex")
+	Expect(err).ToNot(HaveOccurred(), out)
+
+	time.Sleep(10 * time.Second)
```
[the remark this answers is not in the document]
enrichman: We are waiting for Dex to restart and to "stabilize" (getting the conf and get ready).
This was done because we were adding some users to do the tests, but with the latest helm-charts we actually added them in the configuration, so it's possible to authenticate without actually "hack" the configuration and restart it.
I think I can drop this part all together. :smile:

acceptance/login_test.go
enrichman: Yes, it makes sense. Let me try that. :slightly_smiling_face:

internal/auth/user.go
```diff
@@ -47,6 +49,29 @@ func NewUserFromSecret(secret corev1.Secret) User {
 	return user
 }
 
+// NewUserFromIDToken create an Epinio User from an IDToken
+func NewUserFromIDToken(idToken *oidc.IDToken) (User, error) {
+	user := User{}
+
+	// TODO check which claims we need to get
```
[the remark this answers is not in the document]
enrichman: Well, if we are not using it I would say that we can remove it. We can still add it later. :slightly_smiling_face:

internal/cli/server/server.go
```diff
@@ -160,60 +166,131 @@ func initContextMiddleware(logger logr.Logger) gin.HandlerFunc {
...
+var oidcProvider *dex.OIDCProvider
 
-	userMap, err := loadUsersMap(ctx)
-	if err != nil {
-		response.Error(ctx, apierrors.InternalError(err))
-		ctx.Abort()
-		return
+// getOIDCProvider returns a lazy constructed OIDC provider
+func getOIDCProvider(ctx context.Context) (*dex.OIDCProvider, error) {
+	// TODO should this "expire" after a while (key rotations, other auth providers)?
```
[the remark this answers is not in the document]
enrichman: That's true. I still have the feeling that we need to switch the `storage` configuration though. If the Dex installation that we are going to provide is meant to be "production-ready" we should go for the Kubernetes option probably: https://dexidp.io/docs/storage/#configuration

```yaml
storage:
  type: kubernetes
  config:
    inCluster: true
```

We are using the `memory`, and that means that a restart will cause the old tokens to be invalid.

internal/cli/usercmd/login_oidc.go
```diff
@@ -0,0 +1,173 @@
...
+	if err != nil {
+		return errors.Wrap(err, "constructing dexProviderConfig")
+	}
+
+	token, err := c.generateToken(ctx, oidcProvider, prompt)
+	if err != nil {
+		return errors.Wrap(err, "error while asking for token")
+	}
+
+	// TODO store also the RefreshToken and implement refresh flow
```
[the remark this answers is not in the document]
enrichman: Ah yes, it was just an old TODO to remind me about adding the refresh flow. :smile:

internal/dex/code_verifier.go
enrichman: That is part of the RFC. Basically the base64 needs to be "url safe", and so it's actually a `BASE64URL-ENCODE`, not just a `BASE64-ENCODE`.

https://datatracker.ietf.org/doc/html/rfc7636#section-4.2
example: https://www.oauth.com/oauth2-servers/pkce/authorization-request/

internal/dex/dex.go
enrichman: Yes, and also the one above actually!

scripts/login.sh
```diff
@@ -0,0 +1,114 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
enrichman: Yes, I did this before the Go implementation. Then I realized that it was not going to work everywhere and I did the Go script, that is probably more maintanable, and readable. I'll move this in the `docs` just because it took me a lot of time to do it. :laughing:

scripts/prepare-environment-k3d.sh
```diff
@@ -61,7 +61,13 @@ function deploy_epinio_latest_released {
...
   helm repo update
   helm upgrade --wait --install -n epinio --create-namespace epinio epinio/epinio \
-    --set global.domain="$EPINIO_SYSTEM_DOMAIN"
+       --set global.domain="$EPINIO_SYSTEM_DOMAIN" \
+       --set dex.config.issuer="https://auth.$EPINIO_SYSTEM_DOMAIN" \
+       --set "dex.ingress.tls[0].hosts[0]=auth.$EPINIO_SYSTEM_DOMAIN" \
+       --set "dex.ingress.tls[0].secretName=dex-tls" \
+       --set "dex.ingress.hosts[0].host=auth.$EPINIO_SYSTEM_DOMAIN" \
+       --set "dex.ingress.hosts[0].paths[0].path=/" \
+       --set "dex.ingress.hosts[0].paths[0].pathType=Prefix"
```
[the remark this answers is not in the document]
enrichman: Maybe those are the tests that you tried at the beginning, but it seems that they are already defined in the dex.yaml chart as default. I'm going to remove them. :+1:

internal/cli/server/server.go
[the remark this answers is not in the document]
jimmykarily: The type of storage used by dex has nothing to do with this code though, no? The change you refer to, would happen on the helm-chart.

internal/dex/code_verifier.go
jimmykarily: I think you can use [`RawURLEncoding`](https://cs.opensource.google/go/go/+/master:src/encoding/base64/base64.go;l=128) instead of `StdEncoding` above and achieve the same thing.

internal/cli/usercmd/client.go
```diff
@@ -105,7 +107,7 @@ func New() (*EpinioClient, error) {
 		return nil, errors.Wrap(err, "error loading settings")
 	}
 
-	apiClient := epinioapi.New(cfg)
+	apiClient := epinioapi.New(context.Background(), cfg)
```
jimmykarily, line 110: And now we need this `New()` function to also accept a context so we can push down the cobra command context. :thinking:

internal/cli/server/server.go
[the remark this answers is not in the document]
enrichman: Yes, correct, I've changed it to the Kubernetes type. It was just a self reminder because I guess that the key-rotation and failure that I experienced was due to this.

internal/dex/code_verifier.go
enrichman: Ah! That's interesting, thanks! Yes, it seems what we need: https://cs.opensource.google/go/go/+/master:src/encoding/base64/base64.go;l=53;drc=bd8a5b00fcf71fb711ba8996a880b5b07f7b3634

internal/cli/server/server.go
```diff
@@ -160,60 +166,123 @@ func initContextMiddleware(logger logr.Logger) gin.HandlerFunc {
...
-		ctx.Abort()
-		return
+	mainDomain, err := domain.MainDomain(ctx)
+	if err != nil {
+		return nil, errors.Wrap(err, "getting main domain")
 	}
 
-	logger.V(1).Info("Basic auth authentication")
+	issuer := fmt.Sprintf("https://auth.%s", mainDomain)
+	oidcProvider, err = dex.NewOIDCProvider(ctx, issuer, "epinio-cli")
```
jimmykarily, line 183: We are not the `epinio-cli` here, are we? Shouldn't we use an authenticated client whenever possible? We do create a secret for an authenticated client here: https://github.com/epinio/helm-charts/pull/252/files#diff-cd53f64fab50a25f4a05f17fea4f64e7648a1522d85c48decd504d905b6989e8R7

This function is always uses `OutOfBrowserURN` for redirections. I guess it doesn't matter because the server only uses the provider for token validation and not the full login flow (so no redirections happen). Maybe it's fine like this then?

jimmykarily: or we issue the token with the authenticated client in the audience too?

enrichman: Not sure what you mean. The token was issued for the `epinio-cli` client, so the server is checking that the token has the `epinio-cli` audience in it.

jimmykarily: https://dexidp.io/docs/custom-scopes-claims-clients/#scopes

```
audience:server:client_id:( client-id ) Dynamic scope indicating that the ID token should be issued on behalf of another client. See the “Cross-client trust and authorized party” section below.
```

enrichman: The example that they do is the possibility to generate token for another client. i.e. you could have a web page in the Epinio UI that can generate a token, that will have the `epinio-cli` audience, even though it was issued by the `epinio-ui`.
But we still need the clientID, and check that the audience is matching what we expect to match.
So I'm not sure how this could help.

jimmykarily: The `epinio-cli` auth client is generating the tokens but the server auth client (the one we wanted to introduce, the one that has credentials) is the one that checks the validity of the token. Thus, the server auth client is also in the "audience". I think the original request to generate the token, should include the server auth client in the audience so that it can later check the token's validity. At least, that's how I understand this.

enrichman: I'm still not convinced. At the moment we have only one client, that is the `epinio-cli`. The server is not acting as a client, because is not requesting tokens, but it just verifies them.

The audience in the token `MUST contain the OAuth 2.0 client_id of the Relying Party as an audience value. It MAY also contain identifiers for other audiences.` (https://openid.net/specs/openid-connect-core-1_0.html#IDToken).

As far as I understood looking for some other [explanations](https://stackoverflow.com/questions/28418360/jwt-json-web-token-audience-aud-versus-client-id-whats-the-difference) it should be used to identify the resources that the token has access to (i.e: api.epinio.io, and not api3.epinio.io).

Looking at the Dex [doc](https://dexidp.io/docs/custom-scopes-claims-clients/#cross-client-trust-and-authorized-party) they say
> The web app can then use the following scope to request an ID token that’s issued for the command line tool.
> `audience:server:client_id:cli-app`

so in our case we should add a new staticClient (or use/modify the `epinio-ui`?), altough if it's not really a client, and modify the client to ask for an `epinio-server` audience?

```yaml
staticClients:
- id: epinio-server
    secret: {{ $clientSecret }}
    name: 'Epinio Web UI'
    redirectURIs:
    - "https://epinio.{{ .Values.global.domain }}/dex/callback"
    trustedPeers:
    - epinio-cli

- id: epinio-cli
    public: true
    name: 'Epinio cli'
```
so modifying the cli part that asks for the token: https://github.com/epinio/epinio/blob/adc72decd12a82ee717cefe96535e9084dc9edb1/internal/cli/usercmd/login_oidc.go#L53-L56

adding this scope:

```go
oidcProvider.Config.Scopes = append(oidcProvider.Config.Scopes, "audience:server:client_id:epinio-server")
```

But this would simply mean to "hidden" the cli client, and it doesn't seems like the ideal use case.

jimmykarily (approved):

PR #1652, opened 2022-07-21
Remove dependency on `Deployment` resource as entry for app workload

### LINKED ISSUE #1629: Pushing with an application template without a deployment stalls
**Describe the bug**

If I edit the deployment manifest to use a Statefulset instead of a Deployment, the push command doesn't exit after the build and deploy. I can tell from kubectl that the application is running correctly (and can access the app) so the only issue is that the CLI doesn't know that the app is alive.

**To Reproduce**
Add a new application template with my statefulset template: https://chameleons.sfo3.digitaloceanspaces.com/epinio-stateful-application-0.1.15.tgz 

Push an application with this template.

**Expected behavior**
I would expect the CLI to return successfully when the application is running

andreas-kupries:
Fix #1629.

  - With our use of helm charts and --wait'ing for their deployment it is not really necessary any longer.
  - It is specific to kube Deployments.

### DIFF
--- a/internal/api/v1/application/exec.go
+++ b/internal/api/v1/application/exec.go
@@ -45,7 +45,7 @@ func (hc Controller) Exec(c *gin.Context) apierror.APIErrors {
 		return apierror.NewAPIError("Cannot connect to application without workload", http.StatusBadRequest)
 	}
 
-	workload := application.NewWorkload(cluster, app.Meta)
+	workload := application.NewWorkload(cluster, app.Meta, app.Workload.DesiredReplicas)
 	podNames, err := workload.PodNames(ctx)
 	if err != nil {
 		return apierror.InternalError(err)
@@ -71,7 +71,7 @@ func (hc Controller) Exec(c *gin.Context) apierror.APIErrors {
 		podToConnect = podNames[0]
 	}
 
-	appData, err := workload.Get(ctx, 0)
+	appData, err := workload.Get(ctx)
 	if err != nil {
 		return apierror.InternalError(err)
 	}
--- a/internal/api/v1/application/exec.go
+++ b/internal/api/v1/application/exec.go
@@ -71,12 +71,12 @@ func (hc Controller) Exec(c *gin.Context) apierror.APIErrors {
 		podToConnect = podNames[0]
 	}
 
-	deployment, err := workload.Deployment(ctx)
+	appData, err := workload.Get(ctx, 0)
 	if err != nil {
 		return apierror.InternalError(err)
 	}
 
-	proxyRequest(c.Writer, c.Request, podToConnect, namespace, deployment.Name, clientSetHTTP1)
+	proxyRequest(c.Writer, c.Request, podToConnect, namespace, appData.Name, clientSetHTTP1)
 
 	return nil
 }
--- a/internal/api/v1/application/running.go
+++ b/internal/api/v1/application/running.go
@@ -57,7 +57,7 @@ func (hc Controller) Running(c *gin.Context) apierror.APIErrors {
 
 	if app.Workload.DesiredReplicas != app.Workload.ReadyReplicas {
 		err := wait.PollImmediate(time.Second, duration.ToAppBuilt(), func() (bool, error) {
-			podList, err := application.NewWorkload(cluster, app.Meta).Pods(ctx)
+			podList, err := application.NewWorkload(cluster, app.Meta, app.Workload.DesiredReplicas).Pods(ctx)
 			if err != nil {
 				return false, err
 			}
--- a/internal/api/v1/application/running.go
+++ b/internal/api/v1/application/running.go
@@ -1,20 +1,34 @@
 package application
 
 import (
+	"time"
+
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/internal/api/v1/response"
 	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/duration"
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/gin-gonic/gin"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"k8s.io/kubectl/pkg/util/podutils"
 )
 
 // Running handles the API endpoint GET /namespaces/:namespace/applications/:app/running
-// It waits for the specified application to be running (i.e. its
-// deployment to be complete), before it returns. An exception is if
-// the application does not become running without
-// `duration.ToAppBuilt()` (default: 10 minutes). In that case it
-// returns with an error after that time.
+//
+// It waits for the specified application to be running (i.e. its deployment to be
+// complete), before it returns. An exception is if the application does not become ready
+// within `duration.ToAppBuilt()` (default: 3 minutes). In that case it returns with an
+// error after that time.
+//
+// __API PORTABILITY NOTE__
+//
+// With the switch to deployment of apps via helm, and waiting for the helm deployment to
+// be ready before returning this endpoint is technically superfluous, and it should never
+// fail. The command line has been modified to not invoke it anymore.
+//
+// It is kept for older clients still calling on it. Because of that it is also kept
+// functional. Instead of checking for an app `Deployment` and its status it now checks
+// the app `Pod` stati.
 func (hc Controller) Running(c *gin.Context) apierror.APIErrors {
 	ctx := c.Request.Context()
 	namespace := c.Param("namespace")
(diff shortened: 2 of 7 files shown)


andreas-kupries:
Chart PR https://github.com/epinio/helm-charts/pull/253 to support the removal of the dependency on Deployments.

internal/api/v1/application/exec.go
```diff
@@ -71,12 +71,12 @@ func (hc Controller) Exec(c *gin.Context) apierror.APIErrors {
 		podToConnect = podNames[0]
 	}
 
-	deployment, err := workload.Deployment(ctx)
+	appData, err := workload.Get(ctx, 0)
```
jimmykarily, line 74: As we discussed in a call, `Workload` struct seems to be superfluous and can be removed (merged with the `AppDeployment`). Then we can implement a `GetWorkload` method on the app itself instead of this `Get`. We won't need to pass the desired instances then.

andreas-kupries: Larger discussion got triggered ... Defering full resolution into new ticket https://github.com/epinio/epinio/issues/1657

For now a better intermediate change would be to move the `desiredInstances` argument from `Get` to `NewWorkload`.

internal/cli/usercmd/push.go
```diff
@@ -237,10 +237,10 @@ func (c *EpinioClient) Push(ctx context.Context, params PushParams) error { // n
...
 	c.ui.ProgressNote().KeeplineUnder(1).Msg("Creating application resources")
 
-	_, err = c.API.AppRunning(appRef)
-	if err != nil {
-		return errors.Wrap(err, "waiting for app failed")
-	}
+	// _, err = c.API.AppRunning(appRef)
+	// if err != nil {
+	// 	return errors.Wrap(err, "waiting for app failed")
+	// }
```
jimmykarily, line 243:
```suggestion
```

internal/cli/usercmd/push.go
```diff
@@ -237,10 +237,10 @@ func (c *EpinioClient) Push(ctx context.Context, params PushParams) error { // n
...
```
jimmykarily, line 244:
```suggestion
```

jimmykarily: github doesn't let me select the additional empty line that needs to be deleted too :(

jimmykarily:
Would have been nice to have an application chart with `Statefulset` to test this (@agracey [promised one](https://github.com/epinio/epinio/issues/1629#issuecomment-1194032218)). Let's implement the test on another PR.

internal/api/v1/application/exec.go
andreas-kupries: Intermediate change done in commit 6b070caadb31c350cc874eb9fb7cb307562b0bbd

internal/api/v1/application/running.go
```diff
@@ -1,20 +1,34 @@
...
+//
+// __API PORTABILITY NOTE__
+//
+// With the switch to deployment of apps via helm, and waiting for the helm deployment to
+// be ready before returning this endpoint is technically superfluous, and it should never
+// fail. The command line has been modified to not invoke it anymore.
+//
+// It is kept for older clients still calling on it. Because of that it is also kept
+// functional. Instead of checking for an app `Deployment` and its status it now checks
+// the app `Pod` stati.
```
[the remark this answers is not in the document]
andreas-kupries: We can have more than one pod, thus the plural.

internal/application/workload.go
```diff
@@ -185,24 +156,53 @@ func (a *Workload) Replicas(ctx context.Context) (map[string]*models.PodInfo, er
...
+	var createdAt time.Time
+	var stageID string
+	var username string
+	var controllerName string
+
+	if len(podList) > 0 {
+		// Pods found. replace defaults with actual information.
+
+		// Initialize various pieces from the first pod ...
+		createdAt = podList[0].ObjectMeta.CreationTimestamp.Time
```
jimmykarily, line 184: This won't reflect reality anymore. A Deployment may be scaled up or down, causing Pods to appear and disappear. The age of any Pod may have nothing to do with the age of the "workload". That said, the age of the workload is probably nonsense. How old is the "workload" of an application that run 1 Pod for 3 hours, then was scaled down to 0 for an hour or two and then scaled back up to 3 Pods for 10 hours? If we imply that the age of the workload is the time since any Pod was created, then in most cases what we need is the time of creation of the application itself, unless the application was originally created without a workload and stayed like that for while. In that case though, I don't think the creation date of the first Pod is useful in any way.

Unless I'm missing something and this breaks the cli or the UI in a very bad way, I suggest we remove this field from the workload. This information should be provided for the individual Pods.

internal/application/workload.go
```diff
@@ -185,24 +156,53 @@ func (a *Workload) Replicas(ctx context.Context) (map[string]*models.PodInfo, er
...
+		stageID = podList[0].ObjectMeta.Labels["epinio.io/stage-id"]
```
jimmykarily, line 185: This can also be non-true. When pushing an application again, the old Pods co-exist with the new ones for some time (usually for very little). After a while things balance out but I'm not sure this won't lead to wrong assumptions in some cases (e.g. getting the old stage id assuming it's the new one). To suggest a better alternative, we would have to find all consumers. Maybe we can get this information from the application itself (as part of this? https://github.com/epinio/epinio/issues/1657)

internal/application/workload.go
andreas-kupries: I believe you missed the code later in the loop which merges the dates from all the pods, and takes the youngest.
Lines 190-193.

Regarding usage ... The cli uses it for the general `Age` information, see (***) below

```
|         KEY          |                    VALUE                     |
|----------------------|----------------------------------------------|
| Origin               | /home/work/SUSE/dev/epinio/assets/sample-app |
| Created              | 2022-07-26 15:01:14 +0200 CEST               |
| Status               | 1/1                                          |
| Username             | admin                                        |
| Running StageId      | e6b05fbad294fc4b                             |
| Last StageId         | e6b05fbad294fc4b                             |
| Age                  | 6m0s                                         | ***
| Active Routes        |                                              |
|                      | fox-a279f.172.19.0.4.omg.howdoi.website      |
| App Chart            | standard                                     |
| Desired Instances    |                                            1 |
| Bound Configurations |                                              |
| Environment          |                                              |
```

The cli also show per-pod age.
No idea about usage in the UI . @richard-cox ping

Note that a full removal of the field from the structure is an API change now.

For best compatibility we should keep the field, even if we decide to not fill it anymore, i.e. leave it at Go's zero value.

internal/application/workload.go
andreas-kupries: I see, like `desiredReplicas`.

jimmykarily: yes

internal/application/workload.go
jimmykarily: Yes I missed that part but my comment still stands, no Pod age is guaranteed to match a "Deployment age" whatever the latter means.

richard-cox: I might have got confused with the terminology here, but the UI presents app, deployment and 'instances' info as below

image

- App age at the top (from `meta.createdAt`)
- No deployment age (though there would be value in showing, from `deployment.createdAt`)
- 'Instances' Age per row. (from `deployment.replicas` and their `createdAt` property)

I would be nice to keep all of the above, unless the replicaSet age won't differ between entries

andreas-kupries: @richard-cox 
The `deployment age` was from the `Deployment` resource of the app. With us not knowing anymore if the top controller is a `Deployment`/`StatefulSet`/other I am currently trying to synthesize something sensible from the App Pods (I take the ... oldest Pod as the reference) ... @jimmykarily only saw me taking the first Pod.

agracey:
Here's an example chart that should work: 
https://github.com/epinio/helm-charts/tree/additional_application_charts/chart/stateful-appliction 

It's packaged up at: 
https://chameleons.sfo3.digitaloceanspaces.com/epinio-stateful-application-0.1.15.tgz

Let me know if there's any issue with it. I haven't tried it recently so will test shortly once I spin up an environment

andreas-kupries:
Local manual check of the failed ops is ok. Local CI focused on the failing test was ok also. Retrying main CI now.

internal/application/workload.go
andreas-kupries: > match a "Deployment age" whatever the latter means.

The "Deployment age" was literally the `Deployment`'s `MetaData.CreatedAt`.

Does the go kube client stuff have a general way of getting a resource's meta data without knowing the kind of the resource ? I.e. just from the name+namespace ?

If yes, then we can get rid of the code synthesizing something from the pods in favor of pulling from the controller resource again. Note, at that point the new `epinio.io/app-container` label is also important as naming the controller resource.
We could of course separate this and have a `epinio.io/controller` label as well.

andreas-kupries: ... Oh, it might be do-able via `unstructured`

jimmykarily: There is a way to discover what controller the Pod belongs too but I'm not sure if it's guaranteed (e.g. what happens if the application chart simply creates Pods without Deployments or Statefulsets?).

The Pod has an owner field. In the case of Deployments that is a ReplicaSet. The ReplicaSet has an owner, which is the Deployment. I assume there is a similar chain of owners for a Statefulset. We may not even have to go that up, maybe the createdAt field of the first Owner in the chain is enough.

If there is no owner (e.g. a custom application chart creates just Pods for whatever reason), we could opt for the current solution.

I wonder if it's too complicated logic, to get information that makes little sense imho.

andreas-kupries: Or not ... Seems even the dynamic client wants me to tell it the kind of resource ahead of time ... I guess that is because a resource is identified uniquely only by all three of (name, namespace, and kind).

In that case case the chart would have to tell Epinio the kind, another label on the pods.

jimmykarily (approved):

internal/application/workload.go
andreas-kupries: Ok. Stopping here. Putting this all on #1657 ...

andreas-kupries:
> Here's an example chart that should work: https://github.com/epinio/helm-charts/tree/additional_application_charts/chart/stateful-appliction
> 
> It's packaged up at: https://chameleons.sfo3.digitaloceanspaces.com/epinio-stateful-application-0.1.15.tgz
> 
> Let me know if there's any issue with it. I haven't tried it recently so will test shortly once I spin up an environment

New ticket #1658 to handle this.

PR #1659, opened 2022-07-27
Added SecretTypes annotation to Services

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1660
---

To label the secrets as Configurations we are filtering them with labels and by Opaque type. Our secrets were created with Helm, and so we had also the `"app.kubernetes.io/managed-by": "Helm"` label.

```go
secretSelector := labels.Set(map[string]string{
	"app.kubernetes.io/managed-by": "Helm",
	"app.kubernetes.io/instance":   names.ServiceHelmChartName(name, namespace),
}).AsSelector()

listOptions := metav1.ListOptions{
	FieldSelector: "type=Opaque",
	LabelSelector: secretSelector.String(),
}
```

With the Crossplane integration we have seen that the generated secrets are of `connection.crossplane.io/v1alpha1` type, and so they were not labeled as configuration correctly.

This PR relax this filter, adding the possibility to add multiple and different Secret types.

```go
secretSelector := labels.Set(map[string]string{
	"app.kubernetes.io/instance": names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace),
}).AsSelector()

secretTypes := "Opaque"
if len(service.SecretTypes) > 0 {
	secretTypes = strings.Join(service.SecretTypes, ",")
}

listOptions := metav1.ListOptions{
	FieldSelector: "type=" + secretTypes,
	LabelSelector: secretSelector.String(),
}
```

The operator just need to add them as an annotation to the Service Catalog:

```yaml
kind: Service
metadata:
  name: dynamodb-table
  namespace: epinio
  annotations:
    application.epinio.io/catalog-service-secret-types: connection.crossplane.io/v1alpha1
```

WIP :construction: : needs tests

### DIFF
--- a/internal/configurations/configurations.go
+++ b/internal/configurations/configurations.go
@@ -12,6 +12,7 @@ import (
 	"context"
 	"errors"
 	"reflect"
+	"strings"
 
 	"github.com/epinio/epinio/helpers/kubernetes"
 	epinioerrors "github.com/epinio/epinio/internal/errors"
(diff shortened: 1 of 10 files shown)


enrichman:
After a bit of investigation it seems that the field selector does not support the `in` operator, and so I'm going to fetch the secrets by label, and then filter the types manually on the application side. It should be fine since we are not going to have many secrets for a single service.

https://github.com/kubernetes/kubernetes/issues/32946

internal/configurations/configurations.go
```diff
@@ -234,61 +234,85 @@ func (c *Configuration) GetSecret(ctx context.Context) (*v1.Secret, error) {
...
 
 // ForService returns a slice of Secrets matching the given Service.
-func ForService(ctx context.Context, kubeClient *kubernetes.Cluster, namespace, name string) ([]v1.Secret, error) {
+func ForService(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
 	secretSelector := labels.Set(map[string]string{
-		"app.kubernetes.io/managed-by": "Helm",
-		"app.kubernetes.io/instance":   names.ServiceHelmChartName(name, namespace),
-		ConfigurationLabelKey:          "true",
-		ConfigurationTypeLabelKey:      "service",
+		"app.kubernetes.io/instance": names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace),
```
jimmykarily, line 239: Crossplane sets this label to the correct value on the Secrets, right?

jimmykarily (approved):

jimmykarily:
@enrichman  If we merge this, make sure you create an issue to document this.

internal/configurations/configurations.go
enrichman: Yes, you can add any label or annotations on the secrets, so you just need to add them in your Epinio Service:

https://github.com/enrichman/helm-charts/blob/75093b8734bb6c50ff2f13e918038ca56ceb169e/charts/dynamodb-test/templates/dynamodbtable.yaml#L9-L13

Issue #1658, opened 2022-07-26
Add stateful app chart to epinio


andreas-kupries:
https://github.com/epinio/epinio/pull/1652#issuecomment-1195497825

Add the referenced app chart, or something modified fixing issues with it to epinio.
At least for use in testing, to ensure that #1652 works for #1629.
Maybe also as a new standard chart.

~Note: We have this as branch of the helmchart repo ( https://github.com/epinio/helm-charts/tree/additional_application_charts/chart/stateful-appliction ). Not a PR yet. The branch would however be a starting point.~

~Edit: PR is https://github.com/epinio/helm-charts/pull/175, with more app charts than just stateful. Rebased that PR to latest main. TODO comparing the charts to current state (were created in April, between 0.8.1 and 0.9.0 tags).~

Proper PR now is https://github.com/epinio/helm-charts/pull/255. That PR limits itself to the new `stateful` app chart, and leaves the others behind.

~Packaged at https://chameleons.sfo3.digitaloceanspaces.com/epinio-stateful-application-0.1.15.tgz~ Ignore this.

andreas-kupries:
Indeed the new test failing, plus an existing cli test ...
The API test failures are around listing app charts, i.e. they have to be adapted for the additional chart.

andreas-kupries:
Regarding the new test ... Looks to be a server side issue, with the app not coming up properly ... Needs server logs I believe ... First use of

> acceptance/helpers/machine/machine.go:183:func (m *Machine) SaveServerLogs(destination string) {

since that helper was added to the test setup.

andreas-kupries:
Server log contains:

> " "report"="StatefulSet is not ready: namespace-184213286/rapps-638969828-57dab137b6cc0d487f9a0bfcbae58c0e2fd58ad5. 0 out of 1 expected pods have been scheduled"

And now I wonder ... The stateful chart allocates a volume to the app, of size 1Gi, for its state.
If creation of that volume fails then the pod will not be scheduled, waiting for it.

andreas-kupries:
Falsified ...

Looking at the STS events note
[...]

PR #1662, opened 2022-07-28
Add stateful app chart

andreas-kupries:
fix: #1658 

Import https://github.com/epinio/helm-charts/pull/255, plus new test

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -449,14 +449,48 @@ spec:
 	})
 
 	When("pushing as a stateful app", func() {
+		BeforeEach(func() {
+			// Create a custom chart referencing the tarball of the `standard-stateful` chart.
+			// It exists in the set of releases for helm charts.
+			// It is not distributed with epinio however.
+			// At this point in time we use it only internally, for testing.
+
+			chartName = catalog.NewTmpName("chart-")
+			tempFile = catalog.NewTmpName("chart-") + `.yaml`
+			err := os.WriteFile(tempFile, []byte(fmt.Sprintf(`apiVersion: application.epinio.io/v1
+kind: AppChart
+metadata:
+  namespace: epinio
+  name: %s
+  labels:
+    app.kubernetes.io/component: epinio
+    app.kubernetes.io/instance: default
+    app.kubernetes.io/name: epinio-standard-stateful-app-chart
+    app.kubernetes.io/part-of: epinio
+spec:
+  shortDescription: Epinio standard stateful deployment
+  description: Epinio standard support chart for stateful application deployment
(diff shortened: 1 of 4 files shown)


acceptance/apps_test.go
```diff
@@ -447,6 +448,31 @@ spec:
...
+			Expect(err).ToNot(HaveOccurred(), pushLog)
+
+			Eventually(func() string {
+				out, err := env.Epinio("", "app", "list")
+				Expect(err).ToNot(HaveOccurred(), out)
+				return out
+			}, "5m").Should(
+				HaveATable(
+					WithHeaders("NAME", "CREATED", "STATUS", "ROUTES", "CONFIGURATIONS", "STATUS DETAILS"),
+					WithRow(appName, WithDate(), "1/1", appName+".*", "", ""),
```
jimmykarily, line 470: Should we check that what was deployed is really StatefulSets and not Deployments? Just in case `standard-stateful` flag has no effect.

andreas-kupries: Agreed, would be better. Working on that now.

andreas-kupries: Done, commit 2bcab59812aeb6c634f7576252c9d9394de01639

acceptance/apps_test.go
```diff
@@ -470,6 +470,12 @@ spec:
 					WithRow(appName, WithDate(), "1/1", appName+".*", "", ""),
 				),
 			)
+
+			out, err := proc.Kubectl("get", "statefulset",
+				"--namespace", namespace,
+				"--selector=app.kubernetes.io/name="+appName,
+				"-o", "jsonpath={.items[*].spec.template.metadata.labels}")
+			Expect(err).NotTo(HaveOccurred(), out)
```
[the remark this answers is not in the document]
andreas-kupries: Working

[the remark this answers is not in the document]
andreas-kupries: Done, commit 8e3e3dc7bbdd7fa62aee96e7627a7b8fa5edbe87

jimmykarily (approved):

andreas-kupries:
@jimmykarily Thanks for the approval - Did you check https://github.com/epinio/helm-charts/pull/255 also ?

jimmykarily:
> @jimmykarily Thanks for the approval - Did you check [epinio/helm-charts#255](https://github.com/epinio/helm-charts/pull/255) also ?

I checked it briefly. It looks to be mostly a copy of the standard chart, only swapping out Deployments for StatefulSets, is that right? I'm a bit concerned about maintaining these 2 almost identical charts. I'm not sure if there is any simple way to DRY them.

andreas-kupries:
While it is mostly a copy it is not quite either. It has some additional pieces (headless service).
The main reason for the ping was if I could get an approval for it.

jimmykarily (approved):

PR #1677, opened 2022-08-08
1664 architectural diagram

### LINKED ISSUE #1664: Epinio architectural diagram
We need an detailed architectural diagram for Epinio to be used as documentation. 

I started looking at [c4model](https://c4model.com/) through [c4builder](https://adrianvlupu.github.io/C4-Builder/) as an option. In theory it's nice but I can't create a sensible diagram for Epinio. There are some entities that are not represented in a visually sensible way (e.g. Kubernetes itself). Will keep looking.

jimmykarily:
Fixes #1664


jimmykarily:
@agracey there is end to the level of detail we can go. I'm wondering if the image in this PR is already enough. I find the other 3 diagram levels of the c4 model less interesting for our case. Let me know what you think.

agracey:
LGTM! Thank you for this

andreas-kupries (approved):

PR #1678, opened 2022-08-08
Moved Username to annotation

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1679
---

The username was used in some labels, and when using an email then the deploy of an app was going to fail.
This PR moves the username to an annotation, also bumping the helm-charts to the new app chart: https://github.com/epinio/helm-charts/pull/259

Since we are going to use in the future an OIDC provider I've also relaxed the filter for the secret, looking not only for `BasicAuth` type.


internal/api/v1/application/stage.go
```diff
@@ -219,7 +219,7 @@ func (hc Controller) Stage(c *gin.Context) apierror.APIErrors {
 		S3ConnectionDetails: s3ConnectionDetails,
 		Stage:               models.NewStage(uid),
 		PreviousStageID:     previousID,
-		Username:            username,
+		Username:            user.ID,
```
jimmykarily, line 222: Maybe we shouldn't call this "Username" anymore then?

internal/api/v1/deploy/deploy.go
```diff
@@ -60,7 +61,7 @@ func DeployApp(ctx context.Context, cluster *kubernetes.Cluster, app models.AppR
 		Configurations: appObj.Configuration.Configurations,
 		Instances:      *appObj.Configuration.Instances,
 		ImageURL:       imageURL,
-		Username:       username,
+		Username:       user.ID,
```
jimmykarily, line 64: Hmm, changing `Username` here would require us to change the application chart too. I wonder if it's worth it.

internal/application/application.go
```diff
@@ -38,7 +38,7 @@ const EpinioApplicationAreaLabel = "epinio.io/area"
 // Create generates a new kube app resource in the namespace of the
 // namespace. Note that this is the passive resource holding the
 // app's configuration. It is not the active workload
-func Create(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef, username string, routes []string, chart string) error {
+func Create(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef, routes []string, chart string) error {
```
jimmykarily, line 41: I wonder if we should label the app custom resource with the id of the user that created it. Probably another PR.

internal/application/workload.go
```diff
@@ -183,9 +184,19 @@ func (a *Workload) Get(ctx context.Context) (*models.AppDeployment, error) {
...
 		stageID = podList[0].ObjectMeta.Labels["epinio.io/stage-id"]
-		username = podList[0].ObjectMeta.Labels["app.kubernetes.io/created-by"]
 		controllerName = podList[0].ObjectMeta.Labels["epinio.io/app-container"]
 
+		userID := podList[0].ObjectMeta.Labels["app.kubernetes.io/created-by"]
+		authService, err := auth.NewAuthServiceFromContext(ctx)
+		if err != nil {
+			return nil, err
+		}
+		user, err := authService.GetUserByID(ctx, userID)
```
jimmykarily, line 194: Previously, it was fine if the user was no longer there. I think it still is. E.g. A user creates a resource, the operators delete this user, someone tries to fetch the resource. There is no reason why this should fail. Maybe fallback to `[Deleted user]` or something?

jimmykarily (commented):
I wonder if we could avoid this whole PR if we simply hashed the username with a reversible hashing algorithm here: https://github.com/epinio/helm-charts/blob/75b8fae6baf7734aa37012d8b1377e9d8a903313/chart/application/templates/_helpers.tpl#L13

We could "unhash" it in Go space anytime we needed that. The problem was special characters in the name right? The length of the string is less of a problem, it's sane to have a limit on the lenght of usernames anyways.

acceptance/helpers/machine/users.go
```diff
@@ -58,7 +59,8 @@ func (m *Machine) CreateEpinioUser(role string, namespaces []string) (string, st
 
 // DeleteEpinioUser deletes the relevant Kubernetes secret if it exists.
 func (m *Machine) DeleteEpinioUser(username string) error {
-	out, err := proc.Kubectl("delete", "secret", "-n", "epinio", "epinio-user-"+username, "--ignore-not-found")
+	username = names.GenerateResourceName("ruser", "test", username)
+	out, err := proc.Kubectl("delete", "secret", "-n", "epinio", username, "--ignore-not-found")
```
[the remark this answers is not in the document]
enrichman: Done

internal/application/application.go
enrichman: Since it was a complex thing I've added it. :)

enrichman:
Moved the previous implementation to a Base58 encoding of the username.

acceptance/api/v1/application_export_test.go
```diff
@@ -60,8 +61,8 @@ var _ = Describe("AppPart Endpoint", func() {
   stageID: ""
   start: null
   tlsIssuer: epinio-ca
-  username: admin
-`, app)))
```
jimmykarily, line 64: Maybe this deserves a comment on why `base58` was chosen (over base64 for example)

go.mod
```diff
@@ -7,6 +7,7 @@ require (
 	github.com/alron/ginlogr v0.0.4
 	github.com/avast/retry-go v3.0.0+incompatible
 	github.com/briandowns/spinner v1.18.1
+	github.com/btcsuite/btcd/btcutil v1.1.1
```
jimmykarily, line 10: Started mining, are we? :D

internal/api/v1/application/importgit.go
```diff
@@ -131,9 +131,9 @@ func (hc Controller) ImportGit(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err, "creating an S3 manager")
 	}
 
-	username := requestctx.User(ctx).Username
+	user := requestctx.User(ctx)
 	blobUID, err := manager.Upload(ctx, tarball, map[string]string{
-		"app": name, "namespace": namespace, "username": username,
+		"app": name, "namespace": namespace, "username": user.Username,
```
jimmykarily, line 136: These changes are no longer needed? (this and anywhere where you switched to pass a User object instead of just the name).

internal/auth/auth_test.go
```diff
@@ -52,7 +52,7 @@ var _ = Describe("Auth users", func() {
 			})
 		})
 
-		When("kubernetes returns some user secrets", func() {
+		FWhen("kubernetes returns some user secrets", func() {
```
jimmykarily, line 55: oops

enrichman: :upside_down_face:

go.mod
enrichman: Ahah, in every k8s cluster we are going to be deployed!

internal/api/v1/application/importgit.go
jimmykarily: I'm not sure, it complicates this PR for no good reason. Maybe leave it out?

enrichman: Done!

jimmykarily (approved):

Issue #1663, opened 2022-07-29
AWS Application Load Balance


riqueps:
Hello there.
I'm deploying Epinio on AWS EKS.
But I need to use AWS WAF for Epinio and Apps endpoints for security reasons.
But when I deploy Epinio on EKS using traefik or nginx, it is just able to create Classic ELB, even when I add aws-loadbalancer annotations.

jimmykarily:
I'm not sure I'm reading the right docs but, I suppose you are using the [epinio's values](https://github.com/epinio/helm-charts/blob/c097eb4d9604018986daf827a38fe2bfe68b02b3/chart/epinio/values.yaml#L40) to set an annotation:

```
    kubernetes.io/ingress.class: alb
```
[as described here](https://aws.amazon.com/premiumsupport/knowledge-center/eks-alb-ingress-aws-waf/)

You say you deploy Epinio with traefik or nginx but afaict, the alb load balancer is supposed to be the Ingress controller. Is the alb load balancer installed and running on your cluster? Can you try to deploy Epinio without Traefik or nginx and let `alb` implement the Ingress? I've never used AWS WAF before so I could be completely off here.

riqueps:
@jimmykarily Epinio needs to be deployed as a nodePort in order for the ingress to be exposed using the aws load balancer controller. We modified the service helm chart but Epinio is stuck on a loop after changing it to NodePort. Do you know if Epinio supports node port deployment?

![Selection_001](https://user-images.githubusercontent.com/47371890/182233188-a1676d84-0b48-4299-86d9-887ffe9d3236.png)

![Selection_002](https://user-images.githubusercontent.com/47371890/182233304-8a543f92-0927-4e4e-b0dd-93f031fb7d44.png)

jimmykarily:
I don't fully understand your setup but let's start with what I know. The error message you posted indicates that a request landed on the epinio API server for this path: `/` . Such a route is not defined [in the router](https://github.com/epinio/epinio/blob/4809917451358f2a6d104e4ac2a5eac16324d05d/internal/api/v1/router.go#L75-L182) thus the error you see.

In the default Epinio setup (with the Ingress and the load balanced traefik service), the `epinio.your-system-domain-here` url points to the epinio-ui Ingress/service. It seems to me that in your setup this points to the epinio API server service and that's why you see this error.

[...]

PR #1685, opened 2022-08-17
Bump helm-charts to fix Ingress pathTypes

jimmykarily:
Fixes comment here: https://github.com/epinio/epinio/issues/1663#issuecomment-1206834066

There is no reason to use "ImplementationSpecific" when we know we want "Prefix". I'm not sure about the change in the order of rules. Most ingress controllers should prioritize rules, based on the length (e.g. https://kubernetes.github.io/ingress-nginx/user-guide/ingress-path-matching/#path-priority). In any case, it doesn't hurt putting the "catch all" rule at the end.

Sibling PR: https://github.com/epinio/helm-charts/pull/263

### DIFF
--- a/internal/api/v1/application/create.go
+++ b/internal/api/v1/application/create.go
@@ -14,6 +14,7 @@ import (
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
 	"github.com/gin-gonic/gin"
+	networkingv1 "k8s.io/api/networking/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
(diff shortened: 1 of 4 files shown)


jimmykarily:
[Changing the order of rules](https://github.com/epinio/helm-charts/pull/263/files) breaks the route validation because we assume there is only one rule in each ingress:

https://github.com/epinio/epinio/blob/080ad496418795d277e0a2331ff2ae3ff4509081/internal/routes/routes.go#L73


(debugging output from the epinio server [around here](https://github.com/epinio/epinio/blob/080ad496418795d277e0a2331ff2ae3ff4509081/internal/api/v1/application/create.go#L163)):

```
desiredRoutesMap = map[epinio.172.18.0.4.omg.howdoi.website:{}]                                                                                                                                                  ingressRoute = epinio.172.18.0.4.omg.howdoi.website/api 
```

jimmykarily:
Still one failure but it passes locally. It's a flake and unrelated to this PR. We can try to fix it on `main`.

internal/api/v1/application/create.go
```diff
@@ -154,15 +155,35 @@ func validateRoutes(ctx context.Context, cluster *kubernetes.Cluster, appName, n
...
+
+	if len(issues) > 0 {
+		return apierror.NewMultiError(issues)
+	}
+	return nil
+}
+
+// validateIngress checks if the desiredRoutesMap is in conflict with the passed
+// ingress object. Conflict means, the ingress already defines one of the desired
+// routes and it belongs to another or and unknown app.
```
andreas-kupries, line 172:
```suggestion
// routes and it belongs to another or an unknown app.
```

internal/api/v1/application/create.go
```diff
@@ -154,15 +155,35 @@ func validateRoutes(ctx context.Context, cluster *kubernetes.Cluster, appName, n
...
+		return append(issues, apierror.InternalError(err))
+	}
+
+	for _, route := range routes {
+		routeStr := route.String()
 
-		// if a desired route is present within the ingresses then we have to check if it's already owned by the same app
-		if _, found := desiredRoutesMap[ingressRoute]; found {
+		// if a desired route is present within the ingresses then we have to check
+		// if it's already owned by the same app
```
andreas-kupries, line 185:
```suggestion
		// if it is already owned by the same app
```

internal/routes/routes_test.go
```diff
@@ -63,10 +63,10 @@ var _ = Describe("Route", func() {
 											Path: "/api/v1",
 										}}}}}}}}
 		})
-		It("returns a Route object", func() {
+		It("returns a list of Route objects", func() {
 			result, err := FromIngress(routeIngress)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(*result).To(Equal(Route{Domain: "mydomain.org", Path: "/api/v1"}))
+			Expect(result[0]).To(Equal(Route{Domain: "mydomain.org", Path: "/api/v1"}))
```
andreas-kupries, line 69: Add `Expect len(result) == 1 ?`

andreas-kupries (changes requested):
A few small nits

internal/routes/routes_test.go
jimmykarily, line 69:
```suggestion
			Expect(result[0]).To(Equal(Route{Domain: "mydomain.org", Path: "/api/v1"}))
			Expect(len(result)).To(Equal(1))
```

jimmykarily: added

PR #1692, opened 2022-08-22
fix: the string based yaml creation, use proper marshalling to ensure proper quoting in all parts

### LINKED ISSUE #1691: Error parsing a stringify JSON, deployment
Copy from https://github.com/epinio/ui/issues/152
### Current Behavior
While trying to deploy an application and we're setting an `env` variable with a stringify object, the backend will return an error:
```
Server Backtrace: error converting YAML to JSON: yaml: line 3: did not find expected ',' or '}'
```

Request payload:
image

Tested in CLI with an example where it will work and fail:
image

### Expected Behavior

Should correctly parse the object.

### Steps To Reproduce

1. Have an application deployed
2. try to add a new env variable to it
```
epinio apps env set {{YOUR_APP}} foo "{\"usernameOrOrg\":"scures","url":"https://github.com/scures/epinio-sample-app","commit":"3ce7abe14abd849b374eb68729de8c71e9f3a927"}"
```

A server error will occur but the new variable will be set anyway with the correct format.

### More testing:
https://github.com/epinio/ui/issues/152#issuecomment-1222267129

### Relevant parts of the code
- https://github.com/epinio/epinio/blob/main/internal/helm/helm.go#L144-L168
- https://github.com/epinio/epinio/blob/main/internal/helm/helm.go#L94-L99
- https://github.com/epinio/epinio/blob/main/pkg/api/core/v1/models/env.go#L60-L68

andreas-kupries:
fix #1691 

todo: extended tests

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -86,51 +86,51 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
 	// `values.yaml` to hand to helm from the chart parameters.
 
 	type routeParam struct {
-		id     string `yaml:"id"`
-		domain string `yaml:"domain"`
-		path   string `yaml:"path"`
-		secret string `yaml:"secret,omitempty"`
+		Id     string `yaml:"id"`
+		Domain string `yaml:"domain"`
+		Path   string `yaml:"path"`
+		Secret string `yaml:"secret,omitempty"`
 	}
 	type epinioParam struct {
-		appName        string               `yaml:"appName"`
-		env            []models.EnvVariable `yaml:"env"`
-		imageUrl       string               `yaml:"imageURL"`
-		ingress        string               `yaml:"ingress,omitempty"`
-		replicaCount   int32                `yaml:"replicaCount"`
-		routes         []routeParam         `yaml:"routes"`
-		configurations []string             `yaml:"configurations"`
-		stageID        string               `yaml:"stageID"`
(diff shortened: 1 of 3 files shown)


andreas-kupries:
Ignore the last failure ... It is the unrelated test we already bumped the timeout for to two minutes. Which is apparently still not enough, sometimes.

internal/helm/helm.go
```diff
@@ -82,92 +82,93 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
...
+			// Ingress, Start, Routes: see below
+		},
 	}
 
-	environment := `[]`
-	if len(parameters.Environment) > 0 {
-		// TODO: Simplify the chain of conversions. Single `AsYAML` ?
-		environment = fmt.Sprintf(`[ %s ]`, strings.Join(parameters.Environment.List().Assignments(),
-			","))
+	// TODO: Is this properly nulled if the class is not set ?
```
jimmykarily, line 127: Looking at my local deployment of Epinio, `INGRESS_CLASS_NAME` is not set, neither is `--ingress-class-name`  and I see `null` in the values.yaml when I export an app. I guess we can remove the TODO.

jimmykarily (dismissed):

internal/helm/helm.go
andreas-kupries: done: ee24af8df5a42ded79da7556a09482872adc6980

jimmykarily (approved):

PR #1694, opened 2022-08-24
212 e2e tests fix release recreation

jimmykarily:
Fixes https://github.com/epinio/epinio-end-to-end-tests/issues/212

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -273,7 +273,9 @@ func cleanupReleaseIfNeeded(l logr.Logger, c hc.Client, name string) error {
 
 	if r.Info.Status != helmrelease.StatusDeployed {
 		l.Info("Will remove existing release with status: " + string(r.Info.Status))
-		err := c.UninstallRelease(&hc.ChartSpec{ReleaseName: name})
+		err := c.UninstallRelease(&hc.ChartSpec{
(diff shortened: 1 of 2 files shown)


internal/helm/helm.go
```diff
@@ -251,3 +256,30 @@ func GetHelmClient(restConfig *rest.Config, logger logr.Logger, namespace string
...
+func cleanupReleaseIfNeeded(l logr.Logger, c hc.Client, name string) error {
+	r, err := c.GetRelease(name)
+	if err != nil {
+		if err == helmdriver.ErrReleaseNotFound {
+			return nil
+		}
+		return errors.Wrap(err, "getting the helm release")
+	}
+
+	if r.Info.Status != helmrelease.StatusDeployed {
```
jimmykarily, line 274: This is the part I'm not sure about. There is no "sure" way to force a re-deployment of a helm release. Some states of the release won't let helm deploy on top no matter what (e.g. "pending install"). In those cases, one has to remove the release and deploy again (like we do here). I chose to only keep the release if it is in "deployed" state. I wonder if the history of the release is relevant in any way and be removing it we loose something. Currently we don't seem to make any use of the history.

andreas-kupries (approved):

PR #1697, opened 2022-08-25
Return the default builder image as part of the /info response

jimmykarily:
Part of the fix for epinio/ui#154

### DIFF
--- a/acceptance/api/v1/info_test.go
+++ b/acceptance/api/v1/info_test.go
@@ -1,0 +1,34 @@
+package v1_test
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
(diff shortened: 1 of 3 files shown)


jimmykarily:
The fix relies on the [this to be set](https://github.com/epinio/helm-charts/blob/3ece8553d285058bd2cfa7feb62b9e57a51f3b86/chart/epinio/templates/stage-scripts.yaml#L7) but [we rely on the same thing for staging anyway](https://github.com/epinio/epinio/blob/51f702afd715b0814f312f34d88d8cb78a20bb90/internal/api/v1/application/stage.go#L135).

andreas-kupries (approved):

PR #1710, opened 2022-09-01
Remove default route's hash suffix

### LINKED ISSUE #1698: App URL with a279f postfix (app-a279f.domain.com)
Every time I push an app, the app url is created with a postfix. eg: app-a279f.domain.com

Epinio version: 1.1.0
Environment: AWS EKS
Ingress: nginx

[Install epinio on EKS](https://docs.epinio.io/next/howtos/install_epinio_on_eks) to reproduce this behavior.

image

andreas-kupries:
fix #1698
drop md5-based suffix from default route

### DIFF
--- a/acceptance/api/v1/application_deploy_test.go
+++ b/acceptance/api/v1/application_deploy_test.go
@@ -171,7 +171,7 @@ var _ = Describe("AppDeploy Endpoint", func() {
 				deployResponse := deployApplication(appName, namespace, deployRequest)
 
 				// check also the random suffix for the default route: '-[a-z0-9]{5}'
-				Expect(deployResponse.Routes[0]).To(MatchRegexp(appName + `-.*\.omg\.howdoi\.website`))
+				Expect(deployResponse.Routes[0]).To(MatchRegexp(appName + `\..*\.omg\.howdoi\.website`))
 
 				Eventually(func() string {
 					return appFromAPI(namespace, appName).Workload.Status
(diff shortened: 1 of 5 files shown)


andreas-kupries:
CI fail remaining is the known flaky thing

acceptance/api/v1/application_deploy_test.go
```diff
@@ -171,7 +171,7 @@ var _ = Describe("AppDeploy Endpoint", func() {
 				deployResponse := deployApplication(appName, namespace, deployRequest)
 
 				// check also the random suffix for the default route: '-[a-z0-9]{5}'
```
jimmykarily, line 173: remove the comment too?

andreas-kupries: I should, yes

andreas-kupries: Done 1dbc3a45c0dd90bb67c50e92a15e88837cc76ba5

jimmykarily:
While trying to verify that route validation works correctly, I found a "path" that is sub-optimal:

1. `epinio app create sample`
2. `epinio namespace create workspace2`
3. `epinio target workspace2`
4. `epinio app create sample`

(at this point and because no Ingress exists yet, no validation error was fired)

5. `epinio app push -n sample -p assets/sample-app`
6. `epinio target workspace`
7. `epinio app push -n sample -p assets/sample-app`

Eventually, a validation error was returned, but only after staging has happened. This is because we validate routes [on app creation](https://github.com/epinio/epinio/blob/5925739bbc754b47f4a83848db68101dc3aa36a8/internal/api/v1/application/create.go#L89) and [on app deployment](https://github.com/epinio/epinio/blob/5925739bbc754b47f4a83848db68101dc3aa36a8/internal/api/v1/application/deploy.go#L67) but not before we stage an application.
This is an edge case and the end result is still a validation error, so I would say it's ok to keep it as it is.

jimmykarily (approved):

PR #1735, opened 2022-09-12
Block `configuration unbind|delete` for service-owned configurations

### LINKED ISSUE #1733: Unbinding Configuration belonging to a Service should not be allowed
### Issue:
Test done on `v1.2.0-8-g969cfcf7`
Unbinding a configuration belonging to a service command is currently possible.
The service will still appear as bound. The system becomes inconsistent.

```
🚢  Listing applications
Namespace: workspace

✔️  Epinio Applications:
|  NAME   |            CREATED             | STATUS | ROUTES |           CONFIGURATIONS            | STATUS DETAILS |
|---------|--------------------------------|--------|--------|-------------------------------------|----------------|
| testapp | 2022-09-12 09:31:16 +0200 CEST | n/a    | n/a    | x33d052a64bf8afeedb501b9dcf7c-mysql |                |

➜  epinio git:(main) ✗ epinio configuration unbind x33d052a64bf8afeedb501b9dcf7c-mysql testapp

🚢  Unbind Configuration from Application
Configuration: x33d052a64bf8afeedb501b9dcf7c-mysql
Application: testapp
Namespace: workspace

✔️  Configuration Detached From Application.
Configuration: x33d052a64bf8afeedb501b9dcf7c-mysql
Application: testapp
Namespace: workspace
```

#### Note: the application seems still bound to the service:
```
➜  epinio git:(main) ✗ epinio service list                                                    

🚢  Listing Services...
Namespace: workspace

✔️  Details:
|        NAME        |            CREATED             | CATALOG SERVICE | VERSION |  STATUS  | APPLICATIONS |
|--------------------|--------------------------------|-----------------|---------|----------|--------------|
| mycustomsqlservice | 2022-09-12 09:30:11 +0200 CEST | mysql-dev       | 8.0.29  | deployed | testapp      |
```

### Expected behavior:

Unbinding of such configurations should be prevented.

andreas-kupries:
Fix #1733

### DIFF
--- a/acceptance/api/v1/configurations_mutation_test.go
+++ b/acceptance/api/v1/configurations_mutation_test.go
@@ -555,7 +555,6 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 
 		It("returns a 'not found' when the application does not exist", func() {
 			// This requires a valid configuration
-			configuration := catalog.NewConfigurationName()
 			env.MakeConfiguration(configuration)
 
 			response, err := env.Curl("DELETE",
--- a/acceptance/api/v1/configurations_mutation_test.go
+++ b/acceptance/api/v1/configurations_mutation_test.go
@@ -554,6 +554,10 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 		})
 
 		It("returns a 'not found' when the application does not exist", func() {
+			// This requires a valid configuration
+			configuration := catalog.NewConfigurationName()
+			env.MakeConfiguration(configuration)
+
 			response, err := env.Curl("DELETE",
 				fmt.Sprintf("%s%s/namespaces/%s/applications/bogus/configurationbindings/%s",
 					serverURL, api.Root, namespace, configuration),
@@ -569,6 +573,8 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 			json.Unmarshal(bodyBytes, &responseBody)
 			Expect(responseBody["errors"][0].Title).To(
 				Equal("application 'bogus' does not exist"))
+
+			env.CleanupConfiguration(configuration)
 		})
 
 		Context("with application", func() {
--- a/acceptance/api/v1/configurations_mutation_test.go
+++ b/acceptance/api/v1/configurations_mutation_test.go
@@ -556,6 +556,7 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 		It("returns a 'not found' when the application does not exist", func() {
 			// This requires a valid configuration
 			env.MakeConfiguration(configuration)
+			defer env.CleanupConfiguration(configuration)
 
 			response, err := env.Curl("DELETE",
 				fmt.Sprintf("%s%s/namespaces/%s/applications/bogus/configurationbindings/%s",
@@ -572,8 +573,6 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 			json.Unmarshal(bodyBytes, &responseBody)
 			Expect(responseBody["errors"][0].Title).To(
 				Equal("application 'bogus' does not exist"))
-
-			env.CleanupConfiguration(configuration)
 		})
 
 		Context("with application", func() {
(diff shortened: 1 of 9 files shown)


internal/api/v1/configurationbinding/deletebinding.go
```diff
@@ -10,7 +10,7 @@ import (
 	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
 )
 
-func DeleteBinding(ctx context.Context, cluster *kubernetes.Cluster, namespace, appName, configurationName, username string) apierror.APIErrors {
+func DeleteBinding(ctx context.Context, cluster *kubernetes.Cluster, namespace, appName, configurationName, username string, fromService bool) apierror.APIErrors {
```
jimmykarily, line 13: In my opinion, this function has too many responsibilities. I think it shouldn't be responsible to check if the binding should be deleted or not (checking the origin and this additional boolean flag here). It should simply delete the binding. The caller should be responsible to decide whether such a deletion should happen or not.

The caller will need to load the configuration object to be able to check the `config.Origin` value of course but that loading won't need to happen again here, since this is only used for that check (which will now live elsewhere). Even if the configuration object was needed here, we could as well pass it instead of passing just the configuration name.

The reason for my suggestion is:

- keep this function to a single responsibility
- avoid the boolean parameter to keep the code more readable

(relevant random google result: https://understandlegacycode.com/blog/what-is-wrong-with-boolean-parameters/)

internal/cli/usercmd/configuration.go
```diff
@@ -210,16 +210,25 @@ func (c *EpinioClient) DeleteConfiguration(name string, unbind bool) error {
...
 
 			var apiError apierrors.ErrorResponse
 			if err := json.Unmarshal(bodyBytes, &apiError); err != nil {
 				return err
 			}
 
+			// [BELONG] keep in sync with same marker in the server
+			if strings.Contains(apiError.Errors[0].Title, "Configuration belongs to service") {
+				// (2.)
+				return err
```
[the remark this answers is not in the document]
jimmykarily: sorry, my bad. It's not that.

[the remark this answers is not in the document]
andreas-kupries: Huh. Somehow it worked. I will see what happens when I return the apiError instead ...

internal/api/v1/configurationbinding/deletebinding.go
andreas-kupries: Will look into refactoring the code.

internal/cli/usercmd/configuration.go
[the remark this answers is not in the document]
jimmykarily: It's not the unmarshalling error (different scope). It's the argument passed to this function). This code is hard to understand.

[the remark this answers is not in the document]
andreas-kupries: Commit e476e984ab972441b7a9027a358c7e68a0c467ed

internal/api/v1/configurationbinding/deletebinding.go
andreas-kupries: Commit 4032525db1f69d2e54eaca68e50a578efe37adf9

jimmykarily (approved):

acceptance/api/v1/configurations_mutation_test.go
```diff
@@ -554,6 +554,10 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 		})
 
 		It("returns a 'not found' when the application does not exist", func() {
+			// This requires a valid configuration
+			configuration := catalog.NewConfigurationName()
```
jimmykarily, line 558: This is already called in the `BeforeEach` so it's not needed here (?)

acceptance/api/v1/configurations_mutation_test.go
```diff
@@ -569,6 +573,8 @@ var _ = Describe("Configurations API Application Endpoints, Mutations", func() {
 			json.Unmarshal(bodyBytes, &responseBody)
 			Expect(responseBody["errors"][0].Title).To(
 				Equal("application 'bogus' does not exist"))
+
+			env.CleanupConfiguration(configuration)
```
jimmykarily, line 577: `defer` this?

acceptance/api/v1/configurations_mutation_test.go
andreas-kupries: Done. 60a1964edefdd9b1e343980508f61cfd3067a1f7

acceptance/api/v1/configurations_mutation_test.go
andreas-kupries: Done. e5133a021247089698f52f839b0ef9936ec8390e

jimmykarily (approved):

PR #1741, opened 2022-09-15
Trial a `bci-micro` derived image as replacement for `library/bash`.

andreas-kupries:
Coming out of https://github.com/rancher/image-mirror/pull/286/files#r971150514

import helm chart work on the image for staging's unpack step.
trialing an image derived from bci-micro

PR #1747, opened 2022-09-16
added workflow to build and publish the epinio-unpacker image supporting the unpack step of epinio staging

andreas-kupries:
ref: #1741 

chore: workflow to build and publish the docker image for the epinio unpack step of staging

### DIFF
--- a/.github/workflows/unpacker-image.yml
+++ b/.github/workflows/unpacker-image.yml
@@ -1,0 +1,34 @@
+name: build-publish-unpacker-image
+on:
+  workflow_dispatch:
+jobs:
+  build-and-publish:
+    runs-on: self-hosted
+    steps:
+      - name: Checkout Epinio
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@v2
+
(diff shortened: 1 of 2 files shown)


.github/workflows/unpacker-image.yml
```diff
@@ -0,0 +1,34 @@
...
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push unpacker
+        uses: docker/build-push-action@v3
+        with:
+          push: true
+          context: images/.
+          file: images/unpacker-Dockerfile
```
andreas-kupries, line 31: Note: I looked into the [documentation of this action](https://github.com/docker/build-push-action#inputs) and saw no indication that `file` is relative to the `context`. Only the default for `file` is, i.e. when not specified.

jimmykarily: ok let's merge this and we will try to fix it on a branch (where we will be able to run the pipeline)

jimmykarily (approved):

.github/workflows/unpacker-image.yml
andreas-kupries: my thought as well

PR #1749, opened 2022-09-19
Upgrade - Test service creation

### LINKED ISSUE #1704: Bug upgrading Epinio from 1.0.0 to 1.1.0
**Describe the bug**
I'm unable to deploy any service after upgrading my Epinio instance from 1.0.0 to 1.1.0

Tried to deploy using CLI and UI, both return the same error
Tried to deploy using 1.0 and 1.1 versions of CLI, both return the same error
Tried deleting all Epinio CRDs manually, doesn't work either

**To Reproduce**

1. Install an Epinio instance on version 1.0.0
2. Upgrade to 1.1.0 (`helm upgrade epinio -n epinio epinio/epinio --version 1.1.0`)
3. Try to deploy any service (`epinio service create mysql-dev test-db`)
4. CLI returns `❌  error creating service: service create failed: Internal Server Error: error creating helm chart: create not allowed while custom resource definition is terminating`

**Logs**

On Epinio server pod logs, the return is the same: `"msg"="Error" "error"="error creating helm chart: create not allowed while custom resource definition is terminating"`

Describing the CRDs that Epinio creates for Kubernetes does not return any problems, all of them return status active an with no conflicts

**Cluster**
 - Provider: EKS
 - Kubernetes Version: 1.22
 - Load balancer: Nginx

**Additional context**
Already opened an [issue](https://github.com/epinio/epinio/issues/1702) with other problems related to the Epinio upgrade, but those were all visual and not impeditive, as with this one, I can't progress with upgrading my stack because I'm not able to deploy any service on 1.1.0

andreas-kupries:
Ref #1704 ... Extended RKE/GKE/... testing, with upgrade checks ... Reproduced the issue ... The #1764 spike tested if getting rid of helm controller in favor of direct deployment will help ... This was confirmed.

### DIFF
--- a/.github/workflows/aks-upgrade.yml
+++ b/.github/workflows/aks-upgrade.yml
@@ -1,0 +1,168 @@
+name: AKS-CI-UPGRADE
+
+on:
+  workflow_dispatch:
+    inputs:
+      azure_credentials:
+        description: "AZURE_CREDENTIALS"
+        required: false
+        default: ""
+      aks_domain:
+        description: "AKS_DOMAIN to use, managed via Route53's AWS_ZONE_ID"
+        required: false
+        default: ""
+      aws_zone_id:
+        description: "AWS_ZONE_ID"
+        required: false
+        default: ""
+      keep_cluster:
+        description: "Keep the cluster afterwards? (empty/yes)"
+        required: false
+        default: ""
+
+# https://github.com/npm/node-semver#x-ranges-12x-1x-12-
+env:
+  SETUP_GO_VERSION: '1.18.x'
+  GINKGO_NODES: 1
+  FLAKE_ATTEMPTS: 1
+  PUBLIC_CLOUD: 1
+  KUBECONFIG_NAME: 'kubeconfig-epinio-ci'
+  AWS_REGION: 'us-east-2'
+  AKS_MACHINE_TYPE: 'Standard_D3_v2'
+
+jobs:
+  linter:
+    runs-on: self-hosted
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Setup Go
+        uses: actions/setup-go@v3
+        with:
+          go-version: ${{ env.SETUP_GO_VERSION }}
+
+      - name: Cache Tools
+        uses: actions/cache@v3.0.8
+        with:
+          path: ${{ github.workspace }}/tools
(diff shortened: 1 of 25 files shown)


andreas-kupries:
The latest failure in the upgrade test is before it reaches my changes - MetalLB fails to install on account of trouble with a web hook. I am now letting this leave fallow a bit, in case it is rate-limiting thing.

Work on the other public cloud CI flow instead

andreas-kupries:
Reproduction of original issue achieved in GKE.

andreas-kupries:
I am beginning to realize a difference between application deployment with helm charts vs service deployment with helm charts.
Which may be the dividing line between works (apps) and not (services) after upgrading.

As in, wanting to extend the server code to possibly get more data than just the error messages seen in the CI I see that this code seems to __post__ a HelmChart specification into the namespace and then lets __something else__ take over to actually __launch__ the chart (helm controller ?). (And deletion is removing the HC spec).

The application deployment on the other hand uses an __in-server__ helm client (`github.com/mittwald/go-helm-client`) to __directly deploy__ the helm chart (referenced by the app chart) by itself.

And where is the helm controller (HC) on k3s ? k3s integrated (because I see no chart installed) ? Which is a difference to the public cloud setups, where the HC is installed as a separate helm chart and controller. Which might explain the other divide between working (k3s) and not (others).

@epinio/epinio What are the thoughts about reworking the service deployment to also use the mittwald client instead of going through HC ? What were the initial reasons for using/going through HC ?

andreas-kupries:
Ah. To be a bit more precise

> Apps: CRD (class) :arrow_right: CR (instance, config & state) :arrow_right:  HC (active workload)
> Service: CRD (class) :arrow_right: CR (catalog service) :arrow_right: HCspec (service/instance) :arrow_right: HC (active service pods)

The `HCspec` being a CR for a helm-controller CRD. And the HCspec's under epinio's control identified by having the label "application.epinio.io/catalog-service-name", which is the back pointer to the catalog service. Two additional labels given us the catalog "service version", and "secret types" (Ah! `epinio/docs/howtos/create_custom_service.md` explains it. Ok).

It should be possible to use a suitably labeled and annotated secret (or config map) as service instance instead of the HCspec.

andreas-kupries:
The #1764 spike tested if getting rid of helm controller in favor of direct deployment will help ... This was confirmed.
Closing this PR now, without merging. It has fulfilled its purpose.

PR #1762, opened 2022-09-29
1153 cli check new version

### LINKED ISSUE #1153: Upgrade checks for both server and CLI

It would be very useful to let someone using the Epinio CLI to know that they should update their client when the server gets updated. It would also be useful to an admin to show when an update to Epinio as a whole can be upgraded.

A MOTD-like output when there is a mismatch in client/server or an upgrade available to the server. This could also be shown in the UI as well. 

Extra points for a way to let the client update itself in-place.

jimmykarily:
Fixes #1153


internal/application/scale.go
```diff
@@ -33,7 +33,7 @@
 		result = 1
 	}
 
-	return int32(result), nil
+	return int32(result), nil // nolint:gosec // overflow blocked by guards
```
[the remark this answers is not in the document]
jimmykarily: hmm...there is an upper bound check right above, I wish you could detect that CodeQL!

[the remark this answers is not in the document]
enrichman: Maybe the problem is the `Atoi`. Would it pass the check with the `ParseInt`?

```go
i, err := strconv.ParseInt(string(scaleSecret.Data[instanceKey], 10, 32)
if err != nil {
    panic(err)
}
result := int32(i)
```

acceptance/api/v1/healthcheck_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+		defer response.Body.Close()
+		bodyBytes, err := io.ReadAll(response.Body)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(response.StatusCode).To(Equal(http.StatusOK), string(bodyBytes))
+	})
+
+	It("doesn't include the epinio server version in a header (non-authenticated request)", func() {
+		request, err := http.NewRequest("GET", fmt.Sprintf("%s/ready",
+			serverURL), strings.NewReader(""))
```
enrichman, line 33: Same
```suggestion
		request, err := http.NewRequest("GET", fmt.Sprintf("%s/ready", serverURL), nil)
```

acceptance/api/v1/info_test.go
```diff
@@ -31,4 +31,15 @@ var _ = Describe("Info endpoint", func() {
 
 		Expect(info.DefaultBuilderImage).To(Equal("paketobuildpacks/builder:full"))
 	})
+
+	It("includes the epinio server version in a header", func() {
+		response, err := env.Curl("GET", fmt.Sprintf("%s%s/info",
+			serverURL, v1.Root), strings.NewReader(""))
```
enrichman, line 37:
```suggestion
		response, err := env.Curl("GET", fmt.Sprintf("%s%s/info", serverURL, v1.Root), nil)
```

internal/cli/usercmd/clientsync.go
```diff
@@ -0,0 +1,51 @@
...
+	return nil
+}
+
+func getUpdater() (selfupdater.Updater, error) {
+	var updater selfupdater.Updater
+	switch os := runtime.GOOS; os {
+	case "darwin":
+		updater = selfupdater.PosixUpdater{}
+	case "linux":
+		updater = selfupdater.PosixUpdater{}
```
enrichman, line 43: If the updater is the same we can merge the case:
```suggestion
	case "linux", "darwin":
		updater = selfupdater.PosixUpdater{}
```

internal/selfupdater/selfupdater.go
```diff
@@ -0,0 +1,152 @@
...
+	bar := progressbar.DefaultBytes(
+		resp.ContentLength,
+		"Progress",
+	)
+
+	_, err = io.Copy(io.MultiWriter(tmpFile, bar), resp.Body)
+	if err != nil {
+		return "", errors.Wrap(err, "downloading the remote file")
+	}
+	defer resp.Body.Close()
```
enrichman, line 74: This was already deferred, we can remove this line.
```suggestion
```

internal/selfupdater/selfupdater.go
```diff
@@ -0,0 +1,152 @@
...
+
+	if resp.StatusCode != 200 {
+		return "", errors.Errorf("unexpected status code: %d", resp.StatusCode)
+	}
```
enrichman, line 78: Should we do this check before downloading the file? :thinking:

enrichman:
I've left some small comments. The download and check of the updater is basically the same code, except for the `.zip` extension, but it's just 20 lines, so I'm fine with the duplication.

internal/selfupdater/selfupdater.go
jimmykarily: good idea, let me try

enrichman (approved):

PR #1764, opened 2022-09-29
Spike: Deploy services without helm controller, use integrated client, same as for apps

andreas-kupries:
Ref #1704 

This is a test/spike to see if 
1. We can rewrite epinio to use the same integrated client for service instance deployment as we use for apps. And
2. If that change will get rid of the post-upgrade issues with creation of service instances

:warning: :construction: :zap: The tests will very likely fail.

While the manual tests of service instance creation, listing, showing, and deletion worked the API tests for service delete/show may not be fully adapted to the changes yet.

And there is something broken in the test helper function to create a service instance. The integrated helm client seems to time out in 3 seconds with a `cntext deadline exceeded`, and the last commits trying to extand that to 3 minutes does not seem to work.

Currently considering rewriting that part into using the external `helm` cli client instead.


enrichman:
Just referring this comment for a previous spike that we did, maybe it contains some useful hints: https://github.com/epinio/epinio/issues/1281#issuecomment-1085686827

andreas-kupries:
Ok. The code added back for HC-service support has not broken the handling of the new services.
Checking that the HC-service support also works then coming Monday, via extended upgrade tests.

andreas-kupries:
EKS still about

> Creating the PVC
  persistentvolumeclaim/epinio-binary created
  Creating the dummy copier Pod
  pod/epinio-copier created
  Waiting for dummy pod to be ready
  error: timed out waiting for the condition on pods/epinio-copier

Very unrelated to the main work.

internal/configurations/configurations.go
```diff
@@ -235,12 +235,15 @@ func (c *Configuration) GetSecret(ctx context.Context) (*v1.Secret, error) {
 	return secret, nil
 }
 
-// ForService returns a slice of Secrets matching the given Service.
+// ForService returns a slice of configuration secrets matching the given Service.
 func ForService(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
 	secretSelector := labels.Set(map[string]string{
-		"app.kubernetes.io/instance": names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace),
+		"app.kubernetes.io/instance": names.ServiceReleaseName(service.Meta.Name),
```
enrichman, line 241: I haven't tried but the `in` operator ([set-based requirements](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#set-based-requirement)) should help us avoiding two requests and merging the values:

```go
secretSelector := labels.Set(map[string]string{
	ConfigurationLabelKey:        "true",
	ConfigurationTypeLabelKey:    "service",
}).AsSelector()
	
secretSelector.Add(
	labels.NewRequirement(
		"app.kubernetes.io/instance", 
		selection.In, 
		[]string{
			names.ServiceReleaseName(service.Meta.Name),
			names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace)
		},
	),
)
```

internal/configurations/configurations.go
```diff
@@ -251,34 +254,109 @@ func ForService(ctx context.Context, kubeClient *kubernetes.Cluster, service *mo
...
 }
 
-// LabelServiceSecrets will look for the Opaque secrets released with a service, looking for the
-// app.kubernetes.io/instance label, then it will add the Configuration labels to "create" the configurations
-func LabelServiceSecrets(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
+// ForServiceUnlabeled returns a slice of unlabeled secrets matching the given Service
+func ForServiceUnlabeled(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
 	secretSelector := labels.Set(map[string]string{
-		"app.kubernetes.io/instance": names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace),
+		"app.kubernetes.io/instance": names.ServiceReleaseName(service.Meta.Name),
```
enrichman, line 298: Same as before

internal/helm/helm.go
```diff
@@ -75,6 +85,59 @@ func Remove(cluster *kubernetes.Cluster, logger logr.Logger, app models.AppRef)
...
+	return client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
+}
+
+func DeployService(logger logr.Logger, parameters ServiceParameters) error {
+	logger.Info("service helm setup", "parameters", parameters)
+
+	client, err := GetHelmClient(parameters.Cluster.RestConfig, logger, parameters.Namespace)
+	if err != nil {
+		return errors.Wrap(err, "create a helm client")
+	}
```
enrichman, line 103: Just for consistency I would keep the `logger` as the first parameter in the`RemoveService`, and returning a wrapped error, like in the `DeployService`: `errors.Wrap(err, "creating a helm client")`

internal/services/instances.go
```diff
@@ -178,6 +210,195 @@ func (s *ServiceClient) ListInNamespace(ctx context.Context, namespace string) (
...
+	listOpts := metav1.ListOptions{}
+	if namespace == "" {
+		listOpts.LabelSelector = fmt.Sprintf("%s,%s", ServiceNameLabelKey, CatalogServiceLabelKey)
+	} else {
+		listOpts.LabelSelector = fmt.Sprintf(
+			"%s,%s",
+			ServiceNameLabelKey,
+			CatalogServiceLabelKey,
+		)
+	}
```
enrichman, line 222: Are these the same?

internal/services/instances.go
```diff
@@ -178,6 +210,195 @@ func (s *ServiceClient) ListInNamespace(ctx context.Context, namespace string) (
...
+//
+// This is essentially all of the old Get/Delete(All)/List* functions, renamed with an added `HC`
+// suffix (for Helm Controller). The new functions run them in appropriate places.
+//
+// NOTE that `Create` is NOT in this list. We do not create (helm controller)-based services anymore.
+//
+
+// GetHC returns a Service "instance" object if one is exist, or nil otherwise.  Also returns an
+// error if one occurs.
+func (s *ServiceClient) GetHC(ctx context.Context, namespace, name string) (*models.Service, error) {
```
enrichman, line 302: I would avoid shortening too much func names using acronyms. Something like `GetHelmChartService` is clearer.

internal/services/instances.go
```diff
@@ -178,6 +210,195 @@ func (s *ServiceClient) ListInNamespace(ctx context.Context, namespace string) (
...
+	}
+
+	service.Status = models.NewServiceStatusFromHelmRelease(serviceStatus)
+
+	return &service, nil
+}
+
+// DeleteHC deletes the helmcharts that matches the given service which is installed on the
+// namespace (that's the targetNamespace).
+func (s *ServiceClient) DeleteHC(ctx context.Context, namespace, service string) error {
```
enrichman, line 374: Same

internal/services/instances.go
```diff
@@ -178,6 +210,195 @@ func (s *ServiceClient) ListInNamespace(ctx context.Context, namespace string) (
...
+		metav1.DeleteOptions{},
+	)
+
+	return errors.Wrap(err, "error deleting helm charts")
+}
+
+// DeleteAllHC deletes all helmcharts installed on the specified namespace.  It is used to cleanup
+// before a namespace is deleted.  The targetNamespace is not the namespace where the helmchart
+// resource resides (that would be `epinio`) but the `targetNamespace` field of the helmchart.
+func (s *ServiceClient) DeleteAllHC(ctx context.Context, targetNamespace string) error {
```
enrichman, line 386: Same

internal/services/instances.go
```diff
@@ -178,6 +210,195 @@ func (s *ServiceClient) ListInNamespace(ctx context.Context, namespace string) (
...
+			LabelSelector: fmt.Sprintf("%s=%s", TargetNamespaceLabelKey, targetNamespace),
+		},
+	)
+
+	return errors.Wrap(err, "error deleting helm charts")
+}
+
+// listHC will return all the Epinio Services available in the targeted namespace.
+// If the namespace is blank it will return all the instances from all the namespaces
+func (s *ServiceClient) listHC(ctx context.Context, namespace string) (models.ServiceList, error) {
```
enrichman, line 399: Same

enrichman (changes requested):

internal/services/instances.go
andreas-kupries: Maybe? Will check ...

internal/helm/helm.go
andreas-kupries: Will do.

internal/configurations/configurations.go
andreas-kupries: Nice. Will check.

internal/services/instances.go
andreas-kupries: Fixed. Commit 635b0a7778fc6a5afea6d9dd8f0160ef07f86092

internal/services/instances.go
andreas-kupries: Fixed. Commit 635b0a7778fc6a5afea6d9dd8f0160ef07f86092

internal/services/instances.go
andreas-kupries: Fixed. Commit 635b0a7778fc6a5afea6d9dd8f0160ef07f86092

internal/services/instances.go
andreas-kupries: The look indeed the same. Will check if my rewriting the code introduced an error here.

andreas-kupries: The original code is https://github.com/epinio/epinio/blob/15b90264ffb9aa02948bbdb253fe9e54d247fc77/internal/services/instances.go#L182-L191 ... And it was different for namespace vs all, because the helm chart specs were all stored in the epinio namespace, with a label telling us which user namespace it was for (the target namespace).

For the new form the secrets for a service are stored in the namespace they belong to. The change is good, and can be simplified, as you saw. Will do that.

andreas-kupries: Fixed. Commit 773b38c89c0f0c4d0b82f68d6daa41cb20de1427

internal/helm/helm.go
andreas-kupries: Done. Commit ef7781667a767d73ec06f9cfd85e9393830b8e54

internal/configurations/configurations.go
andreas-kupries: Done. Commit 520c57020976e9267154fb21ff968e37371155ad. Let us see what CI thinks of this.

internal/configurations/configurations.go
andreas-kupries: Done. Commit 520c57020976e9267154fb21ff968e37371155ad. Let us see what CI thinks of this.

internal/configurations/configurations.go
andreas-kupries: This has been undone. See commit c1e0a61b4da503bd427fc250310958dc7c89a7f8 for in-code explanations.

enrichman: Having again a look at the code the problem was in my wrong suggestion, that was assuming that the `secretSelector.Add(*multiLabel)` was going to change the `secretSelector`, while instead is returning a new Selector.

Also I've seen that the two functions are basically the same, so we could simplify them using a common `forService` func:

```go
// ForService returns a slice of configuration secrets matching the given Service.
func ForService(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
	// The difference to `ForServiceUnlabeled` is here. We are explicitly looking for
	// the labels attached to the secrets by `LabelServiceSecrets` as we want only
	// proper configurations.
	secretSelector := labels.Set(map[string]string{
		ConfigurationLabelKey:     "true",
		ConfigurationTypeLabelKey: "service",
	}).AsSelector()

	return forService(ctx, kubeClient, service, secretSelector)
}

// ForServiceUnlabeled returns a slice of unlabeled secrets matching the given Service
func ForServiceUnlabeled(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service) ([]v1.Secret, error) {
	// The difference to `ForService` is here. Not looking for the labels attached to
	// the secrets by `LabelServiceSecrets` to turn them into configurations.
	return forService(ctx, kubeClient, service, labels.NewSelector())
}

func forService(ctx context.Context, kubeClient *kubernetes.Cluster, service *models.Service, secretSelector labels.Selector) ([]v1.Secret, error) {

	// Note how we search here for both regular and old-style secrets in one set-based selector
	// and call to come. No need to perform two requests and merge the results.
	//
	// COMPATIBILITY SUPPORT for services from before https://github.com/epinio/epinio/issues/1704 fix
	// Look for secrets referencing a (helm controller)-based service.

	multiLabel, err := labels.NewRequirement(
		"app.kubernetes.io/instance",
		selection.In,
		[]string{
			names.ServiceReleaseName(service.Meta.Name),
			names.ServiceHelmChartName(service.Meta.Name, service.Meta.Namespace),
		},
	)
	if err != nil {
		return nil, err
	}

	secretSelector = secretSelector.Add(*multiLabel)

	listOptions := metav1.ListOptions{
		LabelSelector: secretSelector.String(),
	}

	secretList, err := kubeClient.Kubectl.CoreV1().Secrets(service.Meta.Namespace).List(ctx, listOptions)
	if err != nil {
		return nil, err
	}

	filteredSecrets := filterSecretsByType(secretList.Items, service.SecretTypes)

	return filteredSecrets, nil
}
```

andreas-kupries: I see. Ok. Will trial that.

andreas-kupries: > while instead is returning a new Selector.

And the selector going in was the empty selector, matching everything :facepalm:

andreas-kupries: New trial pushed. Commit be612559d43f318224e4bd052b49e4b5ea549530

enrichman (approved):

thehejik:
Tested on released epinio v1.4.0. For a service installation it shouldn't use an extra `helm-install*` Job anymore but `helm` client from `epinio` pod instead.

I was using simple command from our docs to create a service `epinio service create mysql-dev mydb`. After that I was checking if the related `helm-install*` Job is present in `kubectl get pods -A` output.

Result:
- on epinio 1.3.0 I could see the Job for installing the service is still present:
```
epinio         helm-install-xf964de095d44d6aea92fa3fadcdf-6pxgb                  0/1     Completed   0              3m21s
workspace      xf964de095d44d6aea92fa3fadcdf-mysql-0                             1/1     Running     0              3m7s
```
- on epinio 1.4.0 there was no trace of that Job and the service was running:
```
workspace      xcca9aa0f19a036fb6389474a7be0-mysql-0                             1/1     Running     0              5m51s
```

PR #1778, opened 2022-10-06
1742 application batch deletion

### LINKED ISSUE #1742: Applications batch deletion
Another part of epic : https://github.com/epinio/epinio/issues/1739

Implement the ability to perform bulk deletions of apps from CLI

jimmykarily:
Fixes #1742

### DIFF
--- a/acceptance/api/v1/application_batch_delete_test.go
+++ b/acceptance/api/v1/application_batch_delete_test.go
@@ -61,12 +61,8 @@ func makeApplicationDeleteRequest(namespace, requestBody string, applicationName
 
 	defer response.Body.Close()
 	bodyBytes, err := io.ReadAll(response.Body)
-	Expect(err).ToNot(HaveOccurred())
+	Expect(err).ToNot(HaveOccurred(), string(bodyBytes))
 	Expect(response.StatusCode).To(Equal(http.StatusOK), string(bodyBytes))
-
-	var responseData models.Response
-	err = json.Unmarshal(bodyBytes, &responseData)
-	Expect(err).ToNot(HaveOccurred())
 
 	return bodyBytes
 }
--- a/acceptance/api/v1/application_batch_delete_test.go
+++ b/acceptance/api/v1/application_batch_delete_test.go
@@ -32,7 +32,7 @@ var _ = Describe("AppBatchDelete Endpoint", func() {
 
 			configurationName := *a + "-conf"
 			env.MakeConfiguration(configurationName)
-			env.BindAppConfiguration(app1, configurationName, namespace)
+			env.BindAppConfiguration(*a, configurationName, namespace)
 		}
 	})
 
--- a/acceptance/api/v1/application_batch_delete_test.go
+++ b/acceptance/api/v1/application_batch_delete_test.go
@@ -68,12 +68,10 @@ func makeApplicationDeleteRequest(namespace, requestBody string, applicationName
 }
 
 func validateApplicationDeletionResponse(bodyBytes []byte, configurationNames ...string) {
-	var resp map[string][]string
+	var resp models.ApplicationDeleteResponse
 	err := json.Unmarshal(bodyBytes, &resp)
 	Expect(err).ToNot(HaveOccurred())
-	Expect(resp).To(HaveLen(1))
-	Expect(resp).To(HaveKey("unboundconfigurations"))
-	Expect(resp["unboundconfigurations"]).To(ContainElements(configurationNames))
+	Expect(resp.UnboundConfigurations).To(ContainElements(configurationNames))
 }
 
 func verifyApplicationsDeleted(namespace string, applicationsNames ...string) {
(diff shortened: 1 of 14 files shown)


acceptance/api/v1/application_batch_delete_test.go
```diff
@@ -0,0 +1,106 @@
...
+		env.SetupAndTargetNamespace(namespace)
+
+		for _, a := range []*string{&app1, &app2} {
+			*a = catalog.NewAppName()
+			env.MakeContainerImageApp(*a, 1, containerImageURL)
+
+			configurationName := *a + "-conf"
+			env.MakeConfiguration(configurationName)
+			env.BindAppConfiguration(app1, configurationName, namespace)
+		}
```
enrichman, line 36: Why the pointers?

acceptance/api/v1/application_batch_delete_test.go
```diff
@@ -0,0 +1,106 @@
...
+	Expect(response).ToNot(BeNil())
+
+	defer response.Body.Close()
+	bodyBytes, err := io.ReadAll(response.Body)
+	Expect(err).ToNot(HaveOccurred())
+	Expect(response.StatusCode).To(Equal(http.StatusOK), string(bodyBytes))
+
+	var responseData models.Response
+	err = json.Unmarshal(bodyBytes, &responseData)
+	Expect(err).ToNot(HaveOccurred())
```
enrichman, line 69: Shouldn't we parse an `ApplicationDeleteResponse` and check the parsed resp?

acceptance/api/v1/application_batch_delete_test.go
jimmykarily: Because I need to assign a value to the existing variables on line 30 (maybe this helps: https://go.dev/play/p/dm5q7SQO6_A)

acceptance/api/v1/application_batch_delete_test.go
```diff
@@ -0,0 +1,106 @@
...
+		namespace = catalog.NewNamespaceName()
+		env.SetupAndTargetNamespace(namespace)
+
+		for _, a := range []*string{&app1, &app2} {
+			*a = catalog.NewAppName()
+			env.MakeContainerImageApp(*a, 1, containerImageURL)
+
+			configurationName := *a + "-conf"
+			env.MakeConfiguration(configurationName)
+			env.BindAppConfiguration(app1, configurationName, namespace)
```
andreas-kupries, line 35:
```suggestion
			env.BindAppConfiguration(*a, configurationName, namespace)
```

As is the configurations are all added to app1 only. I suspect you want one config per app.

acceptance/apps_test.go
```diff
@@ -890,6 +890,35 @@ configuration:
...
+			env.MakeContainerImageApp(app1, 1, containerImageURL)
+			app2 := catalog.NewAppName()
+			env.MakeContainerImageApp(app2, 1, containerImageURL)
+
+			var out string
+			var err error
+			Eventually(func() error {
+				out, err = env.Epinio("", "app", "show", app1)
+				return err
+			}, "1m").ShouldNot(HaveOccurred(), out)
```
andreas-kupries, line 904: Do we have a helper function doing this kind of check for apps ? If not, should we create it ?
Does `MakeContainerImageApp` perform this check ? Should the check go there ?

andreas-kupries (changes requested):

acceptance/api/v1/application_batch_delete_test.go
jimmykarily: Actually no, I just need to put the bodyBytes (as a string) in the expectation, to be able to tell what went wrong. No need to parse the response, since we return the response to be parsed in the `verifyApplicationsDeleted` function.

acceptance/api/v1/application_batch_delete_test.go
jimmykarily: Oh true, good catch! My check in the test is too loose to catch this but I'm fixing anyway.

acceptance/apps_test.go
jimmykarily: You are right, `MakeContainerImageApp` already checks this. I'm removing the extra checks from here.

acceptance/api/v1/application_batch_delete_test.go
enrichman: Yes, I was actually seeing that we unmarshal again the response as a map in the `validateApplicationDeletionResponse` though. Should we use the `ApplicationDeleteResponse` there at least, instead of a generic map? Or we usually do this kind of "soft" checks?

jimmykarily: I switched to `ApplicationDeleteResponse` in the `validateApplicationDeletionResponse` function

enrichman (approved):

andreas-kupries (approved):

mmartin24:
Checked working after release `v1.4.0`

```
➜  ~ epinio app delete testapp-1 testapp-2

🚢  Deleting Applications...
Names: testapp-1, testapp-2
Namespace: workspace

✔️  Applications Removed.

➜  ~ epinio version
Epinio Version: v1.4.0
Go Version: go1.18.5
```

PR #1789, opened 2022-10-12
Disable version mismatch warning when doing tab autocompletion

jimmykarily:
The warning was always printed and was messing up the autocompletion. Now we explicitly disable the warning when doing autocompletion. There is still a chance that an out-of-sync client can fail to do the autocompletion but it's better to let it try than break completely.

### DIFF
--- a/internal/cli/commons.go
+++ b/internal/cli/commons.go
@@ -55,8 +55,8 @@ func matchingAppsFinder(cmd *cobra.Command, args []string, toComplete string) ([
 	if err != nil {
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
-
 	app.API.DisableVersionWarning()
+
 	matches := app.AppsMatching(toComplete)
 
 	return matches, cobra.ShellCompDirectiveNoFileComp
@@ -72,8 +72,8 @@ func matchingNamespaceFinder(cmd *cobra.Command, args []string, toComplete strin
 	if err != nil {
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
-
 	app.API.DisableVersionWarning()
+
 	matches := app.NamespacesMatching(toComplete)
 
 	return matches, cobra.ShellCompDirectiveNoFileComp
@@ -89,8 +89,8 @@ func matchingChartFinder(cmd *cobra.Command, args []string, toComplete string) (
 	if err != nil {
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
-
 	app.API.DisableVersionWarning()
+
 	// #args == 0: chart name.
 	matches := app.ChartMatching(toComplete)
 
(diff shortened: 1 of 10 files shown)


internal/cli/usercmd/client.go
```diff
@@ -99,6 +99,9 @@ type APIClient interface {
 	ChartList() ([]models.AppChart, error)
 	ChartShow(name string) (models.AppChart, error)
 	ChartMatch(prefix string) (models.ChartMatchResponse, error)
+
+	DisableVersionWarning()
+	VersionWarningEnabled() bool
```
jimmykarily, line 104: I need to update the "fake" too. On it.

jimmykarily: done

internal/cli/commons.go
```diff
@@ -56,6 +56,7 @@ func matchingAppsFinder(cmd *cobra.Command, args []string, toComplete string) ([
 		return nil, cobra.ShellCompDirectiveNoFileComp
 	}
 
+	app.API.DisableVersionWarning()
```
andreas-kupries, line 59: There are lots of `ValidArgsFunction`s not declared by name, but as a anon function, i.e. `ValidArgsFunction: func(cmd *cobra.Command ... ) { ... }, ...`, directly in the Cmd structures

All of these seem to be missing. `commons.go` only contains the matchers used multiple times.

andreas-kupries (changes requested):
Lots more matchers to fix than done (the anon ones).

internal/cli/commons.go
jimmykarily: Good catch! Fixed here: 8e5b5111

pkg/api/core/v1/client/http.go
```diff
@@ -36,6 +36,18 @@ func wrapResponseError(err error, code int) *responseError {
...
+func (c *Client) DisableVersionWarning() {
+	value := false
+	c.versionWarning = &value
+}
+
+// VersionWarningEnabled returns true if versionWarning field is either not
+// set of true. That makes "true" the default value unless DisableVersionWarning
+// is called.
+func (c *Client) VersionWarningEnabled() bool {
+	return c.versionWarning == nil || *c.versionWarning
```
andreas-kupries, line 48: Name the field `noVersionWarning`, with a default of `false`. `Disable...` then simply sets it to `true`.
That gets rid of all the gymnastics with a pointer. Just a plain bool.

andreas-kupries (approved):
Approved, with a comment regarding the implementation of the flag.

pkg/api/core/v1/client/http.go
jimmykarily: I'm not sure which one I dislike more, double negations or the pointer hacks. I'll try your suggestion.

andreas-kupries (approved):

thehejik:
I was unable to reproduce this issue but I guess I'd need to build epinio CLI from the commit just before merging this PR. Previous `epinio-1.3.0` CLI doesn't contain the warning yet for the version mismatch:
```
⚠️  Epinio server version (v1.3.0) doesn't match the client version (v1.4.0)
```

The autocomplete script works as expected in epinio release 1.4.0 under `bash`.
```
$ source <(epinio completion bash)
$ epinio c<TAB><TAB>
client-sync    completion     configuration
```

Issue #1801, opened 2022-10-18
Mismatching cli version shows the warning multiple times


enrichman:
**Describe the bug**
Running a cli with a version mismatching the server version will sometimes show the warning multiple times (when doing multiple API requests).

![image](https://user-images.githubusercontent.com/1763949/196391756-97d74b10-4810-4470-be12-a3411f602646.png)

**To Reproduce**
1. Install Epinio server and use a different epinio cli version
2. Push an app

**Expected behavior**
The warning should appear only once at the beginning.

mmartin24:
Checked after `v1.4.0` release and ok:

![image](https://user-images.githubusercontent.com/37271841/197746565-b2a68895-cfab-4ff0-bdd2-2e365f838dc7.png)

PR #1804, opened 2022-10-18
Show version mismatch warning only once

enrichman:
Fixes: https://github.com/epinio/epinio/issues/1801

With this PR we disable the warning about the mismatching version after that it was shown.


andreas-kupries (approved):
LGTM (Same as I would have done it)

PR #1807, opened 2022-10-19
Fix conflation of empty and nil slice for routes, in app update handling

### LINKED ISSUE #1799: Remove all the routes from an application
A user can remove all the routes from an application.

Removing the routes from the UI will not work, because is not handled on the BE side.

I would like to delete the routes from a deployed application.

**Additional context**

- https://github.com/epinio/epinio/issues/1775

andreas-kupries:
Fix #1799 

nil   => no change
empty => clear routes

Feat: extend cli with ability send signal for clearing routes (new update-only option --clear-routes, -z)

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -571,6 +571,8 @@ configuration:
 					return err
 				}
 
+				By("ZZZ JSON: ((" + statusJSON + "))")
+
 				var status map[string]string
 				err = json.Unmarshal([]byte(statusJSON), &status)
 				if err != nil {
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -307,6 +307,33 @@ configuration:
 				)
 			})
 
+			It("respects route changes", func() {
+				route := "mycustomdomain.org/api"
+
+				out, err := env.Epinio("", "app", "update", appName, "-r", route)
+				Expect(err).ToNot(HaveOccurred(), out)
+
+				Eventually(func() string {
+					out, err := env.Epinio("", "app", "show", appName)
+					ExpectWithOffset(1, err).ToNot(HaveOccurred(), out)
+					return out
+				}, "1m").Should(HaveATable(WithRow(".*", route)))
+			})
+
+			It("respects complete route removal", func() {
+				out, err := env.Epinio("", "app", "update", appName, "--clear-routes")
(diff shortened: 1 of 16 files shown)


andreas-kupries:
The single failure is a plain push with custom secret - Different from first run, test for overcoming a bad staging.
I suspect that these are transient network and cluster issues. Local execution of these tests is fine.

acceptance/apps_test.go
```diff
@@ -571,6 +631,8 @@ configuration:
 					return err
 				}
 
+				By("ZZZ JSON: ((" + statusJSON + "))")
+
```
enrichman, line 635: I guess this is a leftover.

Just FYI I've found the `GinkgoWriter` pretty handy to log things while testing. It has some convenience methods such as `GinkgoWriter.Println` and so on.

internal/cli/usercmd/push.go
```diff
@@ -261,6 +257,24 @@ func (c *EpinioClient) stageLogs(appRef models.AppRef, stageID string) {
 	}()
 }
 
+func routeMessage(msg *termui.Message, routes []string) *termui.Message {
+	if routes == nil {
+		if len(routes) > 0 {
```
enrichman, line 262: Hm, I'm not sure. How can routes be nil and `>0`?

enrichman (commented):
Just a couple of comments.

acceptance/apps_test.go
andreas-kupries: Yes. :(

internal/cli/usercmd/push.go
andreas-kupries: Will recheck the logic there.

andreas-kupries: Fixed. Commit dd20c7680623195b1487aa257ebed8cb0a8824a2

The logic was inverted. Reworked to reduce indentation and make logic clearer.
This should have a full run of CI to see if there are tests affected by this change to the output.

enrichman: Thank you, much better! :+1:

enrichman:
LGTM, let's wait for CI :+1:

enrichman (approved):
@andreas-kupries LGTM, we need to remember to update the CLI docs :+1:

PR #1817, opened 2022-10-24
Extended app chart taking config names and their mount paths

### LINKED ISSUE #1808: Service mount points should be predictable
**Describe the bug**

If I create an instance of a service from the catalog, I don't know what the name and mount point will be so I can specify that in my code.

For example, if I create a `postgres-dev` service called `postgrestest`, the secret is mounted at `/configurations/x552e4168057e6c22af136063fa72-postgresql`. 

**Expected behavior**
I would expect that the mount point for reading configuration would be predictable. This would likely be the name of the service instance (`/configurations/postgrestest` in the example above

andreas-kupries:
fix #1808

feat: pass in new field `configpath`, config names and mount paths
note: old field `configurations` is still served for use by old app charts

chart pr = https://github.com/epinio/helm-charts/pull/292

- [x] Show the proper mount paths in `configuration show`
- [x] To this end it has to know about sibling configurations from the same origin
- [x] Show these as well
- [x] Implement getting this information in a manner which is not O(n^2) for `configuration list`.

### DIFF
--- a/acceptance/api/v1/appchart_index_test.go
+++ b/acceptance/api/v1/appchart_index_test.go
@@ -54,6 +54,6 @@ var _ = Describe("ChartList Endpoint", func() {
 		Expect(short).Should(ContainElements(
 			"Epinio standard deployment"))
 		Expect(chart).Should(ContainElements(
-			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.22/epinio-application-0.1.22.tgz"))
+			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.23/epinio-application-0.1.23.tgz"))
 	})
 })
--- a/acceptance/api/v1/appchart_index_test.go
+++ b/acceptance/api/v1/appchart_index_test.go
@@ -54,6 +54,6 @@ var _ = Describe("ChartList Endpoint", func() {
 		Expect(short).Should(ContainElements(
 			"Epinio standard deployment"))
 		Expect(chart).Should(ContainElements(
-			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.23/epinio-application-0.1.23.tgz"))
+			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.24/epinio-application-0.1.24.tgz"))
 	})
 })
--- a/acceptance/api/v1/application_export_test.go
+++ b/acceptance/api/v1/application_export_test.go
@@ -52,6 +52,7 @@ var _ = Describe("AppPart Endpoint", func() {
 
 		Expect(string(bodyBytes)).To(Equal(fmt.Sprintf(`epinio:
   appName: %s
+  configpaths: []
   configurations: []
   env: []
   imageURL: splatform/sample-app
--- a/acceptance/apps/rails_test.go
+++ b/acceptance/apps/rails_test.go
@@ -131,8 +131,11 @@ var _ = Describe("RubyOnRails", func() {
 			"--set", "username=myuser")
 		Expect(err).ToNot(HaveOccurred(), out)
 
-		// Change Rails database configuration to match the configuration name
-		out, err = proc.Run(rails.Dir, false, "sed", "-i", fmt.Sprintf("s/mydb/%s/", configurationName), "config/database.yml")
+		// Change Rails database configuration to have the correct mount point in the access paths.
+		// For 1.22 this was the configuration name
+		// For 1.23+ it is the service name instead
+		out, err = proc.Run(rails.Dir, false, "sed", "-i", fmt.Sprintf("s/mydb/%s/", serviceName),
+			"config/database.yml")
 		Expect(err).ToNot(HaveOccurred(), out)
 	})
 
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -1418,6 +1418,7 @@ configuration:
   tuning: speed
 epinio:
   appName: %s
+  configpaths: []
   configurations: []
   env: []
   imageURL: splatform/sample-app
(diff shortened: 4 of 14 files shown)


mmartin24:
Thanks @andreas-kupries  for the explanation on how this worked.

Just FTR this is what was checked on my end at the time being:

- Given a custom service `svc-1` binded to 2 applications `testapp-1, testapp-3`, the hash is only dependent on the custom service name. In the example below `svc-1` yields hash `xd326256210974ceb2fde75b0d087-mysql` in both cases:


```
➜  epinio git:(1808-serviceable-mounts) ✗ epinio app list --all  

🚢  Listing all applications

✔️  Epinio Applications:
| NAMESPACE |   NAME    |            CREATED             | STATUS | ROUTES |           CONFIGURATIONS            | STATUS DETAILS |
|-----------|-----------|--------------------------------|--------|--------|-------------------------------------|----------------|
| workspace | testapp-1 | 2022-10-27 09:37:40 +0200 CEST | n/a    | n/a    | xd326256210974ceb2fde75b0d087-mysql |                |
| workspace | testapp-2 | 2022-10-27 10:32:43 +0200 CEST | n/a    | n/a    | xbceee8a67d6ee59f8e9dd148264c-mysql |                |
| workspace | testapp-3 | 2022-10-27 10:38:02 +0200 CEST | n/a    | n/a    | xd326256210974ceb2fde75b0d087-mysql |                |

```

- Show the mounting paths with `epinio configuration show name-of-config`

```

➜  epinio git:(1808-serviceable-mounts) ✗ epinio configuration show xd326256210974ceb2fde75b0d087-mysql

🚢  Configuration Details
Name: xd326256210974ceb2fde75b0d087-mysql
Namespace: workspace
🚢  
Created: 2022-10-27 09:37:51 +0200 CEST
User: 
Type: service
Origin: svc-1
Used-By: testapp-1, testapp-3
Siblings: 
✔️  
|      PARAMETER      |   VALUE    |                ACCESS PATH                |
|---------------------|------------|-------------------------------------------|
| mysql-password      | 5KISXaf3na | /configurations/svc-1/mysql-password      |
| mysql-root-password | Tpe2BGo4WL | /configurations/svc-1/mysql-root-password |

```

andreas-kupries:
Tested out a compatibility concern. I.e. what happens when
 - An application A is launched with the 1.22 standard app chart
 - Epinio is upgraded to a client and server using the 1.23 standard app chart (note! name is unchanged)
 - A is restarted, whether explicitly (`app restart`), or indirectly (config/service (un)bind, scaling, env changes)

Will the change in the underlying helm chart (1.23 from 1.22) be honored ?
Will the application actually restart properly ?

Well, the restart in itself works. The application smoothly switches to the 1.23 chart and restarts properly, with all that entails.

Which means that when the app has bound services it will break because the mount points for the service secrets will change, and anything in the application's setup referencing the old mount points / access paths will be incorrect after the change. If the relevant setup is done via environment variables then the best thing to do will be do induce the restart by changing these environment variables to the new mount points/access paths. Then the restarted app has the correct setup again as part of the change.

andreas-kupries:
Reworked system.

  - The old key `configurations` left in for backward compat is now put back into full use. It specifies the volumes to mount.
  - The new key `configpaths` specifies the mounts themselves, and can have multiple entries, i.e. mounts, for a specific configuration, i.e. volume.
  - The go code creating the entries for `configpaths` generates entries for the old mounts, for backward compat.
  - The go code computing `configurations` from `configpaths` is extended to dedup the config names so that we do not try to create multiple identical volumns for the configurations with multiple mounts.

This looked ok in a local manual test. IOW the desired (dedup'd) volumes were specified, as well as the multiple mounts. Looking into the pod the mounts were properly executed, i.e. creating copies of the same volume in the desired paths.

An epinio upgrade will now not break a running application, even on restart, because it still has the old mounts and access paths available to it.

Last thing to do !! Have `epinio configuration show` report the multiple mounts, and see if we can put something into `epinio app show` also to make the user aware that an application should migrate to the new access paths.

andreas-kupries:
On the last thing I currently have only unconditional nagging. I.e. when listing configs, or looking at config details it mentions to migrate apps to new access paths ... Because it (the client) actually has no information about the true access paths / mount points used by bound apps. Actually cannot as apps can have many different methods to tell them which access path to use.
Similarly an app only knows the names of the bound configs, and has no knowledge of the true access paths, nor if which of these the app actually uses.

enrichman:
LGTM, it was a tricky one. I've just not fully grasped the siblings configurations. Can you do an example of it?

enrichman (approved):

andreas-kupries:
Sibling configurations ... AFAIK we currently do not have a service which actually creates more than one secret to become a configuration.

Should we have a service `S` creating multiple secrets named, for example, `oritw`, `asl`, and `fkad`, the code will
  - see that they all belong to the same service `S`.
  - order them lexicographically, i.e.
      - `asl`
      - `fkad`
      - `oritw`

    and then give the mount points the names: 

|||
|---|---|
|`asl`|`S`|
|`fkad`|`S-2`|
|`oritw`|`S-3`|

The sorting ensures stability of the mapping. I.e. prevents that `asl` is mapped to `S` at one time and `S-3` some other time.

enrichman:
Hum, I had the case when a service created more than one secret/configuration with Crossplane, but they are going to have different names, so I'm not sure that the siblings was needed at this point, since a service will have always a different name for the secrets. Or maybe I'm still missing something.

The mountpoint are going to be `/configurations/service/secret1`, `/configurations/service/secret2`, `/configurations/service/secret3`.

andreas-kupries:
No. There is no nesting of multiple scerets to the mountpoint.
In the context of my example with service `S` the old vs new mountpoints were/are:

|configuration|new|old|
|---|---|---|
|`asl`|`/configurations/S`|`/configurations/asl`|
|`fkad`|`/configurations/S-2`|`/configurations/fkad`|
|`oritw`|`/configurations/S-3`|`/configurations/oritw`|

andreas-kupries:
It is not about the conflict with the keys of the secret/configuration. If you have multiple secrets you get one volume per secret/config, and one mountpoint per secret. Before the change the unique names of the secrets served to prevent clashes.

enrichman:
Gotcha, thanks for the clarification!

PR #1823, opened 2022-10-28
[CI] Updating actions on aks to avoid ::set-output gh annotation warnings

mmartin24:
Related to https://github.com/epinio/epinio/issues/1797

### Task:
Bump `.github/workflows/aks.yml` to remove annotation warnings related to deprecation of `save-state` and `set-output` commands among others

### Done: 
- Updating `Cache Tools` to use ~`actions/cache@v3.0.8`~ `actions/cache@v3` instead of fixed version
- Updating `Configure AWS credentials for Route53`  to version `node16` to avoid deprecation warning. Explanation [here](https://github.com/aws-actions/configure-aws-credentials#notice-node12-deprecation-warning)
- Changing `::set-output` to new `echo "{name}={value}" >> $GITHUB_OUTPUT`. Explained [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### Results:
[AKS-CI #635 ](https://github.com/epinio/epinio/actions/runs/3342253729), without fix: 4 annotations
[AKS-CI #636](https://github.com/epinio/epinio/actions/runs/3344266782), with fix: 0 annotation warnings, tests passing

image

Issue #1797, opened 2022-10-17
Update CI tests to latest GH actions needs


mmartin24:
Need to update CI tests for latest GH requests.

Example for https://github.com/epinio/epinio/actions/runs/3262226793:

```
[linter](https://github.com/epinio/epinio/actions/runs/3262226793/jobs/5358939885#step:5:14)
The `save-state` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

```
[linter](https://github.com/epinio/epinio/actions/runs/3262226793/jobs/5358939885#step:4:23)
The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
```

```
[acceptance-scenario1](https://github.com/epinio/epinio/actions/runs/3262226793/jobs/5358944416)
Node.js 12 actions are deprecated. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/. Please update the following actions to use Node.js 16: aws-actions/configure-aws-credentials, aws-actions/configure-aws-credentials
```

Tasks to be done:

- [x] Update AKS-CI
- [x] Udpdate AKS-LETSENCRYPT-CI
- [x] Update EKS-CI
- [x] Update GKE-CI
- [x] Update GKE-CI-UPGRADE
- [x] Update GKE-CI-LETSENCRYPT
- [x] Update RKE
- [x] Update RKE-CI-UPGRADE
- [x] Update golang-lint
- [x] Update CI

andreas-kupries:
Looks to be done.

Issue #506, opened 2021-06-21
Multi-user Epinio (same role/permissions), just different creds


kkaempf:
- see also #486

jimmykarily:
I'll close this as a duplicate of #486

PR #1825, opened 2022-10-28
Updating all gke cis  to avoid gh-actions annotation warnings

mmartin24:
Related to https://github.com/epinio/epinio/issues/1797
Doing the same as in https://github.com/epinio/epinio/pull/1823 but for `GKE-CI`,`GKE-CI-UPGRADE` and `GKE-CI-LETSENCRYPT`

### Task:
Bumping  `.github/workflows/gke.yml`, `.github/workflows/gke-upgrade.yml` and `.github/workflows/gke-letsencrypt.yml`  to remove annotation warnings related to deprecation of `save-state` and `set-output` commands among others

### Done: 
- Updating `Cache Tools` to use `actions/cache@v3` instead of fixed version
- Updating `Configure AWS credentials for Route53`  to version `node16` to avoid deprecation warning. Explanation [here](https://github.com/aws-actions/configure-aws-credentials#notice-node12-deprecation-warning)
- Changing `::set-output` to new `echo "{name}={value}" >> $GITHUB_OUTPUT`. Explained [here](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/)

### Results:
[GKE-CI #506](https://github.com/epinio/epinio/actions/runs/3346418869) -> OK. 0 warning annotations
image

### DIFF
--- a/.github/workflows/gke-letsencrypt.yml
+++ b/.github/workflows/gke-letsencrypt.yml
@@ -46,7 +46,7 @@ jobs:
           go-version: ${{ env.SETUP_GO_VERSION }}
 
       - name: Cache Tools
-        uses: actions/cache@v3.0.8
+        uses: actions/cache@v3
         with:
           path: ${{ github.workspace }}/tools
           key: ${{ runner.os }}-tools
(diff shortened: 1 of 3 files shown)


andreas-kupries (approved):

mmartin24:
thank you @andreas-kupries

PR #1826, opened 2022-10-28
Updating actions on rkes yaml  to avoid ::set-output annotation warnings

mmartin24:
Related to https://github.com/epinio/epinio/issues/1797
Doing the same as in https://github.com/epinio/epinio/pull/1823 but for `RKE-CI`,`RKE-CI-UPGRADE` .

### Task:
Bumping  `.github/workflows/rke.yml` and `.github/workflows/rke-upgrade.yml` to remove annotation warnings related to deprecation of `save-state` and `set-output` commands among others

### Done: 
- Updating `Cache Tools` to use `actions/cache@v3` instead of fixed version

### Results:
[RKE-CI #481](https://github.com/epinio/epinio/actions/runs/3346758614) -> 0 annotations after fix
image

### DIFF
--- a/.github/workflows/rke-upgrade.yml
+++ b/.github/workflows/rke-upgrade.yml
@@ -38,7 +38,7 @@ jobs:
           go-version: ${{ env.SETUP_GO_VERSION }}
 
       - name: Cache Tools
-        uses: actions/cache@v3.0.8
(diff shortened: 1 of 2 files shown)


thehejik (approved):
LGTM

Issue #1352, opened 2022-04-14
`epinio app export` extract image too


jimmykarily:
The `epinio export` could be used to migrate the app deployment to a different cluster.
The command is missing the export of the image.

Follow up of: https://github.com/epinio/epinio/issues/1253

# Acceptance Criteria

- `epinio app export` also extract the application image in some format which allows it to be imported to some registry
- [Optional] make it optional with some flag

enrichman:
After the epinio export, the import can be done with helm.

PR #1836, opened 2022-11-02
Added image export to `epinio app export` command

enrichman:
Ref:
- https://github.com/epinio/epinio/issues/1352
- https://github.com/epinio/helm-charts/pull/298
---

This PR adds the the image export to the `app export` command.

I had to add a custom `X-Content-Length` header to let the client knows about the content-length because it seems that the original content-length one is "removed".


internal/api/v1/application/part.go
```diff
@@ -1,28 +1,42 @@
...
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	restclient "k8s.io/client-go/rest"
+	"k8s.io/utils/pointer"
+)
+
+const (
+	SkopeoImage   = "quay.io/skopeo/stable"
+	SkopeoVersion = "v1.10"
```
andreas-kupries, line 39: This has to move into the chart's `values.yaml`, and imported here from a config map or similar. See the staging scripts and their images (awscli, bash, ..._

internal/api/v1/application/part.go
```diff
@@ -111,12 +125,153 @@ func fetchAppChart(c *gin.Context, ctx context.Context, logger logr.Logger, clus
...
+		return apierror.NewInternalError("failed to create job", "error", err.Error())
+	}
+
+	file, err := getFileImageAndJobCleanup(ctx, cluster, jobName, imageOutputFilename)
+	if err != nil {
+		return apierror.NewInternalError("failed waiting for job done", "error", err.Error())
+	}
+
+	defer func() {
+		err := os.Remove("/image-export/" + imageOutputFilename)
```
andreas-kupries, line 159: I see no documentation why this had to be a PVC instead of a simple directory in the server pod.

I believe that the reason is the use of a separate kube job to retrieve the image from the registry. With the PVC both server and job can mount the same (virtual permanent) disk and use it to share data between them.

I strongly believe that this should be spelled out as part of some description/internal documentation of the overall design. This would also include spelling out the reasoning behind the use of a kube job and the introduction of the `skopeo` dependency. Including links to pages explaining skopeo.

andreas-kupries (changes requested):
Needs internal design documentation, and fixes for the skopeo image references. These have to be in the chart values.yaml - See staging scripts and images.

internal/cli/usercmd/app.go
```diff
@@ -227,6 +227,8 @@ func (c *EpinioClient) AppExport(appName string, directory string) error {
 		return errors.Wrapf(err, "failed to create export directory '%s'", directory)
 	}
 
+	fmt.Println()
```
andreas-kupries, line 230: Left over debugging  output ?

pkg/api/core/v1/client/apps.go
```diff
@@ -192,8 +193,21 @@ func (c *Client) AppGetPart(namespace, appName, part, destinationPath string) er
 	}
 	defer out.Close()
 
+	contentLenght := response.ContentLength
```
andreas-kupries, line 196:
```suggestion
	contentLength := response.ContentLength
```

Typo, letter-swap `ht`/`th`

pkg/api/core/v1/client/apps.go
```diff
@@ -27,6 +27,7 @@ import (
 	api "github.com/epinio/epinio/internal/api/v1"
 	"github.com/epinio/epinio/internal/duration"
 	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	progressbar "github.com/schollz/progressbar/v3"
```
andreas-kupries, line 30: The existing internal `helpers/termui` has progress functionality.

pkg/api/core/v1/client/apps.go
```diff
@@ -192,8 +193,21 @@ func (c *Client) AppGetPart(namespace, appName, part, destinationPath string) er
 	}
 	defer out.Close()
 
+	contentLenght := response.ContentLength
+	if response.Header.Get("X-Content-Length") != "" {
+		xContentLenght, err := strconv.ParseInt(response.Header.Get("X-Content-Length"), 10, 64)
```
andreas-kupries, line 198:
```suggestion
		xContentLength, err := strconv.ParseInt(response.Header.Get("X-Content-Length"), 10, 64)
```

andreas-kupries (changes requested):

internal/api/v1/application/part.go
enrichman: I will do that. The main reason was to avoid to modify the epinio server image (that is a scratch image with just the epinio binary on it), and use an external job to do this one-time execution. The reason of using `skopeo copy` instead of the `docker save` was to avoid creating a custom image with the docker daemon, while the skopeo binary can be used as a standalone command. Yes, as you said the PVC is used as a shared disk for this operation.
I'll write down in a doc these considerations.

internal/cli/usercmd/app.go
enrichman: I don't like it but this was done on purpose to add a newline after the "target directory". If there is a better way to do that I will be happy to remove it. Or I can remove the newline if you prefer.

image

pkg/api/core/v1/client/apps.go
enrichman: Everytime. Thanks!

pkg/api/core/v1/client/apps.go
enrichman: Ah, cool, I didn't know. I'll have a look.

internal/cli/usercmd/app.go
andreas-kupries: I see. Ok then.

pkg/api/core/v1/client/apps.go
andreas-kupries: Oh, I see. I had not realized that we already had this package for `client-sync` too.
With that, agreed to keeping as-is.

internal/api/v1/application/part.go
andreas-kupries: The epinio server itself could not be extended to read the image itself ?

docs/explanations/architecture/app-image-export.md
```diff
@@ -0,0 +1,11 @@
+# Application image export
+
+The `epinio app export` command can be used to export all the needed part of an application (the helm chart, its values and the container image).
```
andreas-kupries, line 3:
```suggestion
The `epinio app export` command provides the ability to retrieve all parts of an application (the helm chart, its values and the container image) and save them in the local filesystem.
```

docs/explanations/architecture/app-image-export.md
```diff
@@ -0,0 +1,11 @@
...
+
+The [`skopeo`](https://github.com/containers/skopeo) cli is used to download the image. This was choose over the standard `docker` cli because it doesn't need a deamon to run, it doesn't require root for most of its operations and it's OCI compliants.
```
andreas-kupries, line 5:
```suggestion
The [`skopeo`](https://github.com/containers/skopeo) cli is used to download the image from the registry. This was chosen over the standard `docker` cli because it doesn't need a demon to run, nor does it require root for most of its operations, and it is OCI compliant.
```

docs/explanations/architecture/app-image-export.md
```diff
@@ -0,0 +1,11 @@
...
+
+When a client requests the export of the image of an application to the Epinio server, it will execute a Kubernetes job. This job will run the `skopeo copy` command that will fetch the image from the registry. The image will be downloaded on the PVC shared with the Epinio server. Once downloaded the image will beserved to the client that requested it.
```
andreas-kupries, line 7:
```suggestion
When a client requests the export of the image of an application from the Epinio server, the server first executes a Kubernetes job. This job runs the `skopeo copy` command that fetches the image from the registry. The image is stored on the PVC shared with the Epinio server. Once downloaded the image is served to the client that requested it.
```

docs/explanations/architecture/app-image-export.md
```diff
@@ -0,0 +1,11 @@
...
+
+A cleanup will occur if all the operations will succeed. The job will be not removed if some error occurs, to keep the logs for further investigations.
```
andreas-kupries, line 9:
```suggestion
Cleanup is done only if all the operations succeeded. The job will be not removed if some error occurs, to keep the logs for further investigations.
```

andreas-kupries (changes requested):

enrichman:
Thanks @andreas-kupries !

andreas-kupries (approved):

PR #1841, opened 2022-11-04
Fix gke ci upgrade issues

andreas-kupries:
Debug and fix GKE CI upgrade problems.

### DIFF
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -376,7 +376,7 @@ func (s *ServiceClient) DeleteForHelmController(ctx context.Context, namespace, 
 
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			fmt.Printf("XXX 1\nXXX %s/%s: %s\nXXX\n", namespace, service, err)
+			return nil
 		}
 	}
 
@@ -397,7 +397,7 @@ func (s *ServiceClient) DeleteAllForHelmController(ctx context.Context, targetNa
 
 	if err != nil {
 		if apierrors.IsNotFound(err) {
-			fmt.Printf("XXX 2\nXXX %s: %s\nXXX\n", targetNamespace, err)
+			return nil
 		}
 	}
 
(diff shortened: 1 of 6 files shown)


internal/services/instances.go
```diff
@@ -374,7 +374,13 @@ func (s *ServiceClient) DeleteForHelmController(ctx context.Context, namespace,
 		metav1.DeleteOptions{},
 	)
 
-	return errors.Wrap(err, "error deleting helm charts")
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+	}
```
enrichman, line 381: From the `IsNotFound` doc

```
// It supports wrapped errors and returns false when the error is nil.
```

so I guess that we can drop the `err != nil` check.

```suggestion
	if apierrors.IsNotFound(err) {
		return nil
	}
```

internal/services/instances.go
```diff
@@ -389,7 +395,13 @@ func (s *ServiceClient) DeleteAllForHelmController(ctx context.Context, targetNa
 		},
 	)
 
-	return errors.Wrap(err, "error deleting helm charts")
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			return nil
+		}
+	}
```
enrichman, line 402: Same
```suggestion
	if apierrors.IsNotFound(err) {
		return nil
	}
```

enrichman (commented):
Just a small thing. I didn't know that, maybe we can drop some other if/else in the codebase.

internal/services/instances.go
andreas-kupries: Commit 713797e45437ab627d57046caa6b4d4ab9640365

andreas-kupries:
FOI - The failed tests pass locally. Together with the flakes I suspect CI load and strangeness.

enrichman (approved):

PR #1843, opened 2022-11-07
Fix RKE CI trouble

andreas-kupries:
chore: inspect code verifying config binding via kube commands

### DIFF
--- a/acceptance/helpers/epinio/epinio.go
+++ b/acceptance/helpers/epinio/epinio.go
@@ -87,6 +87,24 @@ func (e *Epinio) Upgrade() {
 }
 
 func (e *Epinio) Install(args ...string) (string, error) {
+
+	// Default is install from local chart
+	chart := "helm-charts/chart/epinio"
+
+	// If requested by the environment, switch to latest release instead, which is older
+	released := os.Getenv("EPINIO_RELEASED")
+	isreleased := released == "true"
+	upgraded := os.Getenv("EPINIO_UPGRADED")
+	isupgraded := upgraded == "true"
+
+	if isupgraded || isreleased {
+		out, err := proc.RunW("helm", "repo", "add", "epinio", "https://epinio.github.io/helm-charts")
+		if err != nil {
+			return out, err
+		}
+		chart = "epinio/epinio"
+	}
+
 	// Update helm repos -- Assumes existence of helm repository providing the helm charts
 	out, err := proc.RunW("helm", "repo", "update")
 	if err != nil {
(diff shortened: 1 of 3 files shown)


andreas-kupries:
Hm ... Per the added inspection either the set of mounts is empty (but we have volumes), or the command retrieving the data is broken. See

```
STEP: verify binding app/config @workspace, a=external-reg-test-rke, c=configuration-4416641967605677923902788657 11/07/22 09:58:15.71
STEP: volumes = ([{"name":"configuration-4416641967605677923902788657","secret":{"defaultMode":420,"secretName":"configuration-4416641967605677923902788657"}}]) 11/07/22 09:58:15.884
STEP: mounts = () 11/07/22 09:58:16.015
```

Local testing of the inspection using `It("doesn't delete a bound configuration", func() { ...` in configurations_test.go
looked ok. ... But was done after attempting to delete the bound config ... Trial the CI steps manually.

Note: CI uses 1.4.0 as the baseline to test upgrading from, and the failure happens for this version.
Trial head and 1.4.0.

andreas-kupries:
Manual trial of head is ok.
Manual trial of 1.4.0 is ok also. Unexpected.
I suspect that I need extended inspection, i.e. look at the entire json instead of just the list of mounts.

andreas-kupries:
Comparing the full state of the deployment for test case and local/manual recreation of testcase ... Mounts is definitely missing in the CI. Both are based on the same app chart (0.1.24).

I have to check if the helm chart and templating is getting the proper app parameters.
Also, check what exact version of the binary is used server side (should be 1.4.0).

andreas-kupries:
Better data now ...
1. install uses app chart 0.1.24 - expects `configurations` and `configpaths` (epinio head, Nov 3, PR 1817)
2. client/server binary is 1.4.0 (Oct 19, requires app chart 0.1.22)
3. We have a server/chart mismatch ... An older server binary drives the newest chart. The chart is not given mount information, and thus there are no mounts, and the test fails, correctly.

This seems to be an issue in how the cluster and epinio installation are set up ... It looks as if the newest build is installed, and the server side binary then downgraded to 1.4.0 ... It should have installed the entire 1.4.0 outright.

Investigating.

andreas-kupries:
The install helper (`acceptance/helper/epinio/epinio.go`:`Install()`) unconditionally uses the local helm chart. No coding to use a released chart ... The regular upgrade test OTOH has this switch happen in the `./scripts/prepare-environment-k3d.sh` script.

andreas-kupries:
And it is a fix.

andreas-kupries:
Rebased, and cleaned up

enrichman (approved):

PR #1851, opened 2022-11-10
Fix various service issues

### LINKED ISSUE #1839: Failure to delete partially created service
**Describe the bug**

With an 1.4.0+ API server it is possible a partially created service which cannot be deleted again.

**To Reproduce**

- Create a service ...
- Interrupt the creation with ^C in the client, before it completes on its own.
- The service should show up in the list, as not ready .
- The show command should work as well.
- Deletion of the partially made service fails.
- This is not the CI issue. Still something to look into.

**Expected behavior**

Proper deletion of the partially created service.

**Logs**

The service is partially created, i.e. its kube secret exists. Its hem release does not, due to the interruption.
The API `ValidateService` checks only for the release, not the service when verifying service existence.
And the `Delete` operation currently wants to fail on a missing service.

### LINKED ISSUE #1849: Bug: Able to create services with the same name if one is from 1.3.0 (helm controller based), and the other is from 1.4.0 (without HC)
Context.

I created in `v.1.3.0` a custom service named mycustom-service and bound to an app testapp . The hash for this version is `x5dc8835923fe6cac2053d8aa18b1-mysql`
I upgrade to `v.1.4.0` and create the same service with hash for `v.1.4.0`: `x8e5ee833a0f2faebaf5c4171baca-mysql `
When I try to bump a second app, `testapp2` it is bound to both services:

```
🚢  Listing configurations
Namespace: workspace

✔️  Epinio Configurations:
|                NAME                 |            CREATED            |  TYPE   |      ORIGIN      |   APPLICATIONS    |
|-------------------------------------|-------------------------------|---------|------------------|-------------------|
| x5dc8835923fe6cac2053d8aa18b1-mysql | 2022-11-10 09:03:51 +0100 CET | service | mycustom-service | testapp, testapp2 |
| x8e5ee833a0f2faebaf5c4171baca-mysql | 2022-11-10 10:04:21 +0100 CET | service | mycustom-service | testapp2          |
➜  epinio git:(main) epinio service list                          

🚢  Listing Services...
Namespace: workspace

✔️  Details:
|       NAME       |            CREATED            | CATALOG SERVICE | VERSION |  STATUS  |   APPLICATIONS    |
|------------------|-------------------------------|-----------------|---------|----------|-------------------|
| mycustom-service | 2022-11-10 10:04:19 +0100 CET | mysql-dev       | 8.0.29  | deployed | testapp, testapp2 |
| mycustom-service | 2022-11-10 09:03:46 +0100 CET | mysql-dev       | 8.0.29  | deployed | testapp, testapp2 |
```

andreas-kupries:
fix #1849 
fix #1839

### DIFF
--- a/internal/api/v1/service/delete.go
+++ b/internal/api/v1/service/delete.go
@@ -53,12 +53,12 @@ func (ctr Controller) Delete(c *gin.Context) apierror.APIErrors {
 
 	theServices := []*models.Service{}
 	for _, serviceName := range serviceNames {
-		service, apiErr := GetService(ctx, cluster, logger, namespace, serviceName)
-		if apiErr != nil {
-			return apiErr
-		}
+		// Note: Validation of the service, i.e. checking for their helm release is
+		// (unfortunately) a step to far in checking. Doing so prevents us from deleting
+		// partially created services, i.e. those whose deployment was interupted after
+		// creastion of the main structure, before the creation of the helm release.
 
-		apiErr = ValidateService(ctx, cluster, logger, service)
+		service, apiErr := GetService(ctx, cluster, logger, namespace, serviceName)
 		if apiErr != nil {
 			return apiErr
 		}
--- a/internal/api/v1/service/delete.go
+++ b/internal/api/v1/service/delete.go
@@ -56,7 +56,7 @@ func (ctr Controller) Delete(c *gin.Context) apierror.APIErrors {
 		// Note: Validation of the service, i.e. checking for their helm release is
 		// (unfortunately) a step to far in checking. Doing so prevents us from deleting
 		// partially created services, i.e. those whose deployment was interupted after
-		// creastion of the main structure, before the creation of the helm release.
+		// creation of the main structure, before the creation of the helm release.
 
 		service, apiErr := GetService(ctx, cluster, logger, namespace, serviceName)
 		if apiErr != nil {
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -96,7 +96,13 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
 		return errors.Wrap(err, "create a helm client")
 	}
 
-	// Ignore errors. The release may not be present (for example due to an aborted deployment)
+	// Ignore errors. The release may not be present. For example due to an aborted
+	// deployment. Note that a multitude of different errors was seen for essentially the same
+	// thing, depending on exact timing of deletion to partial creation. Just ignoring a
+	// specific one is fraught. Report, in case we were to generous and debugging is required.
+	if err != nil {
+		logger.Info("release deletion issue", "error", err)
+	}
 
 	_ = client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
 	return nil
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -96,6 +96,8 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
 		return errors.Wrap(err, "create a helm client")
 	}
 
+	err = client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
+
 	// Ignore errors. The release may not be present. For example due to an aborted
 	// deployment. Note that a multitude of different errors was seen for essentially the same
 	// thing, depending on exact timing of deletion to partial creation. Just ignoring a
@@ -103,8 +105,6 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
 	if err != nil {
 		logger.Info("release deletion issue", "error", err)
 	}
-
-	_ = client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
 	return nil
 }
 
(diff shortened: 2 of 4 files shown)


internal/api/v1/service/delete.go
```diff
@@ -53,12 +53,12 @@ func (ctr Controller) Delete(c *gin.Context) apierror.APIErrors {
...
 	theServices := []*models.Service{}
 	for _, serviceName := range serviceNames {
-		service, apiErr := GetService(ctx, cluster, logger, namespace, serviceName)
-		if apiErr != nil {
-			return apiErr
-		}
+		// Note: Validation of the service, i.e. checking for their helm release is
+		// (unfortunately) a step to far in checking. Doing so prevents us from deleting
+		// partially created services, i.e. those whose deployment was interupted after
+		// creastion of the main structure, before the creation of the helm release.
```
enrichman, line 59:
```suggestion
		// creation of the main structure, before the creation of the helm release.
```

internal/helm/helm.go
```diff
@@ -96,8 +96,10 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
 		return errors.Wrap(err, "create a helm client")
 	}
 
-	// err == nil => passed through unchanged
-	return errors.Wrap(client.UninstallReleaseByName(names.ServiceReleaseName(app.Name)), "cleaning up release")
+	// Ignore errors. The release may not be present (for example due to an aborted deployment)
```
enrichman, line 99: Fine to ignore the error but I would at least log it, to check if something else happened.

The best would be to check if the error is for a not present release but it seems that there are no exported errors to check in the Helm client library, and checking the error string is too brittle. So let's keep it like this and just add a log.

pkg/api/core/v1/errors/errors.go
```diff
@@ -109,7 +109,8 @@ func NewBadRequestError(msg string) APIError {
 	return NewAPIError(msg, http.StatusBadRequest)
 }
 
-// NewBadRequestErrorf constructs an API error for general issues with a request, with a formatted message
+// NewBadRequestErrorf constructs an API error for general issues with a request, with a formatted
+// message
```
enrichman, line 113: Was this needed for the linter? If so I would just add the newline before the comma `,` (nicer imho)

```suggestion
// NewBadRequestErrorf constructs an API error for general issues with a request,
// with a formatted message
```

enrichman (changes requested):
Maybe the same newline for the other comments? Not sure.

Just a couple of small things.

internal/api/v1/service/delete.go
andreas-kupries: oops

internal/helm/helm.go
andreas-kupries: I actually got a multitude of different errors, depending where the server was in the creation/abort of the helm release. `release not found` was the simplest. Other's were about some (deletion?) hook, I think. IOW I found ignoring a specific error to be fraught with peril.

Logging is fine, and will do.

pkg/api/core/v1/errors/errors.go
andreas-kupries: No. Can undo them.

andreas-kupries: Done. Commit 9b1343f364ddc7d955c97ac6dbee1d1ab412fd09

internal/api/v1/service/delete.go
andreas-kupries: Done. Commit 9860077b92793fb9825a84ffa857f5a57a2a0c23

internal/helm/helm.go
andreas-kupries: Done. Commit f5e51745c63f6b23d0de71b04b23b313fd5403a2

andreas-kupries:
The apps :boom: is rails deployment failing. The issue looks to be in staging.
Trying it locally it fails there also.
Trying it locally for `main` instead of the PR ... :boom: 
Does not seem to be an issue of the PR.
From the staging logs It seems that there is some kind of mismatch between buildpack and rails sources now ?
Was there an update of either recently ?

Regardless, does not seem to be blocker for the PR.
Looks to need a separate ticket

internal/helm/helm.go
```diff
@@ -96,8 +96,16 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
...
-	return errors.Wrap(client.UninstallReleaseByName(names.ServiceReleaseName(app.Name)), "cleaning up release")
+	// Ignore errors. The release may not be present. For example due to an aborted
+	// deployment. Note that a multitude of different errors was seen for essentially the same
+	// thing, depending on exact timing of deletion to partial creation. Just ignoring a
+	// specific one is fraught. Report, in case we were to generous and debugging is required.
+	if err != nil {
+		logger.Info("release deletion issue", "error", err)
+	}
+
+	_ = client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
```
enrichman, line 107: This should be the other way around

```suggestion
	err := client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
	if err != nil {
		logger.Info("release deletion issue", "error", err)
	}

```

andreas-kupries: :facepalm:

andreas-kupries: Done. Commit 64a9d9ff58ae235c1ec22c419fd0669292826e7e. Note, `=`, not `:=`.

enrichman: Ups, right!

enrichman (approved):

Issue #1857, opened 2022-11-14
Fix Rails Buildpack


enrichman:
Rails example app build is failing: https://github.com/epinio/epinio/actions/runs/3444532460/jobs/5750665559

andreas-kupries:
A test CI run using paketo `0.2.220-full` is :heavy_check_mark: 
The issue started appearing Thu/Fri (Nov 10/11), pointing at the paketo versions 221+ (current 224).

andreas-kupries:
Opened a ticket upstream: https://github.com/paketo-buildpacks/full-builder/issues/818
It holds the results of today's research (bisecting builder versions).

andreas-kupries:
From the responses I got on the upstream ticket it seems that switching the default Node version from `16.x` to `18.x` is at the heart of the issue, and that setting an env variable BP_NODE_VERSION to `16.*` may prevent it. I.e. forcing the use of the older node version. ... Given that I see similar pinning of the PHP version for the wordpress example I will try this first.

andreas-kupries:
And the proposed mitigation looks good:

```
 Paketo Buildpack for Node Engine 0.19.0 
   Resolving Node Engine version 
     Candidate version sources (in priority order): 
       BP_NODE_VERSION -> "16.*" 
                       -> "" 
       <unknown>       -> "" 
  
     Selected Node Engine version (using BP_NODE_VERSION): 16.18.1 
  
   Executing build process 
     Installing Node Engine 16.18.1 
       Completed in 3.315s 

[...] remainder of log, staging succeeds, no compilation failure as before  
```

PR #1860, opened 2022-11-14
TEST CI with a pinned paketo builder

andreas-kupries:
ref #1857


andreas-kupries:
Abandoning the test work ...

Issue #1809, opened 2022-10-20
[CI] Investigate EKS test failing


mmartin24:
EKS test failing: [EKS-CI](https://github.com/epinio/epinio/actions/runs/3286944443)
We should investigate why.

Error log:
```
<Scenario4> EKS, epinio-ca, on S3 storage
  Installs with loadbalancer IP, custom domain and pushes an app with env vars
  /Users/runner/work/epinio/epinio/acceptance/install/scenario4_test.go:75
STEP: Checking LoadBalancer IP 10/20/22 05:29:40.635
STEP: Updating DNS Entries 10/20/22 05:29:49.669
STEP: Checking that DNS entry is correctly propagated 10/20/22 05:29:53.966
STEP: Installing Epinio 10/20/22 05:33:05.924
Patching Epinio deployment with test binary
------------------------------
• [FAILED] [763.200 seconds]
<Scenario4> EKS, epinio-ca, on S3 storage
/Users/runner/work/epinio/epinio/acceptance/install/scenario4_test.go:21
  [It] Installs with loadbalancer IP, custom domain and pushes an app with env vars
  /Users/runner/work/epinio/epinio/acceptance/install/scenario4_test.go:75

  Begin Captured GinkgoWriter Output >>
    STEP: Checking LoadBalancer IP 10/20/22 05:29:40.635
    STEP: Updating DNS Entries 10/20/22 05:29:49.669
    STEP: Checking that DNS entry is correctly propagated 10/20/22 05:29:53.966
    STEP: Installing Epinio 10/20/22 05:33:05.924
  << End Captured GinkgoWriter Output

  
  Configuration
    - Binary: ./dist/epinio-linux-amd64
    - Tag:    v1.3.0-65-gdba5a096

  Creating the PVC
  persistentvolumeclaim/epinio-binary created
  Creating the dummy copier Pod
  pod/epinio-copier created
  Waiting for dummy pod to be ready
  error: timed out waiting for the condition on pods/epinio-copier

  Unexpected error:
      <*exec.ExitError | 0xc0000be060>: {
          ProcessState: {
              pid: 65019,
              status: 256,
              rusage: {
                  Utime: {
                      Sec: 5,
                      Usec: 38463,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Stime: {
                      Sec: 1,
                      Usec: 736939,
                      Pad_cgo_0: [0, 0, 0, 0],
                  },
                  Maxrss: 80515072,
[...]

PR #1862, opened 2022-11-15
Configure EKS EBS CSI driver for k8s 1.23+

thehejik:
Resolves https://github.com/epinio/epinio/issues/1809

Successful test runs:
https://github.com/epinio/epinio/actions/runs/3513169605
https://github.com/epinio/epinio/actions/runs/3512660788 (failed later on something else)
https://github.com/epinio/epinio/actions/runs/3513477067

To resolve EKS EBS CSI driver [persistent volume's node affinity limitation](https://aws.amazon.com/premiumsupport/knowledge-center/eks-troubleshoot-ebs-volume-mounts/) I had to extend this PR by adding `epinio-copier` pod affinity with `epinio-server` pod. 

That garantees that the `epinio-binary` PV is getting created on the same node (also withtin the same zone) where the `epinio-server` pod is running hence this pod should be always able to mount the PV. I used generic `topologyKey` value `kubernetes.io/hostname` which should be also available out of EKS.

### DIFF
--- a/.github/workflows/eks.yml
+++ b/.github/workflows/eks.yml
@@ -132,6 +132,28 @@ jobs:
           # https://stackoverflow.com/questions/71318743/kubectl-versions-error-exec-plugin-is-configured-to-use-api-version-client-auth
           sed -i .bak -e 's/v1alpha1/v1beta1/' kubeconfig-epinio-ci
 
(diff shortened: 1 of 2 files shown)


mmartin24 (approved):
LGTM.
Several rounds of CI tests passed.

Good work!

PR #1864, opened 2022-11-16
Split tests more

andreas-kupries:
Splitting api/services api tests from the general api tests, via test labeling


andreas-kupries:
Note: Failure is #1857

enrichman (commented):
LGTM, it seems even to take a little less.

Would it make sense to use a `const` for the strings in the `suite_test.go`, since they are repeated many times?

```go
const (
	ConfigurationLabel = "configuration"
	ServiceLabel       = "service"
	...
)
```

andreas-kupries:
Possibly yes. Now wondering if I can constify the entire `Label("...")` expression (To actually shorten the written code).
Well, even if not, I can `var` it.
Will do it tomorrow, as part of a general cleanup (history squash).

andreas-kupries:
Rebased, squashed history, and addressed comment about string repetition. Label (string) decl now done in suite setup. See commit 42792ccb3b8b5b6a3afc1ea5f5f4af1a0dec488c

enrichman (approved):

Issue #1816, opened 2022-10-24
Assign roles based on incoming claims


enrichman:
As discussed [here](https://github.com/epinio/epinio/issues/1758#issuecomment-1285605154) in #1758 it would be nice to have a way to map claims/groups coming from upstream providers to Epinio roles, and assign them automatically.

i.e.: an user having the "Operator:admin" claim/group will be assigned as "admin".

PR #1865, opened 2022-11-17
Added Claims-Roles mapping

enrichman:
Ref:
- #1816 
- https://github.com/epinio/docs/pull/191
---

This PR enable the possibility to assign roles to the Epinio users during the login with an OIDC provider.

Please note that the user role is assigned during the creation of the user, so just after the first login. If the groups are updated these updates will not be reflected in the next login, but you should manually update the role.

The update of the user could be possible, but there are some cases where it's not clear which could be the expected behaviour, so it's better to get a deeper understanding of this.

I.e: what is the expected behaviour when an admin manually change the role of a user? What if the user login with a different provider with different groups/roles, or none?


andreas-kupries (approved):
looks ok to my inexperienced eyes

Issue #1859, opened 2022-11-14
Cleanup golangci-lint configuration


enrichman:
We are using two configurations for golangci-lint but it's something not needed anymore.

We can remove the `schedule` one, and probably update some linters and versions.

PR #1870, opened 2022-11-18
Cleanup golangci-lint config

enrichman:
Ref: 
- #1859
---
Dropping the old `go vet ./...` to use the `golanci-lint` (that runs the `govet` linter as well).  

Removed the old `.golanci-lint-schedule.yml` config file.  

Cleaned up a bit the `.golangci-lint.yml`:
- removed some old deprecated linter:
```
WARN [runner] The linter 'structcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused. 
WARN [runner] The linter 'deadcode' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused. 
WARN [runner] The linter 'varcheck' is deprecated (since v1.49.0) due to: The owner seems to have abandoned the linter.  Replaced by unused. 
```
- ordered linters specifying the default linters and the extra (see doc: https://golangci-lint.run/usage/linters/)

**Note:** there is a TODO item in the config that I'm going to check in a different PR, to avoid cluttering this one. :+1: 
```
# TODO convert the regex to errcheck's new ignore file
```


andreas-kupries (changes requested):
I wanted to disagree with 
> Dropping the old go vet ./... to use the golanci-lint (that runs the govet linter as well).

on the grounds that `go vet` is simpler and quicker to catch the common, i.e. less resource intensive, and that having such a quick check. A quick check told me that difference is about 5s (10s ve, 15s linter), so that argument does not really hold.

Still, the other question which came up for me is not addressed by that.
Why are we keeping the separate golangci-linter flow at all now, given that it has been replicated into all the other flows ?
Or, vice versa, this new setup does not feel DRY at all.

So I still like the old form of our flows, of having a simple linter (`vet`) as the gateway for all the main tests, plus a single flow to perform the deeper linting (`gloangci-lint`).

enrichman:
That's fine, it make sense to keep the `make lint` as a quick check, while still have the `golanci-lint` for a deeper inspection. :slightly_smiling_face: 

I was questing myself the same actually (about why keeping the other action at all).

andreas-kupries (approved):
I assume that the action uses the `.golangci.yml` as default config file ?
(Given that `--config` was removed)

enrichman:
Yes, it's the default config filename. :slightly_smiling_face:

PR #1874, opened 2022-11-21
Supported for `cosign`ed images through goreleaser

### LINKED ISSUE #1011: Signing Epinio images

Operators of Cloud Native platforms are becoming increasingly aware of their upstream software supply chain. To give greater security, we should provide all of the images that we require from a registry that we control as well as sign those images to show that there hasn't been any tampering in transit or caching. 

When we cut a release, we should copy the images that we tested with into our own registry along with signatures from cosign.

andreas-kupries:
Fix #1011


cosign.key
```diff
@@ -0,0 +1,11 @@
+-----BEGIN ENCRYPTED COSIGN PRIVATE KEY-----
```
olblak, line 1: Should we publish the private key?

andreas-kupries: We have to, for gorelease see and use it. Note my comment https://github.com/epinio/epinio/issues/1011#issuecomment-1321814779
On the brighter side, this private key is password protected.

viccuad: can't it be passed via an env value somehow? This seems weird. I suspect that the linked goreleaser docs assumes that the goreleaser job is private and not public.

.goreleaser.yml
```diff
@@ -51,6 +51,10 @@ env:
 snapshot:
   name_template: "{{ .Tag }}-next"
 
+docker_signs:
```
viccuad, line 54: from my limited goreleaser knowledge, I think:
- We need to add signatures of blobs (not of images) to the release and/or archive, see [here](https://goreleaser.com/customization/sign/#with-cosign).
- Possibly, this `docker_signs` needs to be called after the artifacts are built, and not before.
- It may be possible to save `cosign.key` as a GHA secret, and pass it to goreleaser in an env var.

andreas-kupries: I do not believe that the order of clauses in the .goreleaser.yml matters to execution. In the docs about `docker_signs` it says: 

> Note that this pipe will run only at the end of the GoReleaser execution (in its publishing phase), as cosign will change the image in the registry.

IOW we do not specify order, goreleaser does.

GHA secret ... Looked around, found https://docs.github.com/en/actions/security-guides/encrypted-secrets#storing-base64-binary-blobs-as-secrets 
So, looks to be possible, and will go for that. This also means creation of a new key pair to be sure. Password will stay the same, as it was never exposed.

Will look into the blob signing as well (epinio binary).

andreas-kupries: Redone completely per https://blog.goreleaser.com/goreleaser-and-software-supply-chain-security-e8917b26924b
Commit 3659103fdbf6eff14300b838cf8e22ab7bb7e6cf

viccuad (dismissed):
On the cosign side, everything looks fine. Good to see the sboms step thanks to syft.

enrichman:
We can trigger the release-qa workflow on this branch to test it. :+1:

andreas-kupries:
> ... Oh ... That flow (still) has `--skip-sign` in the argument list ... Let me fix that.

Never mind, wrong focus, wrong branch when I looked.

enrichman:
> Note that release-sandbox is already working on the branch.

Oh that's perfect, so we just miss the push, but if the signing is fine we are good!

enrichman (approved):

enrichman (approved):

PR #1879, opened 2022-11-22
Added services' internal routes

### LINKED ISSUE #1867: Show the internal routes of services
When a user deploy a service it's difficult to understand which is the endpoint of the service.

It will be nice to see the internal routes of the service with the `epinio service show` command, so that it could be possible to use them with the application.

It's possible to check the deployed kubernetes services with `kubectl` but a user with no access to the cluster could not do this.

Also sometimes it's "difficult" to know in advance the name of the service, because for example a Redis will deploy a master with some replicas:

```
-> % k get svc -n workspace                                                         
NAME                                                 TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)     AGE
x62215c67064c3cac3413ad9c6536-mongodb                ClusterIP   10.43.173.152   <none>        27017/TCP   31m
xbe737f7c936b454da33f586d0166-redis-headless         ClusterIP   None            <none>        6379/TCP    23m
xbe737f7c936b454da33f586d0166-redis-master           ClusterIP   10.43.155.49    <none>        6379/TCP    23m
xbe737f7c936b454da33f586d0166-redis-replicas         ClusterIP   10.43.106.20    <none>        6379/TCP    23m
```

enrichman:
Fixes #1867

---

This PR adds the `Internal Routes` to the `epinio service show` command. It will look for the Kubernates services matching the Helm release.

image


internal/services/instances.go
```diff
@@ -87,6 +93,29 @@ func (s *ServiceClient) Get(ctx context.Context, namespace, name string) (*model
...
+	if err != nil {
+		return nil, errors.Wrap(err, "fetching the services")
+	}
+
+	internalRoutes := []string{}
+	for _, s := range servicesList.Items {
+		route := fmt.Sprintf("%s.%s.svc.cluster.local", s.Name, s.Namespace)
+		for _, port := range s.Spec.Ports {
+			if port.Port != 80 {
+				route += fmt.Sprintf(":%d", port.Port)
```
andreas-kupries, line 110: So a service `S` with multiple ports `A`, `B`, etc. will look like `S:A:B:...`, right ? Intention ?
Would making a separate entry per port make more sense ?

andreas-kupries (approved):
Mostly happy, one question

internal/services/instances.go
enrichman: Hu, I didn't thought about a service with multiple ports. No, it was not intentional, it would be nicer to show them separately!

Thanks for the gotcha!

andreas-kupries (approved):

PR #1888, opened 2022-11-25
CI tweaks to reduce load

andreas-kupries:
tweak job dependencies to reduce number of concurrent clusters
while keeping ability to quickly redo failed pieces


andreas-kupries:
That looks good with mostly balanced runtimes for the job chains.

enrichman (approved):
Nice graph, and LGTM :)

Issue #1919, opened 2022-12-01
Ruby version `2.7.5` of example-rails app is not supported


enrichman:
**Describe the bug**

The [Rails app test](https://github.com/epinio/epinio/blob/main/acceptance/apps/rails_test.go) started failing because of the unsupported Ruby version.

https://github.com/epinio/epinio/actions/runs/3591825801/jobs/6051174104

```
buildpack failed to satisfy "ruby" dependency version constraint "2.7.5": no compatible versions on "io.buildpacks.stacks.bionic" stack. Supported versions are: [2.7.6, 2.7.7, 3.0.4, 3.0.5, 3.1.2, 3.1.3] 
```

We need to update the https://github.com/epinio/example-rails Ruby version.

andreas-kupries:
The `2.7.5` is mentioned in `.ruby-version`, `Gemfile`, and `Gemfile.lock`. The latter looks to be generated from the second.

andreas-kupries:
Update via `bundle install` ?!

andreas-kupries:
Fixed with https://github.com/epinio/example-rails/pull/6

PR #1922, opened 2022-12-02
Collated dependabot changes for dependent modules

andreas-kupries:


andreas-kupries:
Failure is known rails app issue with ruby version #1919

enrichman (approved):
LGTM :)

PR #1929, opened 2022-12-05
trial - run rails from a test branch

andreas-kupries:
the test branch has minimal bumps to the ruby version, without trying to do a full `bundle install`.


andreas-kupries:
And it is good!

andreas-kupries:
As a trial/test PR this __will not be merged__.
The fix under test here is https://github.com/epinio/example-rails/pull/6

andreas-kupries:
When done simply close and abandon the PR

PR #1933, opened 2022-12-06
Dependabot remainder - dockerfile image ref

andreas-kupries:
Base image for `epinio-unpacker` is bumped in its Dockerfile.

Should bump the unpacker version, and then update the user (epinio helm chart) to match.

With regard to that, users of the dockerfile are:

  - `.github/workflows/unpacker-image.yml` - Explicit tagging with version `1.0`.
  - `.goreleaser.yml` - How is the version number determined ?

### DIFF
--- a/.github/workflows/aks-letsencrypt.yml
+++ b/.github/workflows/aks-letsencrypt.yml
@@ -96,7 +96,7 @@ jobs:
       # The system domain is managed by route53, we need credentials to update
       # it to the loadbalancer's IP
       - name: Configure AWS credentials for Route53
-        uses: aws-actions/configure-aws-credentials@v1-node16
+        uses: aws-actions/configure-aws-credentials@v1.7.0
         with:
           aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
           aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
--- a/.github/workflows/aks.yml
+++ b/.github/workflows/aks.yml
@@ -96,7 +96,7 @@ jobs:
       # The system domain is managed by route53, we need credentials to update
       # it to the loadbalancer's IP
       - name: Configure AWS credentials for Route53
-        uses: aws-actions/configure-aws-credentials@v1-node16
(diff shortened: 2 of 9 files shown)


enrichman:
That was a bit unexpected. Something to add to the release flow document probably.

Basically we have a manual release that is releasing the unpacker image with a fixed tag (the `unpacker-image.yml` GH [action](https://github.com/epinio/epinio/blob/main/.github/workflows/unpacker-image.yml)), and also it's automatically updated during the normal release with Goreleaser: https://github.com/epinio/epinio/blob/main/.goreleaser.yml#L55

https://github.com/epinio/epinio/pkgs/container/epinio-unpacker/versions

The tag is the one used during the release, so we already have the `v1.5.1-amd64`.

andreas-kupries:
Ok. Note however that the helm chart still uses the `1.0`, see

```
  bash:
    registry: ghcr.io/
    repository: epinio/epinio-unpacker
    tag: "1.0"
```

in the `values.yaml`. Without an architecture suffix too. How would that be resolved ?

enrichman:
Yes, I mean that we are releasing an image that we are not using because we are still using the `1.0` with no platform (the one released manually).

I think that we should try to use at least the same tooling and have a single way to release (goreleaser?), then we could think about release the two images at the same time, or maybe having a different tag/workflow with a separate `.goreleaser.yml` file.

If nothing is changed I think that the layers will be the same, and so the tag is just going to be an "alias", right?

andreas-kupries:
> If nothing is changed I think that the layers will be the same, and so the tag is just going to be an "alias", right?

I guess so.

I just noticed the section about docker manifests in goreleaser.
It apparently pulls the platform specific docker images (templates) together into a single thing which is just tagged with the tag, and no architecture.

That is missing for unpacker. I will look into adding such a section now.

andreas-kupries:
For the `epinio-server` the chart leaves the tag empty, and then defaults to the `appVersion`.
I suspect that we can do that for `epinio-unpacker` as well, in the future.

andreas-kupries:
Pushed 882e45aee958b54b2be9db35eaa6602343187c58 with that change

thehejik (approved):
LGTM

PR #1948, opened 2022-12-09
Let epinio consume unpacker image like server image

### LINKED ISSUE #1935: Enable multiarch for epinio-unpacker image from goreleaser and use it in helm chart
We have two pipelines building epinio-unpacker image.

The first one is kind of static and has multiarch support enabled already, it's currently used in helm-charts. We are building it manually by using this [workflow](https://github.com/epinio/epinio/blob/main/.github/workflows/unpacker-image.yml).

The other one which should be resolved by this issue is getting build during a release but for `amd64` only so doesn't have multiarch (arm64 and s390x) support enabled yet. This image is building during a release by using [release.yml](https://github.com/epinio/epinio/blob/main/.github/workflows/release.yml) and [goreleaser](https://github.com/epinio/epinio/blob/main/.goreleaser.yml).

To address this issue we'd need:
- [x] enable multiarch support for `epinio-unpacker:<release-tag>` image via goreleaser.
- [x] use the image in our helm charts

Ref. PR https://github.com/epinio/epinio/pull/1933
Ref. https://github.com/epinio/epinio/issues/1928

andreas-kupries:
Fix #1935 

Chart PR https://github.com/epinio/helm-charts/pull/315 does the main work.
Changes here update CI support code to build the unpacker image they now need, in the same way they need the server image.

### DIFF
--- a/acceptance/helpers/epinio/epinio.go
+++ b/acceptance/helpers/epinio/epinio.go
@@ -41,34 +41,58 @@ func (e *Epinio) Upgrade() {
 	// server image to be assembled in the coming step below. Not doing so causes the new server
 	// image to contain the old binary.
 
+	tag := os.Getenv("EPINIO_CURRENT_TAG")
+	By("Tag: " + tag)
+
 	By("Building server image ...")
 	out, err := proc.Run("../..", false, "docker", "build", "-t", "epinio/epinio-server",
 		"-f", "images/Dockerfile", ".")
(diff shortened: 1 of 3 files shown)


thehejik (approved):
LGTM, CI [Upgrade Test passed](https://github.com/epinio/epinio/actions/runs/3683645167/jobs/6232457335)

PR #1951, opened 2022-12-12
Import proper ingress class usage for dex

### LINKED ISSUE #1821: Helm chart values for Dex ingress.ClassName
Dex ingress.ClassName and Annotations is not getting the value set through `ingress.ingressClassName=....` and `'ingress.annotations.ingress\....' ` flags.

Force to get the same ClassName and Annotations of Epinio or create a new value mapping, eg:`dex.ingress.ingressClassName=....`  ,  `'dex.ingress.annotations.ingress\....`

Tested on epinio 1.3.0

### LINKED ISSUE #1937: [1.5.1] Helm value for `dex.ingress.className` can be set but it's ignored
The key was introduced by https://github.com/epinio/helm-charts/pull/297 but the value is just ignored and not incorporated into dex.yaml template.

Reproducer:
In k3s traefik ingress controller is deployed by default (no ingressClassName set)  so I deployed nginx-ingress in addition from helm which uses `ingressClassName=nginx`.  (Well, it was bad idea because it cannot listen on the same address and ports as traefik but anyway this is not a problem here.)

```bash
$ helm upgrade --install epinio --namespace epinio epinio/epinio  --set skipTraefik=true --set global.domain=10.10.10.16.nip.io --set dex.ingress.className=nginx --create-namespace

# Verify if the class was set
$ kubectl get ingress -n epinio
NAMESPACE   NAME     CLASS    HOSTS                         ADDRESS         PORTS     AGE
epinio      epinio   <none>   epinio.10.10.10.16.nip.io   10.10.10.16   80, 443   20m
epinio      dex      <none>   auth.10.10.10.16.nip.io     10.10.10.16   80, 443   20m
```

andreas-kupries:
Fix #1937 
Actually fix #1821 

Relevant chart PR https://github.com/epinio/helm-charts/pull/316


thehejik (approved):
LGTM - it's including the fixed `ingress.ingressClassName` value in dex template

PR #1985, opened 2023-01-06
New flow for container cleanup

andreas-kupries:
Note: Pattern `v*-g*` for the dev images (they have a `g`-prefix commit id in the tag).  Including the untagged as well

### DIFF
--- a/.github/workflows/cleanup-images.yml
+++ b/.github/workflows/cleanup-images.yml
@@ -18,6 +18,6 @@ jobs:
           account-type: org
           org-name: epinio
           keep-at-least: 1
-          filter-tags: v*-g*
+          filter-tags: v*-[0-9]*-g*
           filter-include-untagged: true
           token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/cleanup-images.yml
```diff
@@ -0,0 +1,23 @@
...
+    steps:
+      - name: Delete untagged and dev containers older than a week
+        uses: snok/container-retention-policy@v1.5.1
+        with:
+          image-names: epinio-server, epinio-unpacker
+          cut-off: A week ago UTC
+          account-type: org
+          org-name: epinio
+          keep-at-least: 1
+          filter-tags: v*-g*
```
enrichman, line 21: Maybe it's not needed but I would check also for the commit number, if any: `v*-[0-9].-g*`

`v1.2.0-78-g535d6393`

I don't think we are going to have an architecture that will start with `g` (like `v1.2.0-g64`) but... who knows.

andreas-kupries: How many digits does the commit number have ? IIRC it is varying, i.e 1/2/... digits, not fixed.
As the pattern is glob-style, not regexp, it would have to be `v*-[0-9]*-g*`, right ? (No `.`)

enrichman (approved):

PR #1986, opened 2023-01-09
fixup: permissions

### LINKED ISSUE #1955: Cleanup untagged/commits `epinio` images
It seems that we have untagged images hanging around, and committish tags (`v1.5.1-42-g2cde6742`).

https://github.com/epinio/epinio/pkgs/container/epinio-server/versions?filters%5Bversion_type%5D=untagged

We should delete the untagged and add some automatic cleanup.

andreas-kupries:
fix #1955

### DIFF
--- a/.github/workflows/cleanup-images.yml
+++ b/.github/workflows/cleanup-images.yml
@@ -5,9 +5,6 @@ on:
   schedule:
     - cron: "0 6 1 * *"  # every day at 6 in the morning
 
-permissions:
-  packages: write
-
 jobs:
   clean-ghcr:
     name: Delete old unused container images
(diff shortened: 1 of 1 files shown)


andreas-kupries:
Sigh. ... GH_TOKEN cannot be configured for proper access, ref https://github.com/snok/container-retention-policy/issues/27
Need a PAT.

andreas-kupries:
With the (classic) PAT set as secret the new flow works.
See for example https://github.com/epinio/epinio/actions/runs/3875264714/jobs/6607553844
Due to the large amount of images we have accrued the flow will catch up over the coming days, incrementally deleting things up to its limit per run.

thehejik (approved):

PR #1995, opened 2023-01-11
Fix server for disabled dex

andreas-kupries:
Fix https://github.com/epinio/helm-charts/issues/325

Import helm chart work fixing templating for when dex is disabled.
Further modified server startup to consider the dex-tls cert file as optional.

### DIFF
--- a/internal/cli/server/server.go
+++ b/internal/cli/server/server.go
@@ -93,8 +93,21 @@ func NewHandler(logger logr.Logger) (*gin.Engine, error) {
 		initContextMiddleware(logger),
 	)
 
-	err := auth.ExtendLocalTrustFromFile("/etc/ssl/certs/dex-tls.pem")
-	if err != nil {
+	// Dex or no dex ?
+
+	dexPEMPath := "/etc/ssl/certs/dex-tls.pem"
+
+	if _, err := os.Stat(dexPEMPath); err == nil {
+		// dex secret is present, load contained cert
+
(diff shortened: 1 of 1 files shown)


enrichman:
Just pointing out an edge case: if someone has logged in with Dex, and then it will be disabled, then the authentication will fail.

This is because in the settings.yaml of the user there still will be the token, so the jwt authentication will be used, but it will fail because of the missing oidcProvider

https://github.com/epinio/epinio/blob/4347a46a099a2742b14ba1fcf523305319f9efd0/internal/cli/server/server.go#L277-L280

The fix would probably be too much work for a so strange case, so I wouldn't bother too much.

andreas-kupries:
> Just pointing out an edge case: if someone has logged in with Dex, and then it will be disabled, then the authentication will fail.
> 
> This is because in the settings.yaml of the user there still will be the token, so the jwt authentication will be used, but it will fail because of the missing oidcProvider
> 
> https://github.com/epinio/epinio/blob/4347a46a099a2742b14ba1fcf523305319f9efd0/internal/cli/server/server.go#L277-L280
> 
> The fix would probably be too much work for a so strange case, so I wouldn't bother too much.

This edge case looks to be about having leftovers in your local setup. I.e. E is installed with Dex, settings contain the token, then E is re-installed without Dex, and the token triggers an error on use. I am ok with not bothering here.

It can be fixed by having the user re-login, right ?

Issue #1956, opened 2022-12-12
Move release notifications after Product Release


enrichman:
Since we would like to trigger the notifications after the publishing of the Product Release we need to move and refine some triggers.

We need to drop the `brew` section from the goreleaser configuration, and add the [brew action](https://github.com/marketplace/actions/bump-homebrew-formula) to this new workflow.

Also the docs should be notified after this release.

Issue #1928, opened 2022-12-05
Epinio Release Roadmap


enrichman:
Define the roadmap and tasks to achieve the "ideal Epinio release flow"

Ref: epinio/epinio#1924 

---

To cleanup the release process we should start with the following steps:

### Epinio

- Split the publishing of the release from the tag of the Epinio:
  https://github.com/epinio/epinio/issues/1952
- Images cleanup:
  https://github.com/epinio/epinio/issues/1955

### Epinio UI

The Epinio UI workflow is still unclear. We should discuss it with @richard-cox to see if we can already try to move the UI to its own repository. This will unblock us to leverage our indipendent workflows.

- Move Epinio dashboard repo
  https://github.com/epinio/ui/issues/178
- Cleanup untagged images:
  https://github.com/epinio/epinio/issues/1953
- Add Release Drafter:
  https://github.com/epinio/epinio/issues/1958

### Release

Since we would like to trigger the notifications after the publishing of the Product Release we need to move and refine some triggers.

Notification and marketplaces:
- https://github.com/epinio/epinio/issues/1956
- https://github.com/epinio/epinio/issues/1957
- https://github.com/epinio/epinio/issues/1960

Release Checklist:
- https://github.com/epinio/epinio/issues/1959

enrichman:
Adding some link to avoid forgetting things:

unpacker image
- https://github.com/epinio/epinio/pull/1933

cleanup untagged images, or daily snapshots tags
- https://github.com/epinio/ui-backend/pkgs/container/epinio-ui/versions

enrichman:
To cleanup the release process we should start with the following steps:

### Epinio

- Split the publishing of the release from the tag of the Epinio:
  https://github.com/epinio/epinio/issues/1952
- Images cleanup:
  https://github.com/epinio/epinio/issues/1955

### Epinio UI

The Epinio UI workflow is still unclear. We should discuss it with @richard-cox to see if we can already try to move the UI to its own repository. This will unblock us to leverage our indipendent workflows.

- Cleanup untagged images:
  https://github.com/epinio/epinio/issues/1953
- Add Release Drafter:
  https://github.com/epinio/epinio/issues/1958

### Release

Since we would like to trigger the notifications after the publishing of the Product Release we need to move and refine some triggers.

Notification:
- https://github.com/epinio/epinio/issues/1956
- https://github.com/epinio/epinio/issues/1957

PR #1996, opened 2023-01-11
Separate Epinio tag/build from release

### LINKED ISSUE #1952: Separate Product Release from Epinio Release
The _Product Release_ should be separate from the _Epinio tag/release_.

We should change the trigger of the release to be triggered from tags, instead of the publishing of the release. https://github.com/epinio/epinio/blob/main/.github/workflows/release.yml#L3-L5

Since by default Goreleaser is going to publish the release we should also remember to set the `draft: true`

Ref: https://goreleaser.com/customization/release/#github

enrichman:
Ref:
- #1928 

Fixes #1952  and #1956

---

This PR changes the trigger for build from release publishing to a tag.
This enables us to tag and build an Epinio version without actually releasing and announcing it as a product version (since we should wait for the Helm charts to be released).

A new action was created. This action will be triggered after the actual publication of a release.
It notifies other repositories about the new Product Release of Epinio.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -1,5 +1,5 @@
-# This action will be triggered when the 'Product Release' will be published.
-# This will notify the external repositories about the new Epinio version
+# This action will be triggered when the 'Product Release' is published.
+# It notifies the external repositories about the new Epinio version
 name: Release Published
 
 on:
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -1,0 +1,51 @@
+# This action will be triggered when the 'Product Release' will be published.
+# This will notify the external repositories about the new Epinio version
+name: Release Published
+
+on:
+  release:
+    types: [published]
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,51 @@
+# This action will be triggered when the 'Product Release' will be published.
```
andreas-kupries, line 1:
```suggestion
# This action will be triggered when the 'Product Release' is published.
```

.github/workflows/release-publish.yml
```diff
@@ -0,0 +1,51 @@
...
+# This will notify the external repositories about the new Epinio version
```
andreas-kupries, line 2:
```suggestion
# It notifies the external repositories about the new Epinio version
```

andreas-kupries (changes requested):
Just comment nits. Approved on the active content

enrichman:
Thanks @andreas-kupries, I missed the notification!

PR #2014, opened 2023-01-18
Update copyrights

andreas-kupries:
edits to add copyright header to project files


enrichman:
I think we should put this at the very top of the files (before the `package` declaration:

- https://github.com/golang/go/blob/master/src/bytes/buffer.go
- https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/endpoint/endpoints_controller.go
- https://github.com/elastic/go-elasticsearch/blob/main/internal/build/main.go

andreas-kupries:
I decided against that location because go looks for the package documentation before the `package` declaration and I did not wish the (c) section it to be mistaken for that.

Checking the doc documentation I now see that the actual documentation has to be without intervening blank line, so the (c) would not be mistaken.

So, ok, I will make the change ... So many files.

enrichman (approved):

PR #2033, opened 2023-01-20
Fixed Epinio Tap (homebrew)

enrichman:
The Github action that we were using to update the official Formula is not working with third-party Taps.

This PR partially revert the https://github.com/epinio/epinio/pull/1996/files

This will keep the bump of our Epinio Tap (https://github.com/epinio/homebrew-tap/blob/main/Formula/epinio.rb) with the tag action. The official Homebrew-Core repository will still need to wait the official publishing of the release.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -47,16 +47,3 @@ jobs:
             Created by https://github.com/mislav/bump-homebrew-formula-action
         env:
           COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
-
-      - name: Bump Homebrew Epinio tap
-        uses: mislav/bump-homebrew-formula-action@v2
-        if: "!contains(github.ref, '-')" # skip prereleases
-        with:
-          homebrew-tap: epinio/homebrew-tap
-          download-url: https://github.com/epinio/epinio/archive/refs/tags/${{ steps.get_tag.outputs.TAG }}.tar.gz
-          commit-message: |
-            {{formulaName}} {{version}}
-
-            Created by https://github.com/mislav/bump-homebrew-formula-action
-        env:
-          COMMITTER_TOKEN: ${{ secrets.COMMITTER_TOKEN }}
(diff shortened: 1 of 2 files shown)


andreas-kupries:
How about instead of moving the brew activity from `release-publish` back to the goreleaser (in release)
using a custom goreleaser step in `release-publish` ? With a goreleaser config limited to just the brew thing ?

enrichman:
> How about instead of moving the brew activity from `release-publish` back to the goreleaser (in release) using a custom goreleaser step in `release-publish` ? With a goreleaser config limited to just the brew thing ?

@andreas-kupries I thought about this option and I wasn't sure because of the availability of binaries built in the previous step. I guess we need to rebuild them again, so the goreleaser config will have also all the building stages, and we should remember to keep them aligned.

Having it after the tag but before the release will let us also test it before releasing it to the public. The Mac installation is simpler with brew.

enrichman:
> > > How about instead of moving the brew activity from `release-publish` back to the goreleaser (in release) using a custom goreleaser step in `release-publish` ? With a goreleaser config limited to just the brew thing ?
> > 
> > 
> > @andreas-kupries I thought about this option and I wasn't sure because of the availability of binaries built in the previous step. I guess we need to rebuild them again, so the goreleaser config will have also all the building stages, and we should remember to keep them aligned.
> 
> Oh, the brew tap puts the actual binaries into the formula ? ... Ok, I suspect that you are right that this then requires the build section for the binaries. However all the images and manifests we should be able to cut out.
> 
> On the topic of keepng them in sync, I just checked, and goreleaser seems to support some kind of include system for the config: https://goreleaser.com/customization/includes/
> 
> That should allow us to have a single spec for the build section shared between the two places..

Unfortunately that's a Pro feature.. :/

Yes, the build part is not that much. I'm just worried about the actual release. The action will start after that we have published the Product Release. So I guess that goreleaser will create a new draft, or maybe it will draft the published release (?). Or maybe nothing. Not sure about the behaviour.

I don't have a strong opinion on this. It has pros and cons. We can say that we already know about the previous approach, and maybe we can try your proposal, and see what is going to happen. What do you think? :slightly_smiling_face:

andreas-kupries:
Ok, let us go with your change now, that way we have something working. Then maybe experiment with my proposal.

andreas-kupries (approved):

PR #2049, opened 2023-01-27
Disable tracking during tests

enrichman:
Disable tracking during tests

### DIFF
--- a/acceptance/helpers/epinio/epinio.go
+++ b/acceptance/helpers/epinio/epinio.go
@@ -156,6 +156,7 @@ func (e *Epinio) Install(args ...string) (string, error) {
 		"epinio",
 		chart,
 		"--wait",
+		"--set server.disableTracking=true", // disable tracking during tests
 	}
 
 	out, err = proc.Run(testenv.Root(), false, "helm", append(opts, args...)...)


acceptance/helpers/epinio/epinio.go
```diff
@@ -156,6 +156,7 @@ func (e *Epinio) Install(args ...string) (string, error) {
 		"epinio",
 		chart,
 		"--wait",
+		"--set server.disableTracking=true", // disable tracking during tests
```
andreas-kupries, line 159: what about the tests going through `scripts/prepare-environment-k3d.sh`, i.e. `make prepare_environment_k3d` ?!

andreas-kupries (changes requested):

acceptance/helpers/epinio/epinio.go
enrichman: Those were already disabled via Helm. :smile: 

https://github.com/epinio/epinio/blob/66ce72bd42654c85cb8945fa7e908ceafea76665/scripts/prepare-environment-k3d.sh#L70-L76

https://github.com/epinio/epinio/blob/66ce72bd42654c85cb8945fa7e908ceafea76665/scripts/prepare-environment-k3d.sh#L97-L102

I missed the programmatic installation for the cloud providers.

andreas-kupries: Ok. Thanks.

andreas-kupries (approved):

PR #2050, opened 2023-01-27
Fix app export for Rancher integration (chart is asset file, not url).

### LINKED ISSUE #1965: Unable to export app when 1.5.1 is installed as Rancher App from dev-v2.x
**Describe the bug**
Rancher:v2.6.9 on k3s v1.24.6+k3s1
Epinio  (100.0.3+up1.5.1) Rancher App installed from `dev-v2.6` rancher repository

and
Rancher:v2.7.0 on k3s v1.24.6+k3s1
Epinio  (101.0.2+up1.5.1) Rancher App installed from `dev-v2.7` rancher repository

After deploying a sample app over webUI and calling `epinio app export sample ./` user is getting only `values.yaml` file. Chart and Image are not exported.

Missing Image was expected as we don't have `skopeo` image mirrored due to a image format incompatability but we should get `chart` in addition which is missing.

```
epinio app export sample ./

🚢  Export application
Namespace: workspace
Application: sample
Target Directory: ./

Downloading app values to 'values.yaml' 100% |█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| (357/357 B, 871 kB/s)        

❌  error exporting app: server status code: Service Unavailable
```

In epinio-server logs (ENV `TRACE_LEVEL=20` in its deployment)  we see that chart endpoints returns status 503 for some reason.

```
2022/12/15 11:08:41 EpinioServer: "level"=1 "msg"="/api/v1/namespaces/workspace/applications/sample/part/chart" "status"=503 "method"="GET" "path"="/api/v1/namespaces/workspace/applications/sample/part/chart" "query"="" "ip"="10.42.0.1" "user-agent"="Go-http-client/1.1" "time"="2022-12-15T11:08:41Z" "latency"="1.33619546s"
```

andreas-kupries:
Fix #1965

fix: make chart export work for rancher integration, where chart is asset file, not remote url
chore: extended logging

### DIFF
--- a/internal/api/v1/application/part.go
+++ b/internal/api/v1/application/part.go
@@ -118,8 +118,38 @@ func fetchAppChart(c *gin.Context, ctx context.Context, logger logr.Logger, clus
 		return apierror.InternalError(err)
 	}
 
+	logger.Info("input", "chart archive", chartArchive)
+
+	// Try to read the archive as local path first, before falling back to retrieval via http.
+
+	file, err := os.Open(chartArchive)
+	if err == nil {
+		logger.Info("input", "chart archive", "is file")
+
+		fileInfo, err := file.Stat()
+		if err == nil {
+			logger.Info("input", "chart archive", "has stat")
+
+			contentLength := fileInfo.Size()
+			contentType := "application/x-gzip"
+
+			logger.Info("input", "chart archive", "returning file")
+
+			c.DataFromReader(http.StatusOK, contentLength, contentType, bufio.NewReader(file),
+				map[string]string{
+					"X-Content-Length": strconv.FormatInt(contentLength, 10),
+				})
+
+			return nil
+		}
+	}
+
+	logger.Info("input", "chart archive", "retrieving by http")
+
 	response, err := http.Get(chartArchive) // nolint:gosec // app chart repo ref
 	if err != nil || response.StatusCode != http.StatusOK {
+		logger.Info("fail, http issue")
+
 		c.Status(http.StatusServiceUnavailable)
 		return nil
 	}
@@ -133,7 +163,7 @@ func fetchAppChart(c *gin.Context, ctx context.Context, logger logr.Logger, clus
 		"returning", fmt.Sprintf("%d bytes %s as is", contentLength, contentType),
 	)
 	c.DataFromReader(http.StatusOK, contentLength, contentType, reader, map[string]string{
-		"X-Content-Length": strconv.FormatInt(response.ContentLength, 10),
+		"X-Content-Length": strconv.FormatInt(contentLength, 10),
 	})
 	return nil
 }


internal/api/v1/application/part.go
```diff
@@ -118,8 +118,38 @@ func fetchAppChart(c *gin.Context, ctx context.Context, logger logr.Logger, clus
...
+	}
+
+	logger.Info("input", "chart archive", "retrieving by http")
+
 	response, err := http.Get(chartArchive) // nolint:gosec // app chart repo ref
 	if err != nil || response.StatusCode != http.StatusOK {
+		logger.Info("fail, http issue")
+
 		c.Status(http.StatusServiceUnavailable)
 		return nil
```
enrichman, line 154: I'm not sure, why we are not returning an error? Is that maybe something that was wrong?

internal/api/v1/application/part.go
```diff
@@ -118,8 +118,38 @@ func fetchAppChart(c *gin.Context, ctx context.Context, logger logr.Logger, clus
...
+	logger.Info("input", "chart archive", chartArchive)
+
+	// Try to read the archive as local path first, before falling back to retrieval via http.
+
+	file, err := os.Open(chartArchive)
+	if err == nil {
+		logger.Info("input", "chart archive", "is file")
+
+		fileInfo, err := file.Stat()
+		if err == nil {
```
enrichman, line 130: Should we handle the err and return? Or we still try to retrieve it via http?

andreas-kupries:
Upgrade and client sync tests use 1.6.1 as baseline, which does not have the dex secret fix, thus trip the new check at the beginning of the install for that issue. We can ignore this, and see it as validation for the new check that it actually captures the issue.

internal/api/v1/application/part.go
andreas-kupries: Not sure actually. And no, that was not what was wrong.
What was wrong was trying to `http.Get()` a file, i.e. `/assets/epinio-application-0.1.24.tgz`.

internal/api/v1/application/part.go
andreas-kupries: I did not wish to use something brittle like a check for the `/assets/...` prefix in the chart archive path to distinguish between a file versus an url.

I simply test if the thing is a path, and if not I assume that it is a proper url.

I.e. for normal deployments (dev, and through chart directly) the `chartArchive` is `https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.24`. This fails the `os.Open()`, and so I go and use `http.Get()` on it.

And for the integration we have `/assets/epinio-application-0.1.24.tgz` which succeeds the `os.Open()` and thus we go and treat it as file.

If we have a file or url which is bad, i.e. does not exist it would simply fail both `os.Open()` and `http.Get()`, and we will get the error from the latter.

enrichman: Yes, I'm fine with the `os.Open()`, but if the `file.Stat()` fails then we are not returning, so we are continuing the flow and go to the `http.Get()`. Or maybe I missed something!

andreas-kupries: Yes. I treat the `file.Stat()` failure same as an `os.Open()` failure. If we cannot stat the file, despite being able to open it I see it as `not a file` too.

enrichman (approved):

PR #2063, opened 2023-02-06
chore: bump Go version in workflows

### LINKED ISSUE #2007: Go 1.19 update
Go 1.19 was released on August and we should update.

andreas-kupries:
Fix #2007


andreas-kupries:
Test failures (update, client-sync) are the known issues with head CI (reject dex/dex-tls fight) and the 1.6.1 release they start with not fixed.

enrichman (approved):
Ok, so everything should be fine! LGTM

enrichman (changes requested):
Are we going to have another PR for the go.mod? :eyes:

enrichman (approved):

enrichman (changes requested):
Having a look at the golangci I've seen we need to update also that file. This PR updates the action but I'm not sure which version will pick up: https://github.com/epinio/epinio/blob/main/.golangci.yml#L13

Also some of the `release*` workflows are missing:
- https://github.com/epinio/epinio/blob/main/.github/workflows/release.yml
- https://github.com/epinio/epinio/blob/main/.github/workflows/release-sandbox.yml
- https://github.com/epinio/epinio/blob/main/.github/workflows/release-qa.yml

I guess that it's because we have not put those in the `SETUP_GO_VERSION` env var. Maybe we can fix this as well.

enrichman (approved):

PR #2066, opened 2023-02-07
Dependabot module updates

andreas-kupries:


andreas-kupries:
While a local run of the `acceptance-cli-apps section` failed as well, it was only a single failure, against 45 passed.
And it took only one hour.
Something looks to be bad with the runner this is on (`Runner name: 'gha04-runner-0'`).

```
Summarizing 1 Failure:
  [TIMEDOUT] Apps when pushing an app multiple times [It] honours the given instance count [application]
  /home/work/SUSE/dev/epinio/acceptance/apps_test.go:699

Ran 46 of 130 Specs in 3623.654 seconds
FAIL! - Suite Timeout Elapsed -- 45 Passed | 1 Failed | 0 Pending | 84 Skipped
--- FAIL: TestAcceptance (3623.66s)
FAIL

Ginkgo ran 1 suite in 1h0m48.362236745s
```

enrichman (approved):

PR #2070, opened 2023-02-08
Fixed release workflow, checklist and minor

enrichman:
During the last release we encountered some small issues:
- the script used to update the Upgrade Responder JSON was not working because of a missing checkout
- the Docker extension was not triggered because of a missing repository dispatch

I've also renamed some of the extension to be easily matched with the workflow filename, and reordered the badges in the README file.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -22,6 +22,12 @@ jobs:
     runs-on: self-hosted
 
     steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: recursive
+          fetch-depth: 0
+
       - name: Get current tag
         id: get_tag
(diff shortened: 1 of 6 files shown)


.github/workflows/release-publish.yml
```diff
@@ -36,6 +42,16 @@ jobs:
           event-type: epinio-release
           client-payload: '{"ref": "${{ steps.get_tag.outputs.TAG }}"}'
 
+      # Allow to automate Docker extension update related to Epinio releases.
```
andreas-kupries, line 45:
```suggestion
      # Automate Docker extension update related to Epinio releases.
```

andreas-kupries (approved):
Small nit in job description. Not a blocker.

andreas-kupries (approved):

PR #2072, opened 2023-02-08
Bump helm charts after v1.6.2 release

thehejik:
Update mainly for our n2n tests which are using `helm-chart` submodule from this repo for installing.


enrichman:
LGTM but it seems that this one is not the v1.6.2 commit. It should be fine anyway, not sure if you want to update it a bit. :slightly_smiling_face:

enrichman (approved):
Nevermind, it was too late.

PR #2073, opened 2023-02-08
Bump helm.sh to 3.11.0, this also gives us containerd 0.6.15

andreas-kupries:
This should address https://github.com/epinio/epinio/security/dependabot/13


olblak (approved):
lgtm

Issue #1926, opened 2022-12-02
Enable access to AWS resources through IRSA (IAM Roles for Service Accounts)


enrichman:
Ref: https://github.com/epinio/epinio/issues/1895

As an operator I would like Epinio to be able to access AWS resources without specifying access keys and secrets but through IRSA (IAM Roles for Service Accounts).

https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

enrichman:
After some tests it seems that we only need to add the `serviceAccount` to the staging jobs.

The Epinio server has already a `epinio-server` service account that can be linked to a IAM Role. I tried with the `eksctl` cli and it was working fine. I need to try again with the AWS cli, because otherwise the eksctl will take the ownership over Helm.

enrichman:
When the PRs are merged it will be possible to associate a serviceaccount to the staging job.

i.e.:
```
helm upgrade --install epinio epinio/epinio \
    --namespace epinio --create-namespace \
    --set global.domain=mydomain.com \
    --set server.stagingServiceAccountName=epinio-staging-service-account
```

Then we just need to create the corresponding Role and ServiceAccount. This can be done with the AWS cli but it will be easier with `eksctl` (we are not managing this with Helm, so no worries with this):

```
eksctl create iamserviceaccount \
    --name epinio-staging-service-account \
    --namespace epinio \
    --cluster ${CLUSTER_NAME} \
    --role-name "epinio-staging-role" \
    --attach-policy-arn "arn:aws:iam::${AWS_ACCOUNT_ID}:policy/EpinioECEKSClusterPolicy" \
    --approve
```

andreas-kupries:
This looks to become a document issue and PR, right ?

thehejik:
Successfully validated on epinio v1.7.0 with EKS v1.23 with EBS addon. Tried only accessing AWS S3. 

Documentation https://docs.epinio.io/howtos/aws_iam_integration#authentication-with-aws-eks-service-account

* deloy EKS v1.23 inc. EBS addon - if on EKS 1.22 run `eksctl utils associate-iam-oidc-provider --cluster=$CLUSTER_NAME --approve`
* deploy epinio - use AWS S3 but do not provide AWS keys:
```
[...]

PR #2074, opened 2023-02-08
Added ServiceAccountName to staging job

enrichman:
Fix: https://github.com/epinio/epinio/issues/1926
Needs: https://github.com/epinio/helm-charts/pull/349

This PR adds the STAGING_SERVICE_ACCOUNT_NAME env var to the Epinio server. This variable is provided to the staging job and can be used to set a specific serviceaccount that will be used during the execution.

It also adds a small refactor of the staging job func, reducing a bit the length of the func (this probably needs a better and more complete refactoring but I didn't want to add to much noise to the PR).

### DIFF
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -48,6 +48,7 @@ type stageParam struct {
 	BuilderImage        string
 	DownloadImage       string
 	UnpackImage         string
+	ServiceAccountName  string
 	Environment         models.EnvVariableList
 	Owner               metav1.OwnerReference
 	RegistryURL         string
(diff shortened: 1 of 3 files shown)


internal/api/v1/application/stage.go
```diff
@@ -573,6 +557,7 @@ func newJobRun(app stageParam) (*batchv1.Job, *corev1.Secret) {
 					},
 				},
 				Spec: corev1.PodSpec{
+					ServiceAccountName: app.ServiceAccountName,
```
andreas-kupries, line 560: This works out (during execution)  even when no service account is set ? (i.e. nothing, chart default). Pod would see empty string, right ?

andreas-kupries (changes requested):
A question about a detail in the SA setup and passing to kube

internal/api/v1/application/stage.go
andreas-kupries: I guess we can see that through the CI

andreas-kupries: And that looks ok.

andreas-kupries (approved):

PR #2077, opened 2023-02-13
Fix loss of certs trusted without user intervention from local settings

### LINKED ISSUE #2003: epinio user (log in on from dex) cant do 'epinio app exec APP-name'
**Describe the bug**
It seems like users that login via dex. eg: `epinio login --oidc <epinio-url>` cant do shell to container.

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. Install epinio version `1.5.2`
2. login using dex user: user - admin@epinio.io/password
3. push an app (we use conda-app via paketo's sample)
4. try container shell via: `epinio app exec <app-name>` . It will fail
5. Now, remove epinio session's file in your current directory
6. login without dex. epinio login  <epinio-url>
7. try  `epinio app exec <app-name>` again... It will work

**Expected behavior**
dex and local user should work the same.

**Logs**
```
╰─$ epinio app exec conda-app

🚢  Executing a shell
Namespace: workspace
Application: conda-app

❌  error getting a shell to application: error sending request: Get "https://epinio.internal.frxxdenv.xxx.com/wapi/v1/namespaces/workspace/applications/conda-app/exec?authtoken=eyJhbGciOiJSUzM4NCIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJlcGluaW8tc2VydmVyIiwiZXhwIjoxNjczNzk4OTc3LCJ1c2VyIjoiYWRtaW5AZXBpbmlvLmlvIn0.yd6Cdu8zRo_LpDNTgmAPhE9RtaoSo1kH0gnlJ_XsmUgAMYZqC7MZOU__ygHNidvO-WAxKUkfIv7UixTMXoc2fn2ME7CwKIB0ypkSRUVPN11XFceWQQqDD0C_6gEdXA2Agl4b0StEHMqjfySCDm-E7fFGLYvIbJNs2axJqwOiBatApv1PB-7maeImk1WUKJiXhZgEqXPRVW_amhRk1vj1DnCBC7ale1I3cs7oPGF-IO8Kv2fXm75h96Ml4JLatoRK3EmM4WT8UGSo_H538SVFRhEWw2s3jrnGVy86fvbO1f_i8UFO9dKXP-ahUSjWDtGRDANCgtBV4Am_ZvQroLh95Q": malformed HTTP response "\x00\x00\x1e\x04\x00\x00\x00\x00\x00\x00\x05\x00\x10\x00\x00\x00\x03\x00\x00\x00\xfa\x00\x06\x00\x10\x01@\x00\x01\x00\x00\x10\x00\x00\x04\x00\x10\x00\x00"
```

**Cluster (please complete the following information):**
 - Provider: K3D
 - Kubernetes Version: [e.g. 1.24]

 - OS: [e.g. Linux, MacOS]
 - Epinio Version: 1.5.2

andreas-kupries:
Fix #2003 
Ref #1964 as the PR introducing the problem.

### DIFF
--- a/internal/cli/usercmd/login.go
+++ b/internal/cli/usercmd/login.go
@@ -137,8 +137,13 @@ func checkAndAskCA(ui *termui.UI, addresses []string, trustCA bool) (string, err
 			if cert == nil {
 				return "", errors.Wrap(err, "error while checking CA")
 			}
-			// add the untrusted certificate
+			// add the untrusted certificate to the list to check
 			certsToCheck = append(certsToCheck, cert)
+		} else {
+			// and regularly trusted certs go directly into the result
+			// This was missing in PR #1964, and demonstrated as bug with issue #2003
+			pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
+			builder.Write(pemCert)
 		}
 	}
 


thehejik (approved):
Tested with trusted/untrusted certs. Both cases ok.

enrichman (commented):
Super, thanks for finding the issue!

Having a look at the code maybe I could have written it in a cleaner way.

I would create one more slice at the top:

```go
// get all the certs to check and the trusted ones
certsToCheck := []*x509.Certificate{}
trustedCerts := []*x509.Certificate{}
```

then the building phase can be moved at the very end, while all the existing "building" will just append the cert to the slice:

FROM:
```golang
pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
builder.Write(pemCert)
```
TO:
```golang
trustedCerts = append(trustedCerts, cert)
```

and a small func to print them:

```golang
func encodeCertsToPEM(ui *termui.UI, certs []*x509.Certificate) string {
	for i, cert := range certsToCheck {
		ui.Success().Msgf("Trusting certificate for address %s...", addresses[i])
		pemCert := pem.EncodeToMemory(&pem.Block{Type: "CERTIFICATE", Bytes: cert.Raw})
			builder.Write(pemCert)
		}
	return builder.String()
}
```

What do you think? I think that it will be easier to read. :slightly_smiling_face:

andreas-kupries:
Well, it is not quite that easy. The proposed function does not have access to `addresses` any longer, and the `i` index does not really match the order of certs anymore. Have to pull the address out of the cert again, or pass structures containing both cert and address.

enrichman:
Ah yes, that's a bit annoying. Nevermind, that would be too complicated I guess. :+1:

enrichman (approved):

PR #2081, opened 2023-02-15
Provide app namespace in file written by `app manifest`

### LINKED ISSUE #2005: Add Namespace as part of Manifest export.
While working on https://github.com/epinio/ui/issues/133 I notice that when we export a `manifest` we don't get any reference to the namespace. Having the namespace would be helpful in some cases like importing it via the UI, without the ns the selected namespace will always default to the first one on the list and it can happen that the given ns won't have all the configurations or services the APP needs.

https://github.com/epinio/epinio/blob/dba5a0962ff5ceeaaa6d6acc638851592373ca17/internal/cli/usercmd/app.go#L244
Could export the `ns`.
For the UI "issue", we can write the namespace to the `manifest.json` on our end and have it present while importing it back in the UI. 

### Example
This is where with the correct namespace reference `sorin` could help with the DX
Jan-16-2023 10-20-07

andreas-kupries:
Fix #2005

:notes: This data is purely informational. It is ignored when pushing the app via this manifest. As before the app goes into the targeted/current namespace.

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -35,6 +35,8 @@ import (
 	"github.com/epinio/epinio/internal/api/v1/application"
 	"github.com/epinio/epinio/internal/names"
 	"github.com/epinio/epinio/internal/routes"
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	"gopkg.in/yaml.v2"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


enrichman (commented):
I was thinking that checking the yaml file with a MatchRegexp is a bit brittle.
Would it make sense to unmarshal the file and check that the yaml is actually valid and contains what we expect to have?

andreas-kupries:
Likely yes.

enrichman (approved):

mmartin24:
Checked CLI in `Epinio Server Version: v1.7.0`

```
➜  epinio git:(main) epinio app manifest testappns2 ~/manifest.json

🚢  Save application details to manifest
Namespace: namespace2
Application: testappns2
Destination: /home/mmartin/manifest.json

✔️  Saved
➜  epinio git:(main) cat ~/manifest.json 
name: testappns2
configuration:
  instances: 1
  routes:
  - testappns2.192.168.16.2.omg.howdoi.website
  appchart: standard
origin:
  git:
    url: ""
  path: sample-app.tar
namespace: namespace2
```

PR #2090, opened 2023-02-21
import new domain validation

andreas-kupries:
Ref https://github.com/epinio/helm-charts/pull/375


andreas-kupries:
Mooted by #2084

PR #2091, opened 2023-02-21
EKS-CI 1.24+ workarounds

thehejik:
Ref. https://github.com/epinio/epinio/issues/2083
Testrun: https://github.com/epinio/epinio/actions/runs/4231512241/jobs/7350140610

This PR is dedicated to CI only, it can be used as a reference for documentation.

~* Add annotation for epinio ingress for enabling websocket communication when using EKS 1.24, ELB and nginx-ingress~ (not needed anymore, the annotation is now set by helm chart via https://github.com/epinio/helm-charts/pull/376)
* Configure containerd on EKS nodes to allow pulling from internal epinio HTTP registry

Issue #2083, opened 2023-02-15
epinio not fully compatible with EKS 1.24


thehejik:
**Describe the bug**
Epinio cli is unable to communicate with epinio-server websocket for unknow reason.

**EDIT:** we identified two issues:
1) Unable to communicate with `epinio-server` websocket when streaming logs - this can be solved by adding `nginx.org/websocket-services: epinio-server` annotation to the epinio ingress (when using nginx-ingress ingress controller)
2) EKS v1.24 dropped dockershim and now using containerd CRI which is pulling images over HTTPS only. But internal epinio registry is using HTTP so EKS is unable to pull images from it. Workaround described [here](https://github.com/epinio/epinio/issues/2083#issuecomment-1434854235).

**To Reproduce**
* deploy EKS 1.24 (eu-central-1) by latest eksctl cli, install cert-manager and nginx-ingress-controller, deploy epinio (1.6.2) in same way as EKS-CI job does where it's failing for a while
* configure route53 to use EXTERNAL-IP hostname value from (elb):  `kubectl get svc -n ingress-nginx ingress-nginx-controller`
* `epinio login ...`,  push an app by `epinio app push --name sample --path ./sample-app/`
```
...
Staging application with code...

🕞  Running staging

🚢  Streaming application logs
Namespace: workspace
Application: sample

❌  failed to tail logs: cannot parse JSON response: 'Bad Request
{"errors":[{"status":500,"title":"websocket: the client is not using the websocket protocol: 'upgrade' token not found in 'Connection' header","details":"\nServer Backtrace: websocket: the client is not using the websocket protocol: 'upgrade' token not found in 'Connection' header"}]}': invalid character 'B' looking for beginning of value

Deploying application ...
............................................................2023/02/15 16:04:10 EpinioApiClient: "msg"="request failed" "error"="Post \"https://epinio.bla.suse.dev/api/v1/namespaces/workspace/applications/sample/deploy\": EOF"

❌  error pushing app to server: can't deploy app: making the request: Post "https://epinio.bla.suse.dev/api/v1/namespaces/workspace/applications/sample/deploy": EOF
```

**Expected behavior**
The app should be deployed without any error message present.

**Additional info**
* It works normally on EKS 1.23
* I checked EKS 1.24 changelog but didn't find anything related to ingress or ELB

andreas-kupries:
[...]

PR #2092, opened 2023-02-21
bump helm-charts (new ingress annotation for EKS)

thehejik:
For #2083 and #2091 we need a helm-charts version bump containing https://github.com/epinio/helm-charts/pull/376


andreas-kupries (approved):
Not sure why not made a part of #2091, but ok

PR #2097, opened 2023-02-22
Revert "Fixed reuse of the existing Release draft"

enrichman:
Reverts epinio/epinio#2034


enrichman:
From a discussion with the author of Goreleaser this seems to be not possible. But since the release is not yet published another round of the Release Drafter will populate again the draft. So it's probably better to add one more run of the release-drafter after the goreleaser action.

Issue #1485, opened 2022-05-24
The application templates should be able to use OCI artifacts for helm


agracey:
If I specify my helm chart in an AppChart using an `oci://` link, it fails with this error:

```
Deploying application ...

❌  error pushing app to server: can't deploy app: Internal Server Error: unable to lookup chart "oci://registry.digitalocean.com/helmtesting/epinio-stateful/epinio-stateful-application:0.1.15", missing registry client
```

agracey:
Here's the line generating the error https://github.com/helm/helm/blob/fa6ba5bb29bd7f8bde3a264ea63ff97a4ff0fe5a/pkg/action/install.go#L679

andreas-kupries:
That is in the `helm/helm` package, i.e. the helm reference implementation. I do not see us fixing that.

Also, you specification is incomplete IMHO. You have two ways of specifying a chart in the AppChart resource.

For one, as a `curl`-able link. Example:

```
spec:
  helmChart: https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.17/epinio-application-0.1.17.tgz
```

For two, as a combination of repository and chart name (plus version). Example:

```
spec:
  helmRepo: https://epinio.github.io/helm-charts
  helmChart: epinio-application:0.1.17
```

Which of these forms did you use, exactly ?

My speculation is:

```
spec:
  helmRepo: oci://registry.digitalocean.com/helmtesting/epinio-stateful
  helmChart: epinio-stateful-application:0.1.15
```

or similar. How badly off am I ?

kkaempf:
no further replies, closing

PR #2101, opened 2023-02-23
Wired up version field in AppChart

agracey:
Currently, you can't specify the needed fields to use an OCI chart as an application template: 
https://github.com/epinio/epinio/issues/1485

Opening as draft since I've confirmed that the code compiles but I ran out of time to test the new functionality :(

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -321,7 +321,7 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
 	}
 
 	helmChart := appChart.HelmChart
-	helmVersion := ""
+	helmVersion := appChart.ChartVersion
 
 	// See also part.go, fetchAppChart
 	if appChart.HelmRepo != "" {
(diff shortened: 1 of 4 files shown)


agracey:
This does work but needs a change to the CRD.

internal/helm/helm.go
```diff
@@ -321,7 +321,7 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
 	}
 
 	helmChart := appChart.HelmChart
-	helmVersion := ""
+	helmVersion := appChart.ChartVersion
```
andreas-kupries, line 325: How do you see this interacting with https://github.com/epinio/epinio/blob/63028b0c605c10784188b8fe13a9f486caa90103/internal/helm/helm.go#L336-L340 just a few lines below ?

IOW when the chart reference is to a helm repository then the version is taking from the chart reference, with chart and version separated by `:`.

Could something similar be done with OCI references ? Assuming that OCI references can be detected as such generally.

Issue #2100, opened 2023-02-22
s3 not accessible when using nginx as ingress


abonillabeeche:
**Describe the bug**
I was successfully using traefik but wanted to switch to nginx, so I reinstalled with
`--set ingress.ingressClassName=nginx`

at that point, I would fail to stage applications as they cannot be uploaded to my s3. Switching back to traefik works fine.

**To Reproduce**
1. Install Epinio 1.6.2 or 1.7.0
2. Push an app 
3. See error

```
Application exists, updating ...

Collecting the application sources ...

Uploading application code ...

❌  error pushing app to server: can't upload archive: server status code: Internal Server Error
[...]

PR #2113, opened 2023-02-27
extended ci flows

andreas-kupries:
Ref #2100 

 - 3a - variant of 3 (RKE, private CA, external registry) where nginx is default ingress controller :heavy_check_mark: 
 - 7 - See 3a, with S3 storage moved to external, AWS S3 bucket. :heavy_check_mark: 
 - 8 - See 7, metallb removed - :boom: 

metallb is an outer lb for dns to point at (magic dns/domain with ip from the rke cluster) - now trialing change to use route53, directly pointing at the nginx IC (see s4), obviating the need


enrichman (changes requested):
When we install/upgrade during tests we need to add also the `server.disableTracking=true` flag. If it's not present also in the other scenarios would you mind adding it?

```
"--set", "server.disableTracking=true", // disable tracking during tests
```

Also I would drop the commented tests if those are just leftovers or copy-paste from the other ones.

andreas-kupries:
There will be a general cleanup of commented parts, likely also a reduction of variants. The various variants is me incrementally working towards the proper test case in small steps.

Will add the `disable tracking`.

andreas-kupries:
Originating issue was solved as issue in the home lab setup.

PR #2121, opened 2023-03-02
Chore: disable version tracking/reporting in the scenario CI flows.

andreas-kupries:
Pulled out of the abandoned #2113.

### DIFF
--- a/.github/workflows/rke-upgrade.yml
+++ b/.github/workflows/rke-upgrade.yml
@@ -130,6 +130,11 @@ jobs:
           sudo sh -c "echo 'disable: rke2-ingress-nginx' > /etc/rancher/rke2/config.yaml.d/00-epinio.yaml"
           # Start RKE cluster
           sudo systemctl enable --now rke2-server
+	  # Save kubeconfig locally and set permissions to prevent kubectl/helm warnings
+	  cp /etc/rancher/rke2/rke2.yaml rke2.yaml
+	  chmod u+rw,go-rwx rke2.yaml
+	  pwd
+	  ls -l rke2.yaml
           # Wait a "little" to let enough time for RKE cluster to be up and running
           sleep 120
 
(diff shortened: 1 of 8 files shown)


andreas-kupries:
Upgrade failure is

>  Error: failed to parse prev-values.yaml: error converting YAML to JSON: yaml: mapping values are not allowed in this context

The referenced `prev-values.yaml` is created by the test case itself, via `helm get values epinio -n epinio`. looking at the result of this command in a local setup shows that the issue is likely the generated header line. See

```
USER-SUPPLIED VALUES:
containerregistry:
  enabled: false
global:
...
```

An additional `-o yaml` seems to fix that.

andreas-kupries:
golangci-lint :boom: yet I see no error messages

andreas-kupries:
Still parsing issues ... Showing actual data in the log:

```
Old values = ((WARNING: Kubernetes configuration file is group-readable. This is insecure. Location: /etc/rancher/rke2/rke2.yaml
  WARNING: Kubernetes configuration file is world-readable. This is insecure. Location: /etc/rancher/rke2/rke2.yaml
  containerregistry:
    enabled: false
  global:
...
```

Lovely. We get the warning written to the yaml we want.

enrichman:
I'm curious to know if we really (still) need that, and why. Maybe it was for an old RKE version? Or a workaround?
- https://github.com/rancher/rke/issues/2395

andreas-kupries:
> I'm curious to know if we really (still) need that, and why. Maybe it was for an old RKE version? Or a workaround?
> 
>     * [RKE cli should generate kube config with 0600 permissions rancher/rke#2395](https://github.com/rancher/rke/issues/2395)

AFAICT from the comments in the script the default is to create the file with secure permissions, i.e `600`.
As this is done by a root process (see all the `sudo`'s) the resulting file will not be readable by helm, kubectl from the rest of the CI steps. Thus the code instructs RKE to make it insecure, to actually have it accessible.

And the warnings are simply ignored

What now happened was that these warning got into the output of `helm get values`. Even if we ask specifically for yaml output. With the warnings in the yaml it was not yaml any longer and the `helm upgrade` broke entirely.

My fix is to make a local copy of the file (possible due to insecure permissions), make that copy secure, and use that as the kubeconfig. No more warnings, no more broken yaml.

enrichman (approved):

Issue #2102, opened 2023-02-23
DEX UI not working with Rancher / Docker Desktop


mmartin24:
### Issue:

Cannot login with OIDC / DEX when deploying Epinio `v1.7.0` on Rancher Desktop `v1.7.0` with Kubernetes version `1.25.4`
Checked on Mac, Windows and Linux. CLI connection is working ok.
Note: verified same error with Docker Desktop

![Image](https://user-images.githubusercontent.com/37271841/220925056-802a9b8c-78bd-4493-a6c4-adc2f2b603f0.png)

### Error logs:

````
ERRO[Thu Feb 23 13:08:51 UTC 2023] Failed to create Dex Client: failed to create dex OIDC provider: Get "https://auth.127.0.0.1.nip.io/.well-known/openid-configuration": dial tcp 127.0.0.1:443: connect: connection refused
creating the provider
github.com/epinio/ui-backend/src/jetstream/dex.NewOIDCProviderWithEndpoint
	/go/src/ui-backend/src/jetstream/dex/dex.go:71
main.(*portalProxy).GetDex
	/go/src/ui-backend/src/jetstream/main.go:1325
github.com/epinio/ui-backend/src/jetstream/plugins/epinio/dex.RedirectUrl
	/go/src/ui-backend/src/jetstream/plugins/epinio/dex/api.go:18
github.com/epinio/ui-backend/src/jetstream/plugins/epinio.(*Epinio).AddRootGroupRoutes.func6
	/go/src/ui-backend/src/jetstream/plugins/epinio/main.go:224
main.(*portalProxy).setSecureCacheContentMiddleware.func1
	/go/src/ui-backend/src/jetstream/middleware.go:242
main.(*portalProxy).setSecureCacheContentMiddleware.func1
	/go/src/ui-backend/src/jetstream/middleware.go:242
github.com/labstack/echo/v4.(*Echo).add.func1
	/root/go/pkg/mod/github.com/labstack/echo/v4@v4.7.2/echo.go:520
main.(*portalProxy).urlCheckMiddleware.func1
	/go/src/ui-backend/src/jetstream/middleware.go:222
main.(*portalProxy).setStaticCacheContentMiddleware.func1
	/go/src/ui-backend/src/jetstream/middleware.go:230
github.com/epinio/ui-backend/src/jetstream/plugins/epinio.(*Epinio).SessionEchoMiddleware.func1
	/go/src/ui-backend/src/jetstream/plugins/epinio/main.go:110
[...]

PR #2123, opened 2023-03-03
Fixed Rancher Desktop Dex issue [update UI]

enrichman:
Fix: https://github.com/epinio/epinio/issues/2102

Updating charts to get latest `ui-backend`.


thehejik (approved):
OIDC login over web now works even in Rancher Desktop.

PR #2130, opened 2023-03-07
Removed `epinio-ui` chart dependency

enrichman:
Fix: https://github.com/epinio/helm-charts/issues/369

This PR update the helm-charts to embed the `epinio-ui` chart into the `epinio` one.

Upgrade tests were fixed using always the "remote" docker image tag. Cannot see a reason to use the local one. :eyes: 
I'm not sure we should even push the `latest` from the tests..


andreas-kupries:
I see no helm chart PR for this ?!

mmartin24 (approved):
Checked UI was ok after deployment both Linux and Mac with Rancher

image (4)

PR #2132, opened 2023-03-08
Performance of `epinio app list`, tab-completion involving app names

### LINKED ISSUE #2087: Very long loading-times for applications.
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Seems that `Request URL:   api/v1/applications` takes a really long time to complete. With ~7 applications it takes from ~13 to 18 seconds for the request to complete. Also, under `https://localhost:8005/epinio/c/default/applications` seems that two requests are sent and I have the feeling the 2nd one is blocked until the first one finishes

image

### Expected Behavior

We need to try and find out why the long response time (infra configuration, backend interaction...) and see why two request are trigger on the application view.

### Steps To Reproduce

1. https://localhost:8005/epinio/c/default/applications
2. Open the Network tab and you should see two requests triggered, you should see two /applications requests

### Environment

```markdown
- OS: 
- Architecture:
- Epinio Version: 1.6.2
- Cluster:
  - Provider: DO
  - Options:
  - Kubernetes Version: K3d with K3s v1.24.4+k3s1
```

### Logs

_No response_

### Anything else?

_No response_

andreas-kupries:
fix #2087

- tab completion - changed from `List` to `ListAppRefs` (just names) 
- most app data is now loaded through a few kube `List` calls, with in-memory fusion into the desired structures
    - 1x app resources (core app information)
    - 1x secrets (aux data scaling, environment, desired routes)
    - 1x pods (basic workload information)
    - 1x pod metrics (cpu load, memory usage)
    - 1x ingresses (actual routes)

the above means that listing the apps now makes `5` kube calls, instead of `1 + n * 7`, where `n` the number of apps.

tab completion is fast now
app listing is faster as well (~0.5 seconds for 10 apps, instead of 7-10 seconds).

Internal changes:

  - split the loader functions for the aux secrets into two pieces ... Actual loader, and data extractor.
  - the modified loader calls the separated extractor now.
  - the new extractors are now also called by the List code as part of fusing the information from the bulk loaded secrets into the application structure
  - the same kind of split is done for ingresses, pods, and pod metrics.


andreas-kupries:
The is something going wrong with the restart count. Main is clean wrt this test, and my branch is not. Still investigating.

andreas-kupries:
Command setting a workload field needed by the `generatePodInfo` got moved during the rewrite to after the call, causing gPI to not count the restarts properly. Moved back, and added comment.

internal/application/application.go
```diff
@@ -494,8 +568,127 @@ func Logs(ctx context.Context, logChan chan tailer.ContainerLogLine, wg *sync.Wa
...
+	appName := appCR.GetName()
+	namespace := appCR.GetNamespace()
+	meta := models.NewAppRef(appName, namespace)
+	app := meta.App()
+
+	// I. Unpack the auxiliary data in the various secrets
+
+	env, found := envs[key]
+	if !found {
+		return nil, errors.New("finding env")
```
enrichman, line 598: Is it an err if no env exists (same for other resources)?

internal/application/application.go
```diff
@@ -494,8 +568,127 @@ func Logs(ctx context.Context, logChan chan tailer.ContainerLogLine, wg *sync.Wa
...
+		status = "failed to get replica details"
+	} else {
+		// extract the metrics for the app, based on the app pods
+		for _, pod := range appPods {
+			m, found := metrics[pod.Name]
+			if found {
+				podMetrics = append(podMetrics, m)
+			}
+		}
+	}
```
enrichman, line 678: I guess we can drop the else check, since it's safe to access a `nil` map (you cannot write but you can check the value on it)

_untested code!_

```go
var status string
if metrics == nil {
	status = "failed to get replica details"
}

// extract the metrics for the app, based on the app pods
podMetrics := []metricsv1beta1.PodMetrics{}
for _, pod := range appPods {
	if m, found := metrics[pod.Name]; found {
		podMetrics = append(podMetrics, m)
	}
}
```

internal/application/application.go
```diff
@@ -247,19 +248,92 @@ func List(ctx context.Context, cluster *kubernetes.Cluster, namespace string) (m
...
+		case "scaling":
+			scales[key] = s
+		case "configuration":
+			bounds[key] = s
+		case "environment":
+			environs[key] = s
+		default:
+			// ignore secret
+		}
+	}
```
enrichman, line 330: I would move this into a separate func, using a dedicated struct instead of multiple maps (`buildAppSecretsMap` or something like that):

```go
type appSecrets struct {
	scaling       *v1.Secret // pointer to check if nil
	configuration *v1.Secret
	environment   *v1.Secret
}

func buildAppSecretsMap(appName, namespace string, secrets secrets.Items) map[string]appSecrets {
	// do stuff
}
```

so we can have a `map[string]appSecrets` to match all the "settings" for an app.

Not sure if we could add also the other info there (pods, routes, metrics).

enrichman (changes requested):
Some comments. And I would remove the changes regarding spaces and such.

Or we should agree on having the same settings and conventions. :smile:

internal/application/configurations.go
```diff
@@ -141,18 +141,22 @@ func BoundAppsNames(ctx context.Context, cluster *kubernetes.Cluster, namespace
...
 	return fmt.Sprintf("%s\000%s", name, namespace)
 }
 
 // DecodeConfigurationKey splits the given key back into name and namespace.
 // The name is the first result, the namespace the second.
+
 func DecodeConfigurationKey(key string) (string, string) {
 	parts := strings.Split(key, "\0000")
 	return parts[0], parts[1]
 }
```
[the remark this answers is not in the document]
andreas-kupries: Broken how ?

internal/application/application.go
andreas-kupries: Yes, even if no env assignments are made, the secret has to exist.
Ditto for the others.

We only might have a temporary condition during app deletion where not all pieces of an app may exist anymore, before it is completely gone.

internal/application/application.go
andreas-kupries: The new code wastes cycles on finding metrics for pod, where we know we have none.
The old code does not waste cycles as skips the loops entirely.
Would prefer to keep that.

internal/application/application.go
andreas-kupries: That should be do-able. Even for the pods, etc. too.

internal/application/configurations.go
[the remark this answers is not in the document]
andreas-kupries: Will see to undo the changes which are purely spacing.

mmartin24:
@andreas-kupries, great job!
I just checked this and it is quite surprising the change: less than  1 second to list 10 apps. Auto-completion equally fast. Just sharing a video to visualize the speed better:

https://user-images.githubusercontent.com/37271841/224312024-3e105aa8-26bb-4fd5-91c9-5abb1e386bad.mov

internal/application/application.go
andreas-kupries: Well, the metrics are keyed by pod name, not by app+namespace as the rest. Thus metrics stay as a separate map.

Done with commit ce4db9642c7cf87e4ea7a1620758276a9e34557f

enrichman:
Agree, nice one! It possibly improved also the tests. :slightly_smiling_face:

enrichman (approved):

andreas-kupries:
```
Summarizing 2 Failures:
  [FAIL] AllApps Endpoints [It] doesn't list applications belonging to non-accessible namespaces [application]
  /home/sles/actions-runner/_work/epinio/epinio/acceptance/api/v1/application_index_all_test.go:111
  [FAIL] AllApps Endpoints [It] lists all applications belonging to all namespaces [application]
  /home/sles/actions-runner/_work/epinio/epinio/acceptance/api/v1/application_index_all_test.go:77
```

I have to check these. The new bulk loading may circumvent some of the permission checks.

andreas-kupries:
They pass locally.

andreas-kupries:
And checking the code shows me that `FiltreResources` is applied to the result of `List`. IOW changes in `List` should not affect the filtering.

andreas-kupries:
Looking deeper into the logs .... Oh

```
 [FAILED] {"errors":[{"status":500,"title":"missing auxiliary data","details":"\nServer Backtrace: [...]
```

and 

```
[FAILED] {"errors":[{"status":500,"title":"finding env","details":"\
```

That is from the new code ... 

Passing locally, failing in CI ... Local check was focused on these two ...
Some concurrency issue ? I.e. picking up a partially deleted app ?

I will rework the handling of missing pieces - I.e. ignore such "applications".

enrichman (approved):

Issue #2131, opened 2023-03-08
Display deleted status in CLI for bulk deletion


mmartin24:
### Issue
Currently when performing bulk deletion in CLI it may take some time until the action is completed and the user does not see if the action is being carried. The only thing displayed are these dots:

![Image](https://user-images.githubusercontent.com/37271841/223687138-b871031e-394e-48ce-b0ab-6d535337768f.png)

### Proposal
As discussed in meeting, perhaps we could improve this by adding a status of the apps that are being removed. Something like:

```🚢  Deleting Applications...
Names: testapp-1, testapp-2, testapp-3, testapp-4, testapp-5, testapp-6, testapp-7
Namespace: workspace

Status:
Application "testapp-1" removed
Application "testapp-2" removed ...

```

mmartin24:
I checked directly it on branch https://github.com/epinio/epinio/pull/2138

- Checked all `applications`, `configurations` and `services` displayed status as they are being deleted:

  ```
  ➜  epinio git:(2114-delete-all) epinio app delete --all

  🚢  Querying Applications for Deletion...
  Namespace: workspace

  🚢  Deleting Applications...
  Names: testapp-1, testapp-10, testapp-2, testapp-3, testapp-4, testapp-5, testapp-6, testapp-7, testapp-8, testapp-9
  Namespace: workspace
  ...
  ✔️  testapp-10 deleted
  ✔️  testapp-2 deleted
  ✔️  testapp-1 deleted
  ✔️  testapp-3 deleted
  ...
  ✔️  testapp-5 deleted
  ✔️  testapp-4 deleted
  ...
  ✔️  testapp-6 deleted
  ✔️  testapp-7 deleted
  ✔️  testapp-8 deleted
  .
  ✔️  Applications Removed.
  ```
  
- Observed particularly in `configurations` does not display status while deleting 10 configs if the deletion part is relatively fast (2-3 seconds). @andreas-kupries  Is this expected?
  ```
  ➜  epinio git:(2114-delete-all) epinio configuration delete --all             

  🚢  Querying Configurations for Deletion...
  Namespace: workspace

  🚢  Deleting Configurations...
  Names: config-1, config-10, config-2, config-3, config-4, config-5, config-6, config-7, config-8, config-9
  Namespace: workspace
  .......
  ✔️  Configurations Removed.
  Names: config-1, config-10, config-2, config-3, config-4, config-5, config-6, config-7, config-8, config-9
  Namespace: workspace
  ```

[...]

PR #2138, opened 2023-03-14
* delete --all

### LINKED ISSUE #2114: Create epinio deleteAll command for Applications, Services, Namespaces, Configurations
### Request: 
implement a deleteAll command targeting `Applications`, `Configurations`, `Namespaces` or `Services` individually to delete all elements within.
So:

- `epinio app deleteAll`
- `epinio configuration deleteAll`
- `epinio services delete All`
- `epinio namespaces deleteAll`

We have implemented already a bulk deletion endpoint for each, but it is still needed to specify the name of each one (ie: if we have 2 apps  `app-1`, `app-2`, we would need to delete them as `epinio app delete app-1 app-2`

The idea is to list all apps and present either in applications, services or configurations and delete them later.

Something like:

Use case would be:

- Say we have 2 apps `app-1`, `app-2`
- We delete them both with 1 command:
```
epinio app delete ALL 
```

- Output:

```
🚢  Deleting Applications...
Names: app-1, app-2
Namespace: workspace

✔️  Applications Removed.
```

andreas-kupries:
Fix #2114 

todo: namespaces  (more complex due to the pre-run question)
todo: #2131

### DIFF
--- a/internal/cli/configuration.go
+++ b/internal/cli/configuration.go
@@ -32,6 +32,7 @@ func init() {
 	CmdConfiguration.AddCommand(CmdConfigurationList)
 
 	CmdConfigurationList.Flags().Bool("all", false, "list all configurations")
+	CmdConfigurationDelete.Flags().Bool("all", false, "delete all configurations")
 
 	changeOptions(CmdConfigurationUpdate)
 }
@@ -94,7 +95,7 @@ var CmdConfigurationDelete = &cobra.Command{
 	Use:   "delete NAME1 [NAME2 ...]",
 	Short: "Delete one or more configurations",
 	Long:  `Delete configurations by name.`,
-	Args:  cobra.MinimumNArgs(1),
+	Args:  cobra.MinimumNArgs(0),
 	RunE:  ConfigurationDelete,
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		epinioClient, err := usercmd.New(cmd.Context())
@@ -240,12 +241,24 @@ func ConfigurationDelete(cmd *cobra.Command, args []string) error {
 		return errors.Wrap(err, "error reading option --unbind")
 	}
 
+	all, err := cmd.Flags().GetBool("all")
+	if err != nil {
+		return errors.Wrap(err, "error reading option --all")
+	}
+
+	if all && len(args) > 0 {
+		return errors.New("Conflict between --all and named configurations")
+	}
+	if !all && len(args) == 0 {
+		return errors.New("No configurations specified for deletion")
+	}
+
 	client, err := usercmd.New(cmd.Context())
 	if err != nil {
 		return errors.Wrap(err, "error initializing cli")
 	}
 
-	err = client.DeleteConfiguration(args, unbind)
+	err = client.DeleteConfiguration(args, unbind, all)
 	if err != nil {
 		return errors.Wrap(err, "error deleting configuration")
 	}
--- a/internal/cli/delete.go
+++ b/internal/cli/delete.go
@@ -23,7 +23,7 @@ var ()
 var CmdAppDelete = &cobra.Command{
 	Use:   "delete NAME1 [NAME2 ...]",
 	Short: "Deletes one or more applications",
-	Args:  cobra.MinimumNArgs(1),
+	Args:  cobra.MinimumNArgs(0),
 	ValidArgsFunction: func(cmd *cobra.Command, args []string, toComplete string) ([]string, cobra.ShellCompDirective) {
 		epinioClient, err := usercmd.New(cmd.Context())
 		if err != nil {
@@ -37,12 +37,24 @@ var CmdAppDelete = &cobra.Command{
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cmd.SilenceUsage = true
 
+		all, err := cmd.Flags().GetBool("all")
+		if err != nil {
+			return errors.Wrap(err, "error reading option --all")
+		}
+
+		if all && len(args) > 0 {
+			return errors.New("Conflict between --all and named applications")
+		}
+		if !all && len(args) == 0 {
+			return errors.New("No applications specified for deletion")
+		}
+
 		client, err := usercmd.New(cmd.Context())
 		if err != nil {
 			return errors.Wrap(err, "error initializing cli")
 		}
 
-		err = client.Delete(cmd.Context(), args)
+		err = client.Delete(cmd.Context(), args, all)
 		if err != nil {
 			return errors.Wrap(err, "error deleting app")
 		}
--- a/internal/cli/apps.go
+++ b/internal/cli/apps.go
@@ -51,6 +51,8 @@ func init() {
 	CmdAppCreate.Flags().String("app-chart", "", "App chart to use for deployment")
 	CmdAppUpdate.Flags().String("app-chart", "", "App chart to use for deployment")
 
+	CmdAppDelete.Flags().Bool("all", false, "Delete all applications")
+
 	CmdApp.AddCommand(CmdAppCreate)
 	CmdApp.AddCommand(CmdAppChart) // See chart.go for implementation
 	CmdApp.AddCommand(CmdAppEnv)   // See env.go for implementation
@@ -63,8 +65,8 @@ func init() {
 	CmdApp.AddCommand(CmdAppShow)
 	CmdApp.AddCommand(CmdAppExport)
 	CmdApp.AddCommand(CmdAppUpdate)
-	CmdApp.AddCommand(CmdAppDelete)
-	CmdApp.AddCommand(CmdAppPush) // See push.go for implementation
+	CmdApp.AddCommand(CmdAppDelete) // See delete.go for implementation
+	CmdApp.AddCommand(CmdAppPush)   // See push.go for implementation
 	CmdApp.AddCommand(CmdAppRestart)
 	CmdApp.AddCommand(CmdAppRestage)
 }
--- a/internal/cli/services.go
+++ b/internal/cli/services.go
@@ -40,6 +40,7 @@ func init() {
 	CmdServices.AddCommand(CmdServiceList)
 
 	CmdServiceList.Flags().Bool("all", false, "list all services")
+	CmdServiceDelete.Flags().Bool("all", false, "delete all services")
 }
 
 var CmdServiceCatalog = &cobra.Command{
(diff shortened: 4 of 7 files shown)


internal/cli/namespaces.go
```diff
@@ -47,6 +47,7 @@ func init() {
 
 	flags := CmdNamespaceDelete.Flags()
 	flags.BoolVarP(&force, "force", "f", false, "force namespace deletion")
+	flags.Bool("all", false, "delete all namespaces")
```
enrichman, line 50: I would try to define a "standard" for our flags. Personally I like the pointers because it will be already defined. You don't have to get the value with the `cmd.Flags().GetBool("all")` and check the error.

internal/cli/namespaces.go
```diff
@@ -101,31 +102,58 @@ var CmdNamespaceCreate = &cobra.Command{
...
 		}
 
-		return nil
-	},
-	RunE: func(cmd *cobra.Command, args []string) error {
-		cmd.SilenceUsage = true
+		force, err := cmd.Flags().GetBool("force")
+		if err != nil {
+			return err
+		}
```
enrichman, line 125: No need to check and get the value of `force`.

internal/cli/usercmd/poll.go
```diff
@@ -0,0 +1,48 @@
...
+
+import "time"
+
+func (c *EpinioClient) trackDeletion(names []string, poller func() []string) {
+
+	// set of tracked names
+	tracked := map[string]bool{}
+	for _, name := range names {
+		tracked[name] = true
+	}
```
enrichman, line 22: I hate the lack of Set in Go. :upside_down_face:

NITP: if you don't need the boolean value you can use an empty struct `struct{}` instead of the `bool`.

internal/cli/delete.go
```diff
@@ -23,7 +23,7 @@ var ()
 var CmdAppDelete = &cobra.Command{
 	Use:   "delete NAME1 [NAME2 ...]",
 	Short: "Deletes one or more applications",
-	Args:  cobra.MinimumNArgs(1),
+	Args:  cobra.MinimumNArgs(0),
```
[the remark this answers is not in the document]
andreas-kupries: Looking at https://github.com/spf13/cobra/blob/main/user_guide.md#positional-and-custom-arguments
it seems that you are correct. `MinimumNArgs(0)` is essentially `ArbitraryArgs`, and that is the default.

internal/cli/namespaces.go
andreas-kupries: The majority is actually not pointers (`Var` modifier) (The `P` modifier` allows for spec of a shorthand flag letter).
Hm. The `force` thing was added with 1844977e1515a59730f6542dddb01138e77c4349. Used `Var` and then later a regular `GetBool` instead of the variable.

Will see how that looks when it is correct.

internal/cli/namespaces.go
andreas-kupries: See above.

internal/cli/usercmd/poll.go
andreas-kupries: Will do. Had forgotten how write this empty struct.

internal/cli/namespaces.go
```diff
@@ -101,31 +102,58 @@ var CmdNamespaceCreate = &cobra.Command{
...
+			} else {
+				names := strings.Join(args, ", ")
+				cmd.Printf("You are about to delete %d namespaces (%s)\nand everything they include,i.e. applications, configurations, etc.\nAre you sure? (y/n): ",
+					len(args), names)
+			}
+
+			if !askConfirmation(cmd) {
+				return errors.New("Cancelled by user")
+			}
+		}
```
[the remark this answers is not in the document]
andreas-kupries: The `c.UI` does not have functions to simply print a line as is needed for the confirmation ... Ah ... It has its own Ask code. Very different however.

The namespace thing read directly from stdin:

```
	reader := bufio.NewReader(os.Stdin)
	for {
		s, _ := reader.ReadString('\n')
		s = strings.TrimSpace(strings.ToLower(s))
		if strings.Compare(s, "n") == 0 {
			return false
		} else if strings.Compare(s, "y") == 0 {
			break
		} else {
			cmd.Printf("Please enter y or n: ")
			continue
		}
	}
	return true
```

The ui code goes through `fmt`:

```
	var value bool
	fmt.Scanf("%b", &value)
	return value
```

Maybe go through `withAskString()` instead. Also `Scanf`, but leaves interpretation to us in the caller.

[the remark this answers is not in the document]
andreas-kupries: Heh. This is the first time the `WithAsk...` functionality of `termui` is actually used, and it looks to be broken

```
🚢  You are about to delete the namespace 'foxlair' and everything
it includes, i.e. applications, configurations, etc.

🚢  Are you sure? (y/n): 
: %!s(*string=0xc000d5add0)

🚢  Are you sure? (y/n): 
: %!s(*string=0xc000d5ae70)

... (infinity ...)

The read thing does not wait ... Working on fixes now ...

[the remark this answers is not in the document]
Comment (andreas-kupries) [reply internal/cli/namespaces.go:154]: Huh. The `WithAsk` declare themselves internally as `show` ... no wonder it loops

File: internal/cli/usercmd/poll.go (hunk 3 above)
Comment (andreas-kupries) [reply internal/cli/usercmd/poll.go]: Done, commit 0d2e4c6b662550ca809b1d48cdf7101c67b508b2

File: internal/cli/namespaces.go (hunk 2 above)
Comment (andreas-kupries) [reply internal/cli/namespaces.go]: Done, commit 529aea3402fe8b101ae32f79858d6e0a50e607e9

File: internal/cli/namespaces.go (hunk 1 above)
Comment (andreas-kupries) [reply internal/cli/namespaces.go]: Done, commit 529aea3402fe8b101ae32f79858d6e0a50e607e9

File: internal/cli/delete.go (hunk 4 above)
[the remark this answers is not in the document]
Comment (andreas-kupries) [reply internal/cli/delete.go:26]: Done, commit 529aea3402fe8b101ae32f79858d6e0a50e607e9

File: internal/cli/namespaces.go (hunk 5 above)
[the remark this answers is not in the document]
Comment (andreas-kupries) [reply internal/cli/namespaces.go]: Done, commit 529aea3402fe8b101ae32f79858d6e0a50e607e9

With termui fixes in 8e6941433590603751f793d3b74cf57d44018000

File: acceptance/configurations_test.go [hunk 6]
```diff
@@ -403,13 +403,13 @@ var _ = Describe("Configurations", LConfiguration, func() {
 		It("doesn't unbind a service-owned configuration", func() {
 			out, err := env.Epinio("", "configuration", "unbind", config, appName)
 			Expect(err).To(HaveOccurred(), out)
-			Expect(out).To(ContainSubstring("Bad Request: Configuration belongs to service"))
+			Expect(out).To(ContainSubstring("Bad Request: Configuration '" + config + "' belongs to service"))
```
enrichman, line 406:
```suggestion
			Expect(out).To(ContainSubstring("Bad Request: Configuration '%s' belongs to service", config))
```

acceptance/configurations_test.go
```diff
@@ -403,13 +403,13 @@ var _ = Describe("Configurations", LConfiguration, func() {
...
 		})
 
 		It("doesn't delete a bound service-owned configuration", func() {
 			out, err := env.Epinio("", "configuration", "delete", config)
 			Expect(err).To(HaveOccurred(), out)
-			Expect(out).To(ContainSubstring("Configuration belongs to service"))
+			Expect(out).To(ContainSubstring("Configuration '" + config + "' belongs to service"))
```
enrichman, line 412:
```suggestion
			Expect(out).To(ContainSubstring("Configuration '%s' belongs to service", config))
```

internal/cli/delete.go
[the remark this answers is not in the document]
enrichman: It seems that the change is not in that commit (also Github is not showing it as `outdated`).

internal/cli/configuration.go
```diff
@@ -94,7 +95,7 @@ var CmdConfigurationDelete = &cobra.Command{
 	Use:   "delete NAME1 [NAME2 ...]",
 	Short: "Delete one or more configurations",
 	Long:  `Delete configurations by name.`,
-	Args:  cobra.MinimumNArgs(1),
+	Args:  cobra.MinimumNArgs(0),
```
enrichman, line 98: Same as the other one (we can drop this).

enrichman (changes requested):
I missed a couple of small things in the last review.

I cannot see any new test, just a couple of fix in the existing ones. I think we should add at least one or two for the new `--all` flag (in the cli and api).

internal/cli/delete.go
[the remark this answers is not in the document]
andreas-kupries: Checking ...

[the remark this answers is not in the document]
andreas-kupries: Indeed, was not done :(

acceptance/configurations_test.go
andreas-kupries: Does `ContainSubstring` handle printf-style placeholders ?
I suspect a `fmt.Sprintf` should be there. Will make the change.

As for tests of `--all`. I left them out intentionally.

Having them are, I strongly believe, a bad idea.

CI has many tests running concurrently on the same E install. A test invoking any kind of `delete --all` will very likely interfere with all tests running in parallel to it, i.e. their services, applications, and configurations, i.e. break them. And note that for a suitable user, i.e. admin, `namespace delete --all` is just one step short of E uninstall.

acceptance/configurations_test.go
```diff
@@ -427,7 +427,7 @@ var _ = Describe("Configurations", LConfiguration, func() {
 
 			out, err = env.Epinio("", "configuration", "delete", config)
 			Expect(err).To(HaveOccurred(), out)
-			Expect(out).To(ContainSubstring("Configuration belongs to service"))
+			Expect(out).To(ContainSubstring("Configuration '" + config + "' belongs to service"))
```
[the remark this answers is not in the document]
andreas-kupries: Done, commit d6da9899c447c965d022a6feb586ddf2e6a028bb

internal/cli/delete.go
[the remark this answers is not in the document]
andreas-kupries: Done. Commit 0f1736a349fed024b42a92d6b29be940e1215fe8

andreas-kupries:
Regarding tests of `--all`. I left them out intentionally.

Having them are, I strongly believe, a bad idea.

CI has many tests running concurrently on the same E install. A test invoking any kind of `delete --all` will very likely interfere with all tests running in parallel to it, i.e. their services, applications, and configurations, i.e. break them. And note that for a suitable user, i.e. admin, `namespace delete --all` is just one step short of E uninstall.

PR #2143, opened 2023-03-15
Fixed missing app status

### LINKED ISSUE #2134: [UI] Deployed apps appear with "Unknown" state in UI
### Issue 
Correctly deployed apps appear in `Unknown` status in UI.

### Steps to reproduce:

- Pull latest version of epino (Currently: `Epinio Server Version: v1.7.0-39-g3d7c647f`)
- Deploy epinio. (I used k3d)

```
➜  epinio git:(main) ✗ make acceptance-cluster-delete  make acceptance-cluster-setup                                                                           
export KUBECONFIG=$PWD/tmp/acceptance-kubeconfig 
make install-cert-manager
make prepare_environment_k3d      
```
- Push an app either from cli or ui. I used:

```
➜  epinio git:(main) ✗ epinio push -n testapp -p assets/sample-app
```

### Observed behavior:
App seems correctly pushed and it is reachable. However UI displays `Unknown` status and several parts or UI seem not working:

Image

- Routes links
- Status is in 'Unkown' state
- Dashboard page recognizes 0 apps running:

Image

### Expected behavior

- State should show `Running`
- Routes link should work
- Dashboard Running apps should match the ones in `Running` state

enrichman:
Fix #2134

Moving the logic from the `fetch` we have missed the `calculateStatus` func that was done AFTER in the Lookup.

Since it doesn't seems fine to me to fetch an app and then calculate somehow the state of it just after, I've moved that logic to inside the `fetch` func. I've then added the missing piece into the  `aggregate` and added a small test for the API.


andreas-kupries:
Oh, so this was caused by the new code to make `app list` fast. :( Thanks for investigation and fix.

internal/application/application.go
```diff
@@ -789,31 +828,25 @@ func fetch(ctx context.Context, cluster *kubernetes.Cluster, app *models.App) er
...
-	if app.Status == models.ApplicationError {
-		return nil
+	if err != nil {
+		err = errors.Wrap(err, "workload loading")
+		app.StatusMessage = err.Error()
+		app.Status = models.ApplicationError
+		return err
 	}
+
 	staging, err := CurrentlyStaging(ctx, cluster, app.Meta.Namespace, app.Meta.Name)
```
andreas-kupries, line 838: This is called per application, right ?
Does it cause a kube API call ?
If yes, could the information be pulled out of the list-loaded data ?
And if not, maybe out of a another list-load ?

enrichman: Ah yes, it does. I guess so, I'll have a look at it. :+1:

andreas-kupries: good luck

internal/application/application.go
```diff
@@ -136,16 +140,37 @@ func Exists(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef)
...
 
-	// Check all jobs for the app for activity.
-	selector := fmt.Sprintf("app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s",
-		appName, namespace)
+// currentlyStaging is a utility func that will load a map of the status of the application's staging jobs
+// If no appName is specified it will load a complete map, otherwise the map will contain just the status of job of the specified app
+func currentlyStaging(ctx context.Context, cluster JobLister, namespace, appName string) (map[string]bool, error) {
+	stagingJobsMap := make(map[string]bool)
+
+	// filter the jobs in the namespace
```
andreas-kupries, line 160: Does this work for `epinio app list --all` ? This listing across all namespaces is signaled by `namespace == ""`.

andreas-kupries (changes requested):

internal/application/application.go
enrichman: The previous implementation was using the same logic with the plain labels (not in a map, but I guess it's the same).
```
selector := fmt.Sprintf("app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s", appName, namespace)
```

 I'll do a manual check to be sure.

We should have some tests for the `--all` flag, anyway. If not, we should add them. :slightly_smiling_face:
I'll have a look also at this. :+1:

andreas-kupries: __Beware__ The list shows the status as `1/1` etc. and that info comes out of the main app pods.
Not from the staging jobs.
What does the `list --all` show while the apps are actually in staging, not running ?

andreas-kupries: You will need multiple terminals, one per app to put into staging, plus the terminal where `list --all` is run.
Best to find some app which takes long(er) to stage.

andreas-kupries (approved):

internal/application/application.go
enrichman: In the end there was an error about namespaces, so thanks for blocking this!

I'll probably open one more PR with some tests to have a wider coverage of this APIs. :+1:

PR #2144, opened 2023-03-16
Fix workflow syntax and test with runner labels

svollath:
Try fixing delete_clusters.yml syntax and test EKS workflow - linter to choose runner by label only


andreas-kupries (approved):
I cannot say much about the differences in the syntax / event input fields.

Issue #1792, opened 2022-10-13
[Epic] Allow to override default port


olblak:
**Is your feature request related to a problem? Please describe.**
Similar to https://github.com/epinio/epinio/issues/1791
I would like to be able to override default application port quickly

**Describe the solution you'd like**
I would like to be able to run `epinio push --name myapplication --port 1313`

**Describe alternatives you've considered**
We discussed the possibility to do that via custom helm Chart but we found it a bit heavy for this kind of setting

**Additional context**

---

Updates (@enrichman): https://github.com/epinio/epinio/issues/1792#issuecomment-1438034128  

---

## Tasks
- [x] https://github.com/epinio/helm-charts/pull/391
- [x] https://github.com/epinio/ui/issues/140
- [x] https://github.com/epinio/epinio/issues/2126

agracey:
I think this can be done more easily by templating the service yaml based on a value to be overwritten using 

`epinio app push -n <some name> -v port 1313`

agracey:
I think there might be several places like this where our default app template could be extended like this without needing to expose it with it's own flag.

enrichman:
I agree on extending our default template. So in order to complete this we will need to:

- [ ] update the [application chart](https://github.com/epinio/helm-charts/tree/main/chart/application), and use the new one as [default](https://github.com/epinio/helm-charts/blob/epinio-application-0.1.24/chart/epinio/templates/default-app-chart.yaml#L15) 

this should be enough since the `-v/--chart-value` flag is already covered, but we need also to

- [ ] add the values to the UI
- [ ] tests for app upgrades from the old app chart to the new one (not sure if we have these)
- [ ] tests for the UI
- [ ] update documentation accordingly

PR #2150, opened 2023-03-21
Allow to override default application port

giubacc:
The port is pushed inside the application container via the environment variable: PORT.
The value defaults to 8080; it can be specified using a custom chart value during the application's push:

epinio push -n 'app-name' --chart-value 'appListeningPort=8087'

Addresses: #1792

### DIFF
--- a/acceptance/api/v1/appchart_index_test.go
+++ b/acceptance/api/v1/appchart_index_test.go
@@ -65,6 +65,6 @@ var _ = Describe("ChartList Endpoint", LAppchart, func() {
 		Expect(short).Should(ContainElements(
 			"Epinio standard deployment"))
 		Expect(chart).Should(ContainElements(
-			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.24/epinio-application-0.1.24.tgz"))
+			"https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.25/epinio-application-0.1.25.tgz"))
 	})
 })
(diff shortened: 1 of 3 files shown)


enrichman (changes requested):
To test this you need to open a matching PR on the `helm-charts` repository with the required changes (the `0.1.25-rc1` chart, and the other things), and update the submodules of this one.

If those are ok we can bump the application chart to a stable release, removing the `-rc1` suffix before merging them.

Do we want to enable even lower port values? What if the standard `80` or `443` ports are needed?

giubacc:
> To test this you need to open a matching PR on the `helm-charts` repository with the required changes (the `0.1.25-rc1` chart, and the other things), and update the submodules of this one.
> 
> If those are ok we can bump the application chart to a stable release, removing the `-rc1` suffix before merging them.
> 

ok

> Do we want to enable even lower port values? What if the standard `80` or `443` ports are needed?

Initially I thought this could be the right thing to do, but then I thought that could be uncommon opening a well-known-port directly on a pod.
On the other hand, it is true that you couldn't have the control on which port you can open on your application (even I find it hard to think of a real case).
I dunno.. @andreas-kupries  what do you think?

enrichman (approved):

thehejik:
Verified fixed in `v1.8.0-rc1`
- over Web UI deployed a simple py http server app listening on port 1234 - entered AppListeningPort = 1234
- ingress endpoint forwards to the app pod correctly and the app works
image


The app:
```
cat index.html
<H1>Hello World</H1>
---
cat Procfile
web: python server.py
---
cat server.py
#!/usr/bin/env python3 -u
import http.server
import socketserver

PORT = 1234

Handler = http.server.SimpleHTTPRequestHandler

with socketserver.TCPServer(("", PORT), Handler) as httpd:
    print("serving at port", PORT)
    httpd.serve_forever()
---
```

PR #2151, opened 2023-03-23
Fixed status information shown for deployed app with zero instances

### LINKED ISSUE #1726: Bad status message for app with zero instances
**Describe the bug**

An application pushed with zero instances, or scaled down to zero instances from running or more, will have a status of `not deployed, staging failed`.

While `not deployed` is arguably correct the `staging failed` is definitely not.

**To Reproduce**

  1. On an installed epinio invoke `epinio push --name foo --path assets/sample-app --instances 0`.
  2. Look at the `Status` line printed by `epinio app show foo`.

**Expected behavior**

A better message. Mainly `staging failed` should not be present.

andreas-kupries:
fix #1726 

When there is no workload, look at desired instances to distinguish between failed staging and scaled to zero

### DIFF
--- a/internal/application/application.go
+++ b/internal/application/application.go
@@ -140,22 +140,28 @@ func Exists(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef)
 	return true, nil
 }
 
-// IsCurrentlyStaging returns true if the app is staging (there is an active Job for this application).
-// If you need this information for more than one app please use the CurrentlyStaging(ctx context.Context, cluster JobLister, namespace string)
+// IsCurrentlyStaging returns true if the named application is staging (there is an active Job for
+// this application).  If this information is needed for more than one application use StagingStati
+// instead.
 func IsCurrentlyStaging(ctx context.Context, cluster JobLister, namespace, appName string) (bool, error) {
-	currentlyStagingApps, err := currentlyStaging(ctx, cluster, namespace, appName)
-	return currentlyStagingApps[EncodeConfigurationKey(appName, namespace)], err
+	staging, err := stagingStatus(ctx, cluster, namespace, appName)
+	if err != nil {
+		return false, err
+	}
+	status := staging[EncodeConfigurationKey(appName, namespace)]
+	return status == models.ApplicationStagingActive, nil
 }
 
-// CurrentlyStaging returns a map of applications that are currently in a staging status
-func CurrentlyStaging(ctx context.Context, cluster JobLister, namespace string) (map[ConfigurationKey]bool, error) {
-	return currentlyStaging(ctx, cluster, namespace, "")
+// StagingStati returns a map of applications and their staging stati
+func StagingStati(ctx context.Context, cluster JobLister, namespace string) (map[ConfigurationKey]models.ApplicationStagingStatus, error) {
+	return stagingStatus(ctx, cluster, namespace, "")
 }
 
-// currentlyStaging is a utility func that will load a map of the status of the application's staging jobs
-// If no appName is specified it will load a complete map, otherwise the map will contain just the status of job of the specified app
-func currentlyStaging(ctx context.Context, cluster JobLister, namespace, appName string) (map[ConfigurationKey]bool, error) {
-	stagingJobsMap := make(map[ConfigurationKey]bool)
+// stagingStatus is a utility function loading a map of the status of the application's staging jobs
+// (active, done, error).  If no appName is specified it will load a complete map, otherwise the map
+// will contain only the status of the job of the specified app
+func stagingStatus(ctx context.Context, cluster JobLister, namespace, appName string) (map[ConfigurationKey]models.ApplicationStagingStatus, error) {
+	stagingJobsMap := make(map[ConfigurationKey]models.ApplicationStagingStatus)
 
 	// filter the jobs in the namespace
 	labelsMap := make(map[string]string)
@@ -182,19 +188,18 @@ func currentlyStaging(ctx context.Context, cluster JobLister, namespace, appName
 		return condition.Status == v1.ConditionTrue && condition.Type == apibatchv1.JobFailed
 	}
 
-	done := func(condition apibatchv1.JobCondition) bool {
-		return completed(condition) || failed(condition)
-	}
-
-	jobStaging := func(job apibatchv1.Job) bool {
+	jobStaging := func(job apibatchv1.Job) models.ApplicationStagingStatus {
 		for _, condition := range job.Status.Conditions {
-			if done(condition) {
+			if failed(condition) {
+				return models.ApplicationStagingFailed
+			}
+			if completed(condition) {
 				// Terminal, not staging
-				return false
+				return models.ApplicationStagingDone
 			}
 		}
 		// No terminal condition found on the job, it is actively staging
-		return true
+		return models.ApplicationStagingActive
 	}
 
 	for _, job := range jobList.Items {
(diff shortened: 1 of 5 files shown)


andreas-kupries:
> Do we have a test where `staging failed` for real? Apart this question, lgtm.

We actually have a test using a failed staging. It tests that re-pushing with a fix clears the failure and succeeds.
See
https://github.com/epinio/epinio/blob/145cda2d4630f5afc473a7e1db24bbb8c6c2fcee/acceptance/apps_test.go#L613
and following lines

I should be able to extend that with a test checking the output of `app show` and `app list`.

giubacc (dismissed):

mmartin24:
Checked in the branch and working ok.
Checked also created from IU with 0 instances and I confirmed the message is correctly displayed.

andreas-kupries:
> Huh:
> 
> ```
> Origin               | <<undefined>>                                                                                       |
>       | Created              | 2023-03-23 17:02:02 +0000 UTC                                                                       |
>       | Status               | 0/1                                                                                                 |
> ...
> ```
> 
> So, with staging failed (in the test) there is a workload, so we see actual/desired info. Ok. Another round of updating the new test. And squashing the fixup history.

Damn ... It can be both. It seems depending on invisible details, maybe a race, we can see either the `0/1` or the `... staging failed` here. :( ... Guess I have to make both acceptable in the RE.

andreas-kupries:
Will fix the conflict here.

andreas-kupries:
Note, due to the merge conflict fixing this really needs CI to be green to be considered ok.

mmartin24:
I just checked at this point  a47678ea (HEAD -> 1726-zero-scale-status, origin/1726-zero-scale-status) fix: allow any status details

**Case 1**.Push correctly deployed with 0 instances (app name: `foo`) -> message  deployed, scaled to zero. This seems ok
**Case 2**. Push staging error app with 0 instances (app name: `errorstgapp-0-inst`) -> status is deployed, scaled to zero. @andreas-kupries  Is this ok? I thought it should say `staging failed`.   Perhaps I am missing anything?
```
➜  epinio git:(1726-zero-scale-status) epinio app show errorstgapp-0-inst 

🚢  Show application details
Namespace: workspace
Application: errorstgapp-0-inst

✔️  Details:
|         KEY          |                      VALUE                       |
|----------------------|--------------------------------------------------|
| Origin               | <<undefined>>                                    |
| Created              | 2023-03-30 11:49:34 +0200 CEST                   |
| Status               | deployed, scaled to zero                         |
| Last StageId         | 06ac983cbed25ea8                                 |
| Desired Routes       |                                                  |
|                      | errorstgapp-0-inst.172.18.0.3.omg.howdoi.website |
| App Chart            | standard                                         |
| Desired Instances    |                                                0 |
| Bound Configurations |                                                  |
| Environment          |                                                  |
| Chart Values         |                                                  |

```

**Case 3 (control)** Push staging error app with 1 instance (app name: `errorstgapp-1-inst`) -> status is not deployed, staging failed - This seems ok


```
➜  epinio git:(1726-zero-scale-status) epinio app list 

🚢  Listing applications
Namespace: workspace

✔️  Epinio Applications:
|        NAME        |            CREATED             | STATUS | ROUTES | CONFIGURATIONS | STATUS DETAILS |
|--------------------|--------------------------------|--------|--------|----------------|----------------|
| errorstgapp-0-inst | 2023-03-30 11:49:34 +0200 CEST | 0/0    | n/a    |                |                |
| errorstgapp-1-inst | 2023-03-30 11:50:01 +0200 CEST | n/a    | n/a    |                |                |
| foo                | 2023-03-30 11:47:25 +0200 CEST | 0/0    | n/a    |                |                |
```

andreas-kupries:
> Push staging error app with 0 instances (app name: errorstgapp-0-inst) -> status is deployed, scaled to zero.
> @andreas-kupries Is this ok? I thought it should say staging failed. Perhaps I am missing anything?

I am not sure how can distinguish this from a plain app with zero instances ...
Both have `desired instances == 0`, and `no workload (i.e. no pods)`.

... thinking ... where is the stage id stored ... maybe that could be used as additional factor

andreas-kupries:
Not sure. Strong preference for static information.

@mmartin24 : What do you use as `staging error app ` ?

mmartin24:
> Not sure. Strong preference for static information.
> 
> @mmartin24 : What do you use as `staging error app ` ?

```
epinio push --name errorstgapp-0-inst --path assets/tests --instances 0
```

andreas-kupries:
Both cases come with a proper stage id ... Only way to distinguish I see now is looking at the status of the stage pod.

andreas-kupries:
Yes, that looks to be doable. ... In the works now ...

internal/application/application.go
```diff
@@ -330,13 +335,13 @@ func List(ctx context.Context, cluster *kubernetes.Cluster, namespace string) (m
 		requestctx.Logger(ctx).Error(err, "metrics not available")
 	}
 
-	// VI. load all the status of the staging jobs
+	// VI. load the stati of all staging jobs
```
enrichman, line 338: Is the plural of status "statuses", or not?

internal/application/application.go
```diff
@@ -140,22 +140,28 @@ func Exists(ctx context.Context, cluster *kubernetes.Cluster, app models.AppRef)
...
+	}
+	status := staging[EncodeConfigurationKey(appName, namespace)]
+	return status == models.ApplicationStagingActive, nil
 }
 
-// CurrentlyStaging returns a map of applications that are currently in a staging status
-func CurrentlyStaging(ctx context.Context, cluster JobLister, namespace string) (map[ConfigurationKey]bool, error) {
-	return currentlyStaging(ctx, cluster, namespace, "")
+// StagingStati returns a map of applications and their staging stati
+func StagingStati(ctx context.Context, cluster JobLister, namespace string) (map[ConfigurationKey]models.ApplicationStagingStatus, error) {
```
enrichman, line 156: Same (stati?)

internal/application/application.go
andreas-kupries: I believe that `stati` is the plural

internal/application/application.go
andreas-kupries: see above

andreas-kupries:
Looking at server logging the client makes patch request when upating an application, like the number of desired instances. This patch request triggers an update of the app helm chart. Except when trying to scale the app from zero instances. There the server receives the patch request and no chart update looks to be triggered.

Investigating that part of the API now.

internal/application/application.go
enrichman: Not sure if this conversation was resolved. Same for the other one :slightly_smiling_face:

mmartin24: Agree what @enrichman : https://www.grammarbook.com/blog/singular-vs-plural/plural-of-status/ 
It also comes to mind more frequently

mmartin24 (approved):
I tested and clarified offline questions about code with @andreas-kupries.
Approving since it LGTM, but I'd suggest attending this comment regarding the function name of `StagingStati` plural https://github.com/epinio/epinio/pull/2151/files#r1153296692

internal/application/application.go
andreas-kupries: Addressing with #2180

internal/application/application.go
andreas-kupries: Addressing with #2180

PR #2180, opened 2023-04-03
chore: stati -> statuses

andreas-kupries:
Addressing https://github.com/epinio/epinio/pull/2151#pullrequestreview-1365176396


andreas-kupries:
I am merging despite the failure. The failures are not in the changed area, and the tests were slow, with a lot of flakes.
I consider this to be still fallout/recovery from the overload of CI yesterday.

PR #2189, opened 2023-04-07
Use `golangci-lint` on lint

enrichman:
This PR drops the `go vet` linter in favor of the more configurable `golangci-lint`.

[Golangci-lint](https://golangci-lint.run/) is a metalinter, with sane [defaults](https://golangci-lint.run/usage/linters/#enabled-by-default).

We are already using `golangci-lint` with also more linters ([.golangci.yml](https://github.com/epinio/epinio/blob/main/.golangci.yml)) but the check was isolated to its action. This PR use this linter for all the other lintings, it updates the version to the `1.52` (`1.50` was actually erroring for some bugs), and it fixes an error for `funlen` (the `Update` func became too long).


giubacc (approved):
lgtm

PR #2198, opened 2023-04-12
fix: security, bump docker/docker directly

### LINKED ISSUE #2176: Address security issues
Resolve reported security issues: https://github.com/epinio/epinio/security/dependabot

andreas-kupries:
Fix #2176

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -91,7 +91,7 @@ require (
 	github.com/dchest/uniuri v1.2.0
 	github.com/docker/cli v20.10.21+incompatible // indirect
 	github.com/docker/distribution v2.8.1+incompatible // indirect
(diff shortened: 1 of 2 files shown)


enrichman (approved):
That's unfortunate that we need to update directly a dependency but having a look at the tree we would need to go too deep asking for updates!

```
-> % go mod why -m github.com/docker/docker\

# github.com/docker/docker
github.com/epinio/epinio/internal/helm
github.com/mittwald/go-helm-client
helm.sh/helm/v3/pkg/registry
oras.land/oras-go/pkg/auth/docker
github.com/docker/docker/api/types
```

:sweat_smile: 

Build is passing and it's a patch, so we should be fine!

andreas-kupries:
> we would need to go too deep asking for updates!

Yes.

PR #2204, opened 2023-04-14
Added logging ADR

enrichman:
This PR adds a new ADR proposal open for discussion.

It should lead to a new feature for the CLI (the `--log-file` flag) and a new coding style document, with initially only some reference and the logging guidelines.


mmartin24 (dismissed):
LGTM.
As commented I'd vote as personal preference for the --log-file option

mmartin24 (approved):

PR #2217, opened 2023-04-17
Added timeouts to setup-go and linting steps

enrichman:
CI is hanging during the setup-go and golangci-lint actions that are downloading a lot of MB.

Adding some timeouts to fail fast.


andreas-kupries (approved):
Heavy heart here. Wondering what would help to keep things quick.

thehejik (approved):
At least our runners wont be blocked anymore.

PR #2229, opened 2023-04-20
Fix runners Go version with patch

enrichman:
We are hitting always the cache limit for our actions: https://github.com/epinio/epinio/actions/caches

This could be the reason for the errors that we are getting while setting up Go. With this we are going to fix the Go version to a specific one, and it should force the runners to have always the same version.


svollath (approved):
LGTM

andreas-kupries (approved):

PR #2233, opened 2023-04-21
Update release checklist with Dashboard and UI backend

enrichman:
Updated Release Checklist to contain also the information on how to tag/release the Epinio UI (dashboard and ui-backend).

Added also info about the Helm charts release candidate.

### DIFF
--- a/.github/ISSUE_TEMPLATE/release_checklist.yaml
+++ b/.github/ISSUE_TEMPLATE/release_checklist.yaml
@@ -27,6 +27,30 @@ body:
         ---
 
   - type: checkboxes
+    id: epinio-ui-checklist
+    attributes:
+      label: Epinio UI [OPTIONAL]
+      options:
+        - label: >
+            **( 📝 Manual step )**  Tag the Epinio Dashboard in the `epinio-dev` branch in the `rancher/dashboard`.
+            The tag needs to be in the `epinio-standalone-vX.Y.Z-a.b.c` format.
+            `git tag -a epinio-standalone-vX.Y.Z-a.b.c -m 'epinio-standalone-vX.Y.Z-a.b.c'`
+            Check the release action, and the generated bundle.
+            [LINK](https://github.com/rancher/dashboard/actions/workflows/release-rancher-epinio-standalone.yml)
+            [TAGS](https://github.com/rancher/dashboard/tags)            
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/release_checklist.yaml
```diff
@@ -26,6 +26,32 @@ body:
...
 
+  - type: checkboxes
+    id: epinio-ui-checklist
+    attributes:
+      label: Epinio UI [OPTIONAL]
+      options:
+        - label: >
+            **( 📝 Manual step )**  Tag the Epinio Dashboard in the `epinio-dev` branch in the `rancher/dashboard`.
+            The tag needs to be in the `epinio-standalone-vX.Y.Z-a.b.c` format.
+            `git tag -a epinio-standalone-vX.Y.Z-a.b.c -m 'epinio-standalone-vX.Y.Z-a.b.c'`
```
andreas-kupries, line 37: Where do the values for `X`, `Y`, `Z`, `a`, `b`, and `c` come from ?

andreas-kupries (changes requested):

.github/ISSUE_TEMPLATE/release_checklist.yaml
andreas-kupries: I did not ask about `why`, only about `where`. The given link helps. Basically `X.Y.Z` is the E release, and `a.b.c` some kind of sequence number within the release. Can this be added to the checklist ?

enrichman: Yes, but also this is not enforced. The dashboard is following its own versioning, so we can agree that it should be this one, but it could also be a completely different one. It happened to had different versions, because of new tags used to fix some issues, while the Epinio was still in the same version.

Also we could have the same dashboard for different Epinio versions, if nothing changed or we didn't update anything on that.

It's better to keep the number closer btw, at least at the moment. So let's use this as a rule of thumb (`X.Y.Z` should be the Epinio release).

enrichman: Added one more line adding the details of this. :+1:

andreas-kupries (approved):

PR #2235, opened 2023-04-21
fix: missing tab completion for `app env unset`.

### LINKED ISSUE #2228: Epinio Autocompletion not working for env vars
## Steps to reproduce

- Deploy an app with env vars. For example:
 ```
epinio push --name testapp-githuburl --git https://github.com/epinio/example-go,main --env='BP_KEEP_FILES=static/*'
```
- Once deployed type on cli type the following and press tab for auto-completion
```
epinio app env unset testapp-githuburl
```

## Observed behavior
The autocompletion does not suggest `BP_KEEP_FILES=static/*` env var that it is using:

```
➜  ~ epinio app show testapp-githuburl                           

🚢  Show application details
Namespace: workspace
Application: testapp-githuburl

✔️  Details:
|         KEY          |                                            VALUE                                             |
|----------------------|----------------------------------------------------------------------------------------------|
| Origin               | https://github.com/epinio/example-go                                                         |
|                      | @ main                                                                                       |
| Created              | 2023-04-20 11:08:40 +0200 CEST                                                               |
| Status               | 1/1                                                                                          |
| Username             | admin                                                                                        |
| Running StageId      | ac41232a19109512                                                                             |
| Last StageId         | ac41232a19109512                                                                             |
| Age                  | 4m13s                                                                                        |
| Internal Route       | rtestapp-githuburl-ba3062a43dfb72953c31ef55908798a69a4b9fe3.workspace.svc.cluster.local:8080 |
| Active Routes        |                                                                                              |
|                      | testapp-githuburl.192.168.16.3.omg.howdoi.website                                            |
| App Chart            | standard                                                                                     |
| Desired Instances    |                                                                                            1 |
| Bound Configurations |                                                                                              |
| Environment          |                                                                                              |
|   - BP_KEEP_FILES    | static/*                                                                                     |
| Chart Values         |                                                                                              |

✔️  Instances: 
|                              NAME                               | READY | MEMORY  | MILLICPUS | RESTARTS |  AGE  |
|-----------------------------------------------------------------|-------|---------|-----------|----------|-------|
| rtestapp-githuburl-ba3062a43dfb72953c31ef55908798a69a4b9feqvlv4 | true  | 1.8 MiB |         0 |        0 | 4m13s |

```

## Expected behavior:
Autocomplete the env vars as well

Tested on:
```
✔️  Epinio Environment
Platform: k3s
Epinio Server Version: v1.7.1-36-g8ee823f4
Epinio Client Version: v1.7.1-36-g8ee823f4
```

### PR DESCRIPTION
**Pull request by andreas-kupries**
fix #2228

### DIFF
--- a/acceptance/appcharts_test.go
+++ b/acceptance/appcharts_test.go
@@ -153,4 +153,29 @@ var _ = Describe("apps chart", LAppchart, func() {
 			Expect(out).To(ContainSubstring("Name: not set, system default applies"))
 		})
 	})
+
+	for _, command := range []string{
+		"default",
+		"show",
+	} {
+		Context(command+" command completion", func() {
+			It("matches empty prefix", func() {
+				out, err := env.Epinio("", "__complete", "app", "chart", command, "")
+				Expect(err).ToNot(HaveOccurred(), out)
+				Expect(out).To(ContainSubstring(chartName))
+			})
+
+			It("does not match unknown prefix", func() {
+				out, err := env.Epinio("", "__complete", "app", "chart", command, "bogus")
+				Expect(err).ToNot(HaveOccurred(), out)
+				Expect(out).ToNot(ContainSubstring("bogus"))
+			})
+
+			It("does not match bogus arguments", func() {
+				out, err := env.Epinio("", "__complete", "app", "chart", command, chartName, "")
+				Expect(err).ToNot(HaveOccurred(), out)
+				Expect(out).ToNot(ContainSubstring(chartName))
+			})
+		})
+	}
 })
--- a/acceptance/apps_env_test.go
+++ b/acceptance/apps_env_test.go
@@ -100,6 +100,50 @@ var _ = Describe("apps env", LApplication, func() {
 
 				Expect(deployedEnv(namespace, appName)).ToNot(ContainSubstring("MYVAR"))
 			})
+
+			Context("command completion", func() {
+				// MYVAX is intentionally different from MYVAR, to avoid possible clashes
+				BeforeEach(func() {
+					out, err := env.Epinio("", "apps", "env", "set", appName, "MYVAX", "myvalue")
+					Expect(err).ToNot(HaveOccurred(), out)
+				})
+
+				AfterEach(func() {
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Review by enrichman — CHANGES_REQUESTED** [review]
This seems to have a small bug. The autocompletion works even after that we have provided the 2 values, suggesting the app name. The check on the lenght should be `>1`.

It's also the same func that we are using for the `app env show`, so it has the same bug. I would extract the two and use the same. A switch case is also probably easier to read:

```go
app, err := usercmd.New(cmd.Context())
if err != nil {
	return nil, cobra.ShellCompDirectiveNoFileComp
}
app.API.DisableVersionWarning()

var matches []string
switch len(args) {
case 0:
	// #args == 0: application name.
	matches = app.AppsMatching(toComplete)
case 1:
	// #args == 1: environment variable name (in application)
	matches = app.EnvMatching(cmd.Context(), args[0], toComplete)
}

return matches, cobra.ShellCompDirectiveNoFileComp
```

Since it's hard to debug this printing out I tried to understand what was happening with my "log file proposal", so I added this small snippet at the beginning of the validation:

```go
f, err := os.OpenFile("cli.log", os.O_RDWR|os.O_APPEND|os.O_CREATE, 0644)
if err != nil {
	log.Fatal(err)
}
fmt.Fprintf(f, "args[%s], toComplete: %s\n", strings.Join(args, ", "), toComplete)
```

It's probably also possible attaching a debugger but I'm not very comfortable with it yet.

enrichman:
I've also seen that we are missing the error checking for some API calls (the `AppsMatching` and `EnvMatching`).
We should fix this here, or adding a new issue.

andreas-kupries:
We generally test the match endpoints, see `acceptance/api/...`.
We currently have no tests for the cli completions build on top of the matcher endpoints.
I will start on adding these, using the hidden `__complete` command as entry.

mmartin24:
Checked autocompletion for env var works for `app env unset`
Checked it does not repeat the name of app nor env
Check also fixed for `epinio app env show`

andreas-kupries:
tab completion tests complete

enrichman (approved):

PR #2241, opened 2023-04-25
Extended application with additional origin information maintained by the web ui, and centralized manifest formatting in the server

### LINKED ISSUE #2224: Get application manifest API
We are missing an API to retrieve/download the application manifest, that it's derived and built from the clients from the "show api".

It will be nice to have this to export the YAML manifest in a clear and centralized way.

**Additional context**
This will fix the UI export in JSON, and it will give us a better way to provide the needed information to remove the `EPINIO_APP_DATA` environment variable.

### LINKED ISSUE #2225: [CLI] Fetch the application manifest from the manifest API
Depends on: https://github.com/epinio/epinio/issues/2224

The CLI will need to fetch the application manifest from the API

### LINKED ISSUE #2237: Implement `ADR extend the application manifest` in the server and CLI
Do the server- and CLI-side implementation of the changes specified by #2221

andreas-kupries:
fix #2237
fix #2224
fix #2225

feat: extended part endpoint to serve a formatted manifest
chore: new test for manifest retrieval
chore: updated client to pull manifest via part API instead of formatting it itself
note: no changes to existing manifest tests

subordinate PRs:

  - application - https://github.com/epinio/application/pull/12
  - helm chart - https://github.com/epinio/helm-charts/pull/419

### DIFF
--- a/acceptance/api/v1/application_export_test.go
+++ b/acceptance/api/v1/application_export_test.go
@@ -100,7 +100,6 @@ configuration:
   appchart: standard
 origin:
   container: splatform/sample-app
-  git: {}
 namespace: %s
 `, app, domain, namespace)
 
--- a/acceptance/api/v1/application_export_test.go
+++ b/acceptance/api/v1/application_export_test.go
@@ -32,6 +32,10 @@ var _ = Describe("AppPart Endpoint", LApplication, func() {
 	)
 	containerImageURL := "splatform/sample-app"
 
+	// The testsuite checks using only part `values` and part `manifest`, as the smallest
+	// possible parts, and also (YAML-formatted) text.  The data returned for parts `chart` and
+	// `image` will be much much larger, and binary.
+
 	BeforeEach(func() {
 		domain = catalog.NewTmpName("exportdomain-") + ".org"
 
@@ -47,10 +51,7 @@ var _ = Describe("AppPart Endpoint", LApplication, func() {
 		env.DeleteNamespace(namespace)
 	})
 
-	It("retrieves the named application part", func() {
-		// The testsuite checks using only part `values`, as the smallest possible, and also text.
-		// The parts `chart` (and, in the future, maybe, `image`) are much larger, and binary.
-
+	It("retrieves the named application part, values", func() {
 		response, err := env.Curl("GET", fmt.Sprintf("%s%s/namespaces/%s/applications/%s/part/values",
 			serverURL, v1.Root, namespace, app), strings.NewReader(""))
 		Expect(err).ToNot(HaveOccurred())
@@ -78,6 +79,35 @@ var _ = Describe("AppPart Endpoint", LApplication, func() {
   tlsIssuer: epinio-ca
   username: admin
 `, app, domain, domain)))
+	})
+
+	It("retrieves the named application part, manifest", func() {
+		response, err := env.Curl("GET", fmt.Sprintf("%s%s/namespaces/%s/applications/%s/part/manifest",
+			serverURL, v1.Root, namespace, app), strings.NewReader(""))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(response).ToNot(BeNil())
+		defer response.Body.Close()
+
+		bodyBytes, err := io.ReadAll(response.Body)
+		Expect(err).ToNot(HaveOccurred())
+		Expect(response.StatusCode).To(Equal(http.StatusOK), string(bodyBytes))
+
+		expecting := fmt.Sprintf(`name: %s
+configuration:
+  instances: 1
+  routes:
+  - %s
+  appchart: standard
+origin:
+  container: splatform/sample-app
+  git:
+    url: ""
+namespace: %s
+`, app, domain, namespace)
+
+		By(string(bodyBytes))
+		By(expecting)
+		Expect(string(bodyBytes)).To(Equal(expecting))
 	})
 
 	It("returns a 404 when the namespace does not exist", func() {
--- a/acceptance/api/v1/application_export_test.go
+++ b/acceptance/api/v1/application_export_test.go
@@ -100,8 +100,7 @@ configuration:
   appchart: standard
 origin:
   container: splatform/sample-app
-  git:
-    url: ""
+  git: {}
 namespace: %s
 `, app, domain, namespace)
 
--- a/internal/api/v1/application/part.go
+++ b/internal/api/v1/application/part.go
@@ -56,8 +56,8 @@ func GetPart(c *gin.Context) apierror.APIErrors {
 	partName := c.Param("part")
 	logger := requestctx.Logger(ctx)
 
-	if partName != "values" && partName != "chart" && partName != "image" {
-		return apierror.NewBadRequestError("unknown part, expected chart, image, or values")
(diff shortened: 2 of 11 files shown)


andreas-kupries:
@richard-cox this looks complete to me, should be ready for testing with the UI.

The one part I was unsure about, how to get the extended origin to the server I found (i.e. got reminded of by the code search) that the entire origin structure is part of the request to the deploy API endpoint. 

With the extended structure and updated swagger json setting the new fields should be no issue for the Web UI.

The CLI has to do nothing, which leaves the new fields at their natural null values (false, empty string), as they should be for the CLI.

acceptance/api/v1/application_export_test.go
```diff
@@ -80,6 +81,35 @@ var _ = Describe("AppPart Endpoint", LApplication, func() {
...
+		expecting := fmt.Sprintf(`name: %s
+configuration:
+  instances: 1
+  routes:
+  - %s
+  appchart: standard
+origin:
+  container: splatform/sample-app
+  git:
+    url: ""
```
enrichman, line 104: Probably this is something "old" but I would not expect to get the `git` field returned if empty. In this case the `origin` is `container` and the Git part is not needed.

internal/api/v1/application/part.go
```diff
@@ -56,8 +56,8 @@ func GetPart(c *gin.Context) apierror.APIErrors {
 	partName := c.Param("part")
 	logger := requestctx.Logger(ctx)
 
-	if partName != "values" && partName != "chart" && partName != "image" {
-		return apierror.NewBadRequestError("unknown part, expected chart, image, or values")
+	if partName != "manifest" && partName != "values" && partName != "chart" && partName != "image" {
+		return apierror.NewBadRequestError("unknown part, expected chart, manifest, image, or values")
```
enrichman, line 60: This `if` statement is becoming a bit long, and it's better to log/return the wrong part.

What about this?

```golang
switch partName {
// valid parts
case "manifest", "values", "chart", "image":
default:
	return apierror.NewBadRequestErrorf("unknown '%s' part, expected chart, manifest, image, or values", partName)
}
```

internal/api/v1/application/part.go
```diff
@@ -74,9 +74,9 @@ func GetPart(c *gin.Context) apierror.APIErrors {
 		return apierror.AppIsNotKnown(appName)
 	}
 
-	if app.Workload == nil {
-		// While the app exists it has no workload, and therefore no chart to
-		// export
+	if partName != "manifest" && app.Workload == nil {
+		// While the app exists it has no workload, and therefore no chart/image/values to
+		// export. Manifest however will be fine.
```
enrichman, line 79: I prefer to early return, because it has less cognitive complexity:

```golang
if partName == "manifest" {
	return fetchAppManifest(c, app)
}

if app.Workload == nil {
	return apierror.NewBadRequestError("no chart available for application without workload")
}
```

internal/api/v1/application/part.go
```diff
@@ -87,9 +87,11 @@ func GetPart(c *gin.Context) apierror.APIErrors {
 		return fetchAppImage(c, ctx, logger, cluster, app.Meta)
 	case "values":
 		return fetchAppValues(c, logger, cluster, app.Meta)
+	case "manifest":
+		return fetchAppManifest(c, app)
 	}
 
-	return apierror.NewBadRequestError("unknown part, expected chart, image, or values")
+	return apierror.NewBadRequestError("unknown part, expected chart, image, manifest, or values")
```
enrichman, line 94: This code should never be reached (we already checked the part). It's probably better to return a `500` (?).

internal/api/v1/application/part.go
```diff
@@ -323,6 +325,26 @@ func fetchAppValues(c *gin.Context, logger logr.Logger, cluster *kubernetes.Clus
...
+		Origin:    app.Origin,
+		Staging:   app.Staging,
+	}
+
+	yaml, err := yaml.Marshal(m)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	response.OKBytes(c, yaml)
```
enrichman, line 344: Can we use an `OKYaml` instead?

```golang
// OKYaml reports a success with some YAM data
func OKYaml(c *gin.Context, response {}interface) {
	requestctx.Logger(c.Request.Context()).Info("OK",
		"origin", c.Request.URL.String(),
		"returning", response,
	)

	c.YAML(http.StatusOK, response)
}
```

It will return the proper content type and we don't even need to marshal it.

enrichman (changes requested):
I wasn't expecting this to be part of the `part` api but it was clever. We don't need to add one more endpoint.

Just some comments,\.

acceptance/api/v1/application_export_test.go
andreas-kupries: Might be something with the serialization annotations (around `omitempty`).

internal/api/v1/application/part.go
andreas-kupries: Agreed.

internal/api/v1/application/part.go
andreas-kupries: Ah, make it separate from the main switch. doable.

internal/api/v1/application/part.go
andreas-kupries: Will try.

internal/api/v1/application/part.go
andreas-kupries: Done, commit 281c55964128a18f8a6bcf3b4d75b06e80395032

internal/api/v1/application/part.go
andreas-kupries: Done, commit 281c55964128a18f8a6bcf3b4d75b06e80395032

internal/api/v1/application/part.go
andreas-kupries: Done, commit 281c55964128a18f8a6bcf3b4d75b06e80395032

internal/api/v1/application/part.go
andreas-kupries: Done, commit 281c55964128a18f8a6bcf3b4d75b06e80395032

acceptance/api/v1/application_export_test.go
andreas-kupries: Managed to reduce the data with commit 6df12eacfbeafad4baaaa44f9d8266c38e711a10.
The `git` does not fully go away :(

enrichman: I had a look again and the issue about the `git` origin is that we are always instanciating it, so that's why the `omitempty` is not working: https://github.com/epinio/epinio/blob/main/internal/application/origin.go#L30-L32

We should create this Origin only for a Git one: https://github.com/epinio/epinio/blob/main/internal/application/origin.go#L79

enrichman (approved):

mmartin24:
Tested by verifying correct funcioning on these UI issues; mainly: [#226](https://github.com/epinio/ui/issues/226), [#228](https://github.com/epinio/ui/issues/228),  [#234](https://github.com/epinio/ui/issues/234),  [#231](https://github.com/epinio/ui/issues/231) . All of them tested and ok in `1.8.1-rc1`

Further checks related to this (https://github.com/epinio/ui/issues/227) to be done in further releases.

PR #2249, opened 2023-04-27
Fixed same suggestions for multiple resources

enrichman:
The `epinio app env set` command was not using the common `matchingAppsFinder` func.

Also some app/service/config commands (`delete`) accepts more than one resource, but our completion were suggesting also the already added resources.

A common `filteredMatchingFinder` func is used.

### DIFF
--- a/internal/cli/commons.go
+++ b/internal/cli/commons.go
@@ -12,7 +12,6 @@
 package cli
 
 import (
-	"context"
 	"fmt"
 	"os"
 
@@ -70,7 +69,7 @@ func matchingConfigurationFinder(cmd *cobra.Command, args []string, toComplete s
 	}
 	app.API.DisableVersionWarning()
 
-	matches := app.ConfigurationMatching(context.Background(), toComplete)
+	matches := app.ConfigurationMatching(toComplete)
 
 	return matches, cobra.ShellCompDirectiveNoFileComp
 }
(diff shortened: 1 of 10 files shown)


internal/cli/commons.go
```diff
@@ -164,3 +163,25 @@ func matchingCatalogFinder(cmd *cobra.Command, args []string, toComplete string)
 
 	return matches, cobra.ShellCompDirectiveNoFileComp
 }
+
+// filteredMatchingFinder will use the finder func to find the resources with the prefix name
+// It will then filter the matches removing the provided args
+func filteredMatchingFinder(args []string, prefix string, finder func(prefix string) []string) []string {
+	// map to check for already selected applications
+	alreadyMatchedApps := map[string]struct{}{}
```
andreas-kupries, line 171:
```suggestion
	alreadyMatched := map[string]struct{}{}
```

Given that this is a generic finder `App` makes no sense.

andreas-kupries: This change will induce updating the other places.

internal/cli/commons.go
```diff
@@ -164,3 +163,25 @@ func matchingCatalogFinder(cmd *cobra.Command, args []string, toComplete string)
...
+	for _, app := range args {
+		alreadyMatchedApps[app] = struct{}{}
+	}
+
+	filteredMatches := []string{}
+
+	matches := finder(prefix)
+	for _, app := range matches {
+		// return only the not matched apps
```
andreas-kupries, line 180:
```suggestion
		// return only the not already matched arguments
```

internal/cli/commons.go
```diff
@@ -164,3 +163,25 @@ func matchingCatalogFinder(cmd *cobra.Command, args []string, toComplete string)
...
+	// map to check for already selected applications
+	alreadyMatchedApps := map[string]struct{}{}
+	for _, app := range args {
+		alreadyMatchedApps[app] = struct{}{}
+	}
+
+	filteredMatches := []string{}
+
+	matches := finder(prefix)
+	for _, app := range matches {
```
andreas-kupries, line 179:
```suggestion
	for _, resource := range matches {
```

andreas-kupries (changes requested):
Leftover variable names indicating start of the generic finder as an app name finder.

andreas-kupries (approved):

mmartin24:
Checked in `v1.8.1-rc1` 
No duplications after autosugestions:
After `epinio configuration delete config-1 config-2` and pressing tab, does not duplicate config-1

PR #2256, opened 2023-04-28
Fix workflows and delete clusters to read ssh-key from file

svollath:
Currently cluster deletion for AWS_RKE2 (rke2-lh-ec2.yml) fails - as well as any dispatch deletion via delete_clusters.yml would - because you can't just use a multiline secret in github env.
Therfore I changed the code to always write the ssh-key to ~/.ssh/, and delete_clusters.go will read it from there.


enrichman (approved):

svollath:
THX Enrico !

PR #2257, opened 2023-04-29
AKS workflow test PR - do not review/merge

svollath:
Quote KUBECONFIG env var when runs-on Windows

PR #2258, opened 2023-04-29
Fix "KUBECONFIG" env var quoting for AKS*-CI (Windows)

svollath:
Fix env var quoting and escaping for windows runners within AKS* workflows.

### DIFF
--- a/.github/workflows/aks-letsencrypt.yml
+++ b/.github/workflows/aks-letsencrypt.yml
@@ -38,9 +38,10 @@ env:
   GINKGO_NODES: 1
   FLAKE_ATTEMPTS: 1
   PUBLIC_CLOUD: 1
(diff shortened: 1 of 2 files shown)


svollath:
When streamlining the KUBECONFIG env var we thought it would work everywhere. However, while it worked for Ubuntu, SLE15 and MacOS, it did not in the Windows case - hence AKS-CI and AKS-LETSENCRYPT-CI do fail.
AKS-CI runs and delete_clusters.go had been tested already within https://github.com/epinio/epinio/pull/2257.
For AKS delete_clusters.go works fine with both workspace paths - when it runs on Windows (CI), or on Ubuntu (delete-clusters.yml).
So, this is good to merge already.

mmartin24 (approved):
Tested against branch and CI passes:
https://github.com/epinio/epinio/actions/runs/4858066533/jobs/8659121040#step:8:253

PR #2270, opened 2023-05-02
Fixed Docker rate limiting

enrichman:
Some of our workflows are hitting the Docker rate limits because the authentication is done in later steps.

Moved docker login before docker setup, and also unified docker login credentials.

PR #2272, opened 2023-05-03
Revert "Fixed Docker rate limiting"

enrichman:
Reverts epinio/epinio#2270

Partial revert, only some tokens. The failure was related to push to `splatform` org in Dockerhub.


andreas-kupries (approved):
looks ok, and hopefully passes CI.

PR #2278, opened 2023-05-04
Chore: Removed namespace controller structure

### LINKED ISSUE #2276: Remove the `namespace.Controller`

andreas-kupries:
fix #2276


internal/api/v1/namespace/controller.go
enrichman, line 1: I think we can remove this file completely

enrichman (changes requested):

enrichman (approved):

PR #2287, opened 2023-05-08
Release CI: Extended the set of artifacts to generate SBOMs for.

### LINKED ISSUE #2271: Missing SBOMs for Epinio binaries (only Windows)
We are not generating SBOMs for the binaries because of the missing `any` artifact type (https://goreleaser.com/customization/sbom/)

https://github.com/epinio/epinio/blob/main/.goreleaser.yml#L309-L317

andreas-kupries:
fix #2271

### DIFF
--- a/.goreleaser-qa.yml
+++ b/.goreleaser-qa.yml
@@ -270,9 +270,11 @@ docker_manifests:
 
 sboms:
   - id: archive 
-    artifacts: any
+    artifacts: archive
   - id: source
     artifacts: source
+  - id: binary
+    artifacts: binary
 
 # signs our docker images
 # https://goreleaser.com/customization/docker_sign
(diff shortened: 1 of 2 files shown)


andreas-kupries:
This seems to have broken the sandbox release

https://github.com/epinio/epinio/actions/runs/4913254983/jobs/8773226807?pr=2287#step:9:56

Retrying to see if that was a transient thing.

andreas-kupries:
Definitely not transient. Now lets try other things.

andreas-kupries:
and the sandbox is :heavy_check_mark: :)

andreas-kupries:
Looking at the release output I cannot say if the result is better than before, or not.

enrichman (approved):

PR #2296, opened 2023-05-11
Added documentation on how to run a local external registry

enrichman:
Added documentation on how to run a local external registry, with selfsigned certificates, and how to load them into Epinio.

### DIFF
--- a/docs/howtos/local_external_registry.md
+++ b/docs/howtos/local_external_registry.md
@@ -1,0 +1,99 @@
+# Setup a local external registry
+
+## Certificates
+
+Before running the registry or the cluster we will need to create the certificates.
+
+We are going to put them in a `docker_reg_certs` folder:
+
+```
+mkdir docker_reg_certs
+
+# create the Certificate Signing Request
+openssl req -new -newkey rsa:4096 -nodes \
+    -config epinio.cfg \
+    -subj "/C=IT/ST=Italy/L=Rome/O=SUSE/OU=Epinio/CN=*.omg.howdoi.website" \
+    -keyout docker_reg_certs/epinio.key \
+    -out docker_reg_certs/epinio.csr
+
+# create the certificate
+openssl x509 -req -sha256 -days 365 \
+    -extfile <(printf "subjectAltName=DNS:172.21.0.5.omg.howdoi.website,DNS:*.omg.howdoi.website") \
+    -in docker_reg_certs/epinio.csr \
+    -signkey docker_reg_certs/epinio.key \
+    -out docker_reg_certs/epinio.pem
+```
+
+**Note:** to check the SAN of the request and the certificate you can run:
+
+```
+openssl req -in docker_reg_certs/epinio.csr -text | grep -A1 'Subject Alternative Name'
+
+openssl x509 -in docker_reg_certs/epinio.pem -text  | grep -A1 'Subject Alternative Name'
+```
+
+## Setup the cluster
+
+With `k3d` we'll need to mount the certificate during the cluster creation, so you will need to mount it adding this flag `--volume $(pwd)/docker_reg_certs/epinio.pem:/etc/ssl/certs/epinio.pem`:
+
+i.e.:
+```
+k3d cluster create --volume /path/to/your/certs.crt:/etc/ssl/certs/yourcert.crt
+```
+
+We need to do that otherwise the kubelet will not be able to pull the created images.
+
+See [k3d doc](https://k3d.io/v5.2.1/faq/faq/#pods-fail-to-start-x509-certificate-signed-by-unknown-authority).
+
+
+## Setup the registry
+
+We can now run our registry:
+
+```
+docker run -d -p 5000:5000 --name registry --rm \
+    -v $PWD/docker_reg_certs:/certs \
+    -v /reg:/var/lib/registry \
+    -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/epinio.pem \
+    -e REGISTRY_HTTP_TLS_KEY=/certs/epinio.key registry:2
+```
+
+Now the registry needs to be accessible from the cluster. We can do that adding it to the same docker network, and check the IP that it got:
+
+```
+docker network connect epinio-acceptance registry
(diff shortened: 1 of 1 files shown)


docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+stateOrProvinceName        = State or Province Name (full name)
+localityName               = Locality Name (eg, city)
+organizationName           = Organization Name (eg, company)
+commonName                 = Common Name (e.g. server FQDN or YOUR name)
+[ req_ext ]
+subjectAltName = @alt_names
+[alt_names]
+DNS.1   = omg.howdoi.website
+DNS.2   = *.omg.howdoi.website
+DNS.3   = 172.21.0.5.omg.howdoi.website
```
andreas-kupries, line 31: I suspect that this entry has to be adapted to the local IP address, right ?
If yes, this should be noted IMHO.

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+we can now run these commands:
+
+```
+# create the Certificate Signing Request
+openssl req -new -newkey rsa:4096 -nodes \
+    -config epinio.cfg \
+    -subj "/C=IT/ST=Italy/L=Rome/O=SUSE/OU=Epinio/CN=*.omg.howdoi.website" \
```
andreas-kupries, line 40: Italy/Rome are also places to customize to the local conditions ? Not sure if the values matter for this scenario.

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+
+# create the certificate
+openssl x509 -req -sha256 -days 365 \
+    -extfile <(printf "subjectAltName=DNS:172.21.0.5.omg.howdoi.website,DNS:*.omg.howdoi.website") \
+    -in docker_reg_certs/epinio.csr \
+    -signkey docker_reg_certs/epinio.key \
+    -out docker_reg_certs/epinio.pem
+```
+
+**Note:** to check the SAN of the request and the certificate you can run:
```
andreas-kupries, line 52:
```suggestion
**Note:** Check the SAN of request and certificate by running:
```

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+
+```
+openssl req -in docker_reg_certs/epinio.csr -text | grep -A1 'Subject Alternative Name'
+
+openssl x509 -in docker_reg_certs/epinio.pem -text  | grep -A1 'Subject Alternative Name'
+```
+
+## Setup the cluster
+
+With `k3d` we'll need to mount the certificate during the cluster creation, so you will need to mount it adding this flag `--volume $(pwd)/docker_reg_certs/epinio.pem:/etc/ssl/certs/epinio.pem`:
```
andreas-kupries, line 62:
```suggestion
With `k3d` we'll need to mount the certificate during the cluster creation. This is done by adding the flag `--volume $(pwd)/docker_reg_certs/epinio.pem:/etc/ssl/certs/epinio.pem`:
```

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+
+i.e.:
+```
+k3d cluster create --volume /path/to/your/certs.crt:/etc/ssl/certs/yourcert.crt
+```
+
+We need to do that otherwise the kubelet will not be able to pull the created images.
```
andreas-kupries, line 69:
```suggestion
If this is not done the kubelet will not be able to pull the created images.
```

andreas-kupries (changes requested):
General phrasing. Also not liking the hard coded IP addresses floating around in the command examples. (I see two different ones, for the system domain, and the registry)

docs/howtos/local_external_registry.md
enrichman: I should check if the wildcard is enough, or if the IP is needed. Maybe we can refine this to the smallest working configuration. I guess that it's full of all the possible combination because I was not sure on what fixed what.

docs/howtos/local_external_registry.md
enrichman: Not really needed, so we can put something generic, such as Nuremberg and DE. What do you think?

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+```
+
+To make Epinio works we need to mount this certificate during the staging job.  
+
+To do so let's create a secret with the certificate, and patch the Epinio deployment to fetch this secret:
+
+```
+kubectl create secret -n epinio generic epinio-external-registry-tls --from-file=tls.crt=docker_reg_certs/epinio.pem
+
+kubectl patch deployment -n epinio epinio-server --patch '{"spec": {"template": {"spec": {"containers": [{"name": "epinio-server","env": [{"name":"REGISTRY_CERTIFICATE_SECRET", "value":"epinio-external-registry-tls"}]}]}}}}'
```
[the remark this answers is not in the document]
enrichman: Yes, absolutely. This is an enhancement that we should/could add. At the moment the secret is used only for our internal registry, but we have seen that the same can be used to add a selfsigned/untrusted cert, so it makes sense to expose this.

docs/howtos/local_external_registry.md
```diff
@@ -0,0 +1,124 @@
...
+
+```
+docker run -d -p 5000:5000 --name registry --rm \
+    -v $PWD/docker_reg_certs:/certs \
+    -v /reg:/var/lib/registry \
+    -e REGISTRY_HTTP_TLS_CERTIFICATE=/certs/epinio.pem \
+    -e REGISTRY_HTTP_TLS_KEY=/certs/epinio.key registry:2
+```
+
+Now the registry needs to be accessible from the cluster. We can do that adding it to the same docker network, and check the IP that it got:
```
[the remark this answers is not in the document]
enrichman: It's the IP of the registry that will be used by Epinio to reach it. As you said I should use something generic, like `<REGISTRY_IP>`.
Maybe even exporting it to an env var.

andreas-kupries (approved):

andreas-kupries (approved):

PR #2297, opened 2023-05-12
Narrow down self-hosted runner usage

svollath:
Narrow down self-hosted runner usage to cases where nested virtualization is needed on the runner itself.

### DIFF
--- a/.github/workflows/release-sandbox.yml
+++ b/.github/workflows/release-sandbox.yml
@@ -30,7 +30,7 @@ env:
 
 jobs:
   release:
-    runs-on: [self-hosted, epinio]
+    runs-on: ubuntu-latest
     permissions:
       id-token: write   # This is the key for OIDC!
     steps:
--- a/.github/workflows/aks-letsencrypt.yml
+++ b/.github/workflows/aks-letsencrypt.yml
@@ -45,7 +45,7 @@ env:
 
 jobs:
   linter:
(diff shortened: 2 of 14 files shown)


andreas-kupries (approved):

.github/workflows/release-sandbox.yml
```diff
@@ -30,7 +30,7 @@ env:
 
 jobs:
   release:
-    runs-on: [self-hosted, epinio]
+    runs-on: ubuntu-latest
```
andreas-kupries, line 33: https://github.com/epinio/epinio/actions/runs/4957350161/jobs/8868917727?pr=2297#step:10:14

Indicates a change in the underlying shell, one which does not understand the `-o pipefail`.
Suspect that the release-publish flow may run into the same.
Might be fixable to changing the script to work without `-o pipefail`.

andreas-kupries (changes requested):

.github/workflows/release-sandbox.yml
svollath: yes realized the same - was reading about problems related to /bin/sh vs. bash in ubuntu - suggesting to run the script (executable) with "sudo" ...

svollath: yep, changing "#!/bin/sh" to "#!/bin/bash" did it.

andreas-kupries (approved):

PR #2298, opened 2023-05-12
Validate git provider strings, client and server

### LINKED ISSUE #2291: Add validation regarding `provider` field
**Describe the bug**

At the moment we don't have any restriction about the `provider` field used to defin the `git` origin.
To avoid misalignment we have agreed to have a fixed set of values.

For now these will suffice:
```
git
github
github_enterprise
gitlab
gitlab_enterprise
```

andreas-kupries:
fix #2291

### DIFF
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -162,11 +162,11 @@ func UpdateSources(manifest models.ApplicationManifest, cmd *cobra.Command) (mod
 		// Standard provider, and conditional override by the user
 		gitRef.Provider = models.ProviderGit
 		if gitProvider != "" {
-			if !models.ProviderIsValid(gitProvider) {
+			provider, err := models.GitProviderFromString(gitProvider)
+			if err != nil {
 				return manifest, errors.New("Bad --git-provider `" + gitProvider + "`")
 			}
-
-			gitRef.Provider = gitProvider
+			gitRef.Provider = provider
 		}
 
 		if origins == 1 {
--- a/internal/manifest/manifest.go
+++ b/internal/manifest/manifest.go
@@ -53,7 +53,7 @@ func UpdateRoutes(manifest models.ApplicationManifest, cmd *cobra.Command) (mode
 }
 
 // UpdateBASN updates the incoming manifest with information pulled from the --builder,
-// sources (--path, --git, and --container-imageurl), --app-chart, and --name options.
+// sources (--path, --git, --git-provider, and --container-image-url), --app-chart, and --name options.
 // Option information replaces any existing information.
 func UpdateBASN(manifest models.ApplicationManifest, cmd *cobra.Command) (models.ApplicationManifest, error) {
 	var err error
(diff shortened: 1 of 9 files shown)


andreas-kupries:
@richard-cox @torchiaf - Note, the set of strings this code considers as legal can be trivially changed to adapt to the needs of the UI.
See https://github.com/epinio/epinio/pull/2298/files#diff-1ce4bd480f02d0d89d05c5733bbdadacc601e586fe8593e4e70ea34e56a9b469
for the relevant chunk and list.

pkg/api/core/v1/models/app.go
```diff
@@ -30,8 +30,31 @@ const (
 	ApplicationStagingActive = "active"
 	ApplicationStagingDone   = "done"
 	ApplicationStagingFailed = "failed"
+
+	ProviderGit              = "git"
+	ProviderGithub           = "github"
+	ProviderGithubEnterprise = "github_enterprise"
+	ProviderGitlab           = "gitlab"
+	ProviderGitlabEnterprise = "gitlab_enterprise"
```
enrichman, line 38: I was thinking that it will be better to have this as a `type` and change the provider to be a `GitProvider` instead of a generic string. This will help us on being stricter during the creation/handling of this value, and be sure to use one of the right one.

Something like:

```golang
type GitProvider string

const (
	ProviderUnknown = GitProvider("unknown")
	ProviderGit     = GitProvider("git")
	ProviderGithub  = GitProvider("github")
)

func GitProviderFromString(provider string) (GitProvider, error) {
	switch provider {
	case "git":
		return ProviderGit, nil
	case "github":
		return ProviderGit, nil
	default:
		return ProviderUnknown, errors.New("unknown provider")
	}
}

```

I really miss enums sometimes.

enrichman (changes requested):

pkg/api/core/v1/models/app.go
andreas-kupries: Certainly doable. The conversion function acts as the validator.

internal/manifest/manifest.go
```diff
@@ -154,6 +159,16 @@ func UpdateSources(manifest models.ApplicationManifest, cmd *cobra.Command) (mod
 		kind = models.OriginGit
 		origins++
 
+		// Standard provider, and conditional override by the user
+		gitRef.Provider = models.ProviderGit
```
enrichman, line 163: I was thinking that maybe we could also check for the standard urls for Github and Gitlab:

```go
func GitProviderFromOriginURL(url string) {
	u, err := url.Parse("https://github.com/ndcsjk/sdhbnis")
	// check for u.Host == github.com or gitlab.com
}
```
and then 
```go
provider, err := models.GitProviderFromOriginURL(gitURL)
if err != nil {
	// log something about the error
	gitRef.Provider = models.ProviderGit
} else {
	gitRef.Provider = provider
}
```

What do you think? The `--git-provider` could be provided anyway to overwrite this behaviour.

andreas-kupries: I feel the beginning of feature creep here. Willing to try this one thing. And CI just went :heavy_check_mark: :(

andreas-kupries: See commit 14309cf6159097d6d95aff0be97dfe35a0bdc498

enrichman (approved):

mmartin24:
Checked new  `--git-provider` flag with `gitlab` and working:

```
epinio push --name testatpp-gitlab-cli-1 --git-provider gitlab --git https://gitlab.com/richard-cox/epinio-sample-app,bb6883114668e9aed4964ed3f60bed3bc623c202
```

Output:
```
─────────────────────────────────────────────────────────── Logs(workspace/rtestatpp-gitlab-cli-1-7ad57b18960c71700e822a7f79d2a3597aeswj67:rtestatpp-gitlab-cli-1-7ad57b18960c71700e822a7f79d2a3597aed6f06)[1m] ───────────────────────────────────────────────────────────┐
│                                                                                                     Autoscroll:On     FullScreen:Off     Timestamps:Off     Wrap:Off                                                                                                      
│ chatty app .... is chatty                                                                                                                                                                                                                                                 
│ chatty app .... is chatty     
```

PR #2300, opened 2023-05-15
Fix mishandling of --all when no resources are found/available.

andreas-kupries:
Properly exist with sensible message.

note: testing this new code path is as difficult as testing --all in general, due to the concurrent executon of multiple tests cases. We generally will not have "no resources".
result: no new tests for this


mmartin24 (approved):
LGTM

PR #2301, opened 2023-05-16
fixed transfer of bad namespace information on login (to new cluster)

### LINKED ISSUE #2299: Remove `target: <namespace>` from `settings.yaml` when logging to another cluster
**Describe the bug**
RKE2 EC2 CI [jobs](https://github.com/epinio/epinio/actions/runs/4954930538/jobs/8863872559#step:15:81) are failing because they run on a shared gh-runner where another CI job wrote the `~/.config/epinio/settings.yaml` file before and did settings of the new default `target` namespace in it.

The EC2 job is then trying to create an app but it uses the non-existing namespace from the previous run hence the app deployment is failing.

The workaround is to delete the `~/.config/epinio/settings.yaml` and start the login from scratch.

**To Reproduce**
Steps to reproduce the behavior, e.g.:

1. add `target: <non-existing-namespace>` into `settings.yaml` file
2. Push an app
3. See the error:
```
  🚢  Create application
  Namespace: namespace-759461032
  Application: apps-149210746

  ❌  error creating app: Not Found: namespace 'namespace-759461032' does not exist
```

**Expected behavior**
I see several options which may address this issue:
* introduce new `epinio login --target-namespace`
* on a new `epinio login` call we may delete the `target:` line esp. when login to a different cluster
* Introduce new `epinio logout` which would delete the settings.yaml file
* support multiple epinio endpoints/clusters in settings.yaml
* remove the `target` when epinio detects a non-existing target within `epinio login`

andreas-kupries:
fix #2299 

I went with the last option: check validity, clear data if not, and report the issue in that case too
(all other options do not really fit into the scope of 1.8.1 to come)

### DIFF
--- a/acceptance/helpers_test.go
+++ b/acceptance/helpers_test.go
@@ -24,7 +24,7 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-func ExpectGoodUserLogin(tmpSettingsPath, password, serverURL string) {
+func ExpectGoodUserLogin(tmpSettingsPath, password, serverURL string) string {
 	By("Regular login")
 	out, err := env.Epinio("", "login", "-u", "epinio", "-p", password,
 		"--trust-ca", "--settings-file", tmpSettingsPath, serverURL)
@@ -34,6 +34,8 @@ func ExpectGoodUserLogin(tmpSettingsPath, password, serverURL string) {
 	Expect(out).To(ContainSubstring(`Trusting certificate`))
 	Expect(out).To(ContainSubstring(`Login successful`))
 	By("Regular login done")
+
+	return out
 }
 
 func ExpectGoodTokenLogin(tmpSettingsPath, serverURL string) {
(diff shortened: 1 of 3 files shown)


enrichman:
I'm not sure on what should be the expected behaviour for a user that has not access to an existing namespace.

Maybe is better to have a more generic message, like `namespace '%s' not found in targeted cluster` instead of `invalid`. That is the behaviour of different website/app (404 instead of errors).

andreas-kupries:
Sure.

thehejik:
Verified fixed in v1.8.1

Created entry `namespace: nonexisting` in settings and performed login:
```
./epinio login --user admin --password password https://epinio.192.168.0.102.sslip.io

🚢  Login to your Epinio cluster [https://epinio.192.168.0.102.sslip.io]

⚠️  Certificate signed by unknown authority

|     KEY     |     VALUE      |
|-------------|----------------|
| Issuer Name | CN=epinio-ca   |
| Common Name | epinio-ca      |
| Expiry      | 2023-August-23 |

Do you want to trust it (y/n): y

✔️  Trusting certificate for address https://epinio.192.168.0.102.sslip.io...

✔️  Login successful

⚠️  Current namespace 'nonexisting' not found in targeted cluster

⚠️  Cleared current namespace

⚠️  Please use `epinio target` to chose a new current namespace
```

Issue #2037, opened 2023-01-23
[Epic] Add port forward for services


enrichman:
**Describe the solution you'd like**
It would be nice to connect to a running service with the port forwarding.

I.e.: connect to a mongoDB with a client, running an `epinio service port-forward mydb 27017`
```[tasklist]
### Tasks
- [ ] https://github.com/epinio/epinio/issues/2365
- [ ] https://github.com/epinio/epinio/pull/2366
- [ ] https://github.com/epinio/epinio/pull/2361
- [ ] https://github.com/epinio/epinio/pull/2390
```

gitricko:
+1

enrichman:
Small bug to fix in the next relase: https://github.com/epinio/epinio/issues/2425

PR #2329, opened 2023-05-24
Add port forward for services (#2037)

giubacc:
Epinio services can now be accessed from the host by executing:

epinio service port-forward SERVICENAME [LOCAL_PORT] [...[LOCAL_PORT_N]] [flags]

Issue: https://github.com/epinio/epinio/issues/2037


giubacc:
This will be instead split into small PRs.

PR #2330, opened 2023-05-25
Target workspace namespace for EKS/EC2 tests

thehejik:
Fixes RKE2-EC2 workflow by targeting a default workspace namespace. This is needed because our self-hosted runner has existing `settings.yaml` with non-existing targeted ns from previous run.

Testrun https://github.com/epinio/epinio/actions/runs/5077065695/jobs/9119968996


mmartin24 (approved):
LGTM and tests passed

enrichman (approved):

Issue #2306, opened 2023-05-17
Use new `1.20` test coverage feature


enrichman:
**Is your feature request related to a problem? Please describe.**
Until Go 1.19 you couldn't build an instrumented test coverage binary, and you had to build with `go test -c` and run a Test.

**Describe the solution you'd like**
We should switch to the new feature introduced with the 1.20 and using the `go build -cover` produced binary.

PR #2334, opened 2023-05-26
Implementation of new go 1.20 cover feature

enrichman:
Ref: https://github.com/epinio/epinio/issues/2306


andreas-kupries (approved):
LGTM

PR #1082, opened 2021-12-17
When updating the Epinio config, trust the cert not the CA

jimmykarily:
because the CA may not be in the secret:

https://github.com/jetstack/cert-manager/issues/2111

PR #2348, opened 2023-05-29
Added CA certs in validation

enrichman:
This PR removes the check/filter for CA certificates introduced in https://github.com/epinio/epinio/pull/1082

It also use directly the paketo functions instead of copying them.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -57,7 +57,6 @@ require (
 	gopkg.in/yaml.v2 v2.4.0
 	helm.sh/helm/v3 v3.11.2
 	k8s.io/api v0.26.3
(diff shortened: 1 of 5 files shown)


thehejik (approved):
LGTM
The PR is verified by checking an app deployment by using:
- external registry with TLS cert signed by subCA
- internal epinio registry (self-signed by epinio/cert-manager)
- external dockerhub registry (signed by trusted CA)

thehejik:
Verified on epinio v1.9.0-rc3 by following https://github.com/epinio/epinio/blob/main/docs/howtos/public_external_registry.md - all good

PR #2361, opened 2023-06-05
move proxy related code to proxy package, add TCPProxy

giubacc:
Added TCPProxy specific implementation to be used with service port forwarding.

- tests to TCPProxy will be added on the PR containing the service port forwarding feature.

Related to: https://github.com/epinio/epinio/issues/2037


internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+	defer cancel()
+
+	var err error
+	if p.ProxyConnection, err = d.DialContext(ctx, "tcp", p.Address); err != nil {
+		logger.Error(err, "dialing service")
+		return err
+	}
+
+	return p.handleConnections()
+}
```
enrichman, line 90: It's better to give the possibility to tweak the Timeout from the caller, and probably also the stop chan could be an argument of this method.

```go
func (p *TCPProxy) Start(ctx context.Context, stopChan <-chan struct{}) error {}
```

With this approach you can stop and start again the Proxy, if needed. Otherwise the closed chan will remain closed, and you need to create a new Proxy in order to Start it again.

internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+func (p *TCPProxy) handleConnections() error {
+	defer p.ProxyConnection.Close()
+	ctx := p.GCtx.Request.Context()
+	logger := requestctx.Logger(ctx).WithName("PortForward")
+	localError := make(chan struct{})
+	remoteDone := make(chan struct{})
+
+	go func() {
+		// Copy from the remote side to the local port.
+		if _, err := io.Copy(p.ProxyConnection, p.UpgradedConnection); err != nil && !strings.Contains(err.Error(), "use of closed network connection") {
```
enrichman, line 101: I think that this can be checked with a `!errors.Is(err, net.ErrClosed)` but I'm not completely sure.

https://pkg.go.dev/net#ErrClosed

internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+			logger.Error(err, "error copying from UpgradedConnection to ProxyConnection")
+			close(localError)
+		}
+		// inform the select below that the remote copy is done
+		close(remoteDone)
+	}()
+
+	go func() {
+		// Copy from the local port to the remote side.
+		if _, err := io.Copy(p.UpgradedConnection, p.ProxyConnection); err != nil && !strings.Contains(err.Error(), "use of closed network connection") {
```
enrichman, line 111: Same as before (`!errors.Is(err, net.ErrClosed)`) (?)

enrichman (changes requested):

internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,126 @@
...
+
+	return nil
+}
+
+type TCPProxy struct {
+	GCtx               *gin.Context
+	Address            string
+	StopChan           <-chan struct{}
+	UpgradedConnection net.Conn
+	ProxyConnection    net.Conn
```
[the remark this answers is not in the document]
giubacc: Here upgradedConnection was to underline that the connection comes from an upgraded socket (websocket).
But I have nothing against changing this in a more generic way.

internal/api/v1/proxy/proxy.go
giubacc: looking better at the code, `stopChan` was actually not used in this context, so I got rid of it.

internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,122 @@
...
+		address += ":80"
+	}
+	return &TCPProxy{
+		Ctx:          ctx,
+		IncomingConn: IncomingConn,
+		Address:      address,
+	}, nil
+}
+
+func (p *TCPProxy) Start(dialTimeout time.Duration) error {
```
enrichman, line 74: I think the context, if needed, should be always propagated. If this func is accepting a `ctx` then it can be used also with different "strategies", maybe with a `context.WithCancel()`, or with an empty Context, to keep the connection open indefinitely.

internal/api/v1/proxy/proxy.go
```diff
@@ -0,0 +1,122 @@
...
+		return err
+	}
+
+	return p.handleConnections()
+}
+
+func (p *TCPProxy) handleConnections() error {
+	defer p.OutgoingConn.Close()
+	logger := requestctx.Logger(p.Ctx).WithName("PortForward")
+	localError := make(chan struct{})
```
enrichman, line 92: Since this function is saying that could error (and it can) I will make this `chan` to return an `error` instead, and reporting it to the caller.

enrichman (changes requested):

internal/api/v1/proxy/proxy.go
giubacc: as discussed, timeout is for dial only.

Issue #2318, opened 2023-05-22
Reproduce error decoding PEM with external registry


enrichman:
In some cases we have an error while trying to decode certificate secrets for external registries.

We should try to reproduce it to understand which could be the issue.

thehejik:
From email communication from FB we found out that they are using subCA certificate with CA:true for their external registry which was not supported by cahash.go, it should be addressed by https://github.com/epinio/epinio/pull/2348

thehejik:
Customer confirmed the fix works for him, it will be part of upcoming release, in meantime ` --set image.epinio.tag=v1.9.0-rc2` can be used.

thehejik:
Issue reproduced, documentation created https://github.com/epinio/epinio/blob/main/docs/howtos/public_external_registry.md

PR #2362, opened 2023-06-05
External registries devel documentation

thehejik:
ref. #2318


thehejik:
@enrichman I tried to make it as close as possible to your "local" external registry deployment doc.

enrichman (approved):

Issue #2195, opened 2023-04-12
Spike: Proper status for deployed and not running service helm chart.


andreas-kupries:
This comes out of comment https://github.com/epinio/epinio/issues/2133#issuecomment-1502928703 and following.
Which is out of scope for the issue in question, actually.

PR #2364, opened 2023-06-05
Fixed unexpected waiting during async service deployment

### LINKED ISSUE #2288: Service deployment with hooks are still waiting
Ref: https://github.com/epinio/epinio/pull/2136

**Describe the bug**
Trying to deploy a service containing a Helm hook will still apparently hang the CLI because of how Helm hooks work: https://helm.sh/docs/topics/charts_hooks/

There is a feature request on Helm on how they will address this, but still under development: https://github.com/helm/helm/issues/10555

We can think about doing the installation in background (goroutine?) without the client waiting.

**To Reproduce**
Try to deploy a service with an hanging Job/hook: https://github.com/enrichman/helm-charts/tree/main/charts/simplejob

```yaml
kind: Service
metadata:
  name: simplejob
  namespace: {{ .Release.Namespace }}
spec:
  name: simplejob
  shortDescription: A simplejob
  description: |
    This simplejob
  chart: simplejob
  chartVersion: 0.0.1
  appVersion: 0.0.1
  helmRepo:
    name: enrichman
    url: https://enrichman.github.io/helm-charts
  values: |-
    {{- template "epinio.catalog-service-values" . }}
```

Installing this service will wait for the Job completion.

**Expected behavior**
Without the `--wait` flag I'm expecting the CLI to exit immediately without waiting for the service deployment.

enrichman:
Fixes #2288 

This PR fixes the async no-wait deployment of services, using a goroutine that will perform the installation of the service in background, even when it's using Helm hooks that are NOT going to be run asynchronously even when no-wait is specified (see the referenced ticket for more info regarding this behavior).

This mean that we should probably perform a better check of the service status.
- https://github.com/epinio/epinio/issues/2195

In this PR I've also fixed the test that will check immediately the existence of the service if run without the `--wait` (because it could take a while before it exists, and before it can be deleted), and added one more test that will not wait if the `--wait` is specified. In this case the creation should be synchronous, so when the API returns the service should exist.

### DIFF
--- a/acceptance/api/v1/service_create_test.go
+++ b/acceptance/api/v1/service_create_test.go
@@ -12,6 +12,7 @@
 package v1_test
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"io"
@@ -88,7 +89,6 @@ var _ = Describe("ServiceCreate Endpoint", LService, func() {
 	})
 
 	When("service exists", func() {
-		var requestBody, serviceName string
 		var catalogService models.CatalogService
 
 		BeforeEach(func() {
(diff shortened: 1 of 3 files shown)


andreas-kupries (approved):
lgtm

PR #2366, opened 2023-06-06
Add port forward for services (#2037)

giubacc:
Epinio services can now be accessed from the host by executing:

epinio service port-forward SERVICENAME [LOCAL_PORT] [...[LOCAL_PORT_N]] [flags]

Issue: https://github.com/epinio/epinio/issues/2037

### DIFF
--- a/pkg/api/core/v1/client/portforwarder.go
+++ b/pkg/api/core/v1/client/portforwarder.go
@@ -1,0 +1,218 @@
+// Copyright © 2021 - 2023 SUSE LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package client
+
+import (
+	"errors"
+	"fmt"
+	"io"
+	"net"
+	"net/url"
+	"strconv"
+
+	. "github.com/epinio/epinio/internal/api/v1/proxy"
+	"github.com/gorilla/websocket"
+)
+
+type ServicePortForwarder struct {
+	Client             *Client
+	Endpoint           string
+	ListeningAddresses []ListenAddress
+	Ports              []ForwardedPort
+	StopChan           <-chan struct{}
+}
+
+func NewServicePortForwarder(client *Client, endpoint string, addresses []string, ports []string, stopChan <-chan struct{}) (*ServicePortForwarder, error) {
+	if len(addresses) == 0 {
+		return nil, errors.New("you must specify at least 1 address")
+	}
+	parsedAddresses, err := ParseAddresses(addresses)
+	if err != nil {
+		return nil, err
+	}
+	if len(ports) == 0 {
+		return nil, errors.New("you must specify at least 1 port")
+	}
+	parsedPorts, err := ParsePorts(ports)
+	if err != nil {
+		return nil, err
+	}
+	return &ServicePortForwarder{
+		Client:             client,
+		Endpoint:           endpoint,
+		ListeningAddresses: parsedAddresses,
+		Ports:              parsedPorts,
+		StopChan:           stopChan,
+	}, nil
+}
+
+func (pf *ServicePortForwarder) ForwardPorts() error {
+	var err error
+
+	listenSuccess := false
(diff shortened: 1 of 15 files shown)


pkg/api/core/v1/client/portforwarder.go
```diff
@@ -0,0 +1,304 @@
...
+
+		forwards = append(forwards, ForwardedPort{uint16(localPort), uint16(remotePort)})
+	}
+
+	return forwards, nil
+}
+
+func (pf *ServicePortForwarder) ForwardPorts() error {
+	return pf.forward()
+}
```
enrichman, line 157: Do we need this? Can we just have either the `Forward()` exported, or the `ForwardPorts()` with the body of the `forward()`?

pkg/api/core/v1/client/portforwarder.go
```diff
@@ -0,0 +1,304 @@
...
+	<-pf.StopChan
+	pf.Client.log.V(1).Info("lost connection to pod")
+
+	return nil
+}
+
+func (pf *ServicePortForwarder) listenOnPort(port *ForwardedPort) error {
+	var errors []error
+	failCounters := make(map[string]int, 2)
+	successCounters := make(map[string]int, 2)
```
enrichman, line 189: I'm a bit confused about the logic and the usage of this counters.
If I understood correctly we are spawing goroutines for every addresses, and start listening.
Maybe we can achieve the same using waitgroups and checking the returned errors.

pkg/api/core/v1/client/portforwarder.go
```diff
@@ -0,0 +1,304 @@
...
+	return &ServicePortForwarder{
+		Client:             client,
+		Endpoint:           endpoint,
+		ListeningAddresses: parsedAddresses,
+		Ports:              parsedPorts,
+		StopChan:           stopChan,
+	}, nil
+}
+
+func parseAddresses(addressesToParse []string) ([]ListenAddress, error) {
```
enrichman, line 70: I would export this function, and put it at the very bottom, or in a different file, together with the `parsePorts`.

They are pretty "long" and can be easily unit tested.
Once we have some tests I think we will be able to simplify a bit the code with more confidence.

enrichman (changes requested):
I still need to have a better look at the `ServicePortForwarder` (that is actually the core of the PR), but trying it it seems that we are not able to return the errors.

I tried to connect to a non existing service, and the API are erroring (checking the log on the server). But still the connection is kept open. I think we are missing a check somewhere, probably within some goroutine.

pkg/api/core/v1/client/portforwarder.go
giubacc: I have exported these functions to the proxy.go file

pkg/api/core/v1/client/portforwarder.go
giubacc: These are simply maps used to count how many failures/successes there are after the user has started epinio client.
They are used to check if the initial listening asset has properly initialized.
This is supposed to fail sync when the user starts epinio client.
I don't see many much added value changing this from this implementation.

giubacc:
> I still need to have a better look at the `ServicePortForwarder` (that is actually the core of the PR), but trying it it seems that we are not able to return the errors.
> 
> I tried to connect to a non existing service, and the API are erroring (checking the log on the server). But still the connection is kept open. I think we are missing a check somewhere, probably within some goroutine.

If epinio client starts listening for a non-existent service I would say this is a bug. Fixing this btw, requires to make a call vs epinio server to check if the service actually exists.
This probably requires a non trivial subsequent implementation.
Given that the PR is already pretty big and I currently have near zero time to follow properly this, could we think to make this enhancement as a separate issue?

enrichman (approved):

mmartin24:
> > I still need to have a better look at the `ServicePortForwarder` (that is actually the core of the PR), but trying it it seems that we are not able to return the errors.
> > I tried to connect to a non existing service, and the API are erroring (checking the log on the server). But still the connection is kept open. I think we are missing a check somewhere, probably within some goroutine.
> 
> If epinio client starts listening for a non-existent service I would say this is a bug. Fixing this btw, requires to make a call vs epinio server to check if the service actually exists. This probably requires a non trivial subsequent implementation. Given that the PR is already pretty big and I currently have near zero time to follow properly this, could we think to make this enhancement as a separate issue?

Sorry I missed this comment. I checked it and I could reproduce the issue. Spawn separate ticket here: https://github.com/epinio/epinio/issues/2425 to be treated aside from the main functionality

mmartin24:
Reviewed [these](https://github.com/epinio/epinio/pull/2366#issuecomment-1578864263) checks in `v1.9.0-rc3` and working



Spawn separate issue from main functionality here: **epinio service port-forward should fail if service does not exist** #2425

PR #2367, opened 2023-06-06
Fixed wrong `upgrade responder` configuration

enrichman:
The Upgrade Responder instance was not starting because of a `null` `ReleaseDate` in the configuration.

This was generated because the usage of the authenticated Github APIs will return the draft as well.

This PR fixes this, filtering the draft releases.


giubacc (approved):
lgtm

PR #2374, opened 2023-06-12
chore: units tests for helper/routes.

andreas-kupries:
refactor: of core path function to have a testable function without panic. panic is in the higher-level caller now.

from before the vacation, was pushed, but not made into PR.

### DIFF
--- a/helpers/routes/routes.go
+++ b/helpers/routes/routes.go
@@ -42,8 +42,7 @@ func NewRoute(method string, path string, h gin.HandlerFunc) Route {
 // https://github.com/gorilla/mux#registered-urls
 type NamedRoutes map[string]Route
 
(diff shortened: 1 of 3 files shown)


enrichman (changes requested):
I was going to approve because the refactoring of the original `Path` function would be too complex, but I've seen that since we're using Ginkgo it's actually possible to test for panics, so we can easily test the `Path` func without adding a wrapper to it.
This will also cover the last lines that Codecov is complaining about.

See https://onsi.github.io/gomega/#panic

It should be something like:

```
It("returns an error for an unknown route", func() {
	unknownPath = func() {
		_ = nr.Path("c")
	}
	Expect(unknownPath).To(Panic())
})
```

enrichman (approved):
:rocket:

PR #2378, opened 2023-06-13
Enable default serviceaccount to use the regcred secret

thehejik:
This is about setting up default SA to use the existing `regcred` secret for dockerhub registries for pulling images.

It may help when CI reach public dockerhub rate-limit.

- [x] Needs to be tested first (any install test and upgrade).
* Upgrade test: https://github.com/epinio/epinio/actions/runs/5257824919
* Install test RKE-CI: https://github.com/epinio/epinio/actions/runs/5257843879

~But seems the `REGISTRY_USERNAME|PASSWORD` vars are not set in CI.~ addressed with this PR

- [ ] TODO retest CI
* Upgrade test
* Upgrade test with bound service
* Install test RKE-CI
* Install test RKE2-EC2-CI

### DIFF
--- a/.github/workflows/rke2-lh-ec2.yml
+++ b/.github/workflows/rke2-lh-ec2.yml
@@ -244,8 +244,8 @@ jobs:
       - name: Installation Acceptance Tests
         env:
           REGEX: Scenario4
-          REGISTRY_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
-          REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
+          REGISTRY_USERNAME: ${{ secrets.CFCIBOT_DOCKERHUB_USERNAME }}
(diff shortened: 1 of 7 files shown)


thehejik:
I just found out that the default SA is namespace scoped so we need to:
1) create the secrets in `epinio` NS
2) patch the default SA in `epinio` NS

Hopefully it won't be needed for other NSes like `workspace` or another custom NSes used for an app deployment testing.

- Other possibility would be to set the registry credentials globally on k3d level, similarly to what we did here https://github.com/epinio/epinio-end-to-end-tests/pull/351/ (k3s there)

thehejik:
NOTE: according to https://github.com/epinio/epinio/pull/2270
```
          REGISTRY_USERNAME: ${{ secrets.DOCKERHUB_USERNAME }}
          REGISTRY_PASSWORD: ${{ secrets.DOCKERHUB_TOKEN }}
```

enrichman:
Closed as too old, stale.

PR #2383, opened 2023-06-16
feat: add new function to get epinio values

### LINKED ISSUE #2188: Get real service name instead of Md5 hash

Working to have some native aws cloud services on epinio service catalog we  want to have the ability to tag the aws service with the real service name and not the md5 hash.
This would help identify the resources from aws point of view.

Through helm-chart values I'm able to have the {{  Release.name }}, which is the md5 hash.
Would be nice to have something like {{ .Values.epinio.servicename }}, where this returns the real service name.

Oziel-Silva:
fix #2188 

This PR adds fixed values as additional values for Helm.
Currently, two values are being added: serviceName and catalogServiceName.
However, in the future, it will be possible to easily add other values.
Using the structure below:
```
type FixedValues struct {
	ServiceName        string `yaml:"serviceName,omitempty"`
	CatalogServiceName string `yaml:"catalogServiceName,omitempty"`
	OrderValue         string `yaml:"otherValue,omitempty"`
}

 ``` 
to test it, use: `"{{ .Values.serviceName }}"` end/or `"{{ .Values.catalogServiceName }}" ` in your Helm chart!

### DIFF
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -157,7 +157,13 @@ func (s *ServiceClient) Create(ctx context.Context, namespace, name string, wait
 		return errors.Wrap(err, "error creating service secret")
 	}
 
-	catalogService.Values += fixedCatalogServiceValues(name, catalogService.Meta.Name)
+	epinioValues, err := getEpinioValues(name, catalogService.Meta.Name)
+	if err != nil {
+		logger := tracelog.NewLogger().WithName("Create")
+		logger.Error(err, "getting epinio values")
+	}
+	catalogService.Values += epinioValues
+
 	err = helm.DeployService(
 		ctx,
 		helm.ServiceParameters{
@@ -537,25 +543,27 @@ func convertUnstructuredListIntoHelmCharts(unstructuredList *unstructured.Unstru
 	return helmChartList, nil
 }
 
-func fixedCatalogServiceValues(serviceName, catalogServiceName string) string {
+func getEpinioValues(serviceName, catalogServiceName string) (string, error) {
+	//epinio:
+	//  serviceName: serviceName
+	//  catalogServiceName: catalogServiceName
 
-	type FixedValues struct {
+	type epinioValues struct {
 		ServiceName        string `yaml:"serviceName,omitempty"`
 		CatalogServiceName string `yaml:"catalogServiceName,omitempty"`
 	}
 
-	fixedValues := FixedValues{
+	extraValues := epinioValues{
 		ServiceName:        serviceName,
 		CatalogServiceName: catalogServiceName,
 	}
 
-	yamlData, err := yaml.Marshal(&fixedValues)
+	yamlData, err := yaml.Marshal(&struct {
+		Epinio epinioValues `yaml:"epinio,omitempty"`
+	}{extraValues})
 	if err != nil {
-		return ""
+		return "", err
 	}
 
-	yamlString := "\n" + string(yamlData)
-
-	// Return the fixed values
-	return yamlString
+	return "\n" + string(yamlData), nil
 }
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -156,7 +156,7 @@ func (s *ServiceClient) Create(ctx context.Context, namespace, name string, wait
 		return errors.Wrap(err, "error creating service secret")
 	}
 
-	catalogService.Values = concatenateCatalogServiceValues(catalogService, name)
+	catalogService.Values += concatenateCatalogServiceValues(name)
 
 	err = helm.DeployService(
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: internal/services/instances.go [hunk 1]
```diff
@@ -534,3 +536,25 @@ func convertUnstructuredListIntoHelmCharts(unstructuredList *unstructured.Unstru
...
+	}
+	var fixedValues FixedValues
+	var builder strings.Builder
+
+	fixedValues.ServiceName = name
+	// Build the concatenated string
+
+	builder.WriteString("\nserviceName: ")
+	builder.WriteString(fixedValues.ServiceName)
+	builder.WriteString("\n")
```
andreas-kupries, line 553: To clarify, you are building a YAML fragment here, right ?

internal/services/instances.go
```diff
@@ -534,3 +536,25 @@ func convertUnstructuredListIntoHelmCharts(unstructuredList *unstructured.Unstru
...
+
+	// Assign the result to catalogService.Values
+	catalogService.Values += builder.String()
```
andreas-kupries, line 556: I am not sure why you are adding this to `catalogService.Values` here, and return `catalogService.Values`, and then in the caller it is assigned to `catalogService.Values` again.

andreas-kupries: I would simply return the `builder.String()` result and do the `+=` in the caller.

internal/services/instances.go
Oziel-Silva: Yes, that's it!

andreas-kupries: In that case you may wish to know about the Go standard library yaml package which can do this for you from properly annotated structures. See https://pkg.go.dev/gopkg.in/yaml.v2#Marshal

andreas-kupries: The main point of this is that the `yaml` package does all the work for you to make the result a proper YAML string, with all the necessary quoting and what not.

... Example code https://dev.to/sagartrimukhe/generate-yaml-files-in-golang-29h1

internal/services/instances.go
andreas-kupries: And with the YAML generation confirmed I would return the string conversion of the `yaml.Marshall` result instead.

internal/services/instances.go
Oziel-Silva: Hi! It was implemented, thank you for your tips and examples, If you prefer I can move the struct `FixedValues` to the models package.

enrichman:
Hi @Oziel-Silva, thanks for the PR. This is actually what we were thinking to do, and the implementation looks pretty good to me now.

I was thinking about having a distinct key such as `epinio` to avoid any clashing with possible already existing keys.

i.e.: `serviceName` can be something related to a Kubernetes service, and this would overwrite any value provided from the user.

The final yaml will be:

```yaml
serviceName: kubernetes-service

# added extra values
serviceName: my-service
catalogServiceName: my-catalog-service
otherValue: foobar
```

Basically I would prefer to have a dedicated struct to have a reserved `epinio` object;

```yaml
name: foo
surname: bar

epinio:
  serviceName: serviceName
  catalogServiceName: catalog
```

To avoid creating an extra struct you can use an anonymous one.


```go
func getEpinioValues(serviceName, catalogServiceName string) (string, error) {
	type epinioValues struct {
		ServiceName        string `yaml:"serviceName,omitempty"`
		CatalogServiceName string `yaml:"catalogServiceName,omitempty"`
	}

	extraValues := epinioValues{
		ServiceName:        serviceName,
		CatalogServiceName: catalogServiceName,
	}

	yamlData, err := yaml.Marshal(&struct {
		Epinio epinioValues `yaml:"epinio,omitempty"`
	}{extraValues})
	if err != nil {
		return "", err
	}

	return "\n" + string(yamlData), nil
}
```

https://play.golang.com/p/zQxxJoAtnP3

Also I don't like suppressing errors, so I would return it, and at least log it. :slightly_smiling_face: 

```go
epinioValues, err := getEpinioValues(name, catalogService.Meta.Name)
if err != nil {
	logger := tracelog.NewLogger().WithName("Create")
	logger.Error(err, "getting epinio values")
}
catalogService.Values += epinioValues
```

What do you think? :smile:

Oziel-Silva:
Hi @enrichman, thank you for your review!!
Yes, you're right. Using only `serviceName` is indeed a concern. I have followed your recommendations (which are so good), and I have also implemented the error management as suggested.

andreas-kupries (approved):

andreas-kupries:
The #2393 is an in-project copy of the PR, for testing with CI and having all the necessary credentials.

enrichman (approved):

PR #2387, opened 2023-06-20
Skips failing tests [needs fix]

enrichman:
CI is broken because of some failing tests.

It seems that the problem is related to the ports used from the `port-forward` not being freed after the tests (not clear)

Needs fix, but this will unblock the CI.

Opened issue: https://github.com/epinio/epinio/issues/2388


enrichman:
Superseded by https://github.com/epinio/epinio/pull/2387

PR #2389, opened 2023-06-20
Support for service custom values

### LINKED ISSUE #1633: Allow setting service chart values with CLI
I would like to be able to set the underlying helm chart values of a service per instance using the CLI. This would allow me to configure say, a different DB size for a specific app that needs a service of that size.

Expansion to the `epinio service create` to allow passing in a values file. Potentially with some templating from epinio for the service name etc.

The only alternative I can think of is having a different custom service with pre-made settings for every possible combination.

**Additional context**
I could be missing something but I don't see this as being possible under the CLI here: https://docs.epinio.io/references/cli/epinio_service_create

Side question... is it not possible for these configuration values after the bind to be exposed as environment variables? From what I can tell they are just mounted according to this: https://docs.epinio.io/references/configurations

andreas-kupries:
Fix #1633 
Ref https://github.com/epinio/helm-charts/pull/441 (service CRD import)
Ref https://github.com/epinio/application/pull/13 (service CRD work)

Done: CRD extension, load from CR, show in CLI, passing into chart, tests

### DIFF
--- a/internal/api/v1/service/create.go
+++ b/internal/api/v1/service/create.go
@@ -14,6 +14,7 @@ package service
 import (
 	"github.com/epinio/epinio/helpers/kubernetes"
 	"github.com/epinio/epinio/internal/api/v1/response"
+	"github.com/epinio/epinio/internal/application"
 	"github.com/epinio/epinio/internal/services"
 	"github.com/gin-gonic/gin"
 
@@ -59,6 +60,22 @@ func Create(c *gin.Context) apierror.APIErrors {
 				WithDetailsf("catalog service %s not found", createRequest.CatalogService)
 		}
 		return apierror.InternalError(err)
+	}
+
+	// Validate the chart values, if any.
+	if len(createRequest.Settings) > 0 {
+		issues := application.ValidateCV(createRequest.Settings, catalogService.Settings)
+		if issues != nil {
+			// Treating all validation failures as internal errors.
+			// I can't find something better at the moment.
+
+			var apiIssues []apierror.APIError
+			for _, err := range issues {
+				apiIssues = append(apiIssues, apierror.NewBadRequestError(err.Error()))
+			}
+
+			return apierror.NewMultiError(apiIssues)
+		}
 	}
 
 	// Now we can (attempt to) create the desired service
--- a/internal/api/v1/service/create.go
+++ b/internal/api/v1/service/create.go
@@ -66,7 +66,7 @@ func Create(c *gin.Context) apierror.APIErrors {
 	if len(createRequest.Settings) > 0 {
 		issues := application.ValidateCV(createRequest.Settings, catalogService.Settings)
 		if issues != nil {
-			// Treating all validation failures as internal errors.
+			// Treating all validation failures as a bad request.
 			// I can't find something better at the moment.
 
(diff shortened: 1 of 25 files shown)


internal/api/v1/service/create.go
```diff
@@ -61,8 +62,25 @@ func Create(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err)
 	}
 
+	// Validate the chart values, if any.
+	if len(createRequest.Settings) > 0 {
+		issues := application.ValidateCV(createRequest.Settings, catalogService.Settings)
+		if issues != nil {
+			// Treating all validation failures as internal errors.
```
enrichman, line 69: Those are treated as bad request (that looks fine to me). Wrong/old comment?

internal/helmchart/settings.go
```diff
@@ -0,0 +1,55 @@
...
+package helmchart
+
+import (
+	"errors"
+
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+)
+
+func SettingsToChart(chart *unstructured.Unstructured) (map[string]models.ChartSetting, error) {
```
enrichman, line 21: I like the conversion from the unstructured object instead of using them in the code (I have seen that), but when we know the type we can use directly the kubernetes provided converter.

See

https://github.com/epinio/epinio/blob/bb5622333bd49c5ffa0d6edd32aeed08f8a12e22/internal/services/catalog.go#L84-L88

pkg/api/core/v1/models/settings.go
```diff
@@ -21,23 +21,23 @@ import (
 
 // This subsection of models provides structures related to the chart values of applications.
 
-// AppSetting represents the Show Response for a chart value variable
-type AppSetting struct {
+// CVSetting represents the Show Response for a chart value variable
+type CVSetting struct {
```
enrichman, line 25: I would avoid using single letter if not really required and clear. `ChartValueSetting` is still fine imo.

internal/helmchart/settings.go
andreas-kupries: Will have a look. Side note: We use a lot of explicit unstructured conversion for both app(charts) and service catalog, I believe. Should look into general simplification ...

internal/api/v1/service/create.go
andreas-kupries: Likely a copy/paste error. And possibly wrong in the origin location also.

pkg/api/core/v1/models/settings.go
andreas-kupries: Done with commit f5b1df5235731c0e44b6a0674c255dc77db412e2

internal/helmchart/settings.go
andreas-kupries: Done with commit fed941e7a2e4b1ac1d3e40af08ddfa6a6f9a9e47
Note: Left the appchart code untouched, i.e. this is still based on unstructured

thehejik (approved):
The feature works as expected, report https://github.com/epinio/epinio/issues/1633#issuecomment-1609171608

PR #2390, opened 2023-06-20
fix port-forward tests

enrichman:


giubacc:
> There is a sleep that we should try to remove, maybe enabling the ingress on the Nginx catalog service.

First of all, thanks for your improvements!

Regarding that sleep; I suspect that that will be not trivial to remove because we don't currently have a reliable way to check for an Epinio service to be ready (do we?).
One approach could be to rely on kubernetes APIs to check that, but I suspect this would be overly convoluted (retrieve all the pieces is a rabbit hole).

acceptance/api/v1/service_portforward_test.go
```diff
@@ -28,37 +27,37 @@ import (
 	. "github.com/onsi/gomega"
 )
 
-var _ = Describe("ServicePortForward Endpoint", Label("ServicePortForward"), func() {
-	var epinioHelper epinio.Epinio
+var _ = FDescribe("ServicePortForward Endpoint", LService, func() {
```
giubacc, line 30: F is a leftover?

giubacc (commented):
Apart the possibly left F, it seems pretty good.

acceptance/api/v1/service_portforward_test.go
enrichman: Yes, thanks! This issue is actually related to the bug that I showed you: https://github.com/onsi/ginkgo/issues/1221

The Focused tests should fail, but it doesn't seems the case when using Labels. So yes, it was a leftover and it was not catched by the CI for this reason. :+1:

enrichman:
Thanks @giubacc. In the end I've added the possibility to provide an ingress to the deployed Nginx.
This can be used to poll and check when the service is available to accept requests.

I've also removed the goroutine for the `cmd.Run`, because the `Run` is a blocking command that is wrapping the `Start` and `Wait`, but since we don't need to wait we can just start the command.

Let's wait for the CI  :smile:

giubacc (approved):
lgtm on CI completion (successfully :smirk:)

PR #2394, opened 2023-06-23
Try to stabilize CI - do not merge

svollath:
Reduce ginkgo nodes for CI


svollath:
As it turned out, our problems had been mostly related to max allowed pulls on dockerhub - so in regard to ginkgo node count it might be still interesting to see whether 6 vs. 8 nodes makes a difference, but in the end that's related to the runner performance and resources - so let's test, when we switched to GCP runners.
On the other hand, bumping the k3s version should be considered anyhow.

PR #2423, opened 2023-07-04
Added proper status check for Service deployments

enrichman:
This PR adds a proper status check for the Service deployments.

From the mittwald client I had to get the underlying Helm client, to run the `helm status` command with the `ShowResources` flag enabled.

This will return an `helm.Release` with the resources involved. Then thanks to the `kube.ReadyChecker` ([see](https://github.com/helm/helm/blob/main/pkg/kube/ready.go#L77-L82)) we are able to check if all the resources are in a ready state.

I had to move the `NewServiceStatusFromHelmRelease` because of a cycle dependency.

### DIFF
--- a/acceptance/helpers/catalog/misc.go
+++ b/acceptance/helpers/catalog/misc.go
@@ -44,10 +44,10 @@ func NginxCatalogService(name string) models.CatalogService {
 		},
 		Values: values,
 		Settings: map[string]models.ChartSetting{
-			"ingress.enabled": models.ChartSetting{
+			"ingress.enabled": {
 				Type: "bool",
 			},
-			"ingress.hostname": models.ChartSetting{
+			"ingress.hostname": {
 				Type: "string",
 			},
 		},
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -412,7 +412,7 @@ const (
 
 // Status will check for the readyness of the release returning an internal status instead of
 // the Helm release status (https://github.com/helm/helm/blob/main/pkg/release/status.go).
-// Helm is not checking for the actual status of the release and also if the resources are still
+// Helm is not checking for the actual status of the release and even if the resources are still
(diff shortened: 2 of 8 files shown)


internal/helm/helm.go
```diff
@@ -377,22 +398,78 @@ func Deploy(logger logr.Logger, parameters ChartParameters) error {
...
+	StatusUnknown ReleaseStatus = "unknown"
+	// StatusReady indicates that all the release's resources are in a ready state.
+	StatusReady ReleaseStatus = "ready"
+	// StatusNotReady indicates that not all the release's resources are in a ready state.
+	StatusNotReady ReleaseStatus = "not-ready"
+)
+
+// Status will check for the readyness of the release returning an internal status instead of
+// the Helm release status (https://github.com/helm/helm/blob/main/pkg/release/status.go).
+// Helm is not checking for the actual status of the release and also if the resources are still
```
andreas-kupries, line 415:
```suggestion
// Helm is not checking for the actual status of the release and even if the resources are still
```

andreas-kupries (changes requested):
comment nit

andreas-kupries (dismissed):

acceptance/helpers/catalog/misc.go
```diff
@@ -44,10 +44,10 @@ func NginxCatalogService(name string) models.CatalogService {
 		},
 		Values: values,
 		Settings: map[string]models.ChartSetting{
-			"ingress.enabled": models.ChartSetting{
+			"ingress.enabled": {
```
andreas-kupries, line 47: Go allows the removal of the type name here ?

enrichman: Yes, it's inherited by the initial map instantiation (my linter-vscode was complaining about it)!

image
image

andreas-kupries (approved):

PR #2426, opened 2023-07-04
Added generic Service delete

enrichman:
This PR change the `DELETE` in the service http client to use the generic delete method. It will use the APIError instead of the callback (see https://github.com/epinio/epinio/pull/2408).

With this approach we can finally drop the now unused `doWithCustomErrorHandling`.


mmartin24 (approved):
LGTM
tests pass in full

PR #2434, opened 2023-07-06
Skip cleanup for non-scheduled run

enrichman:
The cleanup will remove all the containers, also the mirror registry that we are using to cache the pull of our images.

To avoid this we can cleanup only at night during the scheduled workflows.


andreas-kupries (approved):
Let see if it works

Issue #2365, opened 2023-06-06
Display error message for service port-fowarding if user is logged out


mmartin24:
## Issue
Currently, if a user is logged out and a user performs a service port-forwarding, he/she is misled to think it is working, while it is not. This can be noticed once the connection with db is attempted
Related issue: https://github.com/epinio/epinio/pull/2366

## Steps to reproduce
Note: these steps can also be found here: [Connecting service port-forwarding to a database](https://confluence.suse.com/display/Epinio/Connecting+service+port-forwarding+to+a+database)
- Install Epinio
- Create a service (ie. MySQL)
- Log out
- Port-forward it (observe how no error seems to be displayed

![Image](https://user-images.githubusercontent.com/37271841/243640229-fff5e76f-4798-4a3a-b019-0989d3beffc2.png)

- Connect to DB
- Observe error

![Image](https://user-images.githubusercontent.com/37271841/243640370-612be2d5-04ba-4be9-ae42-52c3bcf67130.png)

# Request 
Previous check to API to ensure the user is logged in should be done, displaying some message inviting the user to log in

mmartin24:
Fix verified here: https://github.com/epinio/epinio/pull/2444#issuecomment-1630542097

PR #2444, opened 2023-07-10
Meaningful error message when starting service port-forwarding

giubacc:
Display error message for service port-forwarding if the user is logged out
    
Message displayed if the backend is down or if the client is not authenticated:
    
    "Cannot establish a connection with Epinio. Check both if Epinio is up
    and if the client is logged in."

Related to: https://github.com/epinio/epinio/issues/2365

### DIFF
--- a/internal/cli/usercmd/service.go
+++ b/internal/cli/usercmd/service.go
@@ -447,6 +447,14 @@ func (c *EpinioClient) ServicePortForward(ctx context.Context, serviceName strin
 		return err
 	}
 
+	/*
+		We use ServiceList API to check if we can establish a connection with the Epinio backend service.
+	*/
+	_, err := c.API.ServiceList(c.Settings.Namespace)
+	if err != nil {
+		return errors.Wrap(err, "Cannot establish a connection with Epinio server")
+	}
+
 	opts := client.NewPortForwardOpts(address, ports)
 	return c.API.ServicePortForward(c.Settings.Namespace, serviceName, opts)
 }
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -968,6 +968,34 @@ var _ = Describe("Services", LService, func() {
 			executePortForwardRequest("127.0.0.1", port2)
 		})
 
+		It("Port-forward succeeds after regular login fails after logout", func() {
+			tmpSettingsPath := catalog.NewTmpName("tmpEpinio") + `.yaml`
+			defer proc.Run("", false, "rm", "-f", tmpSettingsPath)
+
+			// login with a valid user
+			ExpectGoodUserLogin(tmpSettingsPath, env.EpinioPassword, serverURL)
+
+			By("port-forward with logged in user, expect success ...")
+
+			cmd := env.EpinioCmd("service", "port-forward", serviceName, "8081", "--settings-file", tmpSettingsPath)
+			err := cmd.Start()
+			Expect(err).ToNot(HaveOccurred())
+
+			DeferCleanup(func() {
+				err := cmd.Process.Kill()
+				Expect(err).ToNot(HaveOccurred())
+			})
+
+			// logout
+			_, err = env.Epinio("", "logout", "--settings-file", tmpSettingsPath)
+			Expect(err).ToNot(HaveOccurred())
+
+			By("port-forward with logged out user, expect failure ...")
+
+			_, err = env.Epinio("", "service", "port-forward", serviceName, "8082", "--settings-file", tmpSettingsPath)
+			Expect(err).To(HaveOccurred())
+		})
+
 		Context("command completion", func() {
 			It("matches empty prefix", func() {
 				out, err := env.Epinio("", "__complete", "service", "port-forward", "")


internal/cli/usercmd/service.go
```diff
@@ -447,6 +447,17 @@ func (c *EpinioClient) ServicePortForward(ctx context.Context, serviceName strin
...
+	/*
+		We use ServiceList API to check if we can establish a connection with the Epinio backend service.
+		In this scenario we are not really interested into retrieve details about the actual error returned;
+		Most common scenarios here are that the backend service is not up and running or that the client is not
+		logged it.
+	*/
+	_, err := c.API.ServiceList(c.Settings.Namespace)
+	if err != nil {
+		return errors.New("Cannot establish a connection with Epinio. Check both if Epinio is up and if the client is logged in.")
+	}
```
enrichman, line 459: From a recent refactor for logic errors we are now getting an `APIError` that can be used to check what kind of error was returned.
It's also better to `Wrap` the error instead of creating a new one, to get all the stack trace.


```suggestion
	/*
		We use ServiceList API to check if we can establish a connection with the Epinio backend service.
		In this scenario we are not really interested into retrieve details about the actual error returned.
		Most common scenarios here are that the backend service is not up and running or that the client is not
		logged in.
	*/
	_, err := c.API.ServiceList(c.Settings.Namespace)
	if err != nil {
		epinioAPIError := &client.APIError{}
		if errors.As(err, &epinioAPIError) {
			statusCode := epinioAPIError.StatusCode
			if statusCode == http.StatusUnauthorized || statusCode == http.StatusForbidden {
				return errors.Wrap(err, "user is not authorized")
			}
		}
		return errors.Wrap(err, "getting services")
	}
```

enrichman (commented):
Small comment/suggestion.

It should be possible also to add a small test for it.

internal/cli/usercmd/service.go
giubacc: Actually, if we wrap the error (as I initially did) we would get this kind of message:

```
🚢  Executing port forwarding
Namespace: workspace
Service: test

❌  error port forwarding to service: Connection with backend failed: making the request: Get "/api/v1/namespaces/workspace/services": unsupported protocol scheme ""

```

I don't like this error because it is cryptic.
What are the most common scenarios of failure here?

1) Server not running
2) Client not authenticated

IMO, the user would prefer a clear human readable error message rather than a too much technical one when the 99.9% of failure scenarios are the 2 above.

No problem changing this btw! :grin:

enrichman: Not really cryptic but it could be nicer. At the moment on the CLI side we are just bubbling up the errors with all the stacktrace, and as you said it's not very nice from an end user perspective.
I would prefer to keep the full ugly stack trace at the moment, and maybe open an enhancement task to improve this.

With the improvement of the HTTP client it's now possible to get the proper status code, and the error, in an easy way, and this could be a way to show proper errors.

Actually the most common scenarios here are "real" failures. If you are trying to interact with an `epinio service port-forward` command I'm expecting the client to be already logged in, with an epinio server running. It's not probably the first command that you are going to execute. :smile:

enrichman (approved):

mmartin24:
Checked on branch and working
Checked with existing and non-existing svc and ok

```
➜  epinio git:(srv-port-fwd-display-error-msg-on-client-logged-out) ✗ epinio service list                      

🚢  Listing Services...
Namespace: workspace

✔️  Details:
|   NAME    |            CREATED             | CATALOG SERVICE | VERSION |  STATUS  | APPLICATIONS |
|-----------|--------------------------------|-----------------|---------|----------|--------------|
| svc-mysql | 2023-07-11 12:01:30 +0200 CEST | mysql-dev       | 8.0.31  | deployed |              |
➜  epinio git:(srv-port-fwd-display-error-msg-on-client-logged-out) ✗ epinio logout      

🚢  Logout from Epinio clusters

✔️  Logout successful
➜  epinio git:(srv-port-fwd-display-error-msg-on-client-logged-out) ✗ epinio service port-forward svc-mysql 3308

🚢  Executing port forwarding
Namespace: workspace
Service: svc-mysql

❌  error port forwarding to service: Cannot establish a connection with Epinio. Check both if Epinio is up and if the client is logged in
```

internal/cli/usercmd/service.go
giubacc: > I would prefer to keep the full ugly stack trace at the moment, and maybe open an enhancement task to improve this.

ok, I will do.
I will try also to implement a test.
Unfortunately, I really dunno if I will be able to address this for this week (so this could actually shift to 1.10).

enrichman (approved):

mmartin24:
Rechecked in `1.9.0` version 

```
✔️  Epinio Environment
Platform: k3s
Epinio Server Version: v1.9.0
Epinio Client Version: v1.9.0-rc5
➜  epinio git:(main) ✗ epinio logout                                                       

🚢  Logout from Epinio clusters

✔️  Logout successful
➜  epinio git:(main) ✗ epinio service port-forward svc-mysql 3309

🚢  Executing port forwarding
Namespace: workspace
Service: svc-mysql

❌  error port forwarding to service: Cannot establish a connection with Epinio server: making the request: Get "/api/v1/namespaces/workspace/services": unsupported protocol scheme ""

```

PR #2446, opened 2023-07-10
Add region flag for eksctl-0.148.0 workaround

svollath:
EKS-CI fails within "Configure EKS EBS CSI storage" - where eksctl relies on AWS_REGION envvar, which was working up to eksctl-0.147.0, but is broken for current 0.148.0.
Fix for https://github.com/weaveworks/eksctl/issues/6774. While the workaround for ekstcl not reading AWS_REGION from env can stay, since the flag uses the var from github env.


svollath:
EKS-CI passed for this branch as well - https://github.com/epinio/epinio/actions/runs/5507614544

andreas-kupries (approved):

PR #2447, opened 2023-07-10
Added error check for release deletion

### LINKED ISSUE #2194: Honor pre-delete hooks for services
I have written a new catalog service that creates a container in a data storage whenever a service (that has been created with the new catalog service) is bound to an app. I added a pre-delete hook in my helm-chart that fails, when this container is not empty. This should prevent the service from being deleted, but Epinio does not honor that. I can see in the events/logs that the delete-hook is working as expected, but Epinio still deleted the service.

Epinio should check the pre-delete hook result and prevent the service from being deleted if it fails. In addition there should be the possibility for the dev to supply a reason message that then can be displayed by the Epinio client (similar to the one that shows when the service is still bound to an app)

There are really no good alternatives to this, the "safest" way would just be to leave the container contents alone, which would create a lot of "orphaned" content. Force-deleting the content is also not an option especially in a prod environment

**Additional context**

enrichman:
Fix #2194

When a release deletion fails we get in an inconsistent state because we are removing anyway the service.

It's probably better to return the error and let the operator trying to understand what's going on.

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -116,15 +116,7 @@ func RemoveService(logger logr.Logger, cluster *kubernetes.Cluster, app models.A
 	}
 
 	err = client.UninstallReleaseByName(names.ServiceReleaseName(app.Name))
-
-	// Ignore errors. The release may not be present. For example due to an aborted
-	// deployment. Note that a multitude of different errors was seen for essentially the same
-	// thing, depending on exact timing of deletion to partial creation. Just ignoring a
-	// specific one is fraught. Report, in case we were to generous and debugging is required.
-	if err != nil {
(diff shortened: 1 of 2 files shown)


andreas-kupries (approved):
The main issue with throwing the error is that can leave a partial service behind, and that then requires manual cleanup outside of E, as E will not be able to remove the remainder anymore.

That said, I see that the order of deletion was changed, i.e. helm release is removed before the kube secret holding E's representation of the instance.

That actually should address this issue. So, overall, looks good.

enrichman:
Yes, exactly. Changing the order of the check will keep the Epinio service if something will fail, so the user could even retry, if it was a transient issue. Thanks, merging this.

PR #2449, opened 2023-07-11
ci: add automation to generate README.md for tests

ldevulder:
This is to have an automated list of QA/integration tests.

It's already done in Elemental project. Not something perfect but could be useful to have a limited QA doc.
It's clearly not perfect and should be improved. The idea is to re-write the script in Go at some point an use it as a library.

### DIFF
--- a/.github/workflows/update-tests-description.yaml
+++ b/.github/workflows/update-tests-description.yaml
@@ -1,0 +1,40 @@
+name: Update tests description
+
+on:
+  schedule:
+    - cron: 0 23 * * *
+  workflow_dispatch:
+
+jobs:
+  update-tests-description:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+        with:
+          # A token is needed to be able to push on main, maybe this can be changed later
+          # with another GHA or with some webhook?
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - name: Generate tests description file
+        id: readme_generator
+        run: |
+          # Generate checksum of current file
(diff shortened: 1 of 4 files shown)


mmartin24:
Thank you very much @ldevulder.
I just checked the outcome and it looks awesome:

```
➜  acceptance git:(add-readme-for-ci) ✗ cat README.md             
# Tests description for acceptance

## `appcharts_test.go`

- **Describe:** apps chart, LAppchart
    - **It:** lists the known app charts
    - **It:** shows the details of the standard app chart
    - **It:** shows the details of the custom chart
    - **It:** fails to show the details of a bogus app chart
    - **It:** shows nothing by default
    - **It:** sets a default
    - **It:** fails to sets a bogus default
    - **It:** unsets a default
      -  **By:** setting default
      -  **By:** unsetting default
  - **Context:** command+ command completion
    - **It:** matches empty prefix
    - **It:** does not match unknown prefix
    - **It:** does not match bogus arguments

## `apps_env_test.go`

- **Describe:** apps env, LApplication
    - **It:** is not shown in the environment listing
    - **It:** is retrieved as empty string with show
    - **It:** is not present in the pushed workload
  - **Context:** command completion
    - **It:** matches empty app prefix
    - **It:** matches empty var prefix
    - **It:** does not match unknown app prefix
    - **It:** does not match unknown var prefix
    - **It:** does not match bogus arguments
    - **It:** creates the relevant secret
    - **It:** is shown in the environment listing

```

Thank you very much for it.

mmartin24 (dismissed):
LGTM

andreas-kupries (approved):

andreas-kupries (approved):

PR #2457, opened 2023-07-12
Added optional `restart` field to `AppUpdate` API

### LINKED ISSUE #2363: Cannot update source from a failed app
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Attempting to update a source from a failed app is not possible

### Expected Behavior

It should be possible to source a failed up and have it into a running state

### Steps To Reproduce

1. Install latest  `epinio-dev` commit (Currently: 23d440198092d8ac60089a7d176e041f2b9b68dd (HEAD -> epinio-dev, origin/epinio-dev) )
2. Deploy a faulty app (ie folder with images)
3. Try to update to a valid app (ie: source type = file with [sample-app.tar.gz](https://github.com/epinio/ui/files/11476924/sample-app.tar.gz))

Observe error:
image

### Environment

```markdown
- OS: Linux
- Architecture:amd64
- Epinio Version: v1.8.0-20-g912b8964
- Cluster:
  - Provider: k3d
  - Options:1 node
  - Kubernetes Version: v1.25.4+k3s1
```

### Logs

_No response_

### Anything else?

_No response_

enrichman:
This PR fix #2363 but it will need some work from the UI, to pass the same `restart: false` flag during the `push`.

After a lot of thoughts the problem was related to an unwanted call to the `DeployApp`, after an `UpdateApp`.

The problem is that if the DeployApp will fail, then the Update will fail as well, and no updated sources are going to be uploaded to try to fix the issue.
To fix this keeping the old behavior for backward compatibility and also to keep the same restart during the `epinio app update` command a new optional `restart` flag can be provided, to force the not-restart (by default it's true).

Changes to the `models.Manifest` struct was needed, because it was using internally the same `models.ApplicationUpdateRequest` as a `Configuration`. While containing the same fields (until now) these two struct are conceptually different, and they could/should be different.  A new similar `ApplicationConfiguration` struct was added, to avoid having the `Restart` flag in the Manifest.

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -773,6 +773,67 @@ var _ = Describe("Apps", LApplication, func() {
 		})
 	})
 
+	When("pushing a failed staging application", func() {
+
+		var tmpDir string
+		var err error
+
+		BeforeEach(func() {
+			By("Pushing an app that will fail")
+			tmpDir, err = os.MkdirTemp("", "epinio-failing-app")
+			Expect(err).ToNot(HaveOccurred())
+
+			err = os.WriteFile(path.Join(tmpDir, "empty"), []byte(""), 0644)
+			Expect(err).ToNot(HaveOccurred())
+
+			// Don't block because this push will only exit if it times out
+			go func() {
+				defer GinkgoRecover()
+				// Ignore any errors. When the main thread pushes the app again,
+				// this command will probably fail with an error because the helm release
+				// will be deleted by the other `push`.
+				_, _ = env.EpinioPush(tmpDir, appName, "--name", appName)
+			}()
+
+			// Wait until previous staging job is complete
+			By("waiting for the old staging job to fail")
+			Eventually(func() string {
+				statusJSON, err := proc.Kubectl(
+					"get", "jobs", "-A",
+					"-l", fmt.Sprintf("app.kubernetes.io/name=%s,app.kubernetes.io/part-of=%s", appName, namespace),
+					"-o", "jsonpath={.items[].status['conditions'][]}",
+				)
+				if err != nil || statusJSON == "" {
+					return ""
+				}
+
+				var status map[string]string
+				err = json.Unmarshal([]byte(statusJSON), &status)
+				Expect(err).ToNot(HaveOccurred())
+
+				return status["type"]
+			}, 3*time.Minute, 3*time.Second).Should(BeEquivalentTo("Failed"))
+
+			DeferCleanup(func() {
+				os.RemoveAll(tmpDir)
+			})
+		})
+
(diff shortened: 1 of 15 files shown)


acceptance/apps_test.go
```diff
@@ -773,6 +773,67 @@ var _ = Describe("Apps", LApplication, func() {
...
+				}
+
+				var status map[string]string
+				err = json.Unmarshal([]byte(statusJSON), &status)
+				Expect(err).ToNot(HaveOccurred())
+
+				return status["type"]
+			}, 3*time.Minute, 3*time.Second).Should(BeEquivalentTo("Failed"))
+
+			DeferCleanup(func() {
```
andreas-kupries, line 817: Maybe move this up to line 785, i.e. immediately after the `tmpdir` is created.

andreas-kupries (dismissed):
Approved with a small nit not affecting things. Just a possibly better location for setting up the dynamic cleanup (nearer to where the resource is created).

acceptance/apps_test.go
enrichman: Done. I'm happy on moving it because I initially had the same thought!

andreas-kupries (approved):

enrichman:
Speaking with @torchiaf this looks to be ok! Mergin' it. :rocket:

mmartin24:
Retested in `epinio-server v1.10.0-rc1`

- Check on CLI same as [here](https://github.com/epinio/epinio/issues/2363#issuecomment-1576881445) and working. 
  - Unzip this faulty-dir: [faulty-dir.zip](https://github.com/epinio/epinio/files/11653271/faulty-dir.zip)
  - Create app with faulty dir:
```
epinio push -n testapp-faulty -p ~/faulty-dir
```
- Push the correct folder but with `-i 1`:
```
epinio push -n testapp-faulty -p assets/sample-app -i 1
```
- No error is shown and app is correctly pushed:

```
➜  epinio git:(main) ✗ epinio push -n testapp-faulty -p assets/sample-app -i 1

Manifest: <<Defaults>>
Name: testapp-faulty
Source Origin: /home/mmartin/suse/epinio-repos/epinio/assets/sample-app
AppChart: 
Target Namespace: workspace
Instances: 1
Routes: <<default>>
```
Output:
```
Deploying application ...
.....
✔️  App is online.
Name: testapp-faulty
Namespace: workspace
Builder Image: 
Routes: 
1: https://testapp-faulty.192.168.16.3.omg.howdoi.website
➜  epinio git:(main) ✗  epinio info

✔️  Epinio Environment
Platform: k3s
Epinio Server Version: v1.10.0-rc1
Epinio Client Version: v1.10.0-rc1
```
--- 
- Rechecked as well in UI that app can be redeployed from faulty apps
image

PR #2460, opened 2023-07-13
Fixed wrong client initialization

enrichman:
Flags are parsed not during the creation of the command but just before the `Run`.

It means that the client initialized at the beginning was not looking for the settings in the specified location but always in the default one. This PR moves the initialization in the `PersistentPreRunE` (run before every command) for all the commands.

### DIFF
--- a/internal/cli/clientsync.go
+++ b/internal/cli/clientsync.go
@@ -12,13 +12,12 @@
 package cli
 
 import (
-	"github.com/epinio/epinio/internal/cli/usercmd"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 // NewClientSyncCmd implements the command: epinio client-sync
-func NewClientSyncCmd(client *usercmd.EpinioClient) *cobra.Command {
+func NewClientSyncCmd() *cobra.Command {
 	return &cobra.Command{
 		Use:   "client-sync",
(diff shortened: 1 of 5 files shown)


giubacc (approved):
lgtm, it should be good thinking to start to cover with tests these flags

enrichman:
> lgtm, it should be good thinking to start to cover with tests these flags

I absolutely agree. After the release and with the upcoming summer it could be something feasible.

mmartin24:
Checked in Epinio `v1.9.0-rc4`
Deleted the `settings.yaml` file and pointed to new `custom_settings.yaml`

Without fix (`v1.8.1`):
```
➜  epinio git:(main) ✗ epinio info --settings-file /home/manuel/.config/epinio/custom_settings.yaml 
2023/07/17 18:00:48 EpinioApiClient: "msg"="request failed" "error"="Get \"/api/v1/info\": unsupported protocol scheme \"\""

❌  error retrieving Epinio environment information: making the request: Get "/api/v1/info": unsupported protocol scheme ""

```

With fix (epinio `v1.9.0-rc4`)
```
➜  epinio git:(main) epinio info --settings-file /home/mmartin/.config/epinio/custom_settings.yaml

✔️  Epinio Environment
Platform: k3s
Epinio Server Version: v1.9.0-rc4
Epinio Client Version: v1.9.0-rc4
```

PR #2481, opened 2023-07-24
Empty commit

enrichman:
Since we have tagged the same commit then the release failed: https://github.com/epinio/epinio/actions/runs/5643293255

We need one more commit (empty)


mmartin24 (approved):

enrichman:
Ref: https://goreleaser.com/cookbooks/set-a-custom-git-tag/

PR #2485, opened 2023-07-26
Support array index syntax and deeper object nesting for custom values of services.

### LINKED ISSUE #2401: Support for list of strings as a value for custom service settings
original ref: https://github.com/epinio/epinio/issues/1633

Currently only standalone values as strings are supported, if you try to use settings for something like `annotations: {}` it will fail.

Reproducer:
* Add settings for existing catalog service:
```bash
$ kubectl edit services.application.epinio.io -n epinio mysql-dev
# add this at values level and save:
  settings:    
    primary.service.annotations:                                              
      type: string
```
* Deploy the service from modified catalog:
```
$ epinio service create mysql-dev mysql6 -v primary.persistence.annotations=first --wait

🚢  Creating Service...
Catalog: mysql-dev
Service: mysql6
Wait For Completion: true

❌  error creating service: service create failed: Internal Server Error: error deploying service helm chart: installing or upgrading service SYNC: 1 error occurred:
	* StatefulSet in version "v1" cannot be handled as a StatefulSet: json: cannot unmarshal string into Go struct field ObjectMeta.spec.volumeClaimTemplates.metadata.annotations of type map[string]string
```

* When index is used `-v primary.persistence.annotations[0]=second`
```bash
$ epinio service create mysql-dev mysql6 -v primary.persistence.annotations[0]=second --wait

🚢  Creating Service...
Catalog: mysql-dev
Service: mysql6
Wait For Completion: true

❌  error creating service: service create failed: Bad Request: Setting "primary.persistence.annotations[0]": Not known
```

andreas-kupries:
fix #2401

also fixes an issue with deleting a partially created service (kube secret, no helm release) exposed during testing.

### Testing

1. Make a copy of one of the services in the catalog, and add the settings below to its CR. Or just modify one of the entries by adding the settings to the CR. Then deploy E with the new or modified service. __Note__: The chosen settings have nothing to do with the settings of the services. That is ok. It means that playing with them will not affect service deployment.
2. Create service instances with various forms of custom values. See below for possibilities to test
3. Use `epinio service show ...` to see the custom values set on creation are returned/shown, and the correct.
4. The command `helm get values ...` can also be used show the full set of values. This has to be pointed at the helm release underlying the service. Note further that all our bitnami services will have a very large `extraDeploy` field in the returned values to confuse things. The customized settings should be at the end, after it.

```
 settings:
    "frobni.blag":
      type: string
```

  - `-v frobni.blag=foo`

    Regular setting, supported before the change, still supported.

  - `-v 'frobni.blag[0]=foo' -v 'frobni.blag[1]=bar'`

    Make `blag` an array holding `foo` and `bar`.

  - `-v 'frobni[0].blag=foo' -v 'frobni[1].blag=bar'`

    Make `frobni`! an array of maps. Each contains a `blag=...` assignment.

  - `-v 'frobni.blag[0]=foo' -v 'frobni[0].blag=bar'`

    __Mismatch__ :boom: First option sets `frobni` to type map, second option then tries to use `frobni` as type array.
    This fails, and leaves a partially created service instance behind (secret, but not helm release).
    __Note__ The first commits of the PR ensure that such a partial service can be deleted again!

:warning: Attention :warning: 

  - `-v 'frobni.blag={foo,bar}'`

    This syntax for array values, while supported by `helm`, is __not__ supported by E. The issue is that the `Cobra` package we use for command line processing does its own handling of the `,` character for list-type flags, which `-v` is. This interferes with what we would have liked to have here.

### DIFF
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -224,7 +224,16 @@ func (s *ServiceClient) Delete(ctx context.Context, namespace, name string) erro
 		models.NewAppRef(name, namespace),
 	)
 	if err != nil {
-		return errors.Wrap(err, "error deleting service helm release")
+		// [NF] NOTE: The err is some nested thing with a `not found` at the bottom.  The
+		// `apierrors.IsNotFound` does not recognize that. Its docs claim that it searches
+		// through the tree of wrapped errors. Maybe the `not found` is not a kube not
+		// found -- Helm ?
+		// ===> For now performing check by string match.
+		if !strings.Contains(err.Error(), "not found") {
+			return errors.Wrap(err, "error deleting service helm release")
+		}
+		// not found -> NAME may be a partially created service, i.e. secret exists, helm release does not.
+		// -> continue to deletion of the secret.
 	}
 
 	err = s.kubeClient.DeleteSecret(ctx, namespace, service)
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -13,6 +13,7 @@ package services
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
@@ -471,13 +472,21 @@ func setServiceStatusAndCustomValues(service *models.Service,
 func getValue(values chartutil.Values, key string, maybetable bool) (string, error) {
 	value, err := values.PathValue(key)
 	if err == nil {
-		valueAsString := fmt.Sprintf("%v", value)
+		data, err := json.Marshal(value)
+		if err != nil {
+			return "", err
+		}
+		valueAsString := string(data)
 		return valueAsString, nil
 	}
 	if maybetable {
 		tvalue, terr := values.Table(key)
 		if terr == nil {
-			valueAsString := fmt.Sprintf("%v", tvalue)
+			data, err := json.Marshal(tvalue)
+			if err != nil {
+				return "", err
+			}
+			valueAsString := string(data)
 			return valueAsString, nil
 		}
 	}
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -267,18 +267,23 @@ func (s *ServiceClient) DeleteAll(ctx context.Context, namespace string) error {
 
 	for _, srv := range services.Items {
 		// Inlined Delete() ... Avoids back and forth conversion between service and secret names
-		err := s.kubeClient.DeleteSecret(ctx, srv.ObjectMeta.Namespace, srv.ObjectMeta.Name)
-		if err != nil {
-			return errors.Wrap(err, "error deleting service secret")
-		}
-
 		service := srv.GetLabels()[ServiceNameLabelKey]
 
 		err = helm.RemoveService(requestctx.Logger(ctx),
 			s.kubeClient,
 			models.NewAppRef(service, srv.ObjectMeta.Namespace))
 		if err != nil {
-			return errors.Wrap(err, "error deleting service helm release")
+			// See [NF] for details
+			if !strings.Contains(err.Error(), "not found") {
+				return errors.Wrap(err, "error deleting service helm release")
+			}
+		}
+		err := s.kubeClient.DeleteSecret(ctx, srv.ObjectMeta.Namespace, srv.ObjectMeta.Name)
+		if err != nil {
+			if apierrors.IsNotFound(err) {
+				return nil
+			}
+			return errors.Wrap(err, "error deleting service secret")
 		}
 	}
 
--- a/internal/services/instances.go
+++ b/internal/services/instances.go
@@ -439,6 +439,24 @@ func setServiceStatusAndCustomValues(service *models.Service,
 			if err != nil {
 				// Not found - This custom value was not customized by the user.
 				// That is ok. Nothing to report.
+				// Actually, it may be customized using array syntax.
+				// Then it is not a pure nested map anymore. Trial this out.
+				// Done by checking shorter keys, starting from the longest.
+
+				pieces := strings.Split(key, ".")
+				pieces = pieces[0 : len(pieces)-1]
+
+				for len(pieces) > 0 {
+					key := strings.Join(pieces, ".")
+					customValue, err := configValues.PathValue(key)
+					if err == nil {
+						customValueAsString := fmt.Sprintf("%v", customValue)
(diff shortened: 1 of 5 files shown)


thehejik:
We found out that the most important (IMO) `map` structure is not supported yet - for example we would need to set multiple `<key>: <value>` within `annotations: {}` yaml by something like `-v primary.service.annotations.foo=bar -v primary.service.annotations.akey=avalue`.

thehejik:
Tested slightly with setting annotations: {} `map` in `mysql-dev` catalog service by adding:

andreas-kupries:
I see `primary.extraEnvVars` specified. The setup uses `primary.extraEnvVars[0].name=TZ`.
In validation this is reduced to `primary.extraEnvVars.name` as key, and that is not known.

I would use `primary.extraEnvVars.name` and `primary.extraEnvVars.value` as fields in the service spec.
Then setting them should be ok.

enrichman:
Functionality is fine, and I was able to use arrays and maps.

I don't like the display/show of the values.  


```
✔️  Settings
|          KEY           |             VALUE              |
|------------------------|--------------------------------|
| auth.password          | mypassword                     |
| auth.username          | enrico                         |
| primary.annotations    | map[]                          |
| primary.extraEnvVars   | [map[name:ENV_EXTRA            |
|                        | value:myvalue]]                |
| primary.podAnnotations | map[foo:bar]                   |
```

It will be nice to see if we can show the yaml/json version of them.
It's clearer for the user.

```
✔️  Settings
|          KEY           |         VALUE         |
|------------------------|-----------------------|
| auth.password          | mypassword            |
| auth.username          | enrico                |
| primary.annotations    |                       |
| primary.extraEnvVars   | - name: ENV_EXTRA     |
|                        |   value: myvalue      |
| primary.podAnnotations | foo: bar              |
```

```
✔️  Settings
|          KEY           |             VALUE                          |
|------------------------|--------------------------------------------|
| auth.password          | mypassword                                 |
| auth.username          | enrico                                     |
| primary.annotations    | {}                                         |
| primary.extraEnvVars   | [{"name":"ENV_EXTRA","value":"myvalue"}]   |
| primary.podAnnotations | {"foo":"bar"}                              |
```

andreas-kupries:
> Functionality is fine, and I was able to use arrays and maps.
> 
> I don't like the display/show of the values.
> It will be nice to see if we can show the yaml/json version of them. It's clearer for the user.
> 
> ```
> ✔️  Settings
> |          KEY           |         VALUE         |
> |------------------------|-----------------------|
> | auth.password          | mypassword            |
> | auth.username          | enrico                |
> | primary.annotations    |                       |
> | primary.extraEnvVars   | - name: ENV_EXTRA     |
> |                        |   value: myvalue      |
> | primary.podAnnotations | foo: bar              |
> ```
> 
> ```
> ✔️  Settings
> |          KEY           |             VALUE                          |
> |------------------------|--------------------------------------------|
> | auth.password          | mypassword                                 |
> | auth.username          | enrico                                     |
> | primary.annotations    | {}                                         |
> | primary.extraEnvVars   | [{"name":"ENV_EXTRA","value":"myvalue"}]   |
> | primary.podAnnotations | {"foo":"bar"}                              |
> ```

YAML requires control of line breaks and indentation, and the table package we use does not give us that (It smoshes everything into a single logical line and then inserts line breaks of its own. Original indentation is lost).

Will have a look into emitting JSON, that looks to be feasible.

internal/services/instances.go
```diff
@@ -437,3 +468,27 @@ func setServiceStatusAndCustomValues(service *models.Service,
...
+	value, err := values.PathValue(key)
+	if err == nil {
+		data, err := json.Marshal(value)
+		if err != nil {
+			return "", err
+		}
+		valueAsString := string(data)
+		return valueAsString, nil
+	}
+	if maybetable {
```
enrichman, line 482: Is it ok to look for this `maybetable` if there was previously an error?

enrichman (approved):
In a previous PR I already commented about splitting the two because it looks complex. It will be nice to have some unit tests for the `setServiceStatusAndCustomValues` because the custom values part looks a bit complicated but... not sure if it's possible.

Approving this btw.

internal/services/instances.go
andreas-kupries: Yes. As I found, `PathValue` may generate an error where `Table` does not.

thehejik:
Verified in `v1.10.0-rc1` in CLI only:

modified existing `mysql-dev` service:
```bash
$ kubectl edit services.application.epinio.io -n epinio mysql-dev
  settings:                                                
    image.pullSecret:
      type: string                         
    auth.rootPassword:                     
      type: string                                                                
    global.storageClass:                                                          
      enum:           
      - default       
      - local-path      
      - none            
      type: string      
    primary.extraEnvVars.name:
      type: string
    primary.extraEnvVars.value:
      type: string            
    primary.persistence.enabled:
      type: bool               
    primary.service.annotations:
      type: map                 
    primary.service.ports.mysql:
      maximum: "3307"           
      minimum: "3306"           
      type: integer
```

service deployed by:
```bash
$ epinio-1.10 service create mysql-dev mysql-custom \
  -v primary.service.annotations.foo=bar -v primary.service.annotations.color=black \
  -v image.pullSecret[0]=secret1 -v image.pullSecret[1]=secret2  \
  -v primary.extraEnvVars[0].name=TZ -v primary.extraEnvVars[0].value=CEST \
  -v primary.extraEnvVars[1].name=FOO -v primary.extraEnvVars[1].value=BAR \
  -v global.storageClass=local-path
```

verified by:
```bash
$ epinio-1.10 service show mysql-custom
✔️  Settings
|                KEY                |   VALUE    |
|-----------------------------------|------------|
| global.storageClass               | local-path |
| image.pullSecret[0]               | secret1    |
| image.pullSecret[1]               | secret2    |
| primary.extraEnvVars[0].name      | TZ         |
| primary.extraEnvVars[0].value     | CEST       |
| primary.extraEnvVars[1].name      | FOO        |
| primary.extraEnvVars[1].value     | BAR        |
| primary.service.annotations.color | black      |
| primary.service.annotations.foo   | bar        |
```
And also k8s resource as service (annotations) and pod (EnvVars) are configured accordingly.

PR #2501, opened 2023-08-01
Dependency updates

andreas-kupries:
- go modules
- docker images
- GHA support


enrichman (approved):
Usually we try to have the GHA, Go deps and Docker images as separated PRs, but LGTM.

We should drop dependabot in favour of updatecli to simplify this!

andreas-kupries:
> We should drop dependabot in favour of updatecli to simplify this!

Make a ticket to discuss this, please.

PR #2503, opened 2023-08-02
Added support for Git configurations

enrichman:
This PR adds the support for Git configurations.

A Git configuration can be used to provide some particular options to a Git instance/provider. It's a secret that holds some information that will be used during the cloning of a repository, such as the `username` and `password` to clone private repositories, the `skipSSL` option to skip the ssl verification of the host, or the `certificate` to load private certificates.

For Github and Gitlab the username is not actually used (but it needs to be provided), and the password is the token (see [issue](https://gitlab.com/gitlab-org/gitlab/-/issues/212953)).

A configuration is a struct holding this informations:

```go
type Configuration struct {
	// ID of the configuration (it maps to the kubernetes secret)
	ID string
	// URL is the full url (schema/host/port) used to match a particular instance
	URL      string
	Provider models.GitProvider
	// Username and Password are used to perform a BasicAuth.
	// For Github/Gitlab the username can be anything (see https://gitlab.com/gitlab-org/gitlab/-/issues/212953).
	Username string
	// The Personal Access Token
	Password string
	// UserOrg is used to specify the username/organization/project
	UserOrg string
	// Repository is used to specify the exact repository
	Repository  string
	SkipSSL     bool
	Certificate []byte
}
```

The only required information is the URL. It should be something like `https://github.com`. A configuration with only this info will apply to the all repositories of `https://github.com`. Otherwise if you want to restrict the configuration to be applied only to a particular organization/project you can specify also the `UserOrg` value. And if you would like to use it only for a specific repo then you can use the `Repository` as well.

If multiple configurations are found then the most specific will be applied.

The secret containing the data needs to have the label ` epinio.io/api-git-credentials: "true"`, and the following fields are available:

```yaml
  url: https://github.com
  provider: github
  username: "myuser" 
  password: "abcde12345" 
  userOrg: epinio 
  repo: helm-charts 
  skipSSL: true 
  certificate: |
    -----BEGIN CERTIFICATE-----
    MIIBaTCCAQ+gAwIBAgIRAN4tvwEOKogvOzT/KccL8t8wCgYIKoZIzj0EAwIwFDES
    ***************
    -----END CERTIFICATE-----
```

Example:

```yaml
apiVersion: v1 
kind: Secret 
type: Opaque 
metadata: 
  labels: 
    epinio.io/api-git-credentials: "true" 
  name: my-github-token 
  namespace: epinio 
stringData:
  url: https://github.com
  provider: github
  username: "myuser" 
  password: "abcde12345" 
  userOrg: epinio 
  repo: helm-charts 
  skipSSL: true 
  certificate: |
    -----BEGIN CERTIFICATE-----
    MIIBaTCCAQ+gAwIBAgIRAN4tvwEOKogvOzT/KccL8t8wCgYIKoZIzj0EAwIwFDES
    ***************
    -----END CERTIFICATE-----
```

Note:
the Git manager was put under the `internal/bridge/git` package. Since the `internal` folder is containing everything maybe it's worth to restructure a bit the organization moving all the "external" clients/managers in this folder (s3, helm, upgraderesponder, and maybe something else).

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -337,7 +337,7 @@ var _ = Describe("Apps", LApplication, func() {
 				Expect(out).To(ContainSubstring("authentication required"))
 			})
 
-			It("pushes the app providing a token", func() {
+			It("pushes the app when providing a proper token", func() {
 				tmpTokenDir, err := os.MkdirTemp("", "tmp-token-dir")
 				Expect(err).ToNot(HaveOccurred())
 
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -53,6 +53,7 @@ var _ = Describe("Apps", LApplication, func() {
 	)
 	containerImageURL := "splatform/sample-app"
 	wordpress := "https://github.com/epinio/example-wordpress"
+	privateRepo := "https://github.com/epinio/example-go-private"
 
 	BeforeEach(func() {
 		namespace = catalog.NewNamespaceName()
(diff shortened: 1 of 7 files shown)


acceptance/apps_test.go
```diff
@@ -328,6 +329,59 @@ var _ = Describe("Apps", LApplication, func() {
...
 
+		When("pushing an app from a private repository", func() {
+
+			It("rejects without a token", func() {
+				out, err := env.Epinio("", "push", "--name", appName, "--git", privateRepo)
+				Expect(err).To(HaveOccurred(), out)
+				Expect(out).To(ContainSubstring("authentication required"))
+			})
+
+			It("pushes the app providing a token", func() {
```
andreas-kupries, line 340:
```suggestion
			It("pushes the app when providing a proper token", func() {
```

internal/bridge/git/git.go
```diff
@@ -0,0 +1,246 @@
...
+
+type Manager struct {
+	logger         logr.Logger
+	SecretLister   SecretLister
+	Configurations []Configuration
+}
+
+// Configuration is used to setup a Git requests to a git provider.
+// The only required field is the URL, needed to check the specific instance to apply the configuration.
+// If also the UserOrg and/or the Repository are specified then the most specific configuration will be used.
```
andreas-kupries, line 43:
```suggestion
// If the UserOrg and/or the Repository are also specified then the most specific configuration will be used.
```

internal/bridge/git/git.go
```diff
@@ -0,0 +1,246 @@
...
+	List(ctx context.Context, opts metav1.ListOptions) (*v1.SecretList, error)
+}
+
+type Manager struct {
+	logger         logr.Logger
+	SecretLister   SecretLister
+	Configurations []Configuration
+}
+
+// Configuration is used to setup a Git requests to a git provider.
```
andreas-kupries, line 41:
```suggestion
// Configuration is used to customize the Git requests for a  specific git provider.
```

andreas-kupries (changes requested):
Comment nits. Functionally it looks ok.

enrichman:
Thanks! I will need a check on the doc as well for sure.

andreas-kupries (approved):

andreas-kupries:
Note, I do not see a doc PR (yet?)

enrichman:
> Note, I do not see a doc PR (yet?)

Yep, still writing it!

PR #2509, opened 2023-08-09
added export to registries

### LINKED ISSUE #1770: Export push directly to OCI registry
I would like to push my application as a set of OCI artifacts so that Fleet can pick it up and run/upgrade the application in many downstream clusters.

I would like an additional subcommand that will pack up a built and deployed application and push it.

`epinio export <app-name> --container <container-OCI-tag> --chart <chart-OCI-tag>` 

This would require auth which could be difficult so it should likely pull down locally then push using the client's auth. It would make sense to just be a second step after https://github.com/epinio/epinio/issues/1352 

**Additional context**
What might be a bit tricky is that the two are obviously linked so the requested container tag might need to be rendered into the chart's values as well.

andreas-kupries:
fix #1770
ref: https://github.com/epinio/docs/pull/304

TODO:
  - tests (unclear how to make a local oci registry in the test environment)

### DIFF
--- a/internal/api/v1/application/export.go
+++ b/internal/api/v1/application/export.go
@@ -201,7 +201,16 @@ func ExportToRegistry(c *gin.Context) apierror.APIErrors {
 
 	destinationURL, err := url.Parse(destination.URL)
 	if err != nil {
-		return apierror.InternalError(err)
+		if !strings.Contains(destination.URL, "://") {
+			// Run again with a scheme - This ensures that host and path are properly separated.
+			// Without a scheme the parser considers everything to be the path.
+			destinationURL, err = url.Parse("oci://" + destination.URL)
+			if err != nil {
+				return apierror.InternalError(err)
+			}
+		} else {
+			return apierror.InternalError(err)
+		}
 	}
 	if destinationURL.Scheme == "" {
 		// Run again with a scheme - This ensures that host and path are properly separated.
--- a/internal/api/v1/application/export.go
+++ b/internal/api/v1/application/export.go
@@ -1,0 +1,652 @@
+// Copyright © 2021 - 2023 SUSE LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package application
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/epinio/epinio/helpers/kubernetes"
+	"github.com/epinio/epinio/internal/api/v1/response"
+	"github.com/epinio/epinio/internal/appchart"
+	"github.com/epinio/epinio/internal/application"
+	"github.com/epinio/epinio/internal/cli/server/requestctx"
+	"github.com/epinio/epinio/internal/helm"
+	"github.com/epinio/epinio/internal/helmchart"
+	"github.com/epinio/epinio/internal/names"
+	"github.com/epinio/epinio/internal/registry"
+	apierror "github.com/epinio/epinio/pkg/api/core/v1/errors"
+	"github.com/epinio/epinio/pkg/api/core/v1/models"
+	"github.com/gin-gonic/gin"
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
+	"helm.sh/helm/v3/pkg/action"
+	"helm.sh/helm/v3/pkg/chart"
+	"helm.sh/helm/v3/pkg/chart/loader"
+	"helm.sh/helm/v3/pkg/chartutil"
+	batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+)
+
+// ExportToRegistry handles the API endpoint GET /namespaces/:namespace/applications/:app/export
+// It exports the named application to the export registry (also specified by name)
+func ExportToRegistry(c *gin.Context) apierror.APIErrors {
+	// //////////////////////////////////////////////////////////////////////////////
+	/// Validate request, and fill in defaults where needed
+
+	ctx := c.Request.Context()
+	namespace := c.Param("namespace")
+	appName := c.Param("app")
+	logger := requestctx.Logger(ctx)
+
+	req := models.AppExportRequest{}
+	if err := c.BindJSON(&req); err != nil {
+		return apierror.NewBadRequestError(err.Error()).
+			WithDetails("failed to unmarshal app export request")
+	}
+
+	cluster, err := kubernetes.GetCluster(ctx)
+	if err != nil {
+		return apierror.InternalError(err, "failed to get access to a kube client")
+	}
+
+	// Get application
+	theApp, err := application.Lookup(ctx, cluster, namespace, appName)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	// destination validation I - do we have a name ?
+	if req.Destination == "" {
+		return apierror.NewBadRequestError("export destination is missing/empty")
+	}
+
+	destination, certSecretName, err := checkDestination(ctx, cluster, req.Destination)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+	logger.Info("OCI export", "destination", destination.URL, "certs@", certSecretName)
+
+	// Make the certs available as file for use by registry clients, if any.
+	certFile := ""
+	if certSecretName != "" {
+		certFile, err = loadCerts(ctx, cluster, certSecretName)
+		if err != nil {
+			return apierror.InternalError(err)
+		}
+		defer func(path string) {
+			cleanupLocalPath(logger, "cert", path)
+		}(certFile)
+	}
+
+	logger.Info("OCI export transport", "cert file", certFile)
+
+	applyDefaults(&req, namespace, appName, theApp.StageID)
+
+	// //////////////////////////////////////////////////////////////////////////////
+	/// Helper information
+	//
+	// ATTENTION: We may see multiple requests for export of the same application, at near the
+	// same time. Using unix nanoseconds as unique file base name to separate these in the file
+	// system.
+
+	// NOTE: Using namespace and appname in the file name is considered by the security checker
+	// as a means of introducing path injection attacks. Not true actually, as the names are
+	// restricted by the kube regexes. Still easier to just use the timestamp.
+
+	trimmedDestination := trim(destination.URL)
+
+	base := fmt.Sprintf("oci-export-%d", time.Now().UnixNano())
+
+	chartLocalFile := base + "-chart.tar.gz"
+	imageLocalFile := base + "-image.tar"
+	imageRemoteFile := trimmedDestination + "/" + req.ImageName + ":" + req.ImageTag
+
+	logger.Info("OCI export local chart", "path", chartLocalFile)
+	logger.Info("OCI export local image", "path", imageLocalFile)
+	logger.Info("OCI export remote image", "path", imageRemoteFile)
+
+	// //////////////////////////////////////////////////////////////////////////////
+	/// Retrieve the app data and save it to local files.
+	//
+	// The base destination path is (const `imageExportVolume`)
+	// See sibling file `part.go` for the definition.
+	// This is the directory for export files, as seen from here, the API server.
+	//
+	// ATTENTION: The image upload job sees the file at a different base path, as per the volume
+	//           and mount declarations.
+
+	logger.Info("OCI export fetch chart archive", "path", chartLocalFile)
+	apierr := fetchAppChartFile(c, ctx, logger, cluster, theApp, chartLocalFile)
+	if apierr != nil {
+		return apierr
+	}
+	defer func(path string) {
+		cleanupLocalPath(logger, "chart", path)
+	}(imageExportVolume + chartLocalFile)
+	// Note: By passing the string as argument instead of through the closure we can change
+	// chartLocalFile without affecting what file is removed by the deferal.
+
+	client, err := helm.GetHelmClient(cluster.RestConfig, logger, namespace)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	logger.Info("OCI export fetch values")
+
+	values, err := client.GetReleaseValues(names.ReleaseName(theApp.Meta.Name), false)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	// And fetch image ...
+
+	logger.Info("OCI export fetch image", "path", imageLocalFile,
+		"origin", "docker://"+theApp.ImageURL)
+	imageFile, err := fetchAppImageFile(ctx, logger, cluster, theApp, imageLocalFile)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+	imageFile.Close() // Note: We do not need the open reader here.
+	defer func(path string) {
+		cleanupLocalPath(logger, "image", path)
+	}(imageExportVolume + imageLocalFile)
+
+	// //////////////////////////////////////////////////////////////////////////////
+	///
+	// Rewrite the chart for the desired name/tag. Also integrate the app configuration,
+	// i.e. the data from app's `values.yaml`. As part of that insert the proper image
+	// reference too.
+
+	values["epinio"].(map[string]interface{})["imageURL"] = imageRemoteFile
+
+	tmp, chartLocalFile, err := rewriteChart(logger,
+		imageExportVolume+chartLocalFile,
+		req.ChartName, req.ChartVersion, values)
+	if tmp != "" {
+		// This removes the tarball as well, as it resides in tmp.
+		defer func(path string) {
+			cleanupLocalPath(logger, "chart repack dir", path)
+		}(tmp)
+	}
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	logger.Info("OCI export rewritten chart", "chart", chartLocalFile)
+
+	// //////////////////////////////////////////////////////////////////////////////
+	/// Upload the parts to the chosen destination
+
+	// chart archive ...
+
+	destinationURL, err := url.Parse(destination.URL)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+	if destinationURL.Scheme == "" {
+		// Run again with a scheme - This ensures that host and path are properly separated.
+		// Without a scheme the parser considers everything to be the path.
+		destinationURL, err = url.Parse("oci://" + destination.URL)
+		if err != nil {
+			return apierror.InternalError(err)
+		}
+	}
+
+	logger.Info("OCI export destination", "url scheme", destinationURL.Scheme)
+	logger.Info("OCI export destination", "url host", destinationURL.Host)
+	logger.Info("OCI export destination", "url path", destinationURL.Path)
+
+	destinationHost := trim(destinationURL.Host)
+
+	logger.Info("OCI export login", "host'", destinationHost)
+
+	rOpts := []action.RegistryLoginOpt{}
+	if certFile != "" {
+		logger.Info("OCI export login", "+certs", certFile)
+		rOpts = append(rOpts, action.WithCAFile(certFile))
+	}
+
+	err = client.RegistryLogin(destinationHost, destination.Username, destination.Password, rOpts...)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	ociRegistry := destination.URL
+	if !strings.HasPrefix(ociRegistry, "oci://") {
+		ociRegistry = "oci://" + trimmedDestination
+	}
+
+	logger.Info("OCI export push", "registry", ociRegistry, "from", chartLocalFile)
+
+	pOpts := []action.PushOpt{}
+	if certFile != "" {
+		logger.Info("OCI export push", "+certs", certFile)
+		pOpts = append(pOpts, action.WithTLSClientConfig("", "", certFile))
+	}
+	_, err = client.Push(chartLocalFile, ociRegistry, pOpts...)
+	// NOTE: Neither chart name nor version are specified here!
+	// See `rewriteChart` above for the place where this information is inserted.
+
+	logger.Info("OCI export upload chart error", "err", err)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	// image ...
+
+	imageJob := createCopyJob(logger,
+		"docker-archive:/workspace/"+imageLocalFile,
+		// NOTE: `/workspace` is the `imageExportVolume` as seen from the job.
+		"docker://"+imageRemoteFile,
+		req.Destination, certSecretName)
+
+	err = runJob("image push", ctx, cluster, logger, imageJob)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	// //////////////////////////////////////////////////////////////////////////////
+	response.OK(c)
+	return nil
+}
+
+func rewriteChart(logger logr.Logger, path, name, version string, params map[string]interface{}) (string, string, error) {
+	logger.Info("OCI export load chart", "chart", path)
+
+	tmpDir, err := os.MkdirTemp("", "oci-export")
+	if err != nil {
+		return "", "", errors.Wrap(err, "can't create temp directory")
+	}
+
+	appChart, err := loader.Load(path)
+	if err != nil {
+		return tmpDir, "", err
+	}
+
+	logger.Info("OCI export chart meta", "chart", appChart.Metadata)
+	logger.Info("OCI export chart values", "chart", appChart.Values)
+
+	logger.Info("OCI export rewrite chart", "name", name, "version", version)
+	logger.Info("OCI export rewrite chart", "values", params)
+
+	appChart.Metadata.Name = name
+	appChart.Metadata.Version = version
+
+	merged := chartutil.CoalesceTables(params, appChart.Values)
+
+	// ATTENTION: While we can directly edit the meta data and see it in the saved tarball the
+	// handling of the values is more complex. The `appChart.Values` is a read-only convenience
+	// field. To make changes it is necessary to edit the `Raw` slice holding the actual file
+	// contents, including the `values.yaml`. Being a slice we have to do a linear search :( And
+	// as the app chart may be a user-specified thing, not our standard consider the possibility
+	// that it does not have a `values.yaml`.
+
+	yaml, err := yaml.Marshal(merged)
+	if err != nil {
+		return tmpDir, "", err
+	}
+	ok := false
+	for index, file := range appChart.Raw {
+		if file.Name == chartutil.ValuesfileName {
+			appChart.Raw[index].Data = yaml
+			ok = true
+			break
+		}
+	}
+	if !ok {
+		// Values file missing, create it.
+		appChart.Raw = append(appChart.Raw, &chart.File{
+			Name: chartutil.ValuesfileName,
+			Data: yaml,
+		})
+	}
+
+	logger.Info("OCI export rewritten chart meta", "chart", appChart.Metadata)
+	logger.Info("OCI export rewritten chart values", "YAML", string(yaml))
+
+	logger.Info("OCI export validate chart")
+	err = appChart.Validate()
+	if err != nil {
+		return tmpDir, "", err
+	}
+
+	logger.Info("OCI export save chart", "tmp", tmpDir)
+	tarball, err := chartutil.Save(appChart, tmpDir)
+	return tmpDir, tarball, err
+}
+
+func applyDefaults(req *models.AppExportRequest, namespace, appName, stageID string) {
+	base := fmt.Sprintf("%s-%s", namespace, appName)
+
+	if req.ImageName == "" {
+		req.ImageName = base + "-image"
+	}
+	if req.ChartName == "" {
+		req.ChartName = base + "-chart"
+	}
+	if req.ImageTag == "" {
+		req.ImageTag = stageID
+	}
+	if req.ChartVersion == "" {
+		req.ChartVersion = "0"
+	}
+}
+
+func checkDestination(ctx context.Context, cluster *kubernetes.Cluster,
+	destination string) (registry.RegistryCredentials, string, error) {
+
+	empty := registry.RegistryCredentials{}
+
+	// destination validation II - do we have a secret for the name ?
+	destinationSecret, err := cluster.GetSecret(ctx, helmchart.Namespace(), destination)
+	if err != nil {
+		return empty, "", errors.Wrap(err, "bad export destination")
+	}
+
+	// destination validation III - is the secret good ?
+	marker, ok := destinationSecret.ObjectMeta.Labels[kubernetes.EpinioAPIExportRegistryLabelKey]
+	if !ok {
+		return empty, "", errors.New("bad export destination: marker label is missing")
+	}
+
+	if marker != "true" {
+		return empty, "", errors.New("bad export destination: bad value for marker label")
+	}
+
+	creds, err := registry.GetRegistryCredentialsFromSecret(*destinationSecret)
+	if err != nil {
+		return empty, "", errors.Wrap(err, "bad export destination: bad url")
+	}
+
+	// Check for and use a cert secret sibling to the auth secret.
+
+	certSecretName := ""
+	certSecret, ok := destinationSecret.Data["certs"]
+	if ok {
+		certSecretName = string(certSecret)
+	}
+
+	return creds, certSecretName, nil
+}
+
+// cleanupLocalPath is invoked in deferals to remove the temporary files and directories used by an
+// export after they are not required any longer.
+func cleanupLocalPath(logger logr.Logger, label, path string) {
+	logger.Info("OCI export cleanup local "+label, "path", path)
+	err := os.RemoveAll(path)
+	if err != nil {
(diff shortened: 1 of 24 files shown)


andreas-kupries:
Research notes wrt ORAS

> --artifact-type ...
--registry-config /path

> https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#defining-a-unique-artifact-type
https://github.com/opencontainers/artifacts/blob/main/artifact-authors.md#visualizing-artifacts

> Artifact 	config.mediaType
OCI Image 	application/vnd.oci.image.config.v1+json
Helm Chart 	application/vnd.cncf.helm.config.v1+json

andreas-kupries:
`oras` looks to be the way to go, despite issues.
upload works and leaves data in the registry I can see from the dashboard.
issues are clearly me not using the proper command line arguments yet for helm charts and container images.
what is in the registry has the type, but not the proper actual content/format.

andreas-kupries:
More investigation and reading of docs showed that `oras`  looks to be too focused on lowl-level OCI artifacts and blobs, and that using it for pushing images and charts would require quite a lot of work, i.e. push layer blobs, config blob, and manifest separately. Was not able to find examples for `oras` pushing an image and/or chart directly from a image save tar and/or chart tarball. Most of the found docs simply direct to usage of `helm` for charts, and `skopeo` and others for images.

Regarding helm charts we have the mittwald client, based on the regular helm packages. I suspect that it is these which actually also pull the oras packages into E (they are in `go.mod` as `// indirect`). Time to try that.

Going back to `skopeo` means getting deeper information on why the push I am doing fails a few layers in on attempting to reuse an existing layer.

andreas-kupries:
@mattfarina

Hoping that you can help. 
The context is #1170 asking E(pinio) to push/export all parts (*) of an application into an OCI registry.

I experimented a bit, as can be seen in the preceding comments on this PR.

Using a `skopeo`  job to upload the image looks to be conceptually ok, is however running into permission issues partway into the upload (apparently when it tries to share a layer which is already found in the registry).

Tried to use the `oras` client in a job for `values.yaml` file (:heavy_check_mark: ) and helm chart (available as tarball) (:boom:). Found no examples for properly uploading a helm chart with the `oras` client. 

I hope that you can refer me to examples for uploading images, helm charts, and files into an OCI registry.

__Note__ While my experimentation was currently based on using `skopeo` and `oras` clients, Epinio does contain the go helm and oras packages (through the https://github.com/mittwald/go-helm-client). This means that examples directly using these are also in the table, as it would us to avoid the spawning of kube jobs.

----
(Ad *) IOW, (docker)/container image holding the runtime, base helm chart, and the `values.yaml` file linking helm chart and image

andreas-kupries:
Testing requires access to an OCI registry. for myself I tested against docker hub, using my own account/org there.
I used the following secret to make the registry known to E (actual credentials blanked). As docker hub is public I had no need for the optional cert secret. If needed the secret\s name has to be set under the key `certs` of the auth secret below.
The cert secret needs a key `tls.cert` which contains the actual certs to use in PEM format. It will be inserted into the skopeo pod directory hierarchy to be seen by standard cert code

 __Note__: As the `helm push` is done via the mittwald client and the regular helm/oras packages these may not see such a cert yet.

As such we really have to test against an OCI registry requiring transport certs to make sure everything is good.

```
---
apiVersion: v1
kind: Secret
type: kubernetes.io/dockerconfigjson
metadata:
  annotations:
    epinio.io/registry-namespace: "a99k"
  labels:
    epinio.io/api-export-registry: "true"
  name: a99k-at-docker
  namespace: epinio
stringData:
  .dockerconfigjson: |-
    {
      "auths": {
        "registry.hub.docker.com": {
          "auth":"************************",
          "username":"a99k",
          "password":"************************"
        }
      }
    }
```

andreas-kupries:
The comment `// Don't use the default registry client if tls options are set.` is indeed the main clue.
Tracing through uploader, pusher ot oci pusher it becomes clear that the TLS settings are only used
if the pusher has to create its own registry client with `newRegistryClient`. This in turn is triggered
by providing `nil` instead of our own RegistryClient.

Changing the push code to

```
	ac := concreteHelmClient.ActionConfig
	ac.RegistryClient = nil

	opts = append(opts, action.WithPushConfig(ac))
```
makes helm happy, and the push happen.

andreas-kupries:
skopeo issue was me having a bogus setup for the `auth` field of the credentials. Had a bogus trailing newline in the encoded value. Causing skopeo to use a bogus password. Fixing that and image upload is happy.

internal/registry/registry.go
```diff
@@ -50,6 +55,156 @@ type ConnectionDetails struct {
...
+	if err != nil {
+		return nil, err
+	}
+
+	registries := []ExportRegistry{}
+	for _, secret := range secretList.Items {
+		url, err := GetRegistryUrlFromSecret(secret)
+		if err != nil {
+			// log the issue, and otherwise ignore the secret
+			log.Error(err, "skipping secret `"+secret.Name+"`")
```
enrichman, line 97: NITP

```suggestion
			log.Error(err, fmt.Sprintf("skipping secret '%s'", secret.Name))
```

internal/registry/registry.go
```diff
@@ -50,6 +55,156 @@ type ConnectionDetails struct {
...
+		})
+	}
+
+	return registries, nil
+}
+
+func GetRegistryUrlFromSecret(secret v1.Secret) (string, error) {
+	namespace, ok := secret.ObjectMeta.Annotations[RegistrySecretNamespaceAnnotationKey]
+	if !ok {
+		return "", errors.New("missing annotation `" + RegistrySecretNamespaceAnnotationKey + "`")
```
enrichman, line 113: NITP

```suggestion
		return "", fmt.Errorf("missing annotation '%s'", RegistrySecretNamespaceAnnotationKey)
```

internal/registry/registry.go
```diff
@@ -50,6 +55,156 @@ type ConnectionDetails struct {
...
+	// Cannot be reached
+	return "", errors.New("registry.go / getRegistryUrlFromSecret - cannot happen")
+}
+
+func GetRegistryCredentialsFromSecret(secret v1.Secret) (RegistryCredentials, error) {
+	empty := RegistryCredentials{}
+
+	namespace, ok := secret.ObjectMeta.Annotations[RegistrySecretNamespaceAnnotationKey]
+	if !ok {
+		return empty, errors.New("missing annotation `" + RegistrySecretNamespaceAnnotationKey + "`")
```
enrichman, line 161: NITP
```suggestion
		return empty, fmt.Errorf("missing annotation '%s'", RegistrySecretNamespaceAnnotationKey)
```

enrichman (commented):
A couple of comments

internal/cli/apps.go
```diff
@@ -143,14 +152,59 @@ var CmdAppShow = &cobra.Command{
 
 // CmdAppExport implements the command: epinio apps export
 var CmdAppExport = &cobra.Command{
-	Use:               "export NAME DIRECTORY",
-	Short:             "Export the named application into the directory",
-	Args:              cobra.ExactArgs(2),
+	Use:               "export [FLAGS] NAME [DIRECTORY]",
```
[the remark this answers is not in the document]
enrichman: See: https://github.com/epinio/docs/pull/304#discussion_r1324478106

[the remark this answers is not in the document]
andreas-kupries:
```suggestion
	Use:               "export NAME [DIRECTORY]",
```

internal/cli/apps.go
```diff
@@ -143,14 +152,59 @@ var CmdAppShow = &cobra.Command{
...
+	Short:             "Export the named application into the directory or flag-specified registry",
+	Args:              cobra.MaximumNArgs(2),
```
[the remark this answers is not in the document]
andreas-kupries: I would keep things separate, i.e. fire two commands. I actually see no case where you truly need both together.

enrichman:
LGTM but I'm not sure why we are using the `.dockerconfigjson` file format to store the secret. This format is usually needed from the Docker daemon to authenticate and pull images.

https://kubernetes.io/docs/reference/generated/kubectl/kubectl-commands#-em-secret-docker-registry-em-

> Dockercfg secrets are used to authenticate against Docker registries.
> [...]
> When creating applications, you may have a Docker registry that requires authentication. **In order for the nodes to pull images on your behalf**, they must have the credentials. You can provide this information by creating a dockercfg secret and attaching it to your service account.

If we are just using it to login and push to OCI registries we could opt for something simpler, and more flexible.  
If we need to provide more options (i.e. a custom certificate, skipSSL..) we are not able to do so with the `type: kubernetes.io/dockerconfigjson` Secret.  
And it's also hard to handle and read since it's a JSON.

In PR https://github.com/epinio/epinio/pull/2530 I've stored the OCI registry credentials for the Services in a plain Secret, with username and password. We could use exactly the same approach, just adding the URL. The same secret could be used to store also the certificates and other information.

This will also almost fix the https://github.com/epinio/epinio/issues/2555.

---

What I would like to do is moving the Secret Type from the `kubernetes.io/dockerconfigjson` to `Opaque`, using the `username`, `password` and the `url` fields.

In my case I was missing the label because the secret was explicitly set into the service, but we can use a more generic `epinio.io/api-oci-registry: "true"`` to fetch these (since we are going maybe to use them also for the services, and other stuff in the future.

enrichman:
I see. I think we could still improve this. I would maybe not doing this in this PR, and because we are in a short time for the release. It seems that skopeo supports a way to use custom certs (--dest-cert-dir ), and it also has a --dest-username and --dest-password that can be used to login, instead of the config file. But that's seems more an improvement.

I would try just to use an Opaque secret with the same key and format.  
This will help us in the future, extending the secret if we need to. I would give it a try, since it's just a matter of changing the type.

andreas-kupries:
Out of scope(o) for this PR, agreed.
Using more direct options to specify credentials and certs - I would like to say this is lesser security, but cannot really. Even when storing things in the secret the data can be dug out with kube commands, if you have the permissions.
Opaque working - I suspect yes.

andreas-kupries:
> I would try just to use an Opaque secret with the same key and format.

The server code does not check the type of the kube secret. Only that it has the expected keys, and the format of their values (i.e. for the `.dockerconfigjson` key that it is a json object of the proper structure/keys).

The skopeo job does not care either, as it only sees the `.dockerconfigjson` key mapped into its filesystem, i.e. a plain file.

A quick test has confirmed this (Using type `kubernetes.io/opaque`).

I will do one small change to the code still, now.
An `OK` message to report successful operation in the terminal.
Just noted during the test that this is missing, and all our other commands have that.

enrichman (changes requested):
I was trying the new functionality and I've seen that I go a `panic` while trying to export twice the image:

```
-> % epinio app export sample --registry myexportregistry

🚢  Export application to registry
Namespace: workspace
Application: sample
Target Registry: myexportregistry

✔️  Ok
enrico@ [11:42:01 AM] [~/Development/SUSE/epinio/epinio] [1770-registry-export *]
-> % epinio app export sample --registry myexportregistry

🚢  Export application to registry
Namespace: workspace
Application: sample
Target Registry: myexportregistry
2023/09/15 11:42:32 EpinioApiClient: "msg"="decoding json error" "error"="unexpected end of JSON input"

❌  error exporting app: parsing error response: unexpected end of JSON input
```

```

2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "destination"="ghcr.io/enrichman" "certs@"=""
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export transport" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "cert file"=""
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export local chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="oci-export-1694770935000419104-chart.tar.gz"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export local image" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="oci-export-1694770935000419104-image.tar"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export remote image" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="ghcr.io/enrichman/workspace-sample-image:58eb9fc6afbe90a8"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export fetch chart archive" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="oci-export-1694770935000419104-chart.tar.gz"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="input" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart archive"="https://github.com/epinio/helm-charts/releases/download/epinio-application-0.1.26/epinio-application-0.1.26.tgz"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="input" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart archive"="retrieving by http"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export fetch values" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249"
2023/09/15 09:42:15 EpinioServer: "level"=1 "msg"="OCI export fetch image" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="oci-export-1694770935000419104-image.tar" "origin"="docker://registry.epinio.svc.cluster.local:5000/apps/workspace-sample:58eb9fc6afbe90a8"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="delete job, done" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "job"="image-export-job-works-a1266fba730acea3b2abd6a9f0a91aa723bc0fab"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export load chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart"="/image-export/oci-export-1694770935000419104-chart.tar.gz"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export chart meta" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart"={"name":"epinio-application","home":"https://github.com/epinio/epinio","sources":["https://github.com/epinio/helm-charts"],"version":"0.1.26","description":"The helm chart for epinio applications to be deployed by","keywords":["epinio","paas"],"maintainers":[{"name":"SUSE","email":"team@epinio.io"}],"icon":"https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png","apiVersion":"v2","annotations":{"artifacthub.io/license":"Apache-2.0"},"type":"application"}
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export chart values" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart"={"userConfig":{},"affinity":{},"epinio":{"username":"user","appName":"placeholder","configurations":null,"env":null,"imageURL":null,"ingress":null,"start":null,"tlsIssuer":null,"replicaCount":1,"routes":null,"stageID":999},"image":{"pullPolicy":"IfNotPresent"},"imagePullSecrets":[],"nodeSelector":{},"resources":{},"tolerations":[]}
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export rewrite chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "name"="workspace-sample-chart" "version"="0"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export rewrite chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "values"={"epinio":{"env":[],"replicaCount":1,"start":null,"appName":"sample","configurations":[],"ingress":null,"routes":[{"domain":"sample.172.19.0.3.omg.howdoi.website","id":"sample.172.19.0.3.omg.howdoi.website","path":"/"}],"stageID":"58eb9fc6afbe90a8","tlsIssuer":"epinio-ca","username":"admin","configpaths":[],"imageURL":"ghcr.io/enrichman/workspace-sample-image:58eb9fc6afbe90a8"}}
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export rewritten chart meta" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart"={"name":"workspace-sample-chart","home":"https://github.com/epinio/epinio","sources":["https://github.com/epinio/helm-charts"],"version":"0","description":"The helm chart for epinio applications to be deployed by","keywords":["epinio","paas"],"maintainers":[{"name":"SUSE","email":"team@epinio.io"}],"icon":"https://raw.githubusercontent.com/epinio/helm-charts/main/assets/epinio.png","apiVersion":"v2","annotations":{"artifacthub.io/license":"Apache-2.0"},"type":"application"}
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export rewritten chart values" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "YAML"="affinity: {}\nepinio:\n  appName: sample\n  configpaths: []\n  configurations: []\n  env: []\n  imageURL: ghcr.io/enrichman/workspace-sample-image:58eb9fc6afbe90a8\n  replicaCount: 1\n  routes:\n  - domain: sample.172.19.0.3.omg.howdoi.website\n    id: sample.172.19.0.3.omg.howdoi.website\n    path: /\n  stageID: 58eb9fc6afbe90a8\n  tlsIssuer: epinio-ca\n  username: admin\nimage:\n  pullPolicy: IfNotPresent\nimagePullSecrets: []\nnodeSelector: {}\nresources: {}\ntolerations: []\nuserConfig: {}\n"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export validate chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export save chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "tmp"="/tmp/oci-export1266142990"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export rewritten chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "chart"="/tmp/oci-export1266142990/workspace-sample-chart-0.tgz"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export destination" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "url scheme"="oci"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export destination" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "url host"="ghcr.io"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export destination" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "url path"="/enrichman"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export login" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "host'"="ghcr.io"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export cleanup local chart repack dir" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="/tmp/oci-export1266142990"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export cleanup local image" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="/image-export/oci-export-1694770935000419104-image.tar"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="OCI export cleanup local chart" "requestId"="151fbad6-32f3-4426-b8d2-a3f8997a7249" "path"="/image-export/oci-export-1694770935000419104-chart.tar.gz"
2023/09/15 09:42:32 EpinioServer: "msg"="[Recovery from panic]" "error"="runtime error: invalid memory address or nil pointer dereference" "time"="2023-09-15T09:42:32Z" "request"="POST /api/v1/namespaces/workspace/applications/sample/export HTTP/1.1\r\nHost: epinio.172.19.0.3.omg.howdoi.website\r\nAccept-Encoding: gzip\r\nAuthorization: Basic YWRtaW46cGFzc3dvcmQ=\r\nContent-Length: 34\r\nUser-Agent: Go-http-client/1.1\r\nX-Forwarded-For: 10.42.0.1\r\nX-Forwarded-Host: epinio.172.19.0.3.omg.howdoi.website\r\nX-Forwarded-Port: 443\r\nX-Forwarded-Proto: https\r\nX-Forwarded-Server: traefik-bb69b68cd-vt4jt\r\nX-Real-Ip: 10.42.0.1\r\n\r\n" "stack"="goroutine 27 [running]:\nruntime/debug.Stack()\n\t/home/enrico/.asdf/installs/golang/1.20.3/go/src/runtime/debug/stack.go:24 +0x65\ngithub.com/alron/ginlogr.RecoveryWithLogr.func1.1()\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/alron/ginlogr@v0.0.4/logr.go:117 +0x534\npanic({0x24104a0, 0x4100ba0})\n\t/home/enrico/.asdf/installs/golang/1.20.3/go/src/runtime/panic.go:884 +0x213\nhelm.sh/helm/v3/pkg/registry.(*Client).Login(0x0, {0xc00110a3ae, 0x7}, {0xc001244d30, 0x3, 0xc000e73180?})\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/helm.sh/helm/v3@v3.12.2/pkg/registry/client.go:201 +0xa4\nhelm.sh/helm/v3/pkg/action.(*RegistryLogin).Run(0xc0006aa940, {0x926a04?, 0x7f573727a3c8?}, {0xc00110a3ae, 0x7}, {0xc000fcaf30, 0x9}, {0xc00107a060, 0x28}, {0xc001244e98, ...})\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/helm.sh/helm/v3@v3.12.2/pkg/action/registry_login.go:83 +0x1f9\ngithub.com/epinio/epinio/internal/helm.(*SynchronizedClient).RegistryLogin(0x278f833?, {0xc00110a3ae, 0x7}, {0xc000fcaf30, 0x9}, {0xc00107a060, 0x28}, {0xc001244e98, 0x0, 0x0})\n\t/home/enrico/Development/SUSE/epinio/epinio/internal/helm/client.go:157 +0x10e\ngithub.com/epinio/epinio/internal/api/v1/application.ExportToRegistry(0xc000618200)\n\t/home/enrico/Development/SUSE/epinio/epinio/internal/api/v1/application/export.go:229 +0x16fe\ngithub.com/epinio/epinio/internal/api/v1.errorHandler.func1(0xc000618200)\n\t/home/enrico/Development/SUSE/epinio/epinio/internal/api/v1/router.go:54 +0x3c\ngithub.com/gin-gonic/gin.(*Context).Next(...)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/context.go:173\ngithub.com/alron/ginlogr.RecoveryWithLogr.func1(0xc000618200)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/alron/ginlogr@v0.0.4/logr.go:129 +0xc2\ngithub.com/gin-gonic/gin.(*Context).Next(...)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/context.go:173\ngithub.com/alron/ginlogr.Ginlogr.func1(0xc000618200)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/alron/ginlogr@v0.0.4/logr.go:34 +0xe7\ngithub.com/gin-gonic/gin.(*Context).Next(...)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/context.go:173\ngithub.com/gin-gonic/gin.CustomRecoveryWithWriter.func1(0xc000618200)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/recovery.go:101 +0x82\ngithub.com/gin-gonic/gin.(*Context).Next(...)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/context.go:173\ngithub.com/gin-gonic/gin.(*Engine).handleHTTPRequest(0xc00050bd40, 0xc000618200)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/gin.go:616 +0x66b\ngithub.com/gin-gonic/gin.(*Engine).ServeHTTP(0xc00050bd40, {0x2c95930?, 0xc00052a0e0}, 0xc000618b00)\n\t/home/enrico/.asdf/installs/golang/1.20.3/packages/pkg/mod/github.com/gin-gonic/gin@v1.8.2/gin.go:572 +0x1dd\nnet/http.serverHandler.ServeHTTP({0x2c84d28?}, {0x2c95930, 0xc00052a0e0}, 0xc000618b00)\n\t/home/enrico/.asdf/installs/golang/1.20.3/go/src/net/http/server.go:2936 +0x316\nnet/http.(*conn).serve(0xc000b1c000, {0x2c980a0, 0xc0001360f0})\n\t/home/enrico/.asdf/installs/golang/1.20.3/go/src/net/http/server.go:1995 +0x612\ncreated by net/http.(*Server).Serve\n\t/home/enrico/.asdf/installs/golang/1.20.3/go/src/net/http/server.go:3089 +0x5ed\n"
2023/09/15 09:42:32 EpinioServer: "level"=1 "msg"="/api/v1/namespaces/workspace/applications/sample/export" "status"=500 "method"="POST" "path"="/api/v1/namespaces/workspace/applications/sample/export" "query"="" "ip"="10.42.0.1" "user-agent"="Go-http-client/1.1" "time"="2023-09-15T09:42:32Z" "latency"="17.307782163s"
```

Also some minor things:
1) I would not use `0` as the default Chart version, maybe a more standard semver such as `0.0.0`.

2) I'm not sure if we should use the `epinio export-registries list` command, instead of just the `epinio export-registries`. We usually have `epinio <resource> <action>` and maybe in the future we would like to handle/manage export-registries via CLI, as we do for other resources.

3) I will comment also a couple of minor things in the doc

andreas-kupries:
Panic happened inside of the helm packages. The message indicates some kind of interruption at the network level, or maybe from the server, causing the code to see incomplete data and breaking on it.

Changed default version. Commit 5ebad2cf454739baa694c6c4359d8848d9b292c0

enrichman:
@andreas-kupries since it was always occuring I had a further investigation adding some logs.

It seems that [this](https://github.com/epinio/epinio/pull/2509#issuecomment-1713580312) is the culprit:

https://github.com/epinio/epinio/blob/5ebad2cf454739baa694c6c4359d8848d9b292c0/internal/helm/client.go#L167

The registryClient was blanked and the next export were going to fail.

I guess that the comment you saw in the library:

```go
	if registry.IsOCI(remote) {
		// Don't use the default registry client if tls options are set.
		c.Options = append(c.Options, pusher.WithRegistryClient(p.cfg.RegistryClient))
	}
```

was just to say that in case of a push the `RegistryClient` set in the configuration is being used, and not that you need to do anything actually.

I removed the line and it seems to work, so I guess the problem you had were unrelated and fixed bu something else.

I would remove that line.

enrichman:
Actually not, I'm using the `ghcr.io` registry, so it's a valid one.

The export will break the second time because I think it will be cleared only after a `Push`.
So if the steps are `login` and then `push`, the second time will break during the second login.

I'll check with Tomas! Thanks.

andreas-kupries:
> The export will break the second time because I think it will be cleared only after a Push.

It is cleared for the Push, correct. In a copy of the ActionConfig. IOW this should not be visible to the general server, just this action. And the helm packages then create their own RC in this field.

`ghcr.io` likely does not need additional certs.

What credentials will I need to talk to ghcr.io ?

enrichman:
You can create a PAT and use it as a password (with package:write/read scopes). You can push to `ghcr.io/andreas-kupries` then.

andreas-kupries:
Fix pushed. The `ActionConfig` in the client is a pointer, and I missed that.
That let the blanking escape the confines of the `Push` operation.
Derefing now to ensure a separate copy of the structure.

andreas-kupries:
> You can create a PAT and ...

:bowing_man: Noted down.

enrichman (dismissed):

internal/api/v1/application/export.go
```diff
@@ -201,7 +201,16 @@ func ExportToRegistry(c *gin.Context) apierror.APIErrors {
...
 	if err != nil {
-		return apierror.InternalError(err)
+		if !strings.Contains(destination.URL, "://") {
+			// Run again with a scheme - This ensures that host and path are properly separated.
+			// Without a scheme the parser considers everything to be the path.
+			destinationURL, err = url.Parse("oci://" + destination.URL)
+			if err != nil {
+				return apierror.InternalError(err)
+			}
+		} else {
```
enrichman, line 211: This could fail providing an url like `/172-18-0-5.nip.io:5000/ak`: https://play.golang.com/p/5Xd3j37kwDj

The Host will be blank, and also the second parse with the Scheme will just add it but the Host will still be blank.

I would also change the comment adding this issue for future reference on why this was needed (comment unclear): https://github.com/golang/go/issues/18824

enrichman (commented):
A small comment.

I was not able to comment later but the `trim` func should probably be called `trimScheme` (I had to scroll to see what it was trimming), and it's not needed (the `Host` doesn't contain the Scheme`.

```go
destinationHost := trim(destinationURL.Host)
```

To avoid any issue I would do a final check for an empty Host, to prevent strange errors like the one that I puit in the comment. If it's empty then we should just fail.

thehejik (approved):
### Test Report
OK - tested against self-hosted (docker) registry:2 with basic auth on domain signed with self signed TLS cert.
OK - tested against official dockehub
Note: values.yaml are incorporated in the chart.tgz with the values from the deployed application

#### Preparation:
Only for self-hosted registry described
* build and patch epinio-server from this PR + CLI
* prepare TLS keys/certificates and deploy the docker registry according to https://github.com/epinio/epinio/blob/main/docs/howtos/public_external_registry.md (docker host uses the same "local" network as k8s node)
* prepare the TLS cert as secret:
```
$ kubectl create secret generic -n epinio export-registry --from-file=tls.crt=tls-registry.pem
```
* prepare the secret for registry:
```
$ kubectl apply -f - <<EOF
---
apiVersion: v1
kind: Secret
type: Opaque
metadata:
  annotations:
    epinio.io/registry-namespace: "thehejik"
  labels:
    epinio.io/api-export-registry: "true"
  name: "export-registry-docker"
  namespace: epinio
stringData:
  certs: "export-registry"
  .dockerconfigjson: |-
    {
      "auths": {
        "1.2.3.4.nip.io": {
          "auth":"b3BlbnN1c2U6cGFzc3dvcmQ=",
          "username":"opensuse",
          "password":"password"
        }
      }
    }
EOF
```

#### Export test
* deploy an app
* export:
```
epinio app export sample-export-oci \
  -r export-registry-docker --image-tag=latest --image-name=sample-php  \
  --chart-name=sample-app-chart --chart-version=1.0.0

🚢  Export application to registry
Namespace: workspace
Application: sample-export-oci
Target Registry: export-registry-docker

✔️  Ok
```
(it can be exported also without custom tags/versions/names - tag will be `stagingID` from `epinio app show <appname>`, chart version will be `0.0.0`)

##### Image validation
```
# docker has to know the CA.pem so do on docker host from where you pull something like `cp CA.pem /etc/pki/trust/anchors/; update-ca-certificates`
$ docker login -u opensuse -p password 1.2.3.4.nip.io
$ docker pull 1.2.3.4.nip.io/thehejik/sample-php:latest
-> OK
```
* when only mandatory flags are used the image can by pulled by:
```
$ epinio app export sample -r export-registry-docker
$ docker pull 1.2.3.4.nip.io/thehejik/workspace-sample-image:00d22dd73fe4c78c # stageID
```
##### Chart validation
```
$ helm registry login 1.2.3.4.nip.io
$ helm pull oci://1.2.3.4.nip.io/thehejik/sample-app-chart --version 1.0.0
Pulled ...
Digest ...
```
* when only the mandatory flags are used the chart can be pulled by:
```
$ epinio app export sample -r export-registry-docker
$ helm pull oci://1.2.3.4.nip.io/thehejik/workspace-sample-chart --version 0.0.0
```

Verify content of pulled chart file and esp values.yaml:
* the `imageURL` points to the registry used for export
* the `env` and `userConfig` settings are set in values accordingly to the deployed app:
```yaml
affinity: {}
epinio:
  appName: sample
  configpaths: []
  configurations: []
  env:
  - name: BAR
    value: string
  - name: FOO
    value: "1"
  imageURL: 1.2.3.4.nip.io/thehejik/workspace-sample-image:00d22dd73fe4c78c
  replicaCount: 1
  routes:
  - domain: sample.1.2.3.10.nip.io
    id: sample.1.2.3.10.nip.io
    path: /
  stageID: 00d22dd73fe4c78c
  tlsIssuer: epinio-ca
  username: admin
image:
  pullPolicy: IfNotPresent
imagePullSecrets: []
nodeSelector: {}
resources: {}
tolerations: []
userConfig:
  appListeningPort: 8080
  myAnnotationEnum: second
  myAnnotationInteger: 1024
  myAnnotationNumber: 1.4
  myAnnotationString: astring
```

internal/api/v1/application/export.go
andreas-kupries: Done with commit 17a8728948067881ba3fc7b593727e13b544b869.
Placed into separate function, and reworked for clarity.

enrichman (approved):

thehejik (approved):
The `app export` retested by using binary built from the latest commit, all good, re-approving

thehejik:
Verified in `v1.10.0-rc2` according to the testcase above

Tested against self-hosted docker registry:2 with self-signed cert. For the export itself only the mandatory flag `-r` was used.

* **OK** released epinio CLI : new subcommand `epinio export-registries`; new flag `app export [..] -r` for exporting to registries verified
* **OK** pulled exported image: `docker pull 1.2.3.4.nip.io/thehejik-rc2/workspace-sample-export-image:a8b1b20b72b63a4e`
* **OK** pulled helm chart: `helm pull oci://1.2.3.4.nip.io/thehejik-rc2/workspace-sample-export-chart --version 0.0.0`

PR #2513, opened 2023-08-17
Show service credentials in detail view

### LINKED ISSUE #2328: Allow the user to get service credentials
Since we are developing: https://github.com/epinio/epinio/issues/2037 that is: exposing services created with epinio to the user; we are required also to equip the user with the credentials needed to access the exposed services.
Currently, there is no such command that the user can invoke to get the credentials for an instance of a service.
The only way the user can obtain the credentials is to search into some kubernetes secretes; since epinio is designed to provide a layer to hide kubernetes concepts, the current available method seems to be logically unacceptable. 

I would like, as user, a command such as:

`epinio service show-credentials mysql`

andreas-kupries:
Fix #2328

### DIFF
--- a/acceptance/services_test.go
+++ b/acceptance/services_test.go
@@ -95,8 +95,8 @@ var _ = Describe("Services", LService, func() {
 		})
 
 		When("Adding a catalog entry", func() {
-			// Already added an nginx catalog service in the top level before block
-			// It's meant to be used by other tests too to make tests faster, because
+			// Note: Already added an nginx catalog service in the top level before the block.
+			// It is meant to be used by other tests too, to make tests faster, because
 			// mysql takes more time to provision.
 
 			It("lists the extended catalog", func() {
@@ -150,7 +150,35 @@ var _ = Describe("Services", LService, func() {
 		})
 	})
 
-	Describe("Show", func() {
+	Describe("Show with credentials", func() {
+		var service string
+
+		BeforeEach(func() {
+			service = catalog.NewServiceName()
+
+			By("create it")
+			out, err := env.Epinio("", "service", "create", "mysql-dev", service, "--wait")
+			Expect(err).ToNot(HaveOccurred(), out)
+		})
+
+		It("shows a service which has credentials", func() {
+			By("show it")
+			out, err := env.Epinio("", "service", "show", service)
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("Showing Service"))
+			Expect(out).To(ContainSubstring("No settings"))
+			Expect(out).To(ContainSubstring("Credentials:"))
+			Expect(out).To(
+				HaveATable(
+					WithHeaders("KEY", "VALUE"),
+					WithRow("mysql-password", ".*"),
+					WithRow("mysql-root-password", ".*"),
+				),
+			)
+		})
+	})
+
+	Describe("Show without credentials", func() {
 		var service string
 
 		BeforeEach(func() {
@@ -161,11 +189,13 @@ var _ = Describe("Services", LService, func() {
 			Expect(err).ToNot(HaveOccurred(), out)
 		})
 
-		It("shows a service", func() {
+		It("shows a service which has no credentials", func() {
 			By("show it")
 			out, err := env.Epinio("", "service", "show", service)
 			Expect(err).ToNot(HaveOccurred(), out)
 			Expect(out).To(ContainSubstring("Showing Service"))
+			Expect(out).To(ContainSubstring("No settings"))
+			Expect(out).To(ContainSubstring("No credentials"))
 			Expect(out).To(
 				HaveATable(
 					WithHeaders("KEY", "VALUE"),
@@ -176,7 +206,6 @@ var _ = Describe("Services", LService, func() {
 					WithRow("Internal Routes", fmt.Sprintf(`.*\.%s\.svc\.cluster\.local`, namespace)),
 				),
 			)
-			Expect(out).To(ContainSubstring("No settings"))
 		})
 
 		Context("customized", func() {
--- a/internal/api/v1/service/show.go
+++ b/internal/api/v1/service/show.go
@@ -12,9 +12,12 @@
 package service
 
 import (
+	"fmt"
+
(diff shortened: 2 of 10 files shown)


acceptance/services_test.go
```diff
@@ -150,7 +150,35 @@ var _ = Describe("Services", LService, func() {
...
+		It("shows a service which has credentials", func() {
+			By("show it")
+			out, err := env.Epinio("", "service", "show", service)
+			Expect(err).ToNot(HaveOccurred(), out)
+			Expect(out).To(ContainSubstring("Showing Service"))
+			Expect(out).To(ContainSubstring("No settings"))
+			Expect(out).To(ContainSubstring("Credentials:"))
+			Expect(out).To(
+				HaveATable(
+					WithHeaders("KEY", "VALUE"),
```
giubacc, line 173: I expect here to see also the username, not only the password; shouldn't the username key/value checked here as well?

internal/cli/usercmd/configuration.go
```diff
@@ -500,7 +500,7 @@ func transformForDisplay(v string) string {
 	// Consider: Count and truncate by runes, not bytes.
 	// - https://pkg.go.dev/unicode/utf8@go1.20.5#RuneCountInString
 	// - https://go.dev/blog/strings (Libraries, foreach/range)
-	limit := 30
+	limit := 70
```
giubacc, line 503: what is the meaning of this limit? oveflowing this value is something tested (or that makes sense to test) ?

internal/api/v1/service/show.go
```diff
@@ -49,9 +52,32 @@ func Show(c *gin.Context) apierror.APIErrors {
 		return apierror.InternalError(err)
 	}
 
+	serviceConfigurations, err := configurations.ForService(ctx, cluster, service)
+	if err != nil {
+		return apierror.InternalError(err)
+	}
+
+	if len(serviceConfigurations) > 0 {
```
giubacc, line 60: hard to grok these bits; can you please explain what are we doing here?

acceptance/services_test.go
andreas-kupries: The change shows all the keys and their values in all the associated K secrets of the service.
The keys in the mysql service secrets do not provide any kind of user information.
I suspect that the helm chart (a bitnami one) uses a fixed username.

internal/cli/usercmd/configuration.go
andreas-kupries: Values larger than that are shown truncated (*). See https://github.com/epinio/epinio/pull/2513/files#diff-db92da3c647e894be7caec46cf94133def888e5c77dbc5a1109a4b9ec6b1614f for an affected test (configurations).

(*) They are likely multi-line as well, something the table formatting will have trouble with.

internal/api/v1/service/show.go
andreas-kupries: I guess you are looking at the inner for loop with the `j` increments ?
Will add code comments.

andreas-kupries: Done with commit b82a5701e8caed448d3e43df07cbe2e35943e01c

giubacc:
Generally lgtm; The codecov is not so good, there are paths not covered in the patch, probably erroneous scenarios. 
If this is not an issue, happy to approve.

andreas-kupries:
I will check codecov and see if I can cover more by tests. Not sure yet. In general it might be error paths which we do/can not trigger.

andreas-kupries:
> I will check codecov and see if I can cover more by tests. Not sure yet. In general it might be error paths which we do/can not trigger.

@giubacc 

Having looked things over the majority of coverage issues are indeed error paths we are simply not triggering in our tests.
And triggering looks to be difficult, likely requiring mocks and such. The one non-error case is the handling of conflicting keys. For that we need a service chart which actually has such, and I do not know of any. (The code was written because I have no proof there is no such chart).

So, no tests for any of these, and while not nice, IMHO good enough.

Please approve.

giubacc (approved):

thehejik:
Verified in `v1.10.0-rc1`, in `v1.9.0` the credentials were not displayed.

```
$ epinio-1.10 service show mysql-sample

🚢  Showing Service...

✔️  Details:
|       KEY       |                                     VALUE                                      |
|-----------------|--------------------------------------------------------------------------------|
| Name            | mysql-sample                                                                   |
| Created         | 2023-09-08 10:16:30 +0200 CEST                                                 |
| Catalog Service | mysql-dev                                                                      |
| Version         | 8.0.31                                                                         |
| Status          | deployed                                                                       |
| Used-By         | sample-app                                                                     |
| Internal Routes | xc11fadea2478b7bb684678d6d8a8-mysql-headless.workspace.svc.cluster.local:3306, |
|                 | xc11fadea2478b7bb684678d6d8a8-mysql.workspace.svc.cluster.local:3306           |

⚠️  No settings

✔️  Credentials:
|         KEY         |   VALUE    |
|---------------------|------------|
| mysql-password      | JblvTfOSnd |
| mysql-root-password | MObIy5IsCD |
```

mmartin24:
rechecked in `1.10.0` and ok:

```

🚢  Showing Service...

✔️  Details:
|       KEY       |                                     VALUE                                      |
|-----------------|--------------------------------------------------------------------------------|
| Name            | svc-mysql                                                                      |
| Created         | 2023-09-25 13:24:59 +0200 CEST                                                 |
| Catalog Service | mysql-dev                                                                      |
| Version         | 8.0.31                                                                         |
| Status          | deployed                                                                       |
| Used-By         | sampleapp2                                                                     |
| Internal Routes | xb097c63b2fb47a525229d1fc3003-mysql-headless.workspace.svc.cluster.local:3306, |
|                 | xb097c63b2fb47a525229d1fc3003-mysql.workspace.svc.cluster.local:3306           |

⚠️  No settings

✔️  Credentials:
|         KEY         |   VALUE    |
|---------------------|------------|
| mysql-password      | pYuMEYQlab |
| mysql-root-password | BYGtnfHhlA |
➜  Downloads epinio info                  

✔️  Epinio Environment
Platform: k3s
Epinio Server Version: v1.10.0
Epinio Client Version: v1.10.0

```

PR #2508, opened 2023-08-07
fixed missing ability configure size of staging pvc/disk like the cpu/memory resources

andreas-kupries:
fix #2504 
ref https://github.com/epinio/helm-charts/pull/478

PR #2514, opened 2023-08-18
fix disk request handling with a default of last resort.

andreas-kupries:
Ref #2508

While an empty disk request should not happen (default from chart) the user could force it empty, or a mismatched chart not set it.


thehejik (approved):
Tested. Now we can use `epinio/epinio` helm chart where `STAGING_RESOURCE_DISK` EV is not set yet even with dev epinio-server binary.

Without the fix it was failing on `epinio app push`, now it works ok

PR #2521, opened 2023-08-22
cli and api for gitconfig management

### LINKED ISSUE #2489: Add git configuration/tokens endpoint
To add on the UI (or also CLI?) the possibility to add/manage the git configuration/token in https://github.com/epinio/epinio/issues/2488 we need to provide some CRUD APIs to manage the configurations.

A token should be updated/deleted only from the user who created it, or by an admin. Also, it should not be possible to see the token value (sensitive), so we need to clear it in the reponse.

andreas-kupries:
fix #2489


thehejik:
So far I tested against my private github repo with a sample php app. I would like to retest also with private gitlab instance with TLS certificates involved but there is a problem with gitconfigs deletion.

**Problems**
1. User can delete any gitconfig entry which was created by another user BUT he has to know the gitconfig name (the  config is just not listed for him)
2. The CLI is ok and doesn't fail when deleting a non-existing gitconfig entry

**Test env**
Epinio with server and CLI built from this branch
```
✔️  Epinio Environment
Platform: k3s
Kubernetes Version: v1.26.6+k3s1
Epinio Server Version: v1.9.0-42-g9338e169
Epinio Client Version: v1.9.0-42-g9338e169
```

**Preparation**
* create a private github repository using a private namespace like https://github.com/thehejik/sample
* create index.php sample app in the repo
* prepare and store a fine-grained PAT token for the private repo only with `Content` and `Metadata` RO permissions
* verify you can't clone the repo without the PAT - but it should work with `git clone https://github_pat_HASH@github.com/thehejik/sample` command

**Create, list, show and delete the gitconfig**
Note: the token/password is visible only when creating the gitconfig as requested
(but you may get it later by `k get secrets -n epinio agitconfigname -o jsonpath='{.data.password}' | base64 -d`)

<details> 
<summary> --help </summary>

```
./dist/epinio-linux-amd64 gitconfig --help
Manage epinio-controlled git configurations

Usage:
  epinio gitconfig [flags]
  epinio gitconfig [command]

Aliases:
  gitconfig, gitconfigs

Available Commands:
  create      Creates an epinio-controlled git configuration
  delete      Deletes an epinio-controlled git configuration
  list        Lists all epinio-controlled git configurations
  show        Shows the details of an epinio-controlled git configuration

Flags:
  -h, --help   help for gitconfig

Global Flags:
  -H, --header stringArray       Add custom header to every request executed
      --settings-file string     (EPINIO_SETTINGS) set path of settings file (default "/home/thehejik/.config/epinio/settings.yaml")
      --trace-file string        (TRACE_FILE) Print trace messages to the specified file
      --trace-level int          (TRACE_LEVEL) Only print trace messages at or above this level (0 to 255, default 0, print nothing)
      --trace-output string      (TRACE_OUTPUT) Sets trace output format [text,json] (default "text")

Use "epinio gitconfig [command] --help" for more information about a command.
```
</details>
<details> 
<summary> create </summary>

```
./dist/epinio-linux-amd64 gitconfig create github-private https://github.com thehejik github_pat_HASH --user-org thehejik --repository sample --git-provider github

🚢  Creating gitconfig...
Name: github-private
Provider: github
Url: https://github.com
User/Org: thehejik
Repository: sample
Username: thehejik
Password: github_pat_HASH
Skip SSL: false
Certificates from: 

✔️  Git configuration created.
```
</details>
<details> 
<summary> list </summary>

```
./dist/epinio-linux-amd64 gitconfig list

🚢  Listing git configurations

✔️  Git configurations:
|       ID       | PROVIDER |        URL         | USER/ORG | REPOSITORY | SKIP SSL | USERNAME |
|----------------|----------|--------------------|----------|------------|----------|----------|
| github-private | github   | https://github.com | thehejik | sample     | false    | thehejik |
```
</details>
<details> 
<summary> show </summary>

```
./dist/epinio-linux-amd64 gitconfig show github-private

🚢  Showing gitconfig...
Name: github-private

✔️  Details:
|    KEY     |       VALUE        |
|------------|--------------------|
| Name       | github-private     |
| Provider   | github             |
| URL        | https://github.com |
| User/Org   | thehejik           |
| Repository | sample             |
| Skip SSL   | false              |
| Username   | thehejik           |
```
</details>
<details> 
<summary> delete </summary>

```
./dist/epinio-linux-amd64 gitconfig delete github-private

🚢  Deleting git configurations...
Git configurations: github-private
.
✔️  Git configurations deleted.
```
</details>

**Push an app from the private Github repo when the gitconfig exists**
```
./dist/epinio-linux-amd64 app push --name sample --git https://github.com/thehejik/sample/
...
Deploying application ...
......
✔️  App is online.
Name: sample2
Namespace: workspace
Builder Image: 
Routes: 
1: https://sample2.BLAH.nip.io

-> OK
```
**Push an app from the private Github repo when the gitconfig DOES NOT exist**
```
./dist/epinio-linux-amd64 app push --name sample2 --git https://github.com/thehejik/sample/

Manifest: <<Defaults>>
Name: sample2
Source Origin: https://github.com/thehejik/sample/
AppChart: 
Target Namespace: workspace
Routes: <<default>>

Create the application resource ...

Importing the application sources from Git ...

❌  error pushing app to server: importing git remote: authentication required
-> OK
```

andreas-kupries:
@thehejik Thanks for testing.

> User can delete any gitconfig entry which was created by another user BUT he has to know the gitconfig name (the config is just not listed for him)

That is IMHO an important enough bug to block merging. I will fix this.

Will also check if this kind of check is also missing for the namespaced resources, as the code come from them.
Of course, it much more likely that I forgot/missed their code for this instead, leaving this hole.

I suspect that you can also do a `gitconfig show` on an entry from another user if the name is known.

@enrichman How about collapsing the whole series of commits into two commits, one for API, the other for the cli. Note that each individual commit is small enough to be quickly understood. (Just so many of them).

thehejik:
> I suspect that you can also do a `gitconfig show` on an entry from another user if the name is known.

@andreas-kupries Indeed, the `show` details are visible as well.

andreas-kupries:
> @enrichman How about collapsing the whole series of commits into two commits, one for API, the other for the cli. Note that each individual commit is small enough to be quickly understood. (Just so many of them).

Having looked deeper I find it is more intertwined than I remembered.
I did the basic list/view/match/delete API first, then the command separately.
After that I did the `create` API and command, in stages (basic, then certs), with slight crossover to delete.
At last I added the auth tracking and stuff across the board.
Very difficult to separate API and command work.

I suspect that the easiest way would be to collapse the entire history into a single commit, and then use that to create the API and CLI PRs by using only the relevant files. That feels quite a bit effort for not much gain.

enrichman:
@andreas-kupries the problem is that I can even go to one commit one by one, but there is just too much code to review.
I would split it into functionalities, and it will be much easier also to test for @thehejik.

I would do the first PR with just the GET endpoint and the CLI command. Then we can extend adding the creation, then the deletion and so on. Do you think it's feasible?

Ref: https://opensource.com/article/18/6/anatomy-perfect-pull-request#:~:text=How%20can%20we%20determine%20the,250%20lines%20of%20code%20changed.

thehejik:
@andreas-kupries and yes, user can also do `namespace delete` for a namespace created by another user. But `namespace show` resists to show foreign ns details with a message.
```
❌  error showing epinio-controlled namespace: user unauthorized
```

andreas-kupries:
> @andreas-kupries and yes, user can also do `namespace delete` for a namespace created by another user. But `namespace show` resists to show foreign ns details with a message.

IMHO should be fixed too.

> 
> ```
> ❌  error showing epinio-controlled namespace: user unauthorized
> ```

So some parts are protected, good.
Note that apps are namespaced resources, may also have this hole.

thehejik:
@enrichman I don't basically mind about the commits cleanup, when I look into code I do it in `Files Changed` pane anyway.

andreas-kupries:
> and yes, user can also do namespace delete for a namespace created by another user.

Now that I am looking through the code this is a bug which started to appear with the introduction of the batch deletion for namespaces. The `AuthorizationMiddleware` looks for a `namespace` parameter in the request, and since the CLI uses the batch API the parameter is `namespaces[]` instead. This leaves `namespace` as the empty string when queries and this bypasses the check, treating the request as authorized.

This should be fixed as a separate PR.

And for gitconfig I have to add another Middleware. This can even prototype the issue with delete, as it also has single and batch APIs.

andreas-kupries:
@enrichman I had some more thoughts about your desire to rewrite to make the review easier.
I still believe that this is actually not really needed.

If we rewrite the thing into a series of commits doing the changes for a single vertical, i.e. api to command and tests, then I will spend effort (1-2d estimated) on doing something the merge will then undo as it squashes the commits again.

Separating the commits into actual separate PRs would keep them from getting squashed, yet also cause additional time spent in CI and testing each again.

On the other side, going to the `Files` tab of the PR I see a tree of the changed files, organized by directory.
This allows the reviewer to go through the files in a self-selected order which matches their needs, and breaking the changes down into digestible pieces, despite the overall largeness of the PR.

For myself, for example, I would go from the bottom to the top, i.e. start in the models, then the server backend code, the server endpoints, router, and then the cli side, up from server client to logic to cobra, and tests. Looks into the supporting changes outside this either at the end or interspersed.

Your order might be different.

thehejik:
I did a retest on private github repo according to https://github.com/epinio/epinio/pull/2521#issuecomment-1697781967 and the problem with deletion of foreign gitconfig and namespace has been resolved.

* `gitconfig delete` and `show` of foreign gitconfigs has been resolved with error message:
```
❌  error deleting|showing epinio-controlled git configuration: user unauthorized for gitconfig github-private
```
* dtto for `namespace delete` and `show`
```
❌  error deleting|showing epinio-controlled namespace: user unauthorized for namespace bla
```

Beyond this PR:
* User can push an app from private github repo via gitconfig made by another user
* `gitconfig delete non-existing` returns 0 when trying to delete non-exsting gitconfig (contrary to app and namespace)

enrichman (changes requested):
Not yet looked at the code, but one thing that we need to change is the `create` command, so I'm just commenting about this to fix it, while tomorrow I'm going to review the rest.

```
-> % epinio gitconfig create
Usage:
  epinio gitconfig create ID URL USER TOKEN [flags]
```

Now the `epinio gitconfig create` accepts 4 args but the only required argument is the URL, as per doc: https://docs.epinio.io/next/references/git_configuration.
We don't need to specify the USER or the TOKEN, because those are optional parameters. A user could use the gitconfig just to specify the skipSSL, or the certificate for example. I would use flags for all the optional arguments (so args for only the ID and URL).  

I would also try to keep in sync with the name used in the secret. They are called username/password because the authentication used is going to be a `Basic` auth.
https://github.com/epinio/epinio/blob/3863a0d523fd65bf8b7fd0c3a041d0c3fbd392eb/internal/api/v1/application/importgit.go#L290-L293  

Maybe in the future we will need to use a different authentication with an actual token, so I would reserve the `TOKEN` keyword for it.

acceptance/helpers/machine/gitconfig.go
enrichman, line 1: Helpers func should have the `GingkoHelper()` func, to be aware of the actual func who called the helper. This is done to avoid to use the `ExpectWithOffset`. See https://onsi.github.io/ginkgo/#mental-model-how-ginkgo-handles-failure

https://github.com/epinio/epinio/blob/3863a0d523fd65bf8b7fd0c3a041d0c3fbd392eb/acceptance/helpers/machine/services.go#L37-L39

andreas-kupries:
> We don't need to specify the USER or the TOKEN, because those are optional parameters. A user could use the gitconfig just to specify the skipSSL, or the certificate for example. I would use flags for all the optional arguments (so args for only the ID and URL).

Ok. I will change that part to flags.

acceptance/helpers/machine/gitconfig.go
andreas-kupries: Thanks. While I had seen this call turning up in code the reason was unclear.

andreas-kupries: Done with commit 9ddb0808336b5ddab0175f6bf8d4e34566a29947

andreas-kupries:
> Ok. I will change that part to flags.

Done with commit a0760de3d378199954882d18423907d6b23ee102

internal/api/v1/auth.go
```diff
@@ -17,66 +17,96 @@ import (
...
+		logger.V(1).WithName("authorizeAdmin").Info("user [admin] is authorized")
+		return
+
 	case "user":
-		authorized = authorizeUser(logger, user, path, namespace)
+		if !unrestrictedPath(logger, path) {
+			response.Error(c, apierrors.NewAPIError("user unauthorized, path restricted",
+				http.StatusForbidden))
+			c.Abort()
+		}
```
enrichman, line 55: The negation of a negative verb is a bit redundant/confusing. Would you agree on change this to:

```go
if restrictedPath(logger, path) {
    // abort
}
```
or
```go
if !allowedPath(logger, path) {
    // abort
}
```

Also, I think we should return after the `c.Abort()`, since the middleware failed.

internal/api/v1/gitconfig/create.go
```diff
@@ -0,0 +1,117 @@
...
+	user := requestctx.User(ctx)
+
+	authService, err := auth.NewAuthServiceFromContext(ctx)
+	if err != nil {
+		return errors.Wrap(err, "error creating auth service")
+	}
+
+	err = authService.AddGitconfigToUser(ctx, user.Username, gitconfig)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("error adding gitconfig [%s] to user [%s]",
```
enrichman, line 113: NITP: the `Wrapf` can be used to format the error message.

internal/api/v1/auth.go
andreas-kupries: Oops and yes.

internal/api/v1/gitconfig/create.go
andreas-kupries: I can retry. IIRC when I used that at first the linter claimed that `Wrapf` does not exist.

internal/api/v1/namespace/create.go
```diff
@@ -87,7 +87,8 @@ func addNamespaceToUser(ctx context.Context, namespace string) error {
 
 	err = authService.AddNamespaceToUser(ctx, user.Username, namespace)
 	if err != nil {
-		return errors.Wrap(err, fmt.Sprintf("error adding namespace [%s] to user [%s]", namespace, user.Username))
+		return errors.Wrap(err, fmt.Sprintf("error adding namespace [%s] to user [%s]",
```
enrichman, line 90: Same for `Wrapf`

internal/auth/auth.go
```diff
@@ -152,6 +152,45 @@ func (s *AuthService) RemoveNamespaceFromUsers(ctx context.Context, namespace st
 	return nil
 }
 
+// AddGitconfigToUser will add the specified gitconfig to the User
+func (s *AuthService) AddGitconfigToUser(ctx context.Context, username, gitconfig string) error {
+	user, err := s.GetUserByUsername(ctx, username)
+	if err != nil {
+		return errors.Wrap(err, fmt.Sprintf("error getting user [%s] by username", username))
```
enrichman, line 159: Same for `Wrapf`

internal/auth/auth.go
```diff
@@ -152,6 +152,45 @@ func (s *AuthService) RemoveNamespaceFromUsers(ctx context.Context, namespace st
...
+	}
+	user.AddGitconfig(gitconfig)
+
+	err = s.updateUserSecret(ctx, user)
+	return errors.Wrap(err, fmt.Sprintf("error updating user secret [%s]", username))
```
enrichman, line 164: `Same for `Wrapf`

internal/api/v1/auth.go
andreas-kupries: Done with commit 87f505aebd12d1d18aa5eeab13042058f94b9b1a

internal/cli/gitconfigs.go
```diff
@@ -0,0 +1,168 @@
...
+)
+
+// CmdGitconfig implements the command: epinio gitconfig
+var CmdGitconfig = &cobra.Command{
+	Use:           "gitconfig",
+	Aliases:       []string{"gitconfigs"},
+	Short:         "Epinio-controlled git configurations",
+	Long:          `Manage epinio-controlled git configurations`,
+	SilenceErrors: true,
+	SilenceUsage:  true,
```
enrichman, line 33: In other "parent" commands we don't have this flag, so the `help` will be shown if no arg is provided.

See `CmdApp` and `CmdService`:

https://github.com/epinio/epinio/blob/3863a0d523fd65bf8b7fd0c3a041d0c3fbd392eb/internal/cli/apps.go#L22-L29

https://github.com/epinio/epinio/blob/3863a0d523fd65bf8b7fd0c3a041d0c3fbd392eb/internal/cli/services.go#L24-L31

internal/cli/gitconfigs.go
```diff
@@ -0,0 +1,168 @@
...
+}
+
+func init() {
+	flags := CmdGitconfigDelete.Flags()
+	flags.BoolVar(&gGitconfigAllFlag, "all", false, "delete all git configurations")
+
+	gitProviderOption(CmdGitconfigCreate)
+	flags = CmdGitconfigCreate.Flags()
+	flags.Bool("skip-ssl", false, "skip ssl")
+	flags.String("user", "", "user for logging into the host")
```
enrichman, line 50: I'm more for `username`, if you are ok with that. :eyes:

internal/cli/gitconfigs.go
```diff
@@ -0,0 +1,168 @@
...
+
+var (
+	gGitconfigAllFlag bool
+)
+
+// CmdGitconfig implements the command: epinio gitconfig
+var CmdGitconfig = &cobra.Command{
+	Use:           "gitconfig",
+	Aliases:       []string{"gitconfigs"},
+	Short:         "Epinio-controlled git configurations",
```
enrichman, line 30: Not sure what "Epinio-controlled" means. Can we rephrase this?

Something like "manage git configurations", or looking at the other commands maybe "Epinio git configurations management"

```
  app           Epinio application features
  client-sync   Downloads a client binary matching the currently logged server
  completion    Generate completion script for a shell
  configuration Epinio configuration features
  gitconfig     Epinio-controlled git configurations
  help          Help about any command
  info          Shows information about the Epinio environment
  login         Epinio login to the server
  logout        Epinio logout from server
  namespace     Epinio-controlled namespaces
  push          Push an application declared in the specified manifest
  server        Starts the Epinio server.
  service       Epinio service management
  settings      Epinio settings management
  target        Targets an epinio-controlled namespace.
  version       Print the version number
```

Probably we need a coherent way for this around all the commands.

pkg/api/core/v1/client/gitconfigs_test.go
```diff
@@ -0,0 +1,82 @@
...
+		})
+
+		DescribeTable("the APIs are returning an error",
+			func(call func() (any, error)) {
+				_, err := call()
+				Expect(err).To(HaveOccurred())
+			},
+			// Entry("gitconfig create", func() (any, error) {
+			// 	return epinioClient.GitconfigCreate(models.GitconfigCreateRequest{})
+			// }),
```
enrichman, line 67: Should this be uncommented or removed?

enrichman (commented):
Some more small comments

pkg/api/core/v1/client/gitconfigs_test.go
andreas-kupries: Will check, likely uncommented.

internal/cli/gitconfigs.go
andreas-kupries: Ok. Would also not be a prefix of `--user-org`.

internal/cli/gitconfigs.go
andreas-kupries: Done with commit 15297132f4d43459d0151bf8e8ca299b399ddb26

internal/cli/gitconfigs.go
andreas-kupries: Done with commit 76ff8f391e5d213d9c2b615d18e2c5825c8f2a8d

pkg/api/core/v1/client/gitconfigs_test.go
andreas-kupries: Done with commit 2c30eeb625f9c113e55c601881b3312e768550ac

internal/cli/gitconfigs.go
andreas-kupries: Done with commit 8104bea6f0c8f43fd6d74351ee9084f41e1868a2

internal/auth/auth.go
andreas-kupries: Done with commit f6bc2eca5f02b7f0634afe443b9e2ea734faeb05

internal/auth/auth.go
andreas-kupries: Done with commit f6bc2eca5f02b7f0634afe443b9e2ea734faeb05

internal/api/v1/namespace/create.go
andreas-kupries: Done with commit f6bc2eca5f02b7f0634afe443b9e2ea734faeb05

internal/api/v1/gitconfig/create.go
andreas-kupries: Done with commit f6bc2eca5f02b7f0634afe443b9e2ea734faeb05

Works now. No idea anymore what was going at the time.

enrichman:
LGTM! And we need to remember to run the `generate-cli-doc` for the new commands. :+1:

enrichman (approved):

enrichman:
Awesome! Thanks for checking it.

thehejik:
Verified in `v1.10.0-rc1` by using gitlab.suse.de as a source for pushing app.

PR #2522, opened 2023-08-25
fixed missing support for edit of service (custom values, update/replace)

### LINKED ISSUE #2518: It's not possible to edit Service chart values/settings
Related issue: [#252 ](https://github.com/epinio/ui/issues/252)

Service/edit APIs to be provided.

andreas-kupries:
fix #2518


enrichman:
There are some conflicts to resolve. One first thing is that I would change the `--remove` flag to `--unset`, to have the `--set/--unset` pair available.

andreas-kupries:
Will resolve the conflicts.

andreas-kupries:
> One first thing is that I would change the `--remove` flag to `--unset`, to have the `--set/--unset` pair available.

Note that making this change would make this different from the `configuration update` command which uses the same `--remove`/`--set` pair.

andreas-kupries:
> > One first thing is that I would change the `--remove` flag to `--unset`, to have the `--set/--unset` pair available.
> 
> Note that making this change would make this different from the `configuration update` command which uses the same `--remove`/`--set` pair.

Finding https://github.com/spf13/pflag#mutating-or-normalizing-flag-names I will see if I can support the old `--remove` while switching to `--unset` going forward. That way we would have backward compatibility.

enrichman (approved):

PR #2533, opened 2023-09-01
Moved some CLI commands to `cmd` package

enrichman:
Since the `settings update-ca` dropped the "admin" operations we can drop the `admincmd` and `usercmd` definitions.

To do so imo it will be nice to have a `cli/cmd` package with all the commands, and leaving the `EpinioClient` in the parent `cli` package.

```
cli/
 |- cmd/ (all the commands here)
 |- termui/ (the "display" library)
 | client.go (all the misc/common stuff in the parent)
```

Having a look at this I've understood the problem of creating the EpinioClient in the root command, and I've fixed it. Since the loading of the flags is done after the command creation we need to create the EpinioClient before, but then finalize its creation (here with an `Init` func) later in the PersistentPreRunE. In this way we are going to load the right settings specified by the user.

### DIFF
--- a/internal/cli/root.go
+++ b/internal/cli/root.go
@@ -131,6 +131,7 @@ func NewRootCmd() (*cobra.Command, error) {
 
 	rootCmd.AddCommand(
 		CmdCompletion,
+		CmdSettings,
 		cmd.NewInfoCmd(client),
 		cmd.NewClientSyncCmd(client),
 		CmdGitconfig,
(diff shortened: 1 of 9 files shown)


internal/cli/root.go
```diff
@@ -125,9 +131,8 @@ func NewRootCmd() (*cobra.Command, error) {
 
 	rootCmd.AddCommand(
 		CmdCompletion,
-		CmdSettings,
```
andreas-kupries, line 128: Is `CmdSettings` gone ?

andreas-kupries (changes requested):

internal/cli/root.go
enrichman: :man_facepalming: fixed with a281f0e9b1193be5f8969521bcbbd309a094a520

andreas-kupries (approved):

PR #2544, opened 2023-09-01
Moved `settings update-ca` command in cmd package

enrichman:
This PR drops the `admincmd` package moving the settings commands in the `cmd` package, and the CLI implementation in the EpinioClient.


internal/cli/cmd/clientsync.go
```diff
@@ -20,12 +20,11 @@ import (
...
 	return &cobra.Command{
-		Use:   "client-sync",
-		Short: "Downloads a client binary matching the currently logged server",
-		Long:  `Synchronizes the epinio client with the server by downloading the matching binary and replacing the current one.`,
+		Use:          "client-sync",
+		Short:        "Downloads a client binary matching the currently logged server",
+		SilenceUsage: true,
+		Long:         `Synchronizes the epinio client with the server by downloading the matching binary and replacing the current one.`,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cmd.SilenceUsage = true
```
andreas-kupries, line 27: IIRC the location of the silenceUsage makes a difference when it applies.
As changed it applies even to argument processing errors (not enough args, etc.)
Used in the command, it silences the usage only for errors coming from the command itself, and not for argument issues thrown by cobra itself.

As a side note, the command is missing a spec telling that it takes 0 arguments.
I suspect that due to the missing spec argument errors currently cannot happen, as it always any number of arguments which are then silently ignored.

internal/cli/cmd/info.go
```diff
@@ -20,17 +20,15 @@ import (
...
 	return &cobra.Command{
-		Use:   "info",
-		Short: "Shows information about the Epinio environment",
-		Long:  "Shows status and versions for epinio's server-side components.",
+		Use:          "info",
+		Short:        "Shows information about the Epinio environment",
+		Long:         "Shows status and versions for epinio's server-side components.",
+		SilenceUsage: true,
 		RunE: func(cmd *cobra.Command, args []string) error {
-			cmd.SilenceUsage = true
```
andreas-kupries, line 27: See above, `client-sync`. 

Just confirmed that `info` can currently be invoked with any number of arguments, which are ignored.

internal/cli/cmd/settings.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// NewSettingsColorsCmd returns a new 'epinio settings colors' command
+func NewSettingsColorsCmd(client *usercmd.EpinioClient) *cobra.Command {
+	return &cobra.Command{
+		Use:          "colors BOOL",
+		Short:        "Manage colored output",
+		Long:         "Enable/Disable colored output",
+		SilenceUsage: true,
+		Args: func(cmd *cobra.Command, args []string) error {
```
andreas-kupries, line 56: See above wrt `SilenceUsage`

internal/cli/cmd/settings.go
```diff
@@ -0,0 +1,131 @@
...
+}
+
+// NewSettingsShowCmd returns a new 'epinio settings show' command
+func NewSettingsShowCmd(client *usercmd.EpinioClient) *cobra.Command {
+	cfg := &SettingsShowConfig{}
+
+	settingsShowCmd := &cobra.Command{
+		Use:          "show",
+		Short:        "Show the current settings",
+		SilenceUsage: true,
```
andreas-kupries, line 96: See above wrt SilenceUsage

internal/cli/cmd/settings.go
```diff
@@ -0,0 +1,131 @@
...
+	return settingsShowCmd
+}
+
+// NewSettingsUpdateCACmd returns a new 'epinio settings update-ca' command
+func NewSettingsUpdateCACmd(client *usercmd.EpinioClient) *cobra.Command {
+	return &cobra.Command{
+		Use:          "update-ca",
+		Short:        "Update the api location and CA certificate",
+		Long:         "Update the api location and CA certificate from the current cluster",
+		SilenceUsage: true,
```
andreas-kupries, line 121: See above wrt SilenceUsage

andreas-kupries (changes requested):

internal/cli/cmd/clientsync.go
andreas-kupries: Confirmed using `epinio info` that my recollection is correct.

Added the `ExactArgs(0)` to the `info` command, and then ran it with `Silence` in the command and in the structure. Result:

In the command, as before the change in placement was made. :heavy_check_mark: 
```
work@tagetarl:~/SUSE/dev/epinio> ep info 9034290 902185r93401
Usage:
  epinio info [flags]

Flags:
  -h, --help   help for info

Global Flags:
  -H, --header stringArray       Add custom header to every request executed
      --settings-file string     (EPINIO_SETTINGS) set path of settings file (default "/home/work/.config/epinio/settings.yaml")
      --trace-file string        (TRACE_FILE) Print trace messages to the specified file
      --trace-level int          (TRACE_LEVEL) Only print trace messages at or above this level (0 to 255, default 0, print nothing)
      --trace-output string      (TRACE_OUTPUT) Sets trace output format [text,json] (default "text")


❌  accepts 0 arg(s), received 2
```

In the structure, i.e. per the change above. :boom: , not desired.
```
work@tagetarl:~/SUSE/dev/epinio> ep info 9034290 902185r93401

❌  accepts 0 arg(s), received 2
```

enrichman: Ah, you are right, I forgot about this. Thanks. Fixing this.

andreas-kupries (approved):

enrichman:
And now I see what is this error about. Now the `info` is expecting no args, and so tests are failing because they were wrong, and providing an extra arg. Basically they were executed with `epinio info info`.

andreas-kupries (approved):

PR #2550, opened 2023-09-05
Added `gitproxy` endpoint

enrichman:
This PR adds the `/gitproxy` endpoint.

This endpoint will proxy some of the Github and Gitlab APIs. If a gitconfig ID is provided then this will be used to enrich the request (with the token, or skipping the SSL verification, or custom certificates).

### DIFF
--- a/acceptance/api/v1/gitproxy_test.go
+++ b/acceptance/api/v1/gitproxy_test.go
@@ -1,0 +1,64 @@
+// Copyright © 2021 - 2023 SUSE LLC
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//     http://www.apache.org/licenses/LICENSE-2.0
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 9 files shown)


enrichman:
As discussed internally we need to check if there was an error regarding the Gitlab search endpoint:

https://github.com/rancher/dashboard/blob/b9180782e23fb67edb7d53dc9769c60573799434/shell/store/gitlab.js#L67-L71

This seems to point to a `projects/${ repo }/repository/branches/${ branch }` but it should probably be `projects/${ username }%2F${ repo }/repository/branches/${ branch }`.

@torchiaf is on it :+1:

enrichman:
Adding one more insight. I was trying to use the Gitlab `/search/repositories` API but it seems to be unknown:
 - https://gitlab.com/api/v4/search/repositories?q=repo:inkscape/inkscape

```json
{"error":"404 Not Found"}
```

Reading the Gitlab doc it should be just `/search`, but it seems to be an authenticated endpoint.  
- https://gitlab.com/api/v4/search?q=repo:inkscape/inkscape

```json
{"message":"401 Unauthorized"}
```

Maybe that's the reason because of the error? Maybe the search on Gitlab needs a token, and so it was not available?

By the way I'm not sure where the `search` API is used (Gitlab or Github). Any help about this? @richard-cox @torchiaf

andreas-kupries (approved):
LGTM

mmartin24 (approved):

PR #2562, opened 2023-09-12
Added support for --output json to configuration list and show

### LINKED ISSUE #2558: Add `--output json` flag to `epinio configuration` command

andreas-kupries:
Fix #2558 

chore: moved configuration ensemble into cmd package


mmartin24:
Tested on branch.
- Checked basic case deploying a simple configuration and a service config and is capable of outputting in `json` :
```
epinio service create mysql-dev svc-mysql-cli  
epinio configuration create config-1 foo bar  
epinio configuration list --output json | jq 
```
Output:

```json
[
  {
    "meta": {
      "name": "config-1",
      "namespace": "workspace",
      "createdAt": "2023-09-13T07:11:21Z"
    },
    "configuration": {
      "user": "admin",
      "details": {
        "foo": "bar"
      },
      "boundapps": null,
      "type": "custom"
    }
  },
  {
    "meta": {
      "name": "xcfefa578ebf738cc24fe12a0f5a3-mysql",
      "namespace": "workspace",
      "createdAt": "2023-09-13T07:13:57Z"
    },
    "configuration": {
      "user": "",
      "details": {
        "mysql-password": "ogKU7cZNKb",
        "mysql-root-password": "xn15bRd3Cw"
      },
      "boundapps": null,
      "type": "service",
      "origin": "svc-mysql-cli"
    }
  }
]
```
Now I noticed that in the `--output json` option the details with service password is shown, whereas in the `--output text` version is not. @andreas-kupries  would we want to enhance this?
```
➜  epinio git:(2558-configuration-json-output) ✗ epinio configuration list --output text     

🚢  Listing configurations
Namespace: workspace

✔️  Epinio Configurations:
|                NAME                 |            CREATED             |  TYPE   |    ORIGIN     | APPLICATIONS |
|-------------------------------------|--------------------------------|---------|---------------|--------------|
| config-1                            | 2023-09-13 09:11:21 +0200 CEST | custom  |               |              |
| xcfefa578ebf738cc24fe12a0f5a3-mysql | 2023-09-13 09:13:57 +0200 CEST | service | svc-mysql-cli |              |
```

---
- Checked that when targeting to a different namespace where a configuration does not exist the `--output json` displays only a `[]`:
To check run:
```
epinio service create mysql-dev svc-mysql-cli  
epinio configuration create config-1 foo bar  
epinio namespace create ns-1
epinio target ns-1 
epinio target ns-1 

# Now compare the outputs of 
epinio configuration show svc-mysql-cli
epinio configuration list --output json | jq


```

Notice the difference with `--output text`
```
➜  epinio git:(2558-configuration-json-output) ✗ epinio configuration list              

🚢  Listing configurations
Namespace: ns-1

✔️  Epinio Configurations:
| NAME | CREATED | TYPE | ORIGIN | APPLICATIONS |
|------|---------|------|--------|--------------|

```
...vs `--output json`:
```
➜  epinio git:(2558-configuration-json-output) ✗ epinio configuration list --output json | jq
[]
```
@andreas-kupries would we want to add a better message in the case of `--output json`?

andreas-kupries:
> Now I noticed that in the --output json option the details with service password is shown, whereas
> in the --output text version is not. @andreas-kupries would we want to enhance this?

Printing the configuration details in the human-readable list is IMHO too much detail. In JSON this is ok, because that is targeted at script processing where all the information is desired, and we do not have to care about formatting and space in the terminal for it.

> @andreas-kupries would we want to add a better message in the case of --output json?

JSON output is the raw data from the API endpoint, to used directly in scripts and the like.
The `[]` is the correct result, an empty JSON array.
If you are talking about putting a human readable message in that output, that would be against the motivation for json output.

internal/cli/termui/ui.go
```diff
@@ -108,10 +108,11 @@ func (u *UI) DisableJSON() {
 	u.jsonEnabled = false
 }
 
-func (u *UI) JSON(value any) {
+func (u *UI) JSON(value any) bool {
 	if u.jsonEnabled {
 		_ = json.NewEncoder(u.output).Encode(value)
 	}
+	return u.jsonEnabled
 }
```
enrichman, line 116: This makes sense, but maybe it's nicer to separate the two. Since the encoder is also returning an error that we are ignoring it's probably better to do something like this:

```go
func (u *UI) JSON(value any) error {
	if u.jsonEnabled {
		return json.NewEncoder(u.output).Encode(value)
	}
	return nil
}

func (u *UI) JSONEnabled() bool {
	return u.jsonEnabled
}
```

and then

```go
if c.ui.JSONEnabled() {
	// Avoid doing all the formatting for human consumption which will not be shown due to json mode
	return c.ui.JSON(configurations)
}
```

internal/cli/cmd/configurations.go
```diff
@@ -0,0 +1,363 @@
...
+func NewConfigurationListCmd(client ConfigurationService, rootCfg *RootConfig) *cobra.Command {
+	cmd := &cobra.Command{
+		Use:   "list [--all]",
+		Short: "Lists configurations",
+		Long:  "Lists configurations in the targeted namespace, or all",
+		Args:  cobra.NoArgs,
+		RunE: func(cmd *cobra.Command, args []string) error {
+			cmd.SilenceUsage = true
+
+			all, err := cmd.Flags().GetBool("all")
```
enrichman, line 81: Moving the other commands I started using structs instead of the `GetFoo()` func, with the `flag.BoolVar()`, instead of the `flag.Bool()`  
 I find it very verbose with many parameters, and when you have many flags that you need to pass to the underlying function if you have a "config struct" it's very convenient to pass everything and leave the logic to it.

I.e. for the [namespace delete command](https://github.com/epinio/epinio/blob/0d7c91219b59d20ed70645f192787f63e93ab566/internal/cli/cmd/namespaces.go#L97-L130) 

This is the Configuration struct:

https://github.com/epinio/epinio/blob/0d7c91219b59d20ed70645f192787f63e93ab566/internal/cli/cmd/namespaces.go#L97-L100

The flags are bound to the fields:

https://github.com/epinio/epinio/blob/0d7c91219b59d20ed70645f192787f63e93ab566/internal/cli/cmd/namespaces.go#L126-L127

and then you can just use them:

https://github.com/epinio/epinio/blob/0d7c91219b59d20ed70645f192787f63e93ab566/internal/cli/cmd/namespaces.go#L113

enrichman (commented):
A couple of comments, but lgtm

internal/cli/cmd/configurations.go
andreas-kupries: It actually also removes quite a lot of error checks (cannot read flag this, that, or the other).

internal/cli/termui/ui.go
andreas-kupries: done with commit 89e4d35b1c0c7a5b15aef82a515c9d95091694c2

internal/cli/cmd/configurations.go
andreas-kupries: done with commit 04b657595d4b19edb345884dfece5b2cb078ad03

enrichman (approved):

mmartin24:
Rechecked in `Epinio Server Version: v1.10.0-rc2` and working:
- Check on service config:
```json
➜  epinio git:(main) ✗ epinio configuration show xcfefa578ebf738cc24fe12a0f5a3-mysql --output json | jq
{
  "meta": {
    "name": "xcfefa578ebf738cc24fe12a0f5a3-mysql",
    "namespace": "workspace",
    "createdAt": "2023-09-20T13:11:18Z"
  },
  "configuration": {
    "user": "",
    "details": {
      "mysql-password": "6Kj6CXDmvF",
      "mysql-root-password": "S8zR2lR18R"
    },
    "boundapps": [],
    "type": "service",
    "origin": "svc-mysql-cli"
  }
}
```
- Check on simple config
```
➜  epinio git:(main) ✗ epinio configuration show config-1 --output json| jq
{
  "meta": {
    "name": "config-1",
    "namespace": "workspace",
    "createdAt": "2023-09-20T13:11:17Z"
  },
  "configuration": {
    "user": "admin",
    "details": {
      "foo": "bar"
    },
    "boundapps": [],
    "type": "custom"
  }
}

```
- Check `list --output json`:
```json
➜  epinio git:(main) ✗ epinio configuration list --output json | jq
[
  {
    "meta": {
      "name": "config-1",
      "namespace": "workspace",
      "createdAt": "2023-09-20T13:11:17Z"
    },
    "configuration": {
      "user": "admin",
      "details": {
        "foo": "bar"
      },
      "boundapps": null,
      "type": "custom"
    }
  },
  {
    "meta": {
      "name": "xcfefa578ebf738cc24fe12a0f5a3-mysql",
      "namespace": "workspace",
      "createdAt": "2023-09-20T13:11:18Z"
    },
    "configuration": {
      "user": "",
      "details": {
        "mysql-password": "6Kj6CXDmvF",
        "mysql-root-password": "S8zR2lR18R"
      },
      "boundapps": null,
      "type": "service",
      "origin": "svc-mysql-cli"
    }
  }
]

```

PR #2586, opened 2023-09-20
Update Epinio version in helm-charts

enrichman:


andreas-kupries (approved):

enrichman:
We found out the original problem about this. This bump should not be needed, and we should try to fix the original issue here: https://github.com/epinio/epinio/issues/2596

Closing.

Issue #2605, opened 2023-09-25
Export of application is failing on multinode cluster with RWO storage


thehejik:
**Describe the bug**
Found when testing v1.10.0 on multinode (1+3) s390x k3s cluster.

Problem is that `epinio-server` pod and `image-export-job-works[...]`, which are both scheduled on different worker nodes, are trying to mount the same PV (claim `epinio/image-export-pvc`) but when using only RWO storage (Longhorn) the mount is impossible and the export job fails.

![image](https://github.com/epinio/epinio/assets/12828077/de0924d2-a8a4-4c5e-afca-efeb8dc2fef0)

```bash
$ k describe pod image-export-job-works-26c44b1dbe9ef2f5811403583d78a58dbe2t8zxr -n epinio
...
Events:
  Type     Reason              Age   From                     Message
  ----     ------              ----  ----                     -------
  Normal   Scheduled           59s   default-scheduler        Successfully assigned epinio/image-export-job-works-26c44b1dbe9ef2f5811403583d78a58dbe2t8zxr to rke2-master-th-ci-k3s-longhorn-0
  Warning  FailedAttachVolume  60s   attachdetach-controller  Multi-Attach error for volume "pvc-ff2d512c-6dcb-44ec-846e-b8c5504ab29f" Volume is already used by pod(s) epinio-server-858b8cc6f-g7ss2
```

This can by solved maybe by a pod affinity to make sure the export job and epinio-server pods are running on the same node.

thehejik:
I successfully simulated the possible `export-job` fix on multinode k8s with RWO Longhorn storage.

The following Job manifest creates a Pod, which will be scheduled on the same node as `epinio-server` Pod. Also verified that the pod is now able to write into `image-export` PV. It is driven by `podAffinity` and label `app.kubernetes.io/name=epinio-server` and `topologyKey: kubernetes.io/hostname` which value comes from the node running `epinio-server` pod.

Testing Job manifest:
```
apiVersion: batch/v1
kind: Job
metadata:
  name: job-with-epinio-server-affinity
  namespace: epinio
spec:
  template:
    spec:
      affinity:
        podAffinity:
          requiredDuringSchedulingIgnoredDuringExecution:
            - labelSelector:
                matchExpressions:
                  - key: app.kubernetes.io/name
                    operator: In
                    values:
                      - epinio-server
              topologyKey: kubernetes.io/hostname
      containers:
      - name: busybox
        image: busybox
        command: 
        - /bin/sh
[...]

PR #2622, opened 2023-10-03
Fixed `app export` on multinode clusters with RWO storage

thehejik:
Fixes. #2605 

Export over webUI tested on multi worker k3s cluster and it seems to work, the export has been always scheduled on the same node as `epinio-server` pod is using.


andreas-kupries (approved):
LGTM.

thehejik:
Verified fixed in `v1.11.0-rc1` on multinode (1+2) k3s cluster with longhorn SC. The export job and its pod is scheduled on the same node as epinio-server pod. Tried 3x in the row.

[multinode2-2023-11-27_16.00.13.webm](https://github.com/epinio/epinio/assets/12828077/4488124f-3575-40ad-b3ec-cfb4929d442d)

PR #2656, opened 2023-10-13
Added new Authorization Roles

enrichman:
See: https://github.com/epinio/epinio/issues/2631
- ADR: https://github.com/epinio/epinio/pull/2677
- Docs: https://github.com/epinio/docs/pull/360
- Charts: https://github.com/epinio/helm-charts/pull/506

---

This PR adds the new authorization roles. Please take a look at the ADR and docs to have a background over the implementation.

An Epinio user may have multiple roles, global or namescoped. A namescoped role will take action only over namescoped endpoints, while for requests that are not-namescoped then a global role is required. For example the "namespace create" action is global, and a global role with the `namespace_write` action is required.

A role has multiple Actions (i.e. `namespace_write`, `app_read`, ...), and every action has multiple Endpoints. An Endpoint has a 1:1 relation within an Epinio endpoint/api. The matching within actions and endpoints is mapped into an `actions.yaml` file, and loaded at the startup. Also the roles are loaded at the beginning.

Some considerations:
- a "default" role has to be defined. This role is the one that will be applied if no roles are found during an oidc authentication, or if a user doesn't have any roles. The `user` role is created as default during the Epinio installation, but another one could be defined setting the `default` field into the Role ConfigMap.
- if a user has a namescoped role then it has also access to the resources of that namespace, if authorized. This allows a dynamic namespace association for user created during the OIDC authentication
- no roles cleanup is done during a namespace delete. This was done to allows restore of previous permissions while recreating previously deleted namespaces

### DIFF
--- a/acceptance/api/v1/namespace_auth_test.go
+++ b/acceptance/api/v1/namespace_auth_test.go
@@ -190,9 +190,9 @@ var _ = Describe("Users Namespace", LNamespace, func() {
 					response.Body.Close()
 				})
 
-				It("doesn't show the other user's namespace", func() {
+				It("shows the other user's namespace", func() {
 					response := showNamespace(user1, passwordUser1, commonNamespace)
-					Expect(response.StatusCode).To(Equal(http.StatusForbidden))
+					Expect(response.StatusCode).To(Equal(http.StatusOK))
 					response.Body.Close()
 				})
 			})
--- a/acceptance/api/v1/namespace_auth_test.go
+++ b/acceptance/api/v1/namespace_auth_test.go
@@ -190,9 +190,9 @@ var _ = Describe("Users Namespace", LNamespace, func() {
 					response.Body.Close()
 				})
 
-				It("shows the other user's namespace", func() {
+				It("doesn't show the other user's namespace", func() {
 					response := showNamespace(user1, passwordUser1, commonNamespace)
-					Expect(response.StatusCode).To(Equal(http.StatusOK))
+					Expect(response.StatusCode).To(Equal(http.StatusForbidden))
 					response.Body.Close()
 				})
 			})
--- a/internal/auth/role.go
+++ b/internal/auth/role.go
@@ -167,10 +167,16 @@ func InitRoles(rolesGetter RolesGetter) error {
 	return nil
 }
 
-func ExtractRoleIDNamespace(idNamespace string) (string, string) {
-	userRoleNamespace := strings.Split(idNamespace, "::")
-	if len(userRoleNamespace) > 1 {
-		return userRoleNamespace[0], userRoleNamespace[1]
+//	will parse the roleID, returning the roleID without the namespace, and the namespace
+//
+// i.e.:
+//
+//	"admin" will return "admin" and ""
+//	"admin::workspace" will return "admin" and "workspace"
+func ParseRoleID(roleID string) (string, string) {
+	roleIDAndNamespace := strings.Split(roleID, "::")
+	if len(roleIDAndNamespace) > 1 {
+		return roleIDAndNamespace[0], roleIDAndNamespace[1]
 	}
-	return userRoleNamespace[0], ""
+	return roleIDAndNamespace[0], ""
 }
--- a/internal/auth/role.go
+++ b/internal/auth/role.go
@@ -10,16 +10,20 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// AdminRole is a special role. It permits all actions.
-var AdminRole = Role{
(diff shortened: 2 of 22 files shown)


internal/auth/role.go
```diff
@@ -0,0 +1,182 @@
...
+	"context"
+	"fmt"
+	"strconv"
+	"strings"
+
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+)
+
+// AdminRole is a special role, that has permission of every action
```
andreas-kupries, line 13:
```suggestion
// AdminRole is a special role. It permits all actions.
```

andreas-kupries (changes requested):

acceptance/api/v1/namespace_auth_test.go
```diff
@@ -190,9 +190,9 @@ var _ = Describe("Users Namespace", LNamespace, func() {
 					response.Body.Close()
 				})
 
-				It("doesn't show the other user's namespace", func() {
+				It("shows the other user's namespace", func() {
 					response := showNamespace(user1, passwordUser1, commonNamespace)
-					Expect(response.StatusCode).To(Equal(http.StatusForbidden))
+					Expect(response.StatusCode).To(Equal(http.StatusOK))
```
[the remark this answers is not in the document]
enrichman: We still have it but I had to change this because of a small change in the actual behavior. This was the difference that I documented in the description, and it's basically due because we are not cleaning up the roles of deleted namespaces.
I'm not sure if this should be the desired behavior or not. When using OIDC auth the roles will be added anyway. Recreation of a namespace with the same name will keep the same roles assigned. Maybe we want this, or maybe not.

I'm thinking that is probably better to cleanup the roles.
Regarding the OIDC auth maybe we should not even assign namespace-scoped roles if the namespace doesn't exists.

WDYT?

These are the existing tests about the namespaces:

https://github.com/epinio/epinio/blob/04de74d4ef482e561d65292c9c515e7c46abcba6/acceptance/api/v1/namespace_auth_test.go#L109-L119

https://github.com/epinio/epinio/blob/04de74d4ef482e561d65292c9c515e7c46abcba6/acceptance/api/v1/namespace_auth_test.go#L122-L135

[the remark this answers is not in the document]
enrichman: Base also on the latest discussion I've added the cleanup of the old namescoped roles. When a namespace is deleted also the user roles that were bound to that namespace are deleted. I can revert the test and keep the original behavior. :slightly_smiling_face:

internal/auth/user.go
```diff
@@ -174,3 +249,80 @@ func updateUserSecretData(user User, userSecret *corev1.Secret) *corev1.Secret {
 
 	return userSecret
 }
+
+// IsUpdateUserNeeded returns whetever a user needs to be updates, and the user with the updated informations
```
andreas-kupries, line 253:
```suggestion
// IsUpdateUserNeeded returns whenever a user needs to be updated, and the user with the updated information
```

internal/auth/user.go
```diff
@@ -174,3 +249,80 @@ func updateUserSecretData(user User, userSecret *corev1.Secret) *corev1.Secret {
...
+			"new", strings.Join(newNamespaces, ","),
+		)
+		updateNeeded = true
+		user.Namespaces = newNamespaces
+	}
+
+	return user, updateNeeded
+}
+
+// isUpdateUserRoleNeeded returns whetever the roles of a user needs to be updated, and the updated roles
```
andreas-kupries, line 282:
```suggestion
// isUpdateUserRoleNeeded returns whenever the roles of a user need to be updated, and the updated roles
```

internal/auth/user.go
```diff
@@ -174,3 +249,80 @@ func updateUserSecretData(user User, userSecret *corev1.Secret) *corev1.Secret {
...
+		// if one is different we can break and return
+		if previousRoles[i] != actualRoles[i] {
+			return actualRoles, true
+		}
+	}
+
+	return previousRoles, false
+}
+
+// isUpdateUserNamespacesNeeded returns whetever the namespaces of a user needs to be updated, and the updated namespaces
```
andreas-kupries, line 304:
```suggestion
// isUpdateUserNamespacesNeeded returns whenever the namespaces of a user need to be updated, and the updated namespaces
```

andreas-kupries (changes requested):
only comment things

andreas-kupries (dismissed):

thehejik (approved):

andreas-kupries (approved):

PR #2660, opened 2023-10-16
Update user roles on each new login session

### LINKED ISSUE #2561: Do the Groups Roles mapping every time a sso user starts a new session
Currently the user role is bound when the user is created and any modification of the groups will not propagate to already existing users.

Do the Groups Roles mapping every time a user starts a new session.

**Additional context**
In large organizations isn't efficient  keep track of all users modifications and keep this up to date on epinio.
By enabling this possibility, users could be fully managed on the IdP and security process, like breaking glass, temporary access, etc.. could be easily put in place.

enrichman:
Fix #2561 
Ref doc: https://github.com/epinio/docs/pull/340

This PR adds the `UpdateUser` method to the AuthService in order to update the User also during the next oidc logins.

It removes also the `AddNamespaceToUser` and `AddGitconfigToUser` methods. These were useless when we already have the `User` available. We can update it directly, without loading again the users and look for the user.

### DIFF
--- a/internal/api/v1/gitconfig/create.go
+++ b/internal/api/v1/gitconfig/create.go
@@ -92,7 +92,8 @@ func Create(c *gin.Context) apierror.APIErrors {
 	}
 
 	// add the gitconfig to the User's gitconfigs
-	err = authService.AddGitconfigToUser(ctx, user.Username, request.ID)
+	user.AddGitconfig(request.ID)
+	user, err = authService.UpdateUser(ctx, user)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


mmartin24:
Checked in `v1.11.0-rc1-8-g1e1f1fba` as detailed here: https://confluence.suse.com/display/Epinio/How+to+check+update+of+user+roles+on+each+new+login+session

Before fix:
before_fix_oidc


After fix:
after_fix_roles_persisted


Note: this was not possible to be checked in epinio `v1.11.0-rc1` due to https://github.com/epinio/epinio/issues/2713#issuecomment-1825252728

mmartin24:
Rechecked in `v1.11.0-rc2` and working:

image

PR #2679, opened 2023-10-23
Update traefik values for CI

thehejik:
Jobs using traefik started to fail, seems they changed their values `ports.web.redirectTo` to `ports.web.redirectTo.port`, anyway the CI is still failing but on something else.

Initial failure visible [here](https://github.com/epinio/epinio/actions/runs/6610814765/job/17953611984#step:12:281).

GKE-up https://github.com/epinio/epinio/actions/runs/6613463682
REK-up https://github.com/epinio/epinio/actions/runs/6613467617


mmartin24 (approved):
LGTM

PR #2689, opened 2023-11-02
Added `EpinioArgValidator` for matchers autocomplete

enrichman:
Added EpinioArgValidator. Removed namespace, configuration and catalog matchers


internal/cli/cmd/args.go
```diff
@@ -92,46 +89,53 @@ type RegistryMatcher interface {
...
+	return NewEpinioArgValidator(NewEpinioMatcher(matcher))
+}
 
-		return matches, cobra.ShellCompDirectiveNoFileComp
-	}
+func AnyArgsValidator(matcher MatcherFunc) ValidArgsFunc {
+	epinioMatcher := NewEpinioMatcher(matcher)
+
+	epinioMatcher.Filter = FilterMatches
+	epinioMatcher.ArgLimit = -1
```
andreas-kupries, line 136: For future PRs extending the set of matchers built on this framework, maybe add `Any()` and `Filter(filter)` modifiers which take an `EpinioMatcher`, make the change, and return the changed struct. Then the next line can be written as
```
return NewEpinioArgValidator(NewEpinioMatcher(matcher).Any().Filter (FilterMatches))
```

andreas-kupries (approved):

internal/cli/cmd/args.go
enrichman: Yes, makes sense! Thanks

andreas-kupries (approved):

PR #2698, opened 2023-11-15
Fixed missing repo update when chart not found

### LINKED ISSUE #2690: Epinio Helm Update Interval?
Hi,

We're giving Epinio a go. Since we have used PCF on premises before, Epinio seems familiar to us, and thus far, we are enjoying it. Continue the fantastic job. But I have two inquiries regarding specifics. 

1. A service CRD is applied and displayed immediately on the service catalog; however, epinio fails to retrieve the helm chart when we attempt to build a service instance, resulting in an error message stating that the helm repo update needs to be executed. The chart is successfully fetched if we remove the Epinio server pod and try the process again. It ought to be able to recognize when the service catalog is updated and either update the helm repos or do so on a regular basis. 

2. How should a service be upgraded properly? Stated differently, when we upgrade the helm chart version of a service crd, instances only can be upgraded by deleting first and recreating them. Any other rolling upgrade scenario would be great. 

Thanks in advance.

enrichman:
Fix #2690 

To fix an issue with the https://github.com/mittwald/go-helm-client/issues/176 a breaking change was introduced: https://github.com/mittwald/go-helm-client/pull/177

Basically it means that the `AddOrUpdateChartRepo` func is not updating the repositories anymore.

This PR fix the issue checking if the requested chart version is available in the repo, and if not found we update the repositories and check again. The `initHelmOCIRegistryOrRepository` was added to remove some code from the main func.

### DIFF
--- a/internal/helm/helm.go
+++ b/internal/helm/helm.go
@@ -37,6 +37,7 @@ import (
 	hc "github.com/mittwald/go-helm-client"
 	"github.com/spf13/viper"
 	"gopkg.in/yaml.v2"
+	"helm.sh/helm/v3/pkg/action"
 	"helm.sh/helm/v3/pkg/kube"
 	"helm.sh/helm/v3/pkg/registry"
 	helmrelease "helm.sh/helm/v3/pkg/release"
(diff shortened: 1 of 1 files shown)


andreas-kupries (approved):

thehejik:
Successfully verified on `v1.11.0-rc1`, working without `epinio-server` restart. No error and the service is running after version bump.

I could reproduce the issue on E v1.10.0. Restart of `epinio-server` to fetch latest chart from custom repo helped. BTW the OCI registries were not affected (there is no index.yaml to be refreshed).

Tested by using custom helm repo exposed on gh-pages.

This is used service manifest:
```yaml
kubectl apply -f - <<EOF
kind: Service
metadata:
  name: simplepod-dev
  namespace: epinio
spec:
  name: simplepod-dev
  shortDescription: Just a simple pod
  description: Just a simple pod
  chart: simplepod
  chartVersion: 0.0.3
  serviceIcon: https://bitnami.com/assets/stacks/nginx/img/nginx-stack-220x234.png
  appVersion: 0.0.1
  helmRepo:
    name: thehejik
    url: https://thehejik.github.io/testing-charts
EOF
```

This is how I was deploying the service:
```bash
$ epinio --verbosity=1 --trace-level=13 service create simplepod-dev asimplepod1 --wait
```

Once the original `chartVersion` was successfully deployed in E I bumped the version on github `charts/simplepod/Chart.yaml` to higher `version` and then triggered manually `Release` action on gh - then I verified that the `index.yaml` has been updated also in `gh-pages` branch.

After changing the service version by `k edit services.application.epinio.io -n epinio simplepod-dev` I tried to create another service:

```bash
$ epinio --verbosity=1 --trace-level=13 service create simplepod-dev asimplepod2 --wait
```

PR #2701, opened 2023-11-16
detect users defined multiple times and reject them

### LINKED ISSUE #2692: Same username should not be possible to be assigned
# Issue
It is possible to create 2 or more users with the same username. Login (and possibly all the other APIs) are not deterministic.

# To reproduce

- Create two users with different metadata name but same `username`

```yaml
cat <<EOF | kubectl apply -f -                                                              
apiVersion: v1
kind: Secret
type: BasicAuth
metadata:
  labels:
    epinio.io/api-user-credentials: "true"
    epinio.io/role: "admin"
  name: my-epinio-user-admin
  namespace: epinio
stringData:
  username: myuser
  namespaces: workspace
  password: "\$2a\$10\$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
EOF
```

```yaml
cat <<EOF | kubectl apply -f -                                                              
apiVersion: v1
kind: Secret
type: BasicAuth
metadata:
  labels:
    epinio.io/api-user-credentials: "true"
    epinio.io/role: "admin"
  name: my-epinio-user
  namespace: epinio
stringData:
  username: myuser
  namespaces: workspace
  password: "\$2a\$10\$6bCi5NMstMK781In7JGiL.B44pgoplUb330FQvm6mVXMppbXBPiXS"
EOF
```
- Go to secrets and decode `my-epinio-user-admin` and `my-epinio-user`

Image

andreas-kupries:
fix #2692 

users defined multiple times are filtered, i.e. seen as not existing, primarily.
enough data is kept to report them as `defined multiple times`, instead of as `not `found`.

### DIFF
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	ErrUserNotFound = errors.New("user not found")
-	ErrUserMultiple = errors.New("user defined multiple times, please talk to the operator")
+	ErrUsernameConflict = errors.New("this user is defined multiple times, please talk to the operator")
 )
 
 //counterfeiter:generate -header ../../LICENSE_HEADER k8s.io/client-go/kubernetes/typed/core/v1.SecretInterface
@@ -157,7 +157,7 @@ func (s *AuthService) GetUserByUsername(ctx context.Context, username string) (U
 	if ok && (count > 1) {
 		s.Logger.V(1).Info("user defined multiple times", "user", username, "count", count)
 
-		return User{}, ErrUserMultiple
+		return User{}, ErrUsernameConflict
 	}
 
 	s.Logger.V(1).Info("user not found")
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -65,10 +65,10 @@ func NewAuthService(logger logr.Logger, cluster *kubernetes.Cluster) *AuthServic
 	}
 }
 
-// getUsers returns all the Epinio users with no conflicting definitions. it further returns a map
+// GetUsers returns all the Epinio users with no conflicting definitions. it further returns a map
 // of definition counts enabling the caller to distinguish between `truly does not exist` versus
 // `has conflicting definitions`.
-func (s *AuthService) getUsers(ctx context.Context) ([]User, DefinitionCount, error) {
+func (s *AuthService) GetUsers(ctx context.Context) ([]User, DefinitionCount, error) {
 	s.Logger.V(1).Info("GetUsers")
 
 	secrets, err := s.getUsersSecrets(ctx)
@@ -145,7 +145,7 @@ func (s *AuthService) GetRoles(ctx context.Context) (Roles, error) {
 func (s *AuthService) GetUserByUsername(ctx context.Context, username string) (User, error) {
 	s.Logger.V(1).Info("GetUserByUsername", "username", username)
 
-	users, counts, err := s.getUsers(ctx)
+	users, counts, err := s.GetUsers(ctx)
 	if err != nil {
 		return User{}, errors.Wrap(err, "error getting users")
 	}
@@ -156,7 +156,7 @@ func (s *AuthService) GetUserByUsername(ctx context.Context, username string) (U
 		}
 	}
 
-	// user not found. check if this was because it was filtered out by getUsers() due to
+	// user not found. check if this was because it was filtered out by GetUsers() due to
 	// conflicting definitions for it.
 
 	count, ok := counts[username]
@@ -212,7 +212,7 @@ func (s *AuthService) UpdateUser(ctx context.Context, user User) (User, error) {
 func (s *AuthService) RemoveNamespaceFromUsers(ctx context.Context, namespace string) error {
 	s.Logger.V(1).Info("RemoveNamespaceFromUsers", "namespace", namespace)
 
-	users, _, err := s.getUsers(ctx)
+	users, _, err := s.GetUsers(ctx)
 	if err != nil {
 		return errors.Wrap(err, "error getting users")
 	}
@@ -242,7 +242,7 @@ func (s *AuthService) RemoveNamespaceFromUsers(ctx context.Context, namespace st
 func (s *AuthService) RemoveGitconfigFromUsers(ctx context.Context, gitconfig string) error {
 	s.Logger.V(1).Info("RemoveGitconfigFromUsers", "gitconfig", gitconfig)
 
-	users, _, err := s.getUsers(ctx)
+	users, _, err := s.GetUsers(ctx)
 	if err != nil {
 		return errors.Wrap(err, "error getting users")
 	}
--- a/internal/auth/auth.go
+++ b/internal/auth/auth.go
@@ -34,7 +34,7 @@ import (
 
 var (
 	ErrUserNotFound     = errors.New("user not found")
-	ErrUsernameConflict = errors.New("this user is defined multiple times, please talk to the operator")
+	ErrUsernameConflict = errors.New("user is defined multiple times")
 )
 
 //counterfeiter:generate -header ../../LICENSE_HEADER k8s.io/client-go/kubernetes/typed/core/v1.SecretInterface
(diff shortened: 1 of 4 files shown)


enrichman (changes requested):
Having a look at this I've seen a couple of things that were probably leftovers.
I would make this changes:

1) Drop the `loadUsersMap` and use the  `authService.GetUserByUsername`

The `loadUsersMap` func doesn't make really sense. IIRC it was used/needed for a Gin automatic detection for the BasicAuth, but we are doing a different check now, so it can be probably removed completely, using directly the `authService.GetUserByUsername`, since we need a specific user.

This will be very convenient, because we can immediately check if the error returned is the `ErrUserMultiple`.

2)  Make the `authService.GetUsers` private -> `authService.getUsers`

The `GetUsers` func at the moment can be used only privately. The only other place where it's used is in the `internal/api/v1/middleware/token.go`, but we need to get the specific user also there (and so eventually trigger the same `ErrUserMultiple` error if needed).

3) `authservice.GetUsers` could return the same ErrUserMultiple error

Probably we should trigger the same error also in the `GetUsers`, but maybe this means that we are interested to know who are the users with this "wrong" configuration.
To do this we can create a more "complex" error type:

```go
type ErrUserMultiple struct {
	UsersCount map[string]int
}

func (e *ErrUserMultiple) Error() string {
	return fmt.Sprintf("found %d users with the same username", len(e.UsersCount))
}
```

This will change the check to something like this:

```go
var errMultiple *ErrUserMultiple
errors.As(err, &errMultiple) {
	// do something
}
```

This could be useful when in the future we would like to know from the `GetUsers` which are the wrong users.

4) Drop the `DefinitionCount`

This field is never cleared, so the `AuthService` will keep the state. It's not also a field that seems to be part of the "object", but scoped to the request. Not sure if we are keeping the `AuthService` instance, but if so that means that we would need to restart Epinio, if that's the case. I would remove it anyway. It's not needed.

5) Rename the error

Change `ErrUserMultiple` to `ErrMultipleUsersFound` or `ErrUsernameConflict`.
If moving to the error struct the convention is to have the `XXXError`.

So:

```go
var ErrMultipleUsersFound = errors.New("dbsjksdjk")
// or
type ErrUserMultiple struct{}
```

6) Does it work?

I have not tested it but the `GetUserByUsername` seems to return early:

```go
for _, user := range users {
	if user.Username == username {
		return user, nil
	}
}

count, ok := s.Counts[username]
if ok && (count > 1) {
	s.Logger.V(1).Info("user defined multiple times", "user", username, "count", count)

	return User{}, ErrUserMultiple
}
```

7) Tests

If possible I would add a test (unit or acceptance). The unit should be pretty easy and fast to add. We already have them setup, and we just need to return the secrets from the FakeSecretInterface.

The acceptance are a bit more annoying.

andreas-kupries:
> I have not tested it but the GetUserByUsername seems to return early:

A pity that you did not attach this comment to the specific code.
Remember: A user defined multiple times is filtered out when the users are loaded. It is effectively 'not found'.
The loop does not find it, and the code reaches the check where it determines of the user truly does not exist, or just because of a multi-definition.

andreas-kupries:
I will now make changes, with commits as separate as possible, and quoting relevant parts of the comment in the commit message. It is too difficult to disentangle them otherwise.

enrichman:
Thank you. Yes, sorry about not commenting the code but unfortunately Github doesn't let you comment the code that was not changed, and most of the comments were not related to the actual changes. :slightly_frowning_face: 

I'll try to comment better in the second round. :+1:

andreas-kupries:
> most of the comments were not related to the actual changes

I am ok with such comments attached to actual changes and refering to the unchanged parts.

andreas-kupries:
> The only other place where it's used is in the internal/api/v1/middleware/token.go

no. Unit testing uses it as well, and cannot test a private function.

```
% golangci-lint run --skip-files docs.go --timeout 5m
internal/auth/action_test.go:1: : # github.com/epinio/epinio/internal/auth_test [github.com/epinio/epinio/internal/auth.test]
internal/auth/auth_test.go:79:39: authService.getUsers undefined (type *"github.com/epinio/epinio/internal/auth".AuthService has no field or method getUsers)
```

andreas-kupries:
Changes addressing comments pushed.

internal/auth/auth.go
```diff
@@ -33,12 +33,15 @@ import (
 )
 
 var (
-	ErrUserNotFound = errors.New("user not found")
+	ErrUserNotFound     = errors.New("user not found")
+	ErrUsernameConflict = errors.New("this user is defined multiple times, please talk to the operator")
```
enrichman, line 37: I would just tell what happened and not what to do:

```suggestion
	ErrUsernameConflict = errors.New("username defined multiple times")
```

internal/auth/auth.go
```diff
@@ -62,19 +65,42 @@ func NewAuthService(logger logr.Logger, cluster *kubernetes.Cluster) *AuthServic
...
+	userCount := DefinitionCount{}
+
+	for _, secret := range secrets {
+		name := string(secret.Data["username"])
+		count, ok := userCount[name]
+		if !ok {
+			userCount[name] = 1
+			continue
+		}
+		userCount[name] = count + 1
```
enrichman, line 92: The zero value of a int is already 0, so this can be changed to:

```suggestion
		count := userCount[name]
		userCount[name] = count + 1
```

internal/auth/auth.go
```diff
@@ -62,19 +65,42 @@ func NewAuthService(logger logr.Logger, cluster *kubernetes.Cluster) *AuthServic
...
+		return nil, nil, errors.Wrap(err, "error getting users secrets")
 	}
 
 	users := []User{}
 	usernames := []string{}
 
+	// map of per-user definition counts, to detect users with conflicting definitions
+	userCount := DefinitionCount{}
+
+	for _, secret := range secrets {
```
enrichman, line 85: I'm commenting here but it's about the two loops. It's probably easier to read if we fetch all the users, mapping the duplicates, and then filter them.

Please be aware that the `newUserFromSecret` is suppose to hide all the "secret-user" internal, to avoid knowing from other parts which data contains what (ie: `secret.Data["username"]`). This is helpful if we change something in the secrets, and to avoid typos.

```go
allUsers :=  []User{}
usernames := []string{}
userCount := DefinitionCount{}

// fetch and convert all the users, mapping the duplicates
for _, secret := range secrets {
	user := newUserFromSecret(s.Logger, secret)
	allUsers = append(allUsers, user)
	usernames = append(usernames, user.Username)

	count := userCount[user.Username]
	userCount[user.Username] = count + 1
}

filteredUsers :=  []User{}

// filter duplicates
for _, user := range allUsers {
	if userCount[user.Username] > 1 {
		s.Logger.V(1).Info("skip duplicate user", "user", user.Username)
		continue
	}
	filteredUsers = append(filteredUsers, user)
}
```

internal/auth/auth.go
```diff
@@ -33,12 +33,15 @@ import (
...
 var (
-	ErrUserNotFound = errors.New("user not found")
+	ErrUserNotFound     = errors.New("user not found")
+	ErrUsernameConflict = errors.New("this user is defined multiple times, please talk to the operator")
 )
 
 //counterfeiter:generate -header ../../LICENSE_HEADER k8s.io/client-go/kubernetes/typed/core/v1.SecretInterface
 //counterfeiter:generate -header ../../LICENSE_HEADER k8s.io/client-go/kubernetes/typed/core/v1.ConfigMapInterface
 
+type DefinitionCount map[string]int
```
enrichman, line 43: Not sure about the name, but I've nothing better to propose.

enrichman (changes requested):
Some small comments. Thank you!

internal/auth/auth.go
andreas-kupries: ok

internal/auth/auth.go
andreas-kupries: Actually, then we do not need a helper variable either. ok

internal/auth/auth.go
andreas-kupries: I had hoped to avoid the construction of `User` structures for the conflicting parts. That said, with the encapsulation argument, it is better to do so. Ok

internal/auth/auth.go
andreas-kupries: done. commit b04ac0d8192ab565f309eb40f53dc5acf34710cb

internal/auth/auth.go
andreas-kupries: done. commit b04ac0d8192ab565f309eb40f53dc5acf34710cb

enrichman (approved):
LGTM

PR #2712, opened 2023-11-21
Magic handling of single app directory in app source archive. Plus CLI support for app archives.

### LINKED ISSUE #2465: When using archive if it as a subfolder it dont compile
Hi when i download git archive from gitlab he create a subfolder inside the archive.
Epinio cannot find the app in the subfolder 

it can be interesting to support by default the gitlab zip export of a repo

Thanks

andreas-kupries:
fix #2465 
imports chart PR https://github.com/epinio/helm-charts/pull/523

### DIFF
--- a/internal/cli/usercmd/push.go
+++ b/internal/cli/usercmd/push.go
@@ -273,23 +273,16 @@ func (c *EpinioClient) uploadSources(log logr.Logger, appRef models.AppRef, sour
 		return "", err
 	}
 
-	var archive string
+	archive := source
 
 	if fileInfo.IsDir() {
 		// package directory as archive/tarball
 		tmpDir, tarball, err := helpers.Tar(source)
-		defer func() {
-			if tmpDir != "" {
-				_ = os.RemoveAll(tmpDir)
-			}
-		}()
+		defer os.RemoveAll(tmpDir)
 		if err != nil {
 			return "", err
 		}
 		archive = tarball
-	} else {
-		// consider file as the archive
-		archive = source
 	}
 
 	c.ui.Normal().Msg("Uploading application code ...")
--- a/internal/cli/usercmd/push.go
+++ b/internal/cli/usercmd/push.go
@@ -268,24 +268,38 @@ func (c *EpinioClient) AppPush(ctx context.Context, manifest models.ApplicationM
 func (c *EpinioClient) uploadSources(log logr.Logger, appRef models.AppRef, source string) (string, error) {
 	c.ui.Normal().Msg("Collecting the application sources ...")
 
-	tmpDir, tarball, err := helpers.Tar(source)
-	defer func() {
-		if tmpDir != "" {
-			_ = os.RemoveAll(tmpDir)
-		}
-	}()
+	fileInfo, err := os.Stat(source)
 	if err != nil {
 		return "", err
+	}
+
+	var archive string
+
+	if fileInfo.IsDir() {
+		// package directory as archive/tarball
+		tmpDir, tarball, err := helpers.Tar(source)
+		defer func() {
+			if tmpDir != "" {
+				_ = os.RemoveAll(tmpDir)
+			}
+		}()
+		if err != nil {
+			return "", err
+		}
+		archive = tarball
+	} else {
+		// consider file as the archive
+		archive = source
 	}
 
 	c.ui.Normal().Msg("Uploading application code ...")
 
 	log.V(1).Info("upload code")
 
-	// open the tarball
-	file, err := os.Open(tarball)
+	// open the archive
+	file, err := os.Open(archive)
 	if err != nil {
-		return "", errors.Wrap(err, "failed to open tarball")
+		return "", errors.Wrap(err, "failed to open archive")
 	}
 	defer file.Close()
 
--- a/acceptance/apps/wordpress_test.go
+++ b/acceptance/apps/wordpress_test.go
@@ -121,7 +121,7 @@ var _ = Describe("Wordpress", func() {
 	It("can deploy Wordpress", func() {
 		out, err := env.EpinioPush(wordpress.Dir, wordpress.Name, "--name", wordpress.Name,
 			"--builder-image", "paketobuildpacks/builder:0.2.443-full",
-			"-e", "BP_PHP_WEB_DIR=wordpress",
+			"-e", "BP_PHP_WEB_DIR=.",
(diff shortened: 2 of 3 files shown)


andreas-kupries:
The PR as is break the wordpress test ...
It seems that for PHP an EV setting like
```
"-e", "BP_PHP_WEB_DIR=wordpress"
```

tells the buildpack to look into the app directory.

I wonder if other buildpacks have equivalent settings.

andreas-kupries:
Deeper research today shows that I am somewhat mistaken about the `BP_PHP_WEB_DIR`.
It specifies the location of the files to serve by the PHP web server, with a default of `htdocs`.
Of course, the `.php` files among these files are the app, so it manages to work out.

However most other languages do not really have equivalents
(See howto section of https://paketo.io/docs/)

`java` and `node.js` may work as well, via `BP_NODE_PROJECT_PATH`, and `.net` possibly too, via `BP_DOTNET_PROJECT_PATH`. The rest I saw nothing similar, not in the howtos, nor in the linked references.

In the end we look to need the magic indeed.

enrichman:
> Deeper research today shows that I am somewhat mistaken about the `BP_PHP_WEB_DIR`. It specifies the location of the files to serve by the PHP web server, with a default of `htdocs`. Of course, the `.php` files among these files are the app, so it manages to work out.
> 
> However most other languages do not really have equivalents (See howto section of https://paketo.io/docs/)
> 
> `java` and `node.js` may work as well, via `BP_NODE_PROJECT_PATH`, and `.net` possibly too, via `BP_DOTNET_PROJECT_PATH`. The rest I saw nothing similar, not in the howtos, nor in the linked references.
> 
> In the end we look to need the magic indeed.

Hm, so.. I'm not sure. Why is it needed now?

Btw, the helm charts update is quite big because it's not containing only the needed changes but also previously merged PRs. Would you mind merging [this PR](https://github.com/epinio/epinio/pull/2718) and then rebasing this one?
So we can have just the related changes of the helm-charts about this.

andreas-kupries:
Approved #2718 

> Hm, so.. I'm not sure. Why is it needed now?

To clarify. The ticket #2465 asks for the magic. I thought for a while yesterday that we could without that magic and simply refer to BP config to allow the submitter achieve the same thing. Due to my research this morning I came to the conclusion that there is not enough in BP configs to do so, making the magic needed for the submitters case.

internal/cli/usercmd/push.go
```diff
@@ -268,24 +268,38 @@ func (c *EpinioClient) AppPush(ctx context.Context, manifest models.ApplicationM
...
+	var archive string
+
+	if fileInfo.IsDir() {
+		// package directory as archive/tarball
+		tmpDir, tarball, err := helpers.Tar(source)
+		defer func() {
+			if tmpDir != "" {
+				_ = os.RemoveAll(tmpDir)
+			}
+		}()
```
enrichman, line 285: if `tmpDir` is empty `""` the `os.RemoveAll` is not going to fail, so it's safe to call it anyway. Also we are not even logging the error, so we can just call it:

```suggestion
		defer os.RemoveAll(tmpDir)
```

internal/cli/usercmd/push.go
```diff
@@ -268,24 +268,38 @@ func (c *EpinioClient) AppPush(ctx context.Context, manifest models.ApplicationM
...
+		if err != nil {
+			return "", err
+		}
+		archive = tarball
+	} else {
+		// consider file as the archive
+		archive = source
+	}
```
enrichman, line 293: If in the `else` we have only this we can remove it completely initializing the `archive` as `source` at the beginning, and then if `fileInfo.IsDir` we can overwrite it:

```go
archive := source

if fileInfo.IsDir() {
	// blabla
	archive = tarball
}

```

enrichman (changes requested):

internal/cli/usercmd/push.go
andreas-kupries: Done. Commit 450458216cfa6b5b756f082194f3a26a246a33c9

internal/cli/usercmd/push.go
andreas-kupries: Done. Commit 450458216cfa6b5b756f082194f3a26a246a33c9

enrichman (dismissed):

andreas-kupries:
While acceptance-apps has failed a local run with additional inspection shows that response is
```
<html>
  <head><title>403 Forbidden</title></head>
  <body>
  <center><h1>403 Forbidden</h1></center>
  <hr><center>nginx</center>
  </body>
</html>
```
instead of 404. The app is deployed, and something prevents access. That does not point to the change here.

andreas-kupries:
Ok. I suspect that the change to `find` broke the lifting code.

andreas-kupries:
Yes. The logs from the local focused test do not show the necessary messages.
Without lifting the files are in `./wordpress`, while BP assumes `.` due to the EVs given.

enrichman (approved):

thehejik:
Verified in `v1.11.0-rc2`. Zip archive with sample app source downloaded from gitlab - the archive contained `sample-main/index.php` and `sample-main/README.md` files. The app has been deployed successfully:


This is related part from staging:
```
🕞  [stage-workspace-sample-c] unpack-blob Archive:  ../e83b6b11-0f6b-4dde-820e-d86ac18dd6de 
🕞  [stage-workspace-sample-c] unpack-blob 9dab8e95a971d6c96ae4f8ec3b2a78e515ada675 
🕞  [stage-workspace-sample-c] unpack-blob    creating: sample-main/ 
🕞  [stage-workspace-sample-c] unpack-blob   inflating: sample-main/README.md 
🕞  [stage-workspace-sample-c] unpack-blob   inflating: sample-main/index.php 
🕞  [stage-workspace-sample-c] unpack-blob OK 
🕞  [stage-workspace-sample-c] unpack-blob Lifting "sample-main" contents ... 
🕞  [stage-workspace-sample-c] unpack-blob renamed 'sample-main/README.md' -> './README.md' 
🕞  [stage-workspace-sample-c] unpack-blob mv: cannot stat 'sample-main/.??*': No such file or directory 
🕞  [stage-workspace-sample-c] unpack-blob renamed 'sample-main/index.php' -> './index.php' 
🕞  [stage-workspace-sample-c] unpack-blob rmdir: removing directory, 'sample-main' 
```

Note for myself: `mv .. 'sample-main/.??*` moves all .hidden files. In this can no hidden files were in the archive.

PR #2717, opened 2023-11-22
drop: standard platform API request

### LINKED ISSUE #2715: Drop hardcoded `CNB_PLATFORM_API=0.4` version
### Is your feature request related to a problem?

Issue:
Currently, if we want to use version `paketobuildpacks/builder-jammy-full:latest`we will face an error like:

```
failed to set platform API: platform API version '0.4' is incompatible with the lifecycle 
```
As to Epinio `v1.11.0-rc1` we are using as default version `paketobuildpacks/builder-jammy-full:0.3.290` which is hardcoded,

To use the `paketobuildpacks/builder-jammy-full:latest` we can use `--env CNB_PLATFORM_API=0.7` as @andreas-kupries  discovered:

```
epinio push -n cat -p /home/mmartin/suse/epinio-repos/epinio/assets/sample-app --builder-image paketobuildpacks/builder-jammy-full:latest --env CNB_PLATFORM_API=0.7
```

We should somehow discuss how to document this for this / future releases

### Solution you'd like

_No response_

### Alternatives you've considered

_No response_

### Anything else?

_No response_

andreas-kupries:
fix #2715

### DIFF
--- a/acceptance/helpers/machine/apps.go
+++ b/acceptance/helpers/machine/apps.go
@@ -76,7 +76,7 @@ func (m *Machine) MakeGolangApp(appName string, instances int, deployFromCurrent
 	return m.MakeAppWithDir(appName, instances, deployFromCurrentDir, appDir)
 }
 
-func (m *Machine) MakeAppWithDir(appName string, instances int, deployFromCurrentDir bool, appDir string) string {
+func (m *Machine) MakeAppWithDir(appName string, instances int, deployFromCurrentDir bool, appDir string, more ...string) string {
 	By("creating app " + appName)
 
 	var pushOutput string
@@ -87,8 +87,10 @@ func (m *Machine) MakeAppWithDir(appName string, instances int, deployFromCurren
 		// This means that the command runs with it as the CWD.
 		pushOutput, err = m.EpinioPush(appDir,
 			appName,
-			"--name", appName,
-			"--instances", strconv.Itoa(instances))
+			append([]string{
+				"--name", appName,
+				"--instances", strconv.Itoa(instances),
+			}, more...)...)
 	} else {
 		// Note: appDir is handed as second argument to the epinio cli.
 		// This means that the command gets the sources from that directory instead of CWD.
--- a/internal/api/v1/application/stage.go
+++ b/internal/api/v1/application/stage.go
@@ -22,6 +22,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	"github.com/spf13/viper"
+	"gopkg.in/yaml.v2"
 	batchv1 "k8s.io/api/batch/v1"
 	"k8s.io/utils/ptr"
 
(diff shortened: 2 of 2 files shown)


andreas-kupries:
The CI breakage shows that we need support in the staging scripts to handle an empty `appenv` directory, i.e. when an app has no EV's at all.

enrichman:
Having a look at the Buildpack Platform spec this seems to be a required value:
- https://github.com/buildpacks/spec/blob/main/platform.md#platform-api-compatibility

They fallback to the 0.3 because it's the older without this need. The latest is the `0.12` ([Platform API v0.12](https://github.com/buildpacks/spec/releases/tag/platform%2Fv0.12)) but this breaking change seems a bit "scary to me":

> The concepts of stacks and mixins are removed in favor of existing constructs in the container image ecosystem such as operating system name, operating system version, and architecture (https://github.com/buildpacks/spec/pull/335, [RFC 096](https://github.com/buildpacks/rfcs/blob/main/text/0096-remove-stacks-mixins.md)).

The Paketo full builder is made with the concept of stacks, so I'm not sure it will work also with this latest API.

Having it hardcoded into Epinio doesn't seem a good option by the way.

Since this is related to the builder used imo makes sense to include it in the default-scripts, so it can be customized with different builder, without the user to explicitly know what to set, and why. It looks more build related, and having the use to explicitly set this basically everytime is very annoying.

enrichman (changes requested):
I would save the user to know which "Platform API" should use, and for which builder, everytime.

Since it's needed and build-related I would make it explicit into the default-scripts.
I would try the `0.11` (not the very latest), and see how it works with our samples. :crossed_fingers:

enrichman:
> > Since this is related to the builder used imo makes sense to include it in the default-scripts, so it can be customized with different builder, without the user to explicitly know what to set, and why. It looks more build related, and having the use to explicitly set this basically everytime is very annoying.
> 
> I will take a look into this possibility tomorrow.

@andreas-kupries thank you (missed the reply) :slightly_smiling_face:

andreas-kupries:
> Since it's needed and build-related I would make it explicit into the default-scripts.

I would prefer to specify it in the config map outside of the script itself. To keep the scripts as shareable as possible.

An `env` key, whose value is a YAML string, so that I can at least pretend to have a properly nested map.

enrichman:
> > Since it's needed and build-related I would make it explicit into the default-scripts.
> 
> I would prefer to specify it in the config map outside of the script itself. To keep the scripts as shareable as possible.
> 
> An `env` key, whose value is a YAML string, so that I can at least pretend to have a properly nested map.

That's even better. I forgot about the new specs.

I was thinking about something like:

```yaml
data:
  builder: "paketobuildpacks/builder-jammy-*:*"
  userID: "1001"
  groupID: "1000"
  base: epinio-stage-scripts
  env:
    - name: "CNB_PLATFORM_API"
      value: "0.11"
    - name: "FOO"
      value: "bar"
```

but then I realized that ConfigMaps are just key-value pairs, so I guess our best bet is going flat, as you said :slightly_smiling_face: 

```yaml
data:
  builder: "paketobuildpacks/builder-jammy-*:*"
  userID: "1001"
  groupID: "1000"
  base: epinio-stage-scripts
  env: |
    CNB_PLATFORM_API=0.11
    FOO=bar
```

andreas-kupries:
Actually

```
env: |
    CNB_PLATFORM_API: "0.11"
    FOO: bar
```

enrichman:
Yes, this encoding is less "standard" for env vars, but much easier to parse and check as a yaml. :smile:

enrichman (approved):
Just a small question but LGTM

mmartin24:
Tested [here](https://github.com/epinio/epinio/issues/2715#issuecomment-1831797748) and working

PR #2721, opened 2023-11-23
epinio-binary grown to 100Mi

thehejik:
RKE2-EC2-CI job started to fail because temporary PV for devel epinio-service binary (used by `make patch-epinio-deployment`) doesn't fit the binary anymore:

```
  pod/epinio-copier condition met
  Copying the binary on the PVC
  mv: write error: No space left on device
  command terminated with exit code 1
```

Its PVC was requesting 100Mi originally, now changed to 150Mi. 

A bit strange is that it doesn't fail on other public clouds, most probably the used `Longhorn` storage provider is more strict or using a different block-size.

Testrun:
https://github.com/epinio/epinio/actions/runs/6971231046

### DIFF
--- a/scripts/patch-epinio-deployment.sh
+++ b/scripts/patch-epinio-deployment.sh
@@ -61,7 +61,7 @@ spec:
     - ReadWriteOnce
   resources:
     requests:
-      storage: 100Mi
+      storage: 150Mi
 EOF
 
 echo "Creating the dummy copier Pod"


andreas-kupries (approved):
Checked my local binary. That is is still at 90M.
Maybe the other clouds simply have not run into the large binary yet.

enrichman:
Wow. For real? That's a lot.

Maybe we should see if we can shrink it a bit. Not for the server, but it's a bit annoying that the users just using the CLI needs to download a ~100Mb binary.

thehejik:
I'd say this is Longhorn specific as the binary can fit when other storage than Longhorn is used - my binary is also 90M only. We are requesting 100Mi PV on every cloud we are testing on.

thehejik:
See LH example about Volumes, point 1, bullet 2 - https://longhorn.io/docs/1.4.1/volumes-and-nodes/volume-size/#example

> There is some meta info in the volume (12Gi)  hence the actual size is 260 Mi and is not exactly 0.

PR #2796, opened 2024-12-20
update splatform references for sample-app and tests, to epinio

krumware:
This updates splatform repo references in the tests and sample apps, and migrates them to the epinio docker org.


vtrenton (approved):
LGTM

PR #2803, opened 2024-12-23
Fix/colpal rate limits

krumware:
comment out colpal since it's requiring extra unauthenticated image pulls as part of dependency building. This should be unnecessary for our ephemeral runners.


vtrenton (approved):
LGTM

PR #2804, opened 2024-12-23
add kubectl download to scripts

krumware:
Kubectl seems to be missing on the base nodes for testing. Let's download it like the other tools.


vtrenton (approved):
LGTM

PR #2810, opened 2024-12-31
Re-add cleanup temporarily for self-hosted runner workaround

krumware:
We need a workaround to get tests running while we work out the ARC runner requirements.
This should allow us to at least test the CI job updates


krumware:
closing for cleanup

PR #2857, opened 2025-06-24
Disable upgrade responder for now

krumware:
Upgrade Responser was used to gather deployment metrics and was previously integrated with a private EKS cluster. This needs to be disabled until it can be set up again.


giubacc (approved):
lg

PR #2874, opened 2025-07-21
Tests: Fix port usage and reference issues

krumware:
* fixes the port check by ensuring that a random port in a safe range is chosen, strips out an existing port (if specified), and doesn't use the existing port.
* use url.Parse instead of string replacement to correct tests when a port is specified.

### DIFF
--- a/acceptance/login_test.go
+++ b/acceptance/login_test.go
@@ -173,7 +173,7 @@ var _ = Describe("Login", LMisc, func() {
 		oldPort := parsed.Port()
 		randomPort := r.Intn(65535-1024) + 1024 // avoid well-known ports
 		if oldPort != fmt.Sprintf("%d", randomPort) {
-			randomPort += 1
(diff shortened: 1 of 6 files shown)


acceptance/login_test.go
```diff
@@ -164,8 +165,19 @@ var _ = Describe("Login", LMisc, func() {
...
 	It("respects the port when one is present [fixed bug]", func() {
-		randomPort := fmt.Sprintf(`:%d`, r.Intn(65536))
-		serverURLWithPort := serverURL + randomPort
+		parsed, err := url.Parse(serverURL)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Get the current port (if any) and strip it
+		host := parsed.Hostname()
+		oldPort := parsed.Port()
+		p := r.Intn(65535-1024) + 1024 // avoid well-known ports
```
dcharles525, line 174: Lets give the variable a name, looks like it was previously names randomPort?

PR #2886, opened 2025-07-31
Replace mholt/archiver with mholt/archives

dcharles525:
The old package was completely deprecated and CVE's assigned to it that were not going to be resolved. 

Testing:
1. push an app from the cli or through Epinio
2. use a git repo from any of the git related options to deploy an application.
3. verify the tar file is created and pushed to s3. 

closes https://github.com/epinio/epinio/security/dependabot/83


krumware (approved):
lgtm

PR #2967, opened 2026-02-11
New rbac implementation

psanghvi17:
### PR Checklist
- [X] Linting Test is passing
- [X] New Unit and Acceptance tests written for the context of the PR
- [X] Unit Tests are passing
- [X] Code is well documented
- [X] If applicable, a PR in the [epinio/docs](https://github.com/epinio/docs) repository has been opened
 

### Summary
Implements the new RBAC role system in the Epinio Helm chart. The chart now installs role ConfigMaps (`application_manager`, `application_developer`, `view_only`, `system_manager`) by default, allowing operators to create users and assign these roles via Kubernetes Secrets. Adds operator documentation for RBAC user management.

- Role resolution from Secrets (`epinio.io/roles` annotation) and role ConfigMaps
- Action-based authorization checks across application, namespace, configuration, and service endpoints
- `/api/v1/me` endpoint returning user, roles, namespaces, and gitconfigs for UI/CLI

### Technical notes summary
- Roles are read from ConfigMaps with label `epinio.io/role: "true"`; each ConfigMap defines `id`, `name`, and `actions`
- User roles are specified via Secret annotation `epinio.io/roles` (e.g. `application_manager,application_manager:namespace`)
- Authorization checks use `internal/auth` action IDs (e.g. `app_create`, `app_delete`, `configuration_read`)

- Login as admin user; verify full access to create/delete apps, configurations, services
- Login as RBAC user (e.g. `application_manager`); verify Create/Delete app work, API returns correct permissions
- Login as view-only user; verify API rejects create/delete requests
- Call `epinio whoami` or `/api/v1/me`; verify roles and namespaces are returned correctly

- **Auth middleware:** Ensure unauthenticated requests still fail correctly
- **Role resolution:** Users with invalid or missing role annotations should not gain unintended permissions
- **Existing admin/default users:** Verify continued access when using default `user`/`admin` roles

### DIFF
--- a/docs/decisions/20231013-authorization-roles.md
+++ b/docs/decisions/20231013-authorization-roles.md
@@ -139,7 +139,7 @@ data:
 
 #### Actions
 
-The actions are **hardcoded** as an embedded yaml file. This simplifies their management, and also enhances flexibility. An action can have some "dependencies", i.e.: the `app` action is a union of the `app_read`, `app_write`, `app_logs`, `app_exec`, and `app_portforward`.
+The actions are **hardcoded** as an embedded yaml file. This simplifies their management, and also enhances flexibility. An action can have some "dependencies", i.e.: the `app` action is a union of granular app actions plus `app_logs`, `app_exec`, and `app_portforward`.
 
 Every action lists the set of endpoints it allows. Note that some Epinio operations are formed from multiple endpoints, i.e. the `app push` consists of a Create, Update and others.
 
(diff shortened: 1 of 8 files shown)


docs/decisions/20231013-authorization-roles.md
grudra7714, line 1: I would recommend creating a new decision document .. the filename is dated so doesnt make sense to put our updates here as it may look like these decisions were made in the past

grudra7714 (commented):
Still reviewing but wanted to go ahead and add this

grudra7714 (approved):

PR #2970, opened 2026-02-16
Add grouped environment to app config

psanghvi17:
### PR Checklist
- [X] Linting Test is passing
- [X] New Unit and Acceptance tests written for the context of the PR
- [X] Unit Tests are passing
- [X] Code is well documented
 

### Summary
Fixes #
Exposes bound service environment variables on the application details endpoint so users can see both user-provided and service-provided environment variables for an app via a single API call.

Application model: Added EnvironmentGrouped *EnvVariableGroupedResponse to ApplicationConfiguration in models.go, providing a structured view of environment variables split into user and service maps.
App show endpoint: Updated fetch in application.go to populate configuration.environment_grouped using GroupedEnvironment, while keeping configuration.environment as the existing user-only map for backward compatibility.
Refactor: Extracted environment-loading logic from fetch into a small helper loadEnvironmentData to keep the function size within linting limits without changing behavior.

### Technical notes summary
Backward compatibility: Existing consumers that rely on configuration.environment are unaffected; it still contains only user-provided environment variables.
New data surface: Clients can now read configuration.environment_grouped.user and configuration.environment_grouped.service on GET /api/v1/namespaces/:namespace/applications/:app, which mirrors the structure of GET /api/v1/namespaces/:namespace/applications/:app/environment?grouped=true.
Error handling: Environment-related errors are wrapped with contextual messages and still update StatusMessage and Status consistently with the rest of fetch.

Happy path:
Create an app, bind at least one service to it, and verify that:
GET /api/v1/namespaces/<ns>/applications/<app>/environment?grouped=true returns non-empty service vars.
GET /api/v1/namespaces/<ns>/applications/<app> includes configuration.environment_grouped.service with the same key/value pairs.
No-services case:
App with no bound services: configuration.environment_grouped.service should be empty while user reflects only user env vars.
CLI integration:
epinio app env list APPNAME still works and shows consistent values with the two endpoints above.
Security / auth:
Accessing the app details endpoint as an unauthorized/unauthenticated user still respects existing auth and does not leak env data beyond what was previously visible.


dcharles525 (approved):
LGTM

PR #3013, opened 2026-05-07
Server-side search for apps, namespaces, services, and configurations

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented

### Summary
Fixes EPINIO-603

Added `?search=` query parameter to all resource list endpoints (applications, namespaces, services, configurations) both namespace-scoped and global variants. Case-insensitive substring match on resource name, applied before pagination. Also added missing pagination to the namespace-scoped service and configuration list endpoints. A `GetSearchParam` helper was added to the `response` package alongside the existing pagination helpers.

### Technical notes summary

Search is opt-in via `?search=<term>` and composes with existing pagination params (`?page=`, `?pageSize=`). Omitting `?search` returns the full list, no change to existing behavior.

- `GET /api/v1/applications?search=foo` — only matching apps returned
- `GET /api/v1/namespaces?search=foo` — only matching namespaces returned
- `GET /api/v1/services?search=foo` — only matching services returned
- `GET /api/v1/configurations?search=foo` — only matching configurations returned
- Composed: `?search=foo&page=1&pageSize=10` — search filters first, pagination applies to filtered set
- No `?search` param — full list returned (backwards-compatible)

Search param is opt-in and all existing callers omit it.

### DIFF
--- a/acceptance/api/v1/application_index_test.go
+++ b/acceptance/api/v1/application_index_test.go
@@ -110,6 +110,41 @@ var _ = Describe("Apps Endpoint", LApplication, func() {
 		Expect(paged.Items).To(HaveLen(1))
 	})
 
+	It("filters applications by search term", func() {
+		app1 := catalog.NewAppName()
+		env.MakeContainerImageApp(app1, 1, containerImageURL)
+		defer env.DeleteApp(app1)
+		app2 := catalog.NewAppName()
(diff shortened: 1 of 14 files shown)


psanghvi17:
LGTM.

PR #3019, opened 2026-05-08
Feat/crud crds

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [ ] If applicable, a PR in the epinio/docs repository has been opened

### Summary
Fixes EPINIO-587

Adds full CRUD over the API for three CRD-backed resources that were previously read-only or seed-only: `AppChart`, `BuilderImage` (renamed from `Buildpack`), and `CatalogService`. Companion CRD rename in `epinio/helm-charts` ships in a paired PR.

> Reviewer note: most of the line count is generated counterfeiter fakes and rename diffs. Focus review on `internal/api/v1/{appchart,builderimage,service}/`, `internal/{appchart,builderimage,services}/`, and the new `*_test.go` files.

- `POST` / `PATCH` / `DELETE` added for `/appcharts/:name`, `/builderimages/:name`, and `/catalogservices/:name`. Existing list/show/match endpoints unchanged.
- Renamed `Buildpack` → `BuilderImage` end-to-end (package, models, errors, routes, auth actions, CRD `Kind`/plural/singular, default seed). The old name was a misnomer — these resources reference builder images, not Paketo buildpacks. Paketo references in stage scripts are untouched.
- Inner `BuilderImage.BuilderImage` field also renamed to `Image` (and JSON/spec key `builderImage` → `image`) to remove the awkward repetition introduced by the type rename.
- Logging pattern (`requestctx.Logger` + `Infow` entry/return/steps) propagated to every new and existing handler in the touched packages.

### Technical notes summary
- Domain-layer functions in `internal/builderimage` and `internal/appchart` now take `dynamic.NamespaceableResourceInterface` directly (instead of `*kubernetes.Cluster`) so they're cleanly unit-testable. Handlers resolve the client themselves via `cluster.Client{AppChart,BuilderImage}()`. Two thin `LookupViaCluster`/`ExistsViaCluster` wrappers preserve the legacy call sites in `internal/api/v1/application/` without churn.
- New shared fakes package `internal/testfakes/k8sdynamic/k8sdynamicfakes/` (counterfeiter, generated) is reused across the appchart/builderimage/services unit suites.
- Auth: `service_write` action group now also gates the catalog write routes; added `builderimage`, `builderimage_read`, `builderimage_write` action groups for parity with `service`/`appchart`.
- Update semantics for primitives: empty-string fields are skipped (matches existing `AppChart` partial-update behavior). For `Settings` and `SecretTypes`, nil leaves the field alone; non-nil replaces.

- `POST /api/v1/builderimages`: validate 201 on first create, 409 on duplicate, 400 when `name` or `image` missing.
- `PATCH /api/v1/builderimages/:name`: confirm partial update doesn't overwrite untouched fields; 404 for unknown name.
- `DELETE /api/v1/builderimages/:name`: 200 on success, 404 if missing.
- Same matrix for `/appcharts/:name` and `/catalogservices/:name`.
- Verify the helm-charts companion PR is deployed before exercising builder image endpoints: the renamed `BuilderImage` CRD must be installed (the `Buildpack` CRD never landed in any release, so no migration needed).
- Spot-check: the seeded catalog services (`postgresql-dev`, `mysql-dev`, etc.) still list correctly via `GET /catalogservices`.

- `appchart` domain function signatures changed (now take a dynamic client). All in-tree callers updated, but any out-of-tree consumer importing these packages directly will need to follow the same pattern.
- `internal/services/catalog.go` gained a public `CatalogServiceExists`; verify nothing shadowed that name.


grudra7714 (approved):
lgtm

PR #3026, opened 2026-05-12
Fix/app config save guard

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
 
### Summary

Backend fix for application saves that were timing out when a workload existed but the app was not in a running state. The change keeps config mutations user-saveable while avoiding synchronous redeploys for non-running apps.

- Updated application save behavior so `PATCH /applications/:app` only performs a synchronous `DeployApp()` restart when:
  - the app is currently `running`, or
  - the requested instance count changed and deployment is genuinely required.
- For apps with an existing workload in a non-running state, config changes are now still persisted, but restart is skipped to avoid blocking on Helm rollout waits.
- Applied the same running-state guard to related mutation endpoints:
  - environment variable set
  - environment variable unset
  - configuration bind
  - configuration unbind
  - app redeploy helper used by configuration/service updates
- Added unit coverage for the instance-change detection helper used by the new restart/deploy logic.

### Technical notes summary
- `internal/api/v1/application/update.go`
  - introduced `appInstancesChanged()` to distinguish pure config saves from true scaling/deploy operations
  - changed restart flow so non-running workloads no longer enter the synchronous Helm deploy path
- `internal/api/v1/application/deploy.go`
  - `Redeploy()` now restarts only running apps; non-running apps keep saved configuration without forcing rollout
- `internal/api/v1/env/set.go` and `internal/api/v1/env/unset.go`
  - save/remove env vars without redeploying non-running workloads
- `internal/api/v1/configurationbinding/create.go` and `internal/api/v1/configurationbinding/deletebinding.go`
  - bind/unbind configuration without redeploying non-running workloads
- `internal/api/v1/application/update_test.go`
  - added tests for instance-change detection logic
- Goal of the change:
  - persist user-requested changes
  - avoid hanging requests caused by Helm waiting on unhealthy/non-running workloads
  - still fail when deployment is actually required and cannot succeed

- This was not browser-tested locally as part of the backend-only change.
- Test from the UI or API that saving app config succeeds quickly when the app has a workload but is not running:
  1. Create or use an app with an existing workload.
  2. Move it into a non-running state (`error`, failed, transitional, etc.).
  3. Change environment variables or settings.
  4. Save.
  5. Verify the request returns successfully without waiting for rollout completion.
  6. Verify the saved config is persisted.
- Test running-app behavior:
  1. Use an app in `running` state.
  2. Change env/settings/config bindings.
  3. Save.
  4. Verify restart/redeploy still happens and failures are still surfaced if rollout fails.
- Test instance-change behavior:
  1. Change instances from `0 -> 1` or another different value.
  2. Save.
  3. Verify deployment still happens and errors remain fatal if deploy cannot complete.
- Test env/config binding endpoints directly:
  - set env
  - unset env
  - bind configuration
  - unbind configuration
  - confirm they skip restart for non-running apps but still persist changes

- Any application mutation path that previously assumed “save always implies immediate redeploy”
- Scaling behavior, especially when instance count changes are submitted alongside other config changes
- Configuration/service binding flows that depend on restart timing to make new values immediately visible in workloads
- Running-app save flows, to ensure valid restart behavior was not weakened
- Redeploy behavior for consumers of `application.Redeploy()` that expect active workloads to restart but inactive workloads to remain untouched


dcharles525 (approved):
LGTM

PR #3033, opened 2026-05-18
Fix/pagination layer move

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented

### Summary
Fixes EPINIO-619: moves pagination to happen before enrichment so expensive Kubernetes queries (pods, ingresses, metrics, staging jobs) are scoped to only the apps/configs/services on the current page, rather than loading everything and slicing after.

- `application/index.go`: paginated path now calls `ListPaginated` instead of `List` + post-slice
- `namespace/index.go`, `configuration/fullindex.go`, `service/fullindex.go`: same pattern, paginate first, then scope binding/app lookups to page namespaces only
- New `ListPaginated` function scopes pods, ingresses, metrics, and staging queries to page app names via a new `appNamesInSelector` helper (builds `in (...)` label selectors for multi-app queries)
- New `BuildPaginatedResponse[T]` helper in `response` package, builds pagination metadata from a pre-sliced result + known total
- New scoped variants: `StagingStatusesForApps`, `BoundAppsNamesForNamespaces`, `ServicesBoundAppsNamesForNamespaces`
- `AddApplicationPods`, `AddActualApplicationRoutes`, `GetPodMetrics` extended with `appNames []string` parameter (nil = fetch all, unchanged behavior)

### Technical notes summary
The previous approach called `PaginateSlice` on a fully-enriched list. For large namespaces this was wasteful, every pod, ingress, and metric for the entire namespace was fetched to serve a single page. The new approach fetches only the CRs to determine the page, then scopes all downstream Kubernetes queries to just those app names. Non-paginated callers pass `nil` and get the original behavior.

- `GET /namespaces/:ns/applications?page=1&pageSize=N` returns `PaginatedResponse` with correct `totalItems`, `totalPages`, and `len(items) == pageSize`
- Same for `GET /namespaces`, `GET /configurations`, `GET /services` with page params
- All four endpoints without page params still return the full flat list (backwards-compatible)
- Page beyond total returns empty `items` with correct `totalItems`

- Any client parsing the list response as a flat array will break when page params are passed, response shape changes from `[]T` to `{items, page, pageSize, totalItems, totalPages}`
- The non-paginated path is unchanged, so existing clients not sending page params are unaffected

### DIFF
--- a/acceptance/api/v1/application_index_test.go
+++ b/acceptance/api/v1/application_index_test.go
@@ -110,6 +110,41 @@ var _ = Describe("Apps Endpoint", LApplication, func() {
 		Expect(paged.Items).To(HaveLen(1))
 	})
 
+	It("filters applications by search term", func() {
+		app1 := catalog.NewAppName()
+		env.MakeContainerImageApp(app1, 1, containerImageURL)
+		defer env.DeleteApp(app1)
+		app2 := catalog.NewAppName()
+		env.MakeContainerImageApp(app2, 1, containerImageURL)
+		defer env.DeleteApp(app2)
+
+		response, err := env.Curl("GET", fmt.Sprintf("%s%s/namespaces/%s/applications?search=%s",
+			serverURL, v1.Root, namespace, app1), strings.NewReader(""))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(response).ToNot(BeNil())
+		defer response.Body.Close()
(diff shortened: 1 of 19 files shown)


psanghvi17:
Moving pagination before enrichment fixes the regression. Approved.

PR #3046, opened 2026-06-03
Remove multiple CI badges from README

dcharles525:
Removed several CI badges for AKS, EKS, GKE, RKE, RKE2, and Let's Encrypt. We don't have accounts and resources for this at the moment.


johnlcos (approved):
LGTM

PR #3047, opened 2026-06-08
Epinio Watch

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [x] If applicable, a PR in the epinio/docs repository has been opened [Docs PR](https://github.com/epinio/docs/pull/479)

### Summary

Adds `epinio app watch`, an inner-loop development workflow that syncs local changes into a running application without a full push per change. After a one-time startup push, saving a file results in the change running in the
cluster in seconds instead of minutes.

**CLI**
- New `epinio app watch NAME` command (`--namespace`, `--path`). Watches the source directory, hashes files on a 500 ms poll, and syncs changes to the running pod.
- Optional `.epinio-sync.yaml` config: `build_cmd` + `binary` enable binary mode for compiled languages; `files_dest`, `binary_dest`, and `process_cmd` override defaults. Without config, files mode is used (interpreted
  languages).
- Respects `.gitignore` and `.epinioignore`, the same exclusion rules as `epinio push`.

**API**
- `PATCH /namespaces/:namespace/applications/:app/source` (AppSourcePatch): replaces the app source blob, restages, and swaps the pod image with a supervisor wrapper, bypassing a full Helm deploy for speed.
- `POST /namespaces/:namespace/applications/:app/sync` (AppSync): streams a tar of changed files ("files" mode) or a compiled binary ("binary" mode) into the running pod via server-side exec, then restarts the app process.
- Both routes registered in `auth/actions.yaml` under `app_stage` and documented in the swagger annotations.
- `S3Manager` gained `Download` (fetching the existing source blob to patch).
- `executeStage` extracted from the `Stage` handler so `SourcePatch` reuses the staging job logic.

**Tests**
- Unit: watch helpers (hashing, ignore rules, config, tar, diff), sync command construction, `findReadyPod`, `swapPodImage` (k8s fakes), `extractTar`, multipart request handler, CLI command wiring.
- Acceptance: endpoint suite (validation errors, source patch + supervisor  install, files sync) in `acceptance/api/v1/application_sync_test.go`, and a full CLI loop test (push, watch, edit, binary sync, route verification) in `acceptance/apps_watch_test.go`.

### Technical notes summary

- The startup patch injects a small shell supervisor as PID 1: it launches the app, records the child PID to `/epinio-sync/pid` (an emptyDir), and relaunches on exit. Syncs deliver content and `kill -9` the child; the supervisor restarts it with the new code in under a second, with no container restart counted by Kubernetes.
- The supervisor discovers the app entrypoint from `/cnb/process`: `web` when the buildpack defines it, otherwise the image's first process symlink (the Paketo Go buildpack names the process after the module, so `web` does not exist there). `process_cmd` overrides this for non-CNB images.
- Files mode `chmod -R u+w`s the app dir before extracting because buildpacks leave staged sources read-only.
- The sync exec uses positional shell args so paths with spaces or special characters are safe without string interpolation.
- Staging after a source patch runs asynchronously; progress is logged server-side with per-step timings. The CLI polls the app until ready.

1. Interpreted app (files mode): push any app, run  `epinio app watch <name>` from its source directory, wait for  "Watching for changes", edit a source file, confirm the change is served within a few seconds.
2. Compiled app (binary mode): add `.epinio-sync.yaml` with `build_cmd: "CGO_ENABLED=0 go build -o ./bin/app ."` and `binary: "./bin/app"`, then repeat the loop; the binary is rebuilt locally and swapped into the pod.
3. State reset: delete `.epinio-patch-state` and rerun watch; a fresh startup push should run. 
4. Regular `epinio push` on an app never used with watch (staging refactor regression check).
5. Focused acceptance runs:
   `ginkgo --focus 'AppSourcePatch and AppSync' acceptance/api/v1/.` and
   `ginkgo --focus 'Apps watch' acceptance/.`

- `internal/api/v1/application/stage.go`: the `Stage` handler body moved into `executeStage` (shared with SourcePatch). Behavior is intended to be identical; the regular push staging path is the area to watch.
- Apps that have used `watch` keep the supervisor wrapper and pinned image until the next regular push or restage, which restores the standard Helm managed deployment.
- Everything else is additive: new routes, new client methods, new CLI command, new tests. Apps never touched by watch follow the existing code paths unchanged.


psanghvi17:
LGTM

grudra7714 (approved):

PR #3048, opened 2026-06-16
Git configuration selection, global configs, and private-repo support

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [x] Does the MCP server need an update? - Yes task created
- [x] If applicable, a PR in the epinio/docs repository has been opened - https://github.com/epinio/docs/pull/489

### Summary
Use the following branches for testing: 
- UI: [1.14.1](https://github.com/epinio/ui)
- Docs: [feat/git-config-updates](https://github.com/epinio/docs/pull/489)
- Helm Chart: [feat/crud-crds](https://github.com/epinio/helm-charts/pull/674)

Reworks git-configuration handling so application sources can be pulled from private or enterprise repositories through explicitly selected git configurations. Git configs can now be marked global (usable by everyone in the org, admin-only to create), and existing apps keep working through a compatibility fallback.

- **Explicit gitconfig selection.** Push/import now uses a `gitconfig` selected by the caller; the server no longer implicitly URL-matches a config at push time. The selected config is stamped onto the app origin (`spec.origin.git.gitconfig`) for later re-clones.
- **Global git configs.** New `global` flag on a git config: global configs are usable by any user with app-creation access; non-global configs remain restricted to their creator plus admins. Only admins may create a global config.
- **Private-repo provider support** for GitHub/GitLab types, plus additional supported git-proxy URLs for repository preview.
- **Auth hardening.** Route-level `GitconfigAuthorization` middleware replaced by inline `auth.CanUseGitconfig` checks at the point of use; `FilterGitconfigResources` governs read visibility.
- **Legacy compatibility fallback.** For apps created before per-app gitconfig stamping, `resolveBlobUID` falls back to URL-matching a config so private-repo re-clones keep working on redeploy.

### Technical notes summary
- Git configs remain labeled Kubernetes Secrets; the Secret schema gains an optional `global` key that reads as `false` when absent (backward compatible with existing configs).
- `cloneAndUpload` no longer resolves a config internally; callers pass an explicit `*gitbridge.Configuration`. Push resolves it via `resolveGitconfig` (rejects unknown names, enforces `CanUseGitconfig`); the system re-clone in `resolveBlobUID` resolves from the app's stored stamp, then falls back to URL match for legacy apps.
- Certificate and password fields remain write-only (never returned on read).

- Push an app from a **private** repo selecting a valid gitconfig: succeeds; the app gains `spec.origin.git.gitconfig`.
- Push selecting a gitconfig the user is **not** permitted to use: 403.
- Create a **global** config as a non-admin: 403; as an admin: succeeds and is usable by other users.
- **Legacy app** (no stamp) whose S3 blob is gone: redeploy triggers a re-clone that still authenticates via URL match (private clone succeeds).
- Existing git-config Secrets (no `global` key) still list and function.

- **Push workflow change (private repos).** Because implicit URL matching at push is removed, an **old CLI, script, or API caller that does not send `gitconfig`** will fail to clone a private repo that previously worked. Mitigation: updated CLI drives the suggest flow; document that private-repo pushes now require selecting a config, and that automation must pass one.
- **Cross-repo dependency (silent pruning).** This PR writes `spec.origin.git.gitconfig`. That App-CRD field ships via the helm chart and reaches existing clusters only through its CRD-upgrade hook. If this server runs against a chart without the field, the apiserver prunes the stamp and re-clones lose their config. Deploy the CRD field/hook **before or with** this change.
- **Legacy re-clone** is covered by the URL-match fallback added here; verify it does not pick an unintended config when multiple match.
- **Auth surface change.** `GitconfigAuthorization` middleware was removed; confirm the gitconfig show/delete routes still enforce authorization via the inline checks (security-review this).


psanghvi17:
LGTM

PR #3051, opened 2026-07-06
Add GET endpoint to retrieve application staging source.

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [x] Does the [MCP](https://github.com/epinio/mcp) server need an update?
 

### Summary
Fixes #
Adds an authenticated API endpoint to download an application’s staging source tarball. Clients (e.g. the Epinio MCP server) can fetch app source through Epinio instead of connecting directly to S3, so normal API authentication and RBAC are enforced.

Added GET /api/v1/namespaces/:namespace/applications/:app/source endpoint (AppSource route).
Implemented GetSource handler in internal/api/v1/application/source.go.
Registered the route in internal/api/v1/router.go.
Added AppGetSource client method in pkg/api/core/v1/client/apps.go.
Added Open() to the S3Manager interface and s3manager.Manager to stream blobs from S3 server-side.
Wired AppSource to the app_export auth action in internal/auth/actions.yaml.
Updated mockS3Manager in internal/application/application_test.go to satisfy the extended interface.
Added acceptance tests in acceptance/api/v1/application_source_test.go.

### Technical notes summary
The handler loads the app CR from Kubernetes and reads spec.blobuid via the existing findPreviousBlobUID helper (same as staging/deploy).
Before returning data, it validates the S3 blob metadata with validateBlob to ensure the blob belongs to the requested app and namespace.
S3 credentials stay on the Epinio server; the client never needs direct S3 access.
The response is streamed as a tarball with Content-Disposition: attachment; filename="<app>-source.tar".
The endpoint is protected under the app_export permission (alongside AppPart and AppExport), requiring app_read as a dependency.
Source is only available after the app has been staged (when spec.blobuid is set on the app resource). Apps that exist but have no stored source return HTTP 400 with "application has no stored source".

Happy path — staged app with source

Create a namespace and application.
Upload source via POST .../applications/:app/store.
Stage the app via POST .../applications/:app/stage.
Call GET .../applications/:app/source.
Expect HTTP 200 and a valid .tar archive in the response body.
App exists but no source

Create an app without uploading/staging.
Call GET .../applications/:app/source.
Expect HTTP 400 with error title "application has no stored source".
Unknown app

Call GET .../applications/bogus/source in a valid namespace.
Expect HTTP 404.
Authorization

Call the endpoint with a user/token that has app_read but not app_export.
Expect HTTP 403.
Call with a user that has app_export on the namespace.
Expect HTTP 200 (for a staged app).
Client library

Verify Client.AppGetSource(namespace, appName) returns the tarball stream and correct content length.
Automated tests

Run acceptance/api/v1/application_source_test.go.
Run unit tests for internal/s3manager, internal/api/v1/application, and pkg/api/core/v1/client.

S3 manager interface — Open() was added to S3Manager; any mock or alternate implementation must implement it (already updated in application_test.go).
Auth / RBAC — New route under app_export; roles that previously could read apps but not export may be denied source download (intentional).
S3 streaming — Large source tarballs are streamed via DataFromReader; verify no memory issues or broken responses for large archives.
Blob validation — Reuses validateBlob from staging; mismatched blob metadata (wrong app/namespace) should return 400, not leak data.
Staging workflow — findPreviousBlobUID depends on spec.blobuid being set during staging; upload-only apps without staging will not return source.


dcharles525 (approved):
LGTM

PR #3076, opened 2026-07-21
Feat/builderimage cr as default source

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [ ] Unit Tests are passing
- [x] Code is well documented
 

### Summary
Fixes #
Use the BuilderImage custom resource as the source of truth for the default builder image, with the existing DEFAULT_BUILDER_IMAGE environment variable retained only as a fallback when no CR is marked as default.

- Added builderimage.Default() to resolve the cluster default from BuilderImage CRs (spec.default: true), with lexicographic name tie-breaking when multiple defaults exist, and nil when none are configured.
- Added resolveDefaultBuilderImage() and wired it into:
- internal/api/v1/application/stage.go
- internal/api/v1/application/sourcepatch.go
- internal/api/v1/application/deployments.go
- Updated /info to report the CR-derived default builder image, falling back to the env var when no default CR exists.
- Replaced direct viper.GetString("default-builder-image") lookups in the above paths with CR-first resolution.
- Added unit tests for builderimage.Default.
- Added an acceptance test verifying staging uses the CR default and that changing the default CR updates both staging behavior and the /info response.

### Technical notes summary

- Resolution order is: explicit request value → application CR spec.builderimage → default BuilderImage CR → DEFAULT_BUILDER_IMAGE env fallback.
- builderimage.Default() reuses the existing List() helper and sorts matching CRs by name for deterministic selection.
- List/client failures are surfaced as internal API errors instead of silently falling back.
- Async deploys retain the existing guard that returns a bad request when no builder can be resolved from any source.
- 

-  Stage an app without specifying a builder image and confirm the default BuilderImage CR image is used in the staging pod.
-  Patch a different BuilderImage CR to spec.default: true (and unset the previous default), then stage a new app and confirm the new default is used.
-  Call GET /api/v1/info and confirm default_builder_image reflects the CR default.

- Staging when no builder is explicitly provided (previously env-only; now CR-first).
- Source patch rebuilds for apps without a stored builder image.
- Async deploy staging path when builder is not specified in the request or app CR.

### DIFF
--- a/acceptance/api/v1/application_stage_test.go
+++ b/acceptance/api/v1/application_stage_test.go
@@ -16,6 +16,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"os"
 	"strings"
 
 	"github.com/epinio/epinio/acceptance/helpers/catalog"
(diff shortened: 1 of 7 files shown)


dcharles525 (approved):
LGTM

PR #3079, opened 2026-07-24
Feat/allow appchart update for active apps

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [ ] Code is well documented
 

### Summary
Fixes #

Allow changing an application’s app chart after deploy via the application update API (API-only; UI is out of scope).
Previously, PATCH rejected app-chart changes for apps with an active workload (unable to change app chart of active application). That block is removed, with validation, redeploy, rollback, and Helm upgrade hardening so chart switches are safe.

1. Application update API (internal/api/v1/application/update.go): app chart can be updated for inactive and running apps.
2. Pre-mutation validation: on chart change, existing (or requested) settings are validated against the target chart before any CRD patch.
3. Forced redeploy on chart change when a workload exists (even if restart=false), so the workload picks up the new Helm chart.

### Technical notes summary

 
- Chart switches are no longer blocked solely because app.Workload != nil.
- Settings validation runs when the chart is changing, or when a running app gets new settings; for chart-only updates, existing settings are checked against the target chart.
- Deploy still re-Lookups the app, so the patched spec.chartname is what Helm uses.
- Chart-name rollback on deploy failure does not reverse other mutations in the same request (settings/env/routes/etc.). Prefer chart-only updates, or treat combined updates carefully in QA.
- shouldReuseHelmValues compares the current release’s chart name-version to the next AppChart’s Helm reference (URL or repo form).

Happy path — running app
Deploy an app on standard, then run:
`epinio app update <app> --app-chart <alternate-chart>`
Expect: success; epinio app show shows the new chart; status returns to 1/1; Helm release uses the new chart package and deployed.

Inactive / no workload
epinio app create with a chart, then app update --app-chart … (still works as before).

Incompatible settings
Set a chart value on the app, switch to a chart that does not declare that setting. Expect: request fails with a bad-request / “setting … not known”; original app chart unchanged.

Unknown chart
app update --app-chart does-not-exist → chart-not-known error; no CRD change.

Application update / redeploy path for any running-app PATCH (restart logic, deploy-after-mutate).
Helm InstallOrUpgradeChart for apps: ReuseValues is now conditional; same-chart updates should still reuse values; cross-package upgrades should not.
Chart-value / settings validation timing and error shape on update.


internal/api/v1/application/update.go
```diff
@@ -95,32 +95,36 @@ func Update(c *gin.Context) apierror.APIErrors { // nolint:gocyclo // simplifica
...
+	if updateRequest.AppChart != "" {
+		targetChartName = updateRequest.AppChart
+	}
 
-		appChart, err := appchart.LookupViaCluster(ctx, cluster, app.Configuration.AppChart)
+	settingsInRequest := len(updateRequest.Settings) > 0
+	// Chart changes always require a settings check (existing or requested). Running apps also
+	// validate when new settings are supplied.
+	if chartChanging || (app.Workload != nil && settingsInRequest) {
+		appChart, err := appchart.Lookup(ctx, cluster, targetChartName)
```
dcharles525: The committed appchart.Lookup(ctx, cluster, targetChartName) no longer compiles after the Lookup/LookupViaCluster split

internal/api/v1/application/update.go
```diff
@@ -199,31 +206,37 @@ func Update(c *gin.Context) apierror.APIErrors { // nolint:gocyclo // simplifica
...
 	}
 
-	// backward compatibility: if no flag provided then restart the app
+	// backward compatibility: if no flag provided then restart the app.
+	// Chart changes on an active app always redeploy: the workload must pick up the new
+	// helm chart. On failure, roll the CRD chart name back so we do not leave a partial update.
 	restart := updateRequest.Restart == nil || *updateRequest.Restart
-	if restart {
+	forceChartRedeploy := chartChanging && app.Workload != nil
+	if restart || forceChartRedeploy {
```
dcharles525: forceChartRedeploy ignores an explicit Restart: false. This will fire a full helm upgrade of the old image onto the new chart before staging, then deploys again. If the first one fails, the new rollback reverts the chart name mid-push.

internal/helm/helm.go
```diff
@@ -401,14 +396,58 @@ func Deploy(parameters ChartParameters) error {
...
 	}
 
 	_, err = client.InstallOrUpgradeChart(context.Background(), &chartSpec, nil)
 
 	return err
 }
 
+// shouldReuseHelmValues is false when an existing release was installed from a different
+// chart package than the one about to be deployed (typical for app chart switches).
+func shouldReuseHelmValues(client hc.Client, releaseName string, appChart *models.AppChartFull) bool {
```
dcharles525: - Substring matching on the chart URL. 0.1.2 matches a 0.1.21 or 0.1.26 URL, so a real version change keeps reuse on. Same for the substring-ing of the name, epinio-application would match epinio-application-gateway-api. 
- Charts whose URL does not encode name-version disable reuse on every deploy, restarts and scaling included.
- Repo-form charts with no :version tag never detect a version change (SplitN yields one piece, the version compare is skipped)

We probably also need to add coverage for two AppChart CRs wrapping the same helm package, bug differing in spec.values or spec.settings.

acceptance/apps_test.go
```diff
@@ -458,18 +458,88 @@ var _ = Describe("Apps", LApplication, func() {
 
 				BeforeEach(func() {
```
dcharles525, line 444: These tests will need to be updated after below changes are resolved.

dcharles525 (changes requested):
While this unblocks the update app chart functionality, a good bit of the written code doesn't function as expected.

dcharles525 (approved):
LGTM

PR #3100, opened 2026-08-10
fix EPINIO-716: show failed builds as error and allow retry

psanghvi17:
### Summary
Fixes #EPINIO-716

Failed builds were shown as `created` in the applications table. This change surfaces staging/deploy failures as `error`, and allows async deploy to retry using a stored source blob (S3/SeaweedFS) or by re-cloning from git when no image/blobuid is provided.

Companion UI PR: https://github.com/epinio/ui/pull/638

- Application status no longer reports `created` when staging failed or when a stage was attempted but no workload is running (including after staging Jobs are removed by TTL).
- Async deploy (`DeploymentsStart`) accepts empty `image` and `blobuid` so a rebuild can resolve source from the app CR blobuid or git origin via `resolveBlobUID`.
- App API model now exposes `blobuid` so clients can tell when a reusable upload/archive source is available.
- Unit tests added for `assignApplicationStatus` (follow-up commit fixes duplicate Ginkgo suite registration).

### Technical notes summary
- Extracted status assignment into `assignApplicationStatus` / `wantsInstances`.
- Sticky failure signal: when `StagingStatus` is empty because the Job is gone, a non-empty `StageID` plus desired instances maps to `error` (`staging failed`) instead of `created`.
- Staging done with no workload and desired instances maps to `error` (`deployment failed`).
- Running workloads still win: an app with an active workload stays `running` even if a later stage failed.
- Retry path reuses existing `resolveBlobUID` (stored blob first; git re-clone fallback when the blob is missing).

1. Push an app that fails staging (bad source / build error). App list/detail status should be `error` (not `created`), with a useful status message.
2. After the staging Job is deleted/TTL-cleaned, status should remain `error` (sticky `stage_id`), not flip back to `created`.
3. Retry/rebuild a failed **git** app with empty image/blobuid (async deploy): should re-clone if needed and restage/deploy.
4. Retry/rebuild a failed **upload/folder/archive** app while the blob still exists in S3/SeaweedFS: should reuse `blobuid` and restage/deploy.
5. Retry when blob is gone and there is **no** git origin: should fail clearly (no silent success).
6. App scaled to 0 instances with a prior `stage_id`: should remain `created`, not `error`.
7. App that is already running, then a restage fails: status should stay `running` while workload exists.
8. Regression: normal create → stage → deploy success path still ends in `running`.
9. Regression: image-based async deploy (image provided, no blob) still deploys without forcing staging.

Local API verification via acceptance/unit tests and manual push against a cluster; pair with UI Rebuild testing in a different browser than the author used.

- Application list/detail status for apps mid-deploy, scaled-to-zero apps, and apps whose staging Jobs have already been garbage-collected.
- Async deploy clients that previously relied on a hard 400 when both `image` and `blobuid` were empty.
- Any client that assumes missing workload always means `created`.
- Restage/rebuild flows that depend on `blobuid` being present on the app response.
- Staging resolution path shared with `resolveBlobUID` (blob validation + git fallback).


dcharles525 (approved):
LGTM

PR #3101, opened 2026-08-10
feat EPINIO-723: service catalog pagination and search

psanghvi17:
### Summary
Fixes #EPINIO-723

On the server only, add optional pagination and name search to `GET /catalogservices`, matching the pattern already used by namespaces and service instances.

- `GET /catalogservices` now accepts optional `search`, `page`, and `pageSize` query parameters.
- Search filters catalog services by name (case-insensitive substring) before pagination, so page metadata reflects the filtered set.
- When `page`/`pageSize` are provided, the response is a paginated payload (`items`, `page`, `pageSize`, `totalItems`, `totalPages`) via the shared `response.PaginateSlice` helper.
- When page params are omitted, the endpoint still returns the full catalog list for backwards compatibility.
- Acceptance coverage added for pagination metadata and search filtering.

### Technical notes summary
- Implementation mirrors service instance listing (`internal/api/v1/service/list.go`): `response.GetSearchParam` → filter → `response.GetPaginationParams` → `response.PaginateSlice`.
- Defaults match other resources (`page=1`, `pageSize=25` when pagination is requested).
- No UI changes in this PR (server-only scope per ticket).

1. `GET /catalogservices` with no query params → full list (legacy shape), unchanged behavior for existing clients.
2. `GET /catalogservices?page=1&pageSize=1` → paginated response with correct `page`, `pageSize`, `totalItems`, `totalPages`, and one item.
3. Request a page beyond the last page → empty `items` with correct totals (same helper behavior as other resources).
4. `GET /catalogservices?search=<exact-or-partial-name>` → only matching catalog services; case-insensitive.
5. Combine `search` + `page`/`pageSize` → pagination totals/pages describe the filtered set, not the unfiltered catalog.
6. Empty search / unknown search term → empty list (or empty page), not an error.
7. Regression: catalog show/create/delete and “in use” / bound-service counts on listed items still work.

- Existing CLI/UI/API clients that parse `GET /catalogservices` as a bare array (must remain valid when page params are absent).
- Any client that starts sending `page`/`pageSize` and must handle the paginated response envelope instead of a raw list.
- Shared pagination/search helpers (`GetSearchParam`, `GetPaginationParams`, `PaginateSlice`) used by namespaces and service instances.
- Catalog list fields such as `BoundServices` / in-use counts when filtering or paging the list.

### DIFF
--- a/acceptance/api/v1/services_catalog_test.go
+++ b/acceptance/api/v1/services_catalog_test.go
@@ -83,4 +83,62 @@ var _ = Describe("ServiceCatalog Endpoint", LService, func() {
 		}
 		Expect(serviceNames).ToNot(ContainElement(catalogService.Meta.Name))
 	})
+
+	It("returns a paginated response when page params are provided", func() {
+		catalog.CreateCatalogService(catalogService)
+		defer catalog.DeleteCatalogService(catalogService.Meta.Name)
+
+		response, err := env.Curl("GET", fmt.Sprintf("%s%s/catalogservices?page=1&pageSize=1",
+			serverURL, v1.Root), strings.NewReader(""))
+		Expect(err).ToNot(HaveOccurred())
+		Expect(response).ToNot(BeNil())
+		defer response.Body.Close()
+		bodyBytes, err := io.ReadAll(response.Body)
+		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


dcharles525 (approved):
LGTM

PR #3103, opened 2026-08-10
feat EPINIO-726: optional PVC deletion on application delete

psanghvi17:
### Summary
Fixes #EPINIO-726

Add an optional `deletePVC` flag to the application delete endpoint (and CLI `--delete-pvc`). Default remains `false` so staging/app-data PVCs are preserved unless explicitly requested.

Companion UI PR: https://github.com/epinio/ui/pull/639

- `ApplicationDeleteRequest` gains `deletePVC` (JSON `deletePVC`, default false).
- `application.Delete` only removes staging PVCs (cache/source blobs) and app-data PVCs (StatefulSet `volumeClaimTemplates`) when `deletePVC` is true.
- App-data PVC names are collected **before** helm uninstall so StatefulSets still exist.
- CLI: `epinio app delete --delete-pvc`.
- API client / usercmd / fakes updated for the new parameter.
- Acceptance: existing cache-PVC deletion test now requires `--delete-pvc`; new test asserts PVC is kept when the flag is omitted.
- Test helpers `DeleteApp`/`CleanupApp` pass `--delete-pvc` so leftover staging PVCs do not accumulate.
- Namespace delete now removes app PVCs when cascading app deletion (`deletePVC=true`) so staging storage is not left behind after namespace removal.

### Technical notes summary
- Behavior change vs older unconditional staging-PVC deletion on app delete: PVCs are now kept by default; callers must opt in.
- Staging PVC deletion still uses existing `deleteCacheStagePVC` / `deleteSourceBlobsStagePVC`.
- App-data PVC deletion uses label-selected StatefulSets (`app.kubernetes.io/name`) and the standard `{vct}-{sts}-{ordinal}` naming.
- Namespace cascade intentionally opts in to PVC cleanup even though single-app delete defaults to false.

1. `epinio app delete <app>` (no flags) → app gone, cache/source/data PVCs still present.
2. `epinio app delete --delete-pvc <app>` → staging PVCs removed; StatefulSet data PVCs removed when present.
3. API delete with body `{deletePVC:false}` / omitted → same as default keep.
4. API delete with `{deletePVC:true}` (alone and with `deleteImage`) → PVCs removed as expected.
5. Delete namespace containing staged apps → apps and their staging PVCs cleaned up.
6. Acceptance apps suite covering the new keep-vs-delete PVC cases.
7. App with no PVCs / already-missing PVCs → delete still succeeds (NotFound ignored).

- Any automation that assumed app delete always removed staging PVCs (now must pass `--delete-pvc` / `deletePVC: true`).
- Bulk/API clients that construct `ApplicationDeleteRequest` (new field is optional/false by default).
- Namespace deletion cleanup volume/cost behavior (now deletes PVCs).
- Helm uninstall ordering vs PVC collection for StatefulSet-backed apps.
- Acceptance helpers always cleaning PVCs may hide accidental PVC leaks in other tests if they relied on leftover volumes.

### DIFF
--- a/acceptance/apps_test.go
+++ b/acceptance/apps_test.go
@@ -1088,7 +1088,7 @@ var _ = Describe("Apps", LApplication, func() {
 			})
 		})
 		When("deleting the app", func() {
-			It("deletes the cache PVC too", func() {
+			It("deletes the cache PVC when --delete-pvc is set", func() {
 				pvcName := names.GenerateResourceName(namespace, "cache", appName)
 				// Wait for build cache PVC to appear.
 				var found bool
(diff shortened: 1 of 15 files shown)


dcharles525 (changes requested):
This is working good, a few things I want to clarify and then will merge. 

- No tests for the app-data PVC discovery, would be good to have that covered since it drives the deletion process. 
- If replicas are scaled up and down the current listAppDataPVCNames won't catch it, use filtering by label selector instead of re-constructing the name.

dcharles525 (approved):
LGTM

PR #3104, opened 2026-08-10
feat(Add namespace scoping to endpoints): For namespace related resou…

dcharles525:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented

### Summary
Fixes EPINIO-732

Adds an optional `?namespaces=` filter to the three cross-namespace list endpoints, so the dashboard can scope one table to a chosen set of namespaces instead of one table per namespace.

- `GET /applications`, `GET /services`, `GET /configurations` accept `?namespaces=ns1,ns2`
- Absent or empty param keeps existing behaviour, so this is backwards compatible

### Technical notes summary
`response.GetNamespacesParam` parses the list, `auth.FilterByNamespaces` applies it to the output of `auth.FilterResources`, before search and pagination.

- test api/v1/applications?namespaces=, api/v1/services?namespaces=, and api/v1/configurations?namespaces=
- only the named namespaces come back
- unknown name returns `[]`, HTTP 200
- a user without access to a named namespace sees nothing from it
- page reset when the selection changes

- These three endpoints only. Namespace-scoped routes and `/applications/grouped` untouched.
- Bound-apps enrichment on configurations and services is page-scoped; confirmed it still resolves under the filter.

### DIFF
--- a/acceptance/api/v1/application_index_all_test.go
+++ b/acceptance/api/v1/application_index_all_test.go
@@ -95,6 +95,122 @@ var _ = Describe("AllApps Endpoints", LApplication, func() {
 			[]string{app2, namespace2}))
 	})
 
+	// listApps GETs /applications with the given query string and returns the
+	// plain, non-paginated list.
+	listApps := func(query string) models.AppList {
+		url := fmt.Sprintf("%s%s/applications?%s", serverURL, v1.Root, query)
+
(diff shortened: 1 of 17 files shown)


psanghvi17:
LGTM

PR #3126, opened 2026-09-01
feat(EPINIO-683): add persistent instance ID for Epinio installs

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [ ] Unit Tests are passing
- [x] Code is well documented

### Summary
Fixes EPINIO-683

Add persistent instance ID for Epinio installs and expose it in API/CLI.

**Companion PR:** epinio/helm-charts#716

- `internal/instance` package to read/create instance ConfigMap
- `/api/v1/info` returns `instance_id` and `install_method`
- `epinio info` shows Instance ID and Install Method
- Unit + acceptance tests added/updated

### Technical notes summary
- ConfigMap name: `epinio-instance`
- Existing ID is never overwritten on upgrade

1. `epinio info` shows Instance ID and Install Method
2. `GET /api/v1/info` returns new fields
3. Upgrade from older install backfills ConfigMap if missing

- `/info` endpoint
- Server startup

### DIFF
--- a/internal/cli/server/server.go
+++ b/internal/cli/server/server.go
@@ -26,6 +26,7 @@ import (
 	"github.com/epinio/epinio/internal/api/v1/response"
 	"github.com/epinio/epinio/internal/auth"
 	"github.com/epinio/epinio/internal/domain"
+	"github.com/epinio/epinio/internal/instance"
 	apierrors "github.com/epinio/epinio/pkg/api/core/v1/errors"
 	"github.com/pkg/errors"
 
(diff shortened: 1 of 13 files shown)


internal/cli/server/server.go
```diff
@@ -168,6 +169,18 @@ func NewHandler() (*gin.Engine, error) {
 	}
 	rolesInitialized = true
 
+	// Ensure a persistent instance id exists (covers upgrades from older charts).
+	if info, err := instance.GetOrCreate(context.Background(), cluster); err != nil {
```
dcharles525: I think this function needs to be split up GetOrCreate is a dual use function and goes against Go's idiomatic single use paradigm. Should check for an instance ID in one func, create if the logic gate says an instance id is needed.

internal/instance/instance.go
```diff
@@ -0,0 +1,196 @@
...
+// GetCached returns the in-memory instance identity when available.
+func GetCached() (Info, bool) {
+	if info := cached.Load(); info != nil && info.ID != "" {
+		return *info, true
+	}
+	return Info{}, false
+}
+
+// GetCachedOrCreate returns the cached identity, loading/creating it once when needed.
+func GetCachedOrCreate(ctx context.Context, cluster *kubernetes.Cluster) (Info, error) {
```
dcharles525: Same thing here as mentioned in the server.go file.

PR #3131, opened 2026-09-07
feat(EPINIO-688): add and configure otelgin tracing middleware

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [x] If applicable, a PR in the [epinio/docs](https://github.com/epinio/docs) repository has been opened (epinio/docs#499)

### Summary
Fixes #EPINIO-688

Instrument the Gin API server with OpenTelemetry (`otelgin`) so request spans export in OTLP for a collector, for request-flow debugging/tracking.

- Added `internal/tracing` package to init OTel TracerProvider (opt-in) and graceful shutdown
- Wired `otelgin.Middleware` into the Gin router
- Added CLI flags / env: `--otel-exporter-otlp-endpoint` / `OTEL_EXPORTER_OTLP_ENDPOINT`, `--otel-exporter-otlp-protocol` / `OTEL_EXPORTER_OTLP_PROTOCOL`
- Unit tests for no-op path and in-memory span export with `epinio-server` resource

### Technical notes summary
- Tracing is **opt-in**: unset endpoint → no-op TracerProvider (unchanged server behavior)
- When set, spans batch via `go.opentelemetry.io/contrib/exporters/autoexport`, resource `service.name=epinio-server`
- `tracing.Init` runs **before** `NewHandler()` because otelgin captures the global TracerProvider at registration
- Tracer shutdown runs **after** HTTP server drain so in-flight spans flush before exporter close

- Unit: `go test ./internal/tracing/... -v`
- No endpoint set → server starts normally, no export
- With Jaeger/collector + `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4318` + `OTEL_EXPORTER_OTLP_PROTOCOL=http/protobuf` → hit `/api/v1/info` → span in collector for service `epinio-server`

- Server startup / graceful shutdown path (`startServerGracefully`)
- Gin middleware order (otelgin is first among common middlewares)
- Deployments without a collector (must remain no-op / no crash)


internal/tracing/tracing.go
```diff
@@ -0,0 +1,100 @@
...
+
+	return tp.Shutdown, nil
+}
+
+// newTracerProvider builds a TracerProvider batching spans through the
+// given exporter, tagged with the given resource. Split out from Init so
+// tests can exercise it directly with an in-memory exporter.
+func newTracerProvider(exporter sdktrace.SpanExporter, res *resource.Resource) *sdktrace.TracerProvider {
+	return sdktrace.NewTracerProvider(
+		sdktrace.WithBatcher(exporter),
```
dcharles525, line 106: Probably want a error handler added here. otel.SetErrorHandler or otel.SetLogger will work here.

internal/cli/server/server.go
```diff
@@ -93,6 +96,7 @@ func NewHandler() (*gin.Engine, error) {
 
 	// Add common middlewares to all the routes declared after
 	router.Use(
```
dcharles525, line 106: We want to filter out web socket traces, they are long running and the trace doesn't make sense in that context.

internal/tracing/tracing_test.go
```diff
@@ -0,0 +1,126 @@
...
+		})
+	})
+
+	When("only the signal-specific traces endpoint is configured", func() {
+		It("enables tracing", func() {
+			globalEndpoint, hadGlobalEndpoint := os.LookupEnv("OTEL_EXPORTER_OTLP_ENDPOINT")
+			tracesEndpoint, hadTracesEndpoint := os.LookupEnv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT")
+			DeferCleanup(func() {
+				if hadGlobalEndpoint {
+					Expect(os.Setenv("OTEL_EXPORTER_OTLP_ENDPOINT", globalEndpoint)).To(Succeed())
```
dcharles525, line 72: Nit, but we need to clean OTEL_EXPORTER_OTLP_PROTOCOL as well here since it will leak in later specs.

dcharles525 (commented):
A few things in general. 

1) There needs to be docs so people know how to utilize this, the docs check on the PR is not checked. 
2) There aren't any tests that cover the otelgin path via the gin router.

internal/tracing/tracing.go
psanghvi17: Added: otel.SetErrorHandler now routes exporter errors through the zap logger instead of stderr.

internal/cli/server/server.go
psanghvi17: Added: otelgin.WithFilter(tracedPath) now excludes websocket routes from tracing.

internal/tracing/tracing_test.go
psanghvi17: Fixed, and generalized further: a BeforeEach + preserveEnv helper now saves/restores all four relevant OTEL_* env vars (including this one) around every spec in this Describe block.

PR #3132, opened 2026-09-10
feat(EPINIO-689): instrument k8s API calls with otelhttp

psanghvi17:
### PR Checklist
- [x] Linting Test is passing
- [x] Unit Tests are passing
- [x] Code is well documented
- [x] If applicable, a PR in the [epinio/docs](https://github.com/epinio/docs) repository has been opened (epinio/docs#499)

### Summary
Fixes #EPINIO-689

Instrument outbound Kubernetes API HTTP traffic with OpenTelemetry (`otelhttp`) so client spans and W3C trace context are emitted for request-flow and resource-usage tracing. This builds on the EPINIO-688 server tracing foundation.

- Added `tracing.WrapHTTPRoundTripper` using `otelhttp.NewTransport` with an explicit W3C TraceContext propagator. Baggage is deliberately **not** propagated to the Kubernetes API — see Technical notes.
- Applied the wrapper via `rest.Config.Wrap(...)` in `helpers/kubernetes/config`, so all consumers of `KubeConfig` (clientset, dynamic client, Helm, metrics) inherit the same instrumentation.
- Added unit tests that assert a client span is recorded, the `traceparent` header is injected on outbound requests, and `baggage` is not.
- Added dependency `go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp`.

No API contract changes. Swagger regeneration is not required. MCP updates are not required for this PR; documentation covering this alongside EPINIO-688/690 has been opened at epinio/docs#499.

### Technical notes summary
- Safe when tracing is disabled: if a no-op `TracerProvider` is installed (OTLP endpoint unset, EPINIO-688 behavior), spans are discarded and server behavior remains unchanged.
- **Baggage is intentionally excluded from the propagator set here** (`otelhttp.WithPropagators(propagation.TraceContext{})`, not a composite with `propagation.Baggage{}`) — this differs from the global propagator `tracing.Init` sets for the Gin side, and deliberately so. Baggage is arbitrary caller-controlled key/value metadata; forwarding it onto requests to the privileged Kubernetes API would let a caller inject headers into that traffic. Only trace context (which carries no caller-supplied values, just IDs) is propagated. An earlier version of this PR's description inaccurately said both were propagated — the code was already correct, this has now been corrected to match.
- Propagators are set at wrap time (not read from globals later), so injection still works if wrap runs before `Init`, and a later `Init` cannot leave an already-wrapped transport on a no-op propagator.
- Follows the Kubernetes component-base pattern: `rest.Config.Wrap(otelhttp.NewTransport)`.
- This PR depends on EPINIO-688 (`otelgin` + tracer `Init`). Prefer opening against the EPINIO-688 branch, or rebase onto `development` after EPINIO-688 is merged.

1. Tracing disabled (no `OTEL_EXPORTER_OTLP_ENDPOINT` / `--otel-exporter-otlp-endpoint`): start `epinio-server` and run normal app/namespace Kubernetes operations. Behavior should be unchanged and no spans should be exported.
2. Tracing enabled: set an OTLP endpoint (collector / Grafana Alloy / Tempo), then trigger API actions that call the Kubernetes API (for example app list/create or namespace operations).
3. In Tempo/Jaeger/collector UI, confirm client spans for Kubernetes API HTTP calls appear under service `epinio-server`, with correct parent/child linkage to Gin request spans from EPINIO-688.
4. Confirm outbound requests carry a `traceparent` header but no `baggage` header (covered by unit tests; optional live check via collector debug logs).
5. Run unit tests for `internal/tracing`, especially the `WrapHTTPRoundTripper` suite.

- Any code path that builds Kubernetes clients from `KubeConfig` (clientset, dynamic, Helm, metrics), because RoundTripper wrapping sits in the shared HTTP transport chain.
- Trace context propagation / header injection on requests to kube-apiserver.
- Server startup and shutdown ordering relative to tracer `Init` (EPINIO-688) when both Gin middleware and the HTTP RoundTripper wrapper are active.
- Performance under heavy Kubernetes API traffic when the exporter is enabled, due to extra span creation per HTTP round-trip.


internal/tracing/tracing.go
```diff
@@ -98,3 +100,23 @@ func newTracerProvider(exporter sdktrace.SpanExporter, res *resource.Resource) *
...
+//
+// Safe to install unconditionally: when Init installed a no-op
+// TracerProvider the spans are discarded; with an exporter configured they
+// are exported. Matches the Kubernetes component-base tracing pattern of
+// rest.Config.Wrap(otelhttp.NewTransport).
+//
+// Propagators are set explicitly (not read from the global at wrap time) so
+// injection works even if this runs before Init, and so a later Init cannot
+// accidentally leave an already-wrapped transport on a no-op propagator.
+func WrapHTTPRoundTripper(rt http.RoundTripper) http.RoundTripper {
```
dcharles525, line 128: In order to understand all of the children in the trace, we'd want to add a formatter deriving the span name from the k8s resource path

dcharles525 (commented):
The PR body says that the wrapper uses "explicit W3C TraceContext and Baggage propagators", but in the code it doesn't do that? What is correct here?

psanghvi17:
@dcharles525 re: your review comment on the Baggage propagators — the code was already correct, the PR description was just wrong. `WrapHTTPRoundTripper` only sets `otelhttp.WithPropagators(propagation.TraceContext{})`, no Baggage. I've corrected the description now.

Baggage is deliberately excluded here: it's arbitrary caller-controlled key/value metadata, and forwarding it onto requests to the privileged Kubernetes API would let a caller inject headers into that traffic. Only trace context (IDs, no caller-supplied values) is propagated. The existing `WrapHTTPRoundTripper` test already asserts the outbound `baggage` header is empty while `traceparent` is present, so this is covered.

internal/tracing/tracing.go
psanghvi17: Added: k8sSpanName derives the span name from the request path, e.g. "GET namespaces/default/pods" instead of "HTTP GET" for every call. Falls back to the default for anything that isn't a /api or /apis path. Table-driven unit tests cover core v1, grouped APIs, cluster-scoped resources, subresources, and the fallback case.

